PR #4, opened 2015-06-04
renamed readme, added built status, logo and headings

hackerkid:
renamed README to README.md
Added tumux logo in README.md
Added built status in README.md
Added titles in README.md


nicm:
Sorry, I do not want to clutter up README. It is not a long document so markup is not required.

PR #15, opened 2015-06-06
Markdownify README

yvan-sraka:


nicm:
We are not going to go with markdown for files that ship with tmux thanks

alotaibi1967:
Sex

Issue #17, opened 2015-06-07
Switching to another session blows away target session's environment variables!


unphased:
_I was surprised to find my email to tmux-users mailing list bounced. But GitHub is a great new home for tmux. Welcome to GitHub!_

I had a workflow that I quite liked which was that I would create **grouped** sessions and treat them in this way: each session "belongs" to a piece of physical hardware. 

My laptop's session `mac` can stay on window 0, and my desktop's large portrait-oriented monitor can be used in session `putty` can stay on window 1. The beauty is that the list of windows are shared in the sessions and the dimensional constraints are lifted whenever the sessions focus on different windows. 

I began to leverage more behavior based on sessions having their own environments. This is covered well in the manpage under ENVIRONMENT. What I took a liking to was that I could set `$GIT_AUTHOR_NAME` to a different value per session. This is really great because I can now train my shell to call `tmux showenv` and differentiate at the command level whether I executed the command from the tmux session corresponding to my laptop or the tmux session corresponding to my desktop. Indeed, setting the `$GIT_AUTHOR_NAME` accomplishes this very easily for Git. 

OK. so this still works as well as it ever did -- that is to say, really well. 

But in recent git builds, I have been noticing something sinister. Now, when I use `choose-tree` or `switch-client` to switch from session A to session B, session B's environment is **completely overwritten** by session A's session environment. This happens whether A and B are grouped or not.

Now this isn't a really unforgivable behavior, and it does certainly seem to fall outside of specified behavior, as far as the man page is concerned. 

However, I do have to say that it is very useful to be able to temporarily switch to another session just to e.g. move it off to another window, in order to get screen space back, for instance. Now I do imagine that if I remain disciplined (maybe bind a key that does something tricksy to cause the other session to change its focused window, but without actually switching to it to blow away its environment) I can avoid the environment-nuking. But I think I have sufficiently described the use case at this point.

Thanks!

P.S. I have now also tested this without loading my (rather elaborate) `tmux.conf`, and also without running my (also elaborate) zsh shell with tmux by reverting my shell to bash. So, the behavior seems to be easily reproducible and this all but eliminates the possibility that other configuration is causing the behavior. Commit `f7598b8`, OS X 10.10.3.

nicm:
Please give instructions to reproduce this from a fresh server (started
with "tmux -Ltest -f/dev/null new").

unphased:
Okay, I guess even though I did a lot of testing I still did not do deep enough testing. 

[...]

PR #20, opened 2015-06-07
This is for issue #17

unphased:


ThomasAdam:
Even if this were OK, you need to update the man page anyway.

unphased:
Yeah I was figuring out how to do that :stuck_out_tongue_closed_eyes:

nicm:
No, we are not going to do this. If you don't want something to be updated, don't put it in update-environment...

unphased:
@nicm There is a real use case for this. Please see #17. I want them to update when attaching. But not when switching sessions from a client.

nicm:
I may accept a flag that bypasses update-environment entirely, but it will need to be added to all commands that use it. -d is a poor choice, prhaps -E.

unphased:
@nicm Very sensible! I'll have a look at the rest of the commands.

unphased:
I would go ahead and do this since it's trivial now that I understand the code (because it's so beautiful :joy:)...

But there's one thing -- `choose-tree.c` runs `switch-client`, so I'm unsure if this means I ought to also add a `-E` flag option to the `choose-tree` command expressly for the purpose of passing it onto `switch-client`.

nicm:
No, leave choose-tree as it is, the command can be specified by the user already.

unphased:
Well that oughta do it

nicm:
Applied with some fixes (flags must be in alphabetical order, new sentence new line in man page, etc).

PR #23, opened 2015-06-10
Fixed a potential resource leak.

orbitcowboy:


nicm:
Applied, thanks, although you also need to save and restore errno around close.

PR #30, opened 2015-06-17
Enable silent build by default

yegorich:
Signed-off-by: Yegor Yefremov <yegorslists@googlemail.com>


ThomasAdam:
Not going to happen.  See this:

http://thread.gmane.org/gmane.comp.terminal-emulators.tmux.user/7305

PR #38, opened 2015-07-02
Add additional window-status-*-format options

jeroenseegers:
Options added for custom formatting of window-status-activity,
window-status-bell and window-status-last.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -662,6 +662,11 @@ const struct options_table_entry window_options_table[] = {
 	  .style = "window-status-activity-style"
 	},
 
+	{ .name = "window-status-activity-format",
+	  .type = OPTIONS_TABLE_STRING,
+	  .default_str = "#I:#W#{?window_flags,#{window_flags}, }"
+	},
+
 	{ .name = "window-status-activity-style",
 	  .type = OPTIONS_TABLE_STYLE,
 	  .default_str = "reverse"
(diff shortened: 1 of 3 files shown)


ThomasAdam:
Possibly; not sure how useful this is, given that when we get hook supported, this can be done via that.

jeroenseegers:
How would this work via hook support? I could only find [this discussion](http://comments.gmane.org/gmane.comp.terminal-emulators.tmux.user/3462) on it and to me it's not very clear on how this would work through a hook.

When is support for hooks expected to drop?

nicm:
You should be able to do this already with #{?window_bell_flag,foo,bar}, I don't think we need extra options.

jeroenseegers:
With `#{?window_bell_flag,foo,bar}` you can only change a single property as far as I understand. Using these new options you could do something along the lines of
```shell
set-window-option -g window-status-last-format '#[bg=colour10]#[fg=colour0]>#[bg=colour10]#[fg=white] #I: #W#F #[default]#[fg=colour10]>#[default]'
```
which would look something like this
screen shot 2015-07-09 at 08 27 20

It allows for greater customisation of your native tmux installation without the bloat of something like powerline.

nicm:
New options may make this easier, but it is possible to do this without using #{?} so I don't think they are necessary.

jeroenseegers:
Could you provide an example on how to do the above with a `#{?}` command?

nicm:
Try something like:

set -g window-status-format '#{?window_bell_flag,#[bg=red]foo,#{?window_activity_flag,#[bg=yellow]bar,#[bg=blue]xyz}}'

jeroenseegers:
That seems to work indeed, the syntax for it seems to be a bit fuzzy though, but thanks!

Any reason in particular why `window-status-format` and `window-status-current-format` were included, but the ones in this PR weren't? Seems to be a lot easier and consistent to have these included too.

nicm:
They were added before we already had 80 other options and before we had #{?}.

PR #45, opened 2015-07-09
More precise color comparison

bacher09:
I implemented more precise color comparison using [CIE76 ΔE][CIE76E]. This metric is used to select more similar color.

To compare difference between tmux algorithm and CIE76 ΔE I made few [tests]. There're two groups of tests, one that can be [viewed in web-browser][web_test] and another that can be viewed in terminal that supports 24-bit colors. Terminal tests can be viewed by this command:

    $ curl http://xonotic.in.ua/slava/color/term_test4.txt

All tests have 3 columns: first - colors that will be returned by tmux's algorithm, second - original colors, that's why terminal tests require terminal that support 24-bit colors, and last - colors that will be nearest by ΔE metric.

Also, I made attempt to recalculate 256to16 color mapping table using ΔE comparison, but seems in most cases this produces worse results. These tests can be viewed in any terminal that supports 256 xterm colors via such command:

    $ curl http://xonotic.in.ua/slava/color/ansi16_test.txt

Actually, I'm not confident it's the right feature to be in tmux. Anyway, I'd like to share it with you as I've already done it.

[CIE76E]: https://en.wikipedia.org/wiki/Color_difference#CIE76
[tests]: http://xonotic.in.ua/slava/color/
[web_test]: http://xonotic.in.ua/slava/color/lab_d65_3.html


nicm:
Thanks. The code itself looks good but this will not go in at the moment. Firstly, it is a lot of extra code and a large table for a relatively esoteric feature, I think few users care about having more than 256 colours and fewer about the accuracy of their computation - and in any case they can tweak what they put in until they get the colour they want. Secondly, we currently match how xterm does the mapping and I'm happy to keep it that way. Still, perhaps we may want it in future so it is nice that you have done this, thanks.

PR #53, opened 2015-07-17
server_client_create: initialize FD to -1

bpowers:
Without this, if a client fails to call MSG_IDENTIFY_STDIN, on client
disconnect FD 0 will be closed, eventually leading to corruption.


ThomasAdam:
This is already in Git.

nicm:
It's not, I only added the c->cwd = -1 bit. I'm not sure this can actually happen but applied anyway, thanks.

PR #57, opened 2015-07-19
find: rename relative left/right to avoid ambiguity

mathstuf:
---
Feel free to use something more descriptive than `r` as well when applying (or I can respin).

### DIFF
--- a/cmd-find.c
+++ b/cmd-find.c
@@ -103,8 +103,8 @@ const char *cmd_find_pane_table[][2] = {
 	{ "{bottom-right}", "bottom-right" },
 	{ "{up}", "{up}" },
 	{ "{down}", "{down}" },
-	{ "{left}", "{left}" },
-	{ "{right}", "{right}" },
+	{ "{rleft}", "{left}" },
+	{ "{rright}", "{right}" },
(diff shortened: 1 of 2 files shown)


nicm:
I think this is more confusing, and we call them left/right/up/down in other places too.

mathstuf:
True, but there's also `{left}` meaning left-most pane. The one meaning left-of-focused never gets hit.

nicm:
Ah yes, that is a problem. I don't like rleft and rright though, maybe
the top/bottom/left/right/top-left/etc should be prefixed with window, so like
window-top etc?

mathstuf:
That can work, but seems a little wordy. What about `{left-of}` for the relative one (`{up}` and `{down}` are already obviously relative to me)?

nicm:
Yes *-of sounds ok but I think they should all be consistent.

mathstuf:
Renamed.

mathstuf:
Also, do you still prefer `format-patch` and `send-email`?

nicm:
Email is better or link me to the patch please.

mathstuf:
Patch sent to the list.

PR #84, opened 2015-08-16
Double- and triple-click selection of whole words and lines

osandov:
The lack of this feature makes it more convenient to bypass copy-mode with Shift-click in many cases, but it'd be awesome to have this natively in tmux. Here's my go at implementing it.

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -59,6 +59,8 @@ cmd_copy_mode_exec(struct cmd *self, struct cmd_q *cmdq)
 	struct client		*c = cmdq->client;
 	struct session		*s;
 	struct window_pane	*wp = cmdq->state.tflag.wp;
+	u_int			 b;
+	int			 scroll_exit;
 
 	if (args_has(args, 'M')) {
 		if ((wp = cmd_mouse_pane(&cmdq->item->mouse, &s, NULL)) == NULL)
@@ -72,18 +74,21 @@ cmd_copy_mode_exec(struct cmd *self, struct cmd_q *cmdq)
 		return (CMD_RETURN_NORMAL);
 	}
 
-	if (wp->mode != &window_copy_mode) {
-		if (window_pane_set_mode(wp, &window_copy_mode) != 0)
-			return (CMD_RETURN_NORMAL);
-		window_copy_init_from_pane(wp, args_has(self->args, 'e'));
-	}
+	scroll_exit = args_has(self->args, 'e');
 	if (args_has(args, 'M')) {
-		if (wp->mode != NULL && wp->mode != &window_copy_mode)
+		b = cmdq->item->mouse.b;
+		if (MOUSE_DRAG(b))
+			window_copy_start_drag(c, &cmdq->item->mouse, scroll_exit);
+		else if (!MOUSE_WHEEL(b) && !MOUSE_RELEASE(b))
+			window_copy_mouse_down(c, &cmdq->item->mouse, scroll_exit);
+		else
+			return (CMD_RETURN_ERROR);
+	} else {
+		if (window_enter_copy_mode(wp, scroll_exit) != 0)
 			return (CMD_RETURN_NORMAL);
-		window_copy_start_drag(c, &cmdq->item->mouse);
+		if (args_has(self->args, 'u'))
+			window_copy_pageup(wp, 0);
 	}
-	if (wp->mode == &window_copy_mode && args_has(self->args, 'u'))
-		window_copy_pageup(wp, 0);
 
 	return (CMD_RETURN_NORMAL);
 }
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -224,7 +224,7 @@ key_bindings_init(void)
 		"bind -r C-Down resize-pane -D",
 		"bind -r C-Left resize-pane -L",
 		"bind -r C-Right resize-pane -R",
-		"bind -n MouseDown1Pane select-pane -t=\\; send-keys -M",
+		"bind -n MouseDown1Pane select-pane -t=\\; if -Ft= '#{mouse_any_flag}' 'if -Ft= \"#{pane_in_mode}\" \"copy-mode -M\" \"send-keys -M\"' 'copy-mode -M'",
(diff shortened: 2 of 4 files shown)


nicm:
Thanks. This looks great but there is one problem:

When I double click, I don't want the command bound to Mouse1Down fired, I just want the Double click key fired, so there needs to be a way to achieve this.

I don't know what the best way is, if we wait for the timer to expire before figuring out what key to fire (Down, Up+Down, Double Click, Triple Click) then there will be a delay before anything happens. Maybe a delay would be acceptable, it's hard to tell without trying it. It works fine for Escape.

Perhaps a better option is to make Up/Down work as you have them (always fired as well as Double Click or Triple Click), but also add Single Click which is subject to the timer. So we fire Up/Down/etc as they happen, but also fire Single Click, Double Click, Triple Click if there are 1, 2 or 3 Down/Up pairs within 1 timer period for Single/Double or 2 timer periods (for Triple). So Down/Up is a direct mouse event, *Click is a count of Down/Up pairs using the timer.

I have some concerns about the actual copy mode bits but I think it would be better not to worry about that part until the mouse click behaviour is settled.

osandov:
Thanks for taking a look! I threw something together to see how it this would feel with a timeout and put it here https://github.com/osandov/tmux/tree/double-click-select2. That doesn't wire up any actual events or anything, it justs logs when a click event would in theory be fired.

My opinion is that it feels too unresponsive, and it'd feel especially weird since other applications seem to trigger an event for all three clicks. As an example, I looked at Vim. Try doing a split then triple-clicking a line in the split that doesn't have focus. The first click changes the focus, the second click selects a word, and the third click selects the whole line. The same goes for Chrome. I think it'd be most natural to follow the existing convention and fire all three events, what do you think? And maybe it'd still be a good idea to separate single-, double-, and triple-click events from mouse down and mouse up events regardless?

nicm:
Hmm. It is not very useful for users if they bind commands to single, double and triple click to have all three commands fired when they do a triple click.  I don't think it matters what vim does, it is not a useful parallel to tmux.

I think the best thing would be if we did make it consistent and added a single click as well also using the timer, and also change it so that the down event is fired for every mouse down - so for a single click I get a mouse down, a mouse up and a single click once the timer expires; for a double click I get mouse  down/mouse up/mouse down/mouse up and then a double click; and so on. That way if something is harmless when used with my triple click command (like select-pane) I can bind it to mouse down, if it is something that must be exclusive I can bind it to single click and put up with the timer. If I don't care about clicks, I can just unbind them and not have the timer interfere with my mouse down events.

The other thing you will need to do is remove the options, we do not need any new ones for this. For the timeout, pick a sensible default and add a #define KEYC_CLICK_TIMEOUT or something and for the separators please use the existing word-separators option.

I think your timeout of 300 is too long, 100 or 200 would be better.

nicm:
I realise that you want double click selects word then one more click selects line - but you should be able to do that anyway, once double click selects a word it should be possible to arrange it so that another click (or mouse down) toggles to select the line.

I'm thinking of being able to do things like:

bind -n DoubleClick1Pane select-pane -Z
bind -n TripleClick1Pane rotate-window

Where it is more useful (and more obvious to the user) to have double and triple click entirely separate.

osandov:
Okay, I see what you're getting at about generic double- and triple-click events, for those use cases what you describe is definitely more intuitive. I don't see a good way to resolve that, because I really want the instant behavior for selection. So, instead of  adding generic double- and triple-click events, we can just do this at the copy mode level: copy-mode -M will handle MouseDown and do the selection from there. How does that sound?

nicm:
Sure you could make it so that one MouseDown moves the cursor or enters copy mode, and the next on the cursor selects word, and the next selects line, and so on, so it is not a double or triple click or a timer but just toggling the selection. I'm not sure we could bind it by default though because it would interfere with selecting a pane, I want one click to select a pane not to enter copy mode.

I think it'd be better in the long run to have click/double-click/triple-click separate from mouse up/down as I mentioned before (so you get down/up AND click).

osandov:
Hi, sorry, just picking up on this again. Could you take a look at what I just pushed? I implemented what I described above. I used a timer because doing it any other way felt weird to use. The default select-pane behavior is not interfered with. Overall, I feel that this behavior feels the most natural, and it doesn't preclude adding generic double-/triple-click events in the future.

nicm:
Can you send me a diff or give me a URL please? This issue is still
pointing to the old patch.

Thanks

osandov:
It's the same branch as in the issue, I've just updated it (https://github.com/osandov/tmux/tree/double-click-select). If that doesn't work for you, I'm happy to send a patch instead.

nicm:
Please send a patch, thanks

nicm:
Never mind turns out I can do https://github.com/tmux/tmux/compare/master...osandov:double-click-select.patch

I'll look at it in a bit

Thanks

nicm:
This diff is good as far as it goes but it sucks that we have double/triple clicks without letting people bind them, I need to think about it.

nicm:
I think I would like this better if the timer was in window-copy.c and the flags in window_copy_mode_data. Can you do that instead?

osandov:
Hey, I picked this back up and rebased on top of master. There were quite a few conflicts but it should all be cleaned up now. Let me know if there's anything else I can do, I'd still love to see this get in.

nicm:
I'm busy at the moment but I'll try to look at it again at some point.

osandov:
Rebased again. I can send a patch to the mailing list instead if that's more convenient.

nicm:
Hi. I don't like making this all part of copy mode, I think it is much better to have generic DoubleClick and TripleClick and bind them to select-word and select-line commands, even if it needs a few extra clicks to swap between different selections. I think I will go with this: [tmux-clicks.diff.txt](https://github.com/tmux/tmux/files/520002/tmux-clicks.diff.txt)

nicm:
Soon you will be able to bind full commands as keys in copy mode, so it will be possible to make these more sophisticated (check current selection perhaps).

nicm:
Applied this now.

PR #95, opened 2015-08-28
Fix session->name memory leak

kcwu:


nicm:
Fixed in OpenBSD, will be in Git soon, thanks.

Issue #34, opened 2015-06-28
Old Issues From SourceForge?


jfelchner:
There was a very long thread about an issue that I was interested in on SourceForge.  I'd be happy to convert it over, but it seems as though the link is no longer valid.

Does one of the maintainers still have access to the old SourceForge issues?

jfelchner:
It was specifically issue number 140 which was the issue about truecolor support in Tmux.

The old URL was: http://sourceforge.net/p/tmux/tickets/140/

jfelchner:
@ThomasAdam what is going on?

jfelchner:
@ThomasAdam why did you close this?

ThomasAdam:
Hmm.  There was supposed to be a comment from me.  Not sure what happened there.

We had a patch for this in the past.  I'm unsure how useful 24bit colour support is, it's certainly not something happening by default, and only a handful of applications support this.  Until such time it's more ubiquitous, this isn't worth it.

jfelchner:
@ThomasAdam that's everything that was in that ticket.  Evidence that it is currently _very_ ubiquitous.  Almost every major terminal currently supports it.  That ticket had links to tickets on all of them outlining their support for it.

jfelchner:
I'd rather not retype all of that if someone can get access to the old ticket.

ThomasAdam:
You mean this:  https://www.mail-archive.com/tmux-users@lists.sourceforge.net/msg06296.html

jfelchner:
That's the first part to it yes. :grinning: thanks!

ThomasAdam:
There never was a follow-up.  But it shows that libvte supports it, and almost no applications.  What XTerm does, is what tmux does now with 256 colour support, in terms of choosing the nearest colour match.

jfelchner:
@ThomasAdam it may have been a different issue then that had all the followups.  There were a lot including some from Nicholas.  Is there a way for me to search the archives?

jfelchner:
Here's what iTerm does:

![image](https://cloud.githubusercontent.com/assets/285582/8394546/76a1556e-1d02-11e5-8cc7-3c59a54a1d29.png)

Here's what Tmux does:

![image](https://cloud.githubusercontent.com/assets/285582/8394548/8512cfce-1d02-11e5-9bdd-a3cf9b8bcf49.png)

ThomasAdam:
[...]

PR #112, opened 2015-09-17
add support for 24-bit RGB color escape sequences

sunaku:
This patch adds support for 24-bit RGB color escape sequences, requested
in issue #34, by adjusting Koichi Shiraishi's patch [2], which derives
from Christian Hopps's patch [3], which in turn derives from Arnis
Lapsa's patch [4], according to feedback given by Nicholas Marriott [1].

[1]: https://github.com/tmux/tmux/issues/34#issuecomment-132557915
[2]: https://gist.github.com/zchee/9f6f2ca17acf49e04088
[3]: https://gist.github.com/choppsv1/dd00858d4f7f356ce2cf
[4]: https://gist.github.com/ArnisL/6156593

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -37,7 +37,7 @@
 
 /* Default grid cell data. */
 const struct grid_cell grid_default_cell = {
-	0, 0, 8, 8, { { ' ' }, 0, 1, 1 }
+	0, 0, { .fg = 8 }, { .bg = 8 }, { { ' ' }, 0, 1, 1 }
 };
 const struct grid_cell_entry grid_default_entry = {
 	0, { .data = { 0, 8, 8, ' ' } }
@@ -284,6 +284,7 @@ grid_set_cell(struct grid *gd, u_int px, u_int py, const struct grid_cell *gc)
 	struct grid_line	*gl;
 	struct grid_cell_entry	*gce;
 	struct grid_cell 	*gcp;
+	int			 extended;
 
 	if (grid_check_y(gd, py) != 0)
 		return;
@@ -293,8 +294,12 @@ grid_set_cell(struct grid *gd, u_int px, u_int py, const struct grid_cell *gc)
 	gl = &gd->linedata[py];
 	gce = &gl->celldata[px];
 
-	if ((gce->flags & GRID_FLAG_EXTENDED) || gc->data.size != 1 ||
-	    gc->data.width != 1) {
+	extended = (gce->flags & GRID_FLAG_EXTENDED);
+	if (!extended && (gc->data.size != 1 || gc->data.width != 1))
+		extended = 1;
+	if (!extended && (gc->flags & (GRID_FLAG_FGRGB|GRID_FLAG_BGRGB)))
+		extended = 1;
+	if (extended) {
 		if (~gce->flags & GRID_FLAG_EXTENDED) {
 			gl->extddata = xreallocarray(gl->extddata,
 			    gl->extdsize + 1, sizeof *gl->extddata);
--- a/input.c
+++ b/input.c
@@ -1629,18 +1629,20 @@ input_csi_dispatch_sgr_256(struct input_ctx *ictx, int fgbg, u_int *i)
 	c = input_get(ictx, *i, 0, -1);
 	if (c == -1) {
 		if (fgbg == 38) {
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = 8;
 		} else if (fgbg == 48) {
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = 8;
 		}
 	} else {
 		if (fgbg == 38) {
 			gc->flags |= GRID_FLAG_FG256;
+			gc->flags &= ~GRID_FLAG_FGRGB;
 			gc->fg = c;
 		} else if (fgbg == 48) {
 			gc->flags |= GRID_FLAG_BG256;
+			gc->flags &= ~GRID_FLAG_BGRGB;
 			gc->bg = c;
 		}
 	}
@@ -1651,7 +1653,7 @@ void
 input_csi_dispatch_sgr_rgb(struct input_ctx *ictx, int fgbg, u_int *i)
 {
 	struct grid_cell	*gc = &ictx->cell.cell;
-	int			 c, r, g, b;
+	int			 r, g, b;
 
 	(*i)++;
 	r = input_get(ictx, *i, 0, -1);
@@ -1666,13 +1668,18 @@ input_csi_dispatch_sgr_rgb(struct input_ctx *ictx, int fgbg, u_int *i)
 	if (b == -1 || b > 255)
 		return;
 
-	c = colour_find_rgb(r, g, b);
 	if (fgbg == 38) {
-		gc->flags |= GRID_FLAG_FG256;
-		gc->fg = c;
+		gc->flags &= ~GRID_FLAG_FG256;
+		gc->flags |= GRID_FLAG_FGRGB;
+		gc->fg_rgb.r = r;
+		gc->fg_rgb.g = g;
+		gc->fg_rgb.b = b;
 	} else if (fgbg == 48) {
-		gc->flags |= GRID_FLAG_BG256;
-		gc->bg = c;
+		gc->flags &= ~GRID_FLAG_BG256;
+		gc->flags |= GRID_FLAG_BGRGB;
+		gc->bg_rgb.r = r;
+		gc->bg_rgb.g = g;
+		gc->bg_rgb.b = b;
 	}
 }
 
@@ -1754,11 +1761,11 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 35:
 		case 36:
 		case 37:
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = n - 30;
 			break;
 		case 39:
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = 8;
 			break;
 		case 40:
@@ -1769,11 +1776,11 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 45:
 		case 46:
 		case 47:
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = n - 40;
 			break;
 		case 49:
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = 8;
 			break;
 		case 90:
@@ -1784,7 +1791,7 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 95:
 		case 96:
 		case 97:
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = n;
 			break;
 		case 100:
@@ -1795,7 +1802,7 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 105:
 		case 106:
 		case 107:
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = n - 10;
 			break;
 		}
--- a/tmux.1
+++ b/tmux.1
@@ -2925,7 +2925,7 @@ and poor for interactive programs such as shells.
 .Op Ic on | off
 .Xc
 Allow programs to change the window name using a terminal escape
-sequence (\\033k...\\033\\\\).
+sequence (\eek...\ee\e\e).
 The default is on.
 .Pp
 .It Xo Ic alternate-screen
@@ -4029,7 +4029,7 @@ This command only works from outside
 .El
 .Sh TERMINFO EXTENSIONS
 .Nm
-understands some extensions to
+understands some unofficial extensions to
 .Xr terminfo 5 :
 .Bl -tag -width Ds
 .It Em Cs , Cr
@@ -4053,10 +4053,12 @@ $ printf '\e033[4 q'
 If
 .Em Se
 is not set, \&Ss with argument 0 will be used to reset the cursor style instead.
+.It Em \&Tc
+Indicate that the terminal supports the
+.Ql direct colour
+RGB escape sequence (for example, \ee[38;2;255;255;255m).
 .It Em \&Ms
-This sequence can be used by
-.Nm
-to store the current buffer in the host terminal's selection (clipboard).
+Store the current buffer in the host terminal's selection (clipboard).
 See the
 .Em set-clipboard
 option above and the
--- a/tmux.h
+++ b/tmux.h
@@ -387,6 +387,7 @@ enum tty_code_code {
 	TTYC_SMSO,	/* enter_standout_mode, so */
 	TTYC_SMUL,	/* enter_underline_mode, us */
 	TTYC_SS,	/* set cursor style, Ss */
+	TTYC_TC,	/* 24-bit "true" colour, Tc */
 	TTYC_TSL,	/* to_status_line, tsl */
 	TTYC_VPA,	/* row_address, cv */
 	TTYC_XENL,	/* eat_newline_glitch, xn */
@@ -643,16 +644,31 @@ enum utf8_state {
 #define GRID_FLAG_BG256 0x2
 #define GRID_FLAG_PADDING 0x4
 #define GRID_FLAG_EXTENDED 0x8
+#define GRID_FLAG_FGRGB 0x10
+#define GRID_FLAG_BGRGB 0x20
 
 /* Grid line flags. */
 #define GRID_LINE_WRAPPED 0x1
+
+/* Grid cell RGB colours. */
+struct grid_cell_rgb {
+	u_char	r;
+	u_char	g;
+	u_char	b;
+};
 
 /* Grid cell data. */
 struct grid_cell {
 	u_char			flags;
 	u_char			attr;
-	u_char			fg;
-	u_char			bg;
+	union {
+		u_char		fg;
+		struct grid_cell_rgb	fg_rgb;
+	};
+	union {
+		u_char		bg;
+		struct grid_cell_rgb	bg_rgb;
+	};
 	struct utf8_data	data;
 
 };
--- a/tty-term.c
+++ b/tty-term.c
@@ -254,6 +254,7 @@ const struct tty_term_code_entry tty_term_codes[] = {
 	[TTYC_SMSO] = { TTYCODE_STRING, "smso" },
 	[TTYC_SMUL] = { TTYCODE_STRING, "smul" },
 	[TTYC_SS] = { TTYCODE_STRING, "Ss" },
+	[TTYC_TC] = { TTYCODE_FLAG, "Tc" },
 	[TTYC_TSL] = { TTYCODE_STRING, "tsl" },
 	[TTYC_VPA] = { TTYCODE_STRING, "vpa" },
 	[TTYC_XENL] = { TTYCODE_FLAG, "xenl" },
--- a/tty.c
+++ b/tty.c
@@ -36,8 +36,15 @@ static int tty_log_fd = -1;
 void	tty_read_callback(struct bufferevent *, void *);
 void	tty_error_callback(struct bufferevent *, short, void *);
 
+static int tty_same_fg(const struct grid_cell *, const struct grid_cell *);
+static int tty_same_bg(const struct grid_cell *, const struct grid_cell *);
+static int tty_same_colours(const struct grid_cell *, const struct grid_cell *);
+static int tty_is_fg(const struct grid_cell *, int);
+static int tty_is_bg(const struct grid_cell *, int);
+
 void	tty_set_italics(struct tty *);
 int	tty_try_256(struct tty *, u_char, const char *);
+int	tty_try_rgb(struct tty *, const struct grid_cell_rgb *, const char *);
 
 void	tty_colours(struct tty *, const struct grid_cell *);
 void	tty_check_fg(struct tty *, struct grid_cell *);
@@ -60,6 +67,74 @@ void	tty_default_colours(struct grid_cell *, const struct window_pane *);
 
 #define tty_pane_full_width(tty, ctx) \
 	((ctx)->xoff == 0 && screen_size_x((ctx)->wp->screen) >= (tty)->sx)
+
+static int
+tty_same_fg(const struct grid_cell *gc1, const struct grid_cell *gc2)
+{
+	int	flags1, flags2;
+
+	flags1 = (gc1->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
+	flags2 = (gc2->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
+
+	if (flags1 != flags2)
+	    return (0);
+
+	if (flags1 & GRID_FLAG_FGRGB) {
+		if (gc1->fg_rgb.r != gc2->fg_rgb.r)
+			return (0);
+		if (gc1->fg_rgb.g != gc2->fg_rgb.g)
+			return (0);
+		if (gc1->fg_rgb.b != gc2->fg_rgb.b)
+			return (0);
+		return (1);
+	}
+	return (gc1->fg == gc2->fg);
+}
+
+static int
+tty_same_bg(const struct grid_cell *gc1, const struct grid_cell *gc2)
+{
+	int	flags1, flags2;
+
+	flags1 = (gc1->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
+	flags2 = (gc2->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
+
+	if (flags1 != flags2)
+	    return (0);
+
+	if (flags1 & GRID_FLAG_BGRGB) {
+		if (gc1->bg_rgb.r != gc2->bg_rgb.r)
+			return (0);
+		if (gc1->bg_rgb.g != gc2->bg_rgb.g)
+			return (0);
+		if (gc1->bg_rgb.b != gc2->bg_rgb.b)
+			return (0);
+		return (1);
+	}
+	return (gc1->bg == gc2->bg);
+}
+
+static int
+tty_same_colours(const struct grid_cell *gc1, const struct grid_cell *gc2)
+{
+	return (tty_same_fg(gc1, gc2) && tty_same_bg(gc1, gc2));
+}
+
+static int
+tty_is_fg(const struct grid_cell *gc, int c)
+{
+	if (gc->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB))
+		return (0);
+	return (gc->fg == c);
+}
+
+static int
+tty_is_bg(const struct grid_cell *gc, int c)
+{
+	if (gc->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB))
+		return (0);
+	return (gc->bg == c);
+}
 
 void
 tty_create_log(void)
@@ -1423,12 +1498,10 @@ void
 tty_colours(struct tty *tty, const struct grid_cell *gc)
 {
 	struct grid_cell	*tc = &tty->cell;
-	u_char			 fg = gc->fg, bg = gc->bg, flags = gc->flags;
 	int			 have_ax, fg_default, bg_default;
 
 	/* No changes? Nothing is necessary. */
-	if (fg == tc->fg && bg == tc->bg &&
-	    ((flags ^ tc->flags) & (GRID_FLAG_FG256|GRID_FLAG_BG256)) == 0)
+	if (tty_same_colours(gc, tc))
 		return;
 
 	/*
@@ -1437,8 +1510,8 @@ tty_colours(struct tty *tty, const struct grid_cell *gc)
 	 * case if only one is default need to fall onward to set the other
 	 * colour.
 	 */
-	fg_default = (fg == 8 && !(flags & GRID_FLAG_FG256));
-	bg_default = (bg == 8 && !(flags & GRID_FLAG_BG256));
+	fg_default = tty_is_fg(gc, 8);
+	bg_default = tty_is_bg(gc, 8);
 	if (fg_default || bg_default) {
 		/*
 		 * If don't have AX but do have op, send sgr0 (op can't
@@ -1451,48 +1524,52 @@ tty_colours(struct tty *tty, const struct grid_cell *gc)
 		if (!have_ax && tty_term_has(tty->term, TTYC_OP))
 			tty_reset(tty);
 		else {
-			if (fg_default &&
-			    (tc->fg != 8 || tc->flags & GRID_FLAG_FG256)) {
+			if (fg_default && !tty_is_fg(tc, 8)) {
 				if (have_ax)
 					tty_puts(tty, "\033[39m");
-				else if (tc->fg != 7 ||
-				    tc->flags & GRID_FLAG_FG256)
+				else if (!tty_is_fg(tc, 7))
 					tty_putcode1(tty, TTYC_SETAF, 7);
 				tc->fg = 8;
-				tc->flags &= ~GRID_FLAG_FG256;
+				tc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			}
-			if (bg_default &&
-			    (tc->bg != 8 || tc->flags & GRID_FLAG_BG256)) {
+			if (bg_default && !tty_is_bg(tc, 8)) {
 				if (have_ax)
 					tty_puts(tty, "\033[49m");
-				else if (tc->bg != 0 ||
-				    tc->flags & GRID_FLAG_BG256)
+				else if (!tty_is_bg(tc, 0))
 					tty_putcode1(tty, TTYC_SETAB, 0);
 				tc->bg = 8;
-				tc->flags &= ~GRID_FLAG_BG256;
+				tc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			}
 		}
 	}
 
 	/* Set the foreground colour. */
-	if (!fg_default && (fg != tc->fg ||
-	    ((flags & GRID_FLAG_FG256) != (tc->flags & GRID_FLAG_FG256))))
+	if (!fg_default && !tty_same_fg(gc, tc))
 		tty_colours_fg(tty, gc);
 
 	/*
 	 * Set the background colour. This must come after the foreground as
 	 * tty_colour_fg() can call tty_reset().
 	 */
-	if (!bg_default && (bg != tc->bg ||
-	    ((flags & GRID_FLAG_BG256) != (tc->flags & GRID_FLAG_BG256))))
+	if (!bg_default && !tty_same_bg(gc, tc))
 		tty_colours_bg(tty, gc);
 }
 
 void
 tty_check_fg(struct tty *tty, struct grid_cell *gc)
 {
-	u_int	colours;
+	struct grid_cell_rgb	*rgb = &gc->fg_rgb;
+	u_int			 colours;
 
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_FGRGB) {
+		/* Not a 24-bit terminal? Translate to 256-colour palette. */
+		if (!tty_term_flag(tty->term, TTYC_TC)) {
+			gc->flags &= ~GRID_FLAG_FGRGB;
+			gc->flags |= GRID_FLAG_FG256;
+			gc->fg = colour_find_rgb(rgb->r, rgb->g, rgb->b);
+		}
+	}
 	colours = tty_term_number(tty->term, TTYC_COLORS);
 
 	/* Is this a 256-colour colour? */
@@ -1524,8 +1601,18 @@ tty_check_fg(struct tty *tty, struct grid_cell *gc)
 void
 tty_check_bg(struct tty *tty, struct grid_cell *gc)
 {
-	u_int	colours;
+	struct grid_cell_rgb	*rgb = &gc->bg_rgb;
+	u_int			 colours;
 
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_BGRGB) {
+		/* Not a 24-bit terminal? Translate to 256-colour palette. */
+		if (!tty_term_flag(tty->term, TTYC_TC)) {
+			gc->flags &= ~GRID_FLAG_BGRGB;
+			gc->flags |= GRID_FLAG_BG256;
+			gc->bg = colour_find_rgb(rgb->r, rgb->g, rgb->b);
+		}
+	}
 	colours = tty_term_number(tty->term, TTYC_COLORS);
 
 	/* Is this a 256-colour colour? */
@@ -1560,12 +1647,21 @@ tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
 	u_char			 fg = gc->fg;
 	char			 s[32];
 
+	tc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
+
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_FGRGB) {
+		if (tty_try_rgb(tty, &gc->fg_rgb, "38") == 0)
+			goto save_fg;
+		/* Should not get here, already converted in tty_check_fg. */
+		return;
+	}
+
 	/* Is this a 256-colour colour? */
 	if (gc->flags & GRID_FLAG_FG256) {
-		/* Try as 256 colours. */
 		if (tty_try_256(tty, fg, "38") == 0)
 			goto save_fg;
-		/* Else already handled by tty_check_fg. */
+		/* Should not get here, already converted in tty_check_fg. */
 		return;
 	}
 
@@ -1581,9 +1677,12 @@ tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
 
 save_fg:
 	/* Save the new values in the terminal current cell. */
-	tc->fg = fg;
-	tc->flags &= ~GRID_FLAG_FG256;
-	tc->flags |= gc->flags & GRID_FLAG_FG256;
+	if (gc->flags & GRID_FLAG_FGRGB)
+		memcpy(&tc->fg_rgb, &gc->fg_rgb, sizeof tc->fg_rgb);
+	else
+		tc->fg = fg;
+	tc->flags &= ~(GRID_FLAG_FGRGB|GRID_FLAG_FG256);
+	tc->flags |= (gc->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
 }
 
 void
@@ -1593,12 +1692,19 @@ tty_colours_bg(struct tty *tty, const struct grid_cell *gc)
 	u_char			 bg = gc->bg;
 	char			 s[32];
 
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_BGRGB) {
+		if (tty_try_rgb(tty, &gc->bg_rgb, "48") == 0)
+			goto save_bg;
+		/* Should not get here, already converted in tty_check_bg. */
+		return;
+	}
+
 	/* Is this a 256-colour colour? */
 	if (gc->flags & GRID_FLAG_BG256) {
-		/* Try as 256 colours. */
 		if (tty_try_256(tty, bg, "48") == 0)
 			goto save_bg;
-		/* Else already handled by tty_check_bg. */
+		/* Should not get here, already converted in tty_check_bg. */
 		return;
 	}
 
@@ -1614,9 +1720,12 @@ tty_colours_bg(struct tty *tty, const struct grid_cell *gc)
 
 save_bg:
 	/* Save the new values in the terminal current cell. */
-	tc->bg = bg;
-	tc->flags &= ~GRID_FLAG_BG256;
-	tc->flags |= gc->flags & GRID_FLAG_BG256;
+	if (gc->flags & GRID_FLAG_BGRGB)
+		memcpy(&tc->bg_rgb, &gc->bg_rgb, sizeof tc->bg_rgb);
+	else
+		tc->bg = bg;
+	tc->flags &= ~(GRID_FLAG_BGRGB|GRID_FLAG_BG256);
+	tc->flags |= (gc->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
 }
 
 int
@@ -1652,6 +1761,20 @@ tty_try_256(struct tty *tty, u_char colour, const char *type)
 
 fallback:
 	xsnprintf(s, sizeof s, "\033[%s;5;%hhum", type, colour);
+	tty_puts(tty, s);
+	return (0);
+}
+
+int
+tty_try_rgb(struct tty *tty, const struct grid_cell_rgb *rgb, const char *type)
+{
+	char	s[32];
+
+	if (!tty_term_flag(tty->term, TTYC_TC))
+		return (-1);
+
+	xsnprintf(s, sizeof s, "\033[%s;2;%hhu;%hhu;%hhum", type, rgb->r,
+	    rgb->g, rgb->b);
 	tty_puts(tty, s);
 	return (0);
 }


nicm:
This is something at least, although you didn't read my previous
comments :-). Here are some comments:

- I do not want to support : as a separator for all escape sequences, or
  in fact for any, including the RGB sequence. So that part of the diff
  can be removed.

- This is adding too many bytes to grid_cell. You do not need to use the
  entire colour_rgb struct or to call colour_find_rgb, you should reuse
  the existing fg and bg either by using them for one component manually
  or by doing something like:

  struct grid_cell {
        u_char  attr;
	u_char  flags;
	union {
		u_char		fg;
		struct rgb_fg {
			u_char	r;
			u_char	g;
			u_char	b;
		};
	};
	union {
		u_char		bg;
		struct rgb_fg {
			u_char	r;
			u_char	g;
			u_char	b;
		};
	};
	...

- Use memcpy not struct assignment please (so memcpy instead of
  gc->bg_rgb = rgb - although this will go away when you fix the last
  point).

- Don't send diffs with FIXME comments please... you will need make tmux
  look for a Tc terminfo entry (that can initially be added with
  terminal-overrides). Please don't add a command line flag like -2.

- Likewise what is the code commented out in input_csi_dispatch_sgr_rgb
  for? You probably do need to reset the 256 colour flags there.

- You forgot about grid_string_cells.

- Ultimately I think this will need to replace 256 colours but I am
  happy to see that done after this diff is good.

dequis:
Thanks @sunaku for picking this up! Also thanks @nicm for the helpful review comments!

>you will need make tmux look for a Tc terminfo entry (that can initially be added with terminal-overrides).

Interesting idea, would that mean bypassing the problem of the terminfo maintainer? Should other applications check for that Tc entry too?

nicm:
We will need some way to detect it, and there is no point in proposing a flag to Thomas Dickey that we might need, or might not, and we aren't even using. Once we have code and know that adding a flag is the best solution for tmux, we can see what he thinks - perhaps he will suggest something else, or perhaps we will have to keep it as an extension, or do something else. There is no problem yet.

sunaku:
On Thu, 17 Sep 2015 14:20:25 -0700, Nicholas Marriott wrote:
> This is something at least, although you didn't read my previous
> comments :-).   

Sorry about that!  This patch was indeed a work in progress; I was
mainly focused on reducing unnecessary elements from the prior one.

> Here are some comments:
> 
> - I do not want to support : as a separator for all escape sequences, or
>   in fact for any, including the RGB sequence. So that part of the diff
>   can be removed.  

Agreed, will do.

> - This is adding too many bytes to grid_cell. You do not need to use the
>   entire colour_rgb struct or to call colour_find_rgb, you should reuse
>   the existing fg and bg either by using them for one component manually
>   or by doing something like:
> 
>   struct grid_cell {
>         u_char  attr;
> 	u_char  flags;
> 	union {
> 		u_char		fg;
> 		struct rgb_fg {
> 			u_char	r;
> 			u_char	g;
> 			u_char	b;
> 		};
> 	};
> 	union {
> 		u_char		bg;
> 		struct rgb_fg {
> 			u_char	r;
> 			u_char	g;
> 			u_char	b;
> 		};
> 	};
> 	...
> 
> - Use memcpy not struct assignment please (so memcpy instead of
>   gc->bg_rgb = rgb - although this will go away when you fix the last
>   point).  

How about storing colors in 32-bit unsigned integers as 0xRRGGBBII?

The upper bytes will hold the red, green, and blue channel values.

The lowest byte will hold either (1) the 256-color index or (2) the
native 16-color, 8-color, or monochrome color value.  To distinguish
these two use-cases, we'll decipher the upper byte values as follows:

* All zero upper bytes (0x000000II) indicate II is a native color.
* All 0xFF upper bytes (0xFFFFFFII) indicate II is a 256-color index.
* Non-zero upper bytes (0x------II) indicate II is nearest 256-color.

It seems that this scheme allows us to forgo storing the grid_cell
flags as well because colors now provide both RGB and indexed values, 
so we can decide (at emission time) which to use based on terminal
capabilities (as is currently done with the aid of grid_cell flags).

In addition, this scheme will eliminate struct copying and simplify
RGB color equality checking (which is done channel-by-channel now).

> - Don't send diffs with FIXME comments please... you will need make tmux
>   look for a Tc terminfo entry (that can initially be added with
>   terminal-overrides). Please don't add a command line flag like -2.  

Thanks, I was searching the web for the name of an existing 24-bit
terminfo entry but only found that no such thing officially exists!

> - Likewise what is the code commented out in input_csi_dispatch_sgr_rgb
>   for? You probably do need to reset the 256 colour flags there.  

Since we first try emitting RGB and fall back to 256-color anyway,
do RGB and 256-color flags really need to be mutually exclusive?

> - You forgot about grid_string_cells.  

Thanks, I'll look into it.

> - Ultimately I think this will need to replace 256 colours but I am
>   happy to see that done after this diff is good.  

Thanks for your feedback and patience.  I will push additional
commits to this pull request and, once the net result is
satisfactory, I'll squash them and force-push a single final patch.

nicm:
I don't think it will be better to use 32 bits for the colour. It is a
nice idea, but the main problem with RGB colour is that it adds 4 bytes
to struct grid_cell (so 30% more memory use) for something very few
people will use, and most of those who do will use it for the minority
of cells. Adding 6 bytes instead is less helpful. If you can put the
flags in there too it'd be fine, but then you will have bit shifting and
masks everywhere so it'll be just as bad.

This is a real concern - for example I have 6 MB of history here with
relatively few windows. I am not particularly interested in Internet
hipsters running tmux on the VAX or some other truly ancient, dead
thing, but there are newer (like SPARC) and even modern platforms (like
ARM) that often have limited RAM too.

I did plan to have a separate array of long cells where if the character
was more than one byte, the cell would be a offset into the array
instead, that way cells could mostly be four or six bytes with a few
larger ones being an offset instead of the cell data. That's why there
are accessors for the character data in grid-cell.c. But I didn't get
round to doing it. If we had that, we could use it for cells with RBG
colours too, although there would need to be accessors for getting
colours and attributes too (at the moment there are only accessors for
character data).

As far as 256 and RGB being mutually exclusive - you should never have a
cell with both flags set, there can only be one colour at a time.

zchee:
@sunaku Thanks for rework my patch.
@nicm Thanks the helpful review comments.

I'm not good at C language, I will watch for a while this PR.

jfelchner:
@sunaku and you _so much_ for taking the time to help make this happen. There are quite a few people who are waiting with baited breath for this to land. :grinning: 

@nicm your comments are awesome. Thanks for taking such time to help point @sunaku to something that you'll be willing to merge. :ship:

sunaku:
On Sun, 20 Sep 2015 01:36:27 -0700, Nicholas Marriott wrote:
> I don't think it will be better to use 32 bits for the colour. It
> is a nice idea, but the main problem with RGB colour is that it
> adds 4 bytes to struct grid_cell (so 30% more memory use) for
> something very few people will use, and most of those who do will
> use it for the minority of cells. Adding 6 bytes instead is less
> helpful. If you can put the flags in there too it'd be fine, but
> then you will have bit shifting and masks everywhere so it'll be
> just as bad.
> 
> This is a real concern - for example I have 6 MB of history here
> with relatively few windows. I am not particularly interested in
> Internet hipsters running tmux on the VAX or some other truly
> ancient, dead thing, but there are newer (like SPARC) and even
> modern platforms (like ARM) that often have limited RAM too.

That makes sense; agreed.

> I did plan to have a separate array of long cells where if the
> character was more than one byte, the cell would be a offset into
> the array instead, that way cells could mostly be four or six
> bytes with a few larger ones being an offset instead of the cell
> data. That's why there are accessors for the character data in
> grid-cell.c. But I didn't get round to doing it. If we had that,
> we could use it for cells with RBG colours too, although there
> would need to be accessors for getting colours and attributes too
> (at the moment there are only accessors for character data).

Would you like me to pursue this route?  Or perhaps you might
refactor toward it later, after this patch has been reworked?

> As far as 256 and RGB being mutually exclusive - you should never
> have a cell with both flags set, there can only be one colour at a
> time.

In that case, what should happen when a 24-bit color is specified
for a grid_cell, but the current TTY lacks 24-bit color capability?

Do we dynamically degrade the 24-bit color down to the next highest
level of color fidelity (256-color, 88-color, 16-color, 8-color,
2-color, 1-color) supported by the current TTY at emission time?

nicm:
On Tue, Sep 22, 2015 at 10:20:33PM -0700, Suraj N. Kurapati wrote:
>    > I did plan to have a separate array of long cells where if the
>    > character was more than one byte, the cell would be a offset into
>    > the array instead, that way cells could mostly be four or six
>    > bytes with a few larger ones being an offset instead of the cell
>    > data. That's why there are accessors for the character data in
>    > grid-cell.c. But I didn't get round to doing it. If we had that,
>    > we could use it for cells with RBG colours too, although there
>    > would need to be accessors for getting colours and attributes too
>    > (at the moment there are only accessors for character data).
> 
>    Would you like me to pursue this route? Or perhaps you might
>    refactor toward it later, after this patch has been reworked?

I suggest doing it later but for now keep grid_cell as small as
practical.

> 
>    > As far as 256 and RGB being mutually exclusive - you should never
>    > have a cell with both flags set, there can only be one colour at a
>    > time.
> 
>    In that case, what should happen when a 24-bit color is specified
>    for a grid_cell, but the current TTY lacks 24-bit color capability?
> 
>    Do we dynamically degrade the 24-bit color down to the next highest
>    level of color fidelity (256-color, 88-color, 16-color, 8-color,
>    2-color, 1-color) supported by the current TTY at emission time?

Yes, except only 256 and 16/8 colours. You probably just need to add
something using colour_find_rgb in tty_check_fg and tty_check_bg.

sunaku:
On Wed, 23 Sep 2015 00:44:58 -0700, Nicholas Marriott wrote:
> On Tue, Sep 22, 2015 at 10:20:33PM -0700, Suraj N. Kurapati wrote:
> >    Do we dynamically degrade the 24-bit color down to the next highest
> >    level of color fidelity (256-color, 88-color, 16-color, 8-color,
> >    2-color, 1-color) supported by the current TTY at emission time?
> 
> Yes, except only 256 and 16/8 colours. You probably just need to add
> something using colour_find_rgb in tty_check_fg and tty_check_bg.

Hmm, in my review of tty.c, that doesn't exactly seem to be the case: 
tty_colours() and its children tty_colours_fg() and tty_colours_bg()
modify the colour values in `gc`, the grid_cell, itself rather than
`tc`, the current terminal grid_cell, during graceful degradation.

As a result, we may lose colour information when, say, specifying a
24-bit colour whilst connected to a non-true-colour terminal and
then reattaching tmux to a true-colour terminal.  The original
24-bit colour would have already been degraded down to a 256-colour
or 16-colour representation before the latter terminal sees it.

Is this correct?

P.S. I have force-pushed an updated patch implementing `Tc` terminfo
lookup and tty_check_[fb]g() consideration, but colour_fromstring()
is still pending.  Please have a (cursory) look to keep me on track
with your vision of the final patch.  Thanks for your consideration.

nicm:
No: tty_check_fg, tty_check_bg and tty_colours all operate on a
grid_cell on the stack. Look at tty_attributes. Please link me to an
updated patch and I will take a look.

jfelchner:
@nicm on github when you force push a PR the existing PR is updated with the new overwritten commits.

sunaku:
On Sat, 26 Sep 2015 16:25:12 -0700, Nicholas Marriott wrote:
> tty_check_fg, tty_check_bg and tty_colours all operate on a
> grid_cell on the stack. Look at tty_attributes. 

Thanks for the hint!  Will do.

> Please link me to an updated patch and I will take a look.

Sure, here is the colourful web version: 
https://github.com/tmux/tmux/pull/112/files

Or if you like, the traditional version:
https://patch-diff.githubusercontent.com/raw/tmux/tmux/pull/112.patch

These links always reflect the current state of the patch / PR.

nicm:
Hi

Thanks. Some comments:

You don't need the changes to the input_transition structs.

I think where you have:

  	      gc->flags &= ~GRID_FLAG_BG256;
	      gc->flags &= ~GRID_FLAG_BGRGB;

It would be better just as gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB).

I don't like defining the struct inside the union, define it outside
then just use it inside. Also grid_cell_colour_rgb is too long, maybe
just grid_cell_rgb?

I think the big check at the start of tty_colours could be broken up to
be more readable, maybe using a temporary flag, something like (not
tested):

changed = 0;
if (flags ^ tc->flags) & (GRID_FLAG_FG256|GRID_FLAG_BG256)) != 0)
	changed = 1;
if (flags ^ tc->flags) & (GRID_FLAG_FGRGB|GRID_FLAG_BGRGB)) != 0)
	changed = 1;
if (!changed && (flags & GRID_FLAG_FG256)) {
	if (gc->fg != tc->fg)
		changed = 1;
} else if (!changed && (flags & GRID_FLAG_FGRGB)) {
	if (gc->fg_rgb.r != tc->fg_rgb.r)
		changed = 1;
   ... and so on ...
}
... same for bg ...
if (!changed)
	return;

Perhaps this bit could set multiple flags since basically the same
checks are later done under "/* Set the foreground colour. */". Not sure
if that would make it too hard to read though.

Otherwise looks good so far.

bergman:
I tried this on 2.1 and current master and it did not work. Colors in 256 color mode worked fine but 24bit/truecolor was black/white. There was also a gray color coming in sometimes, like when splitting, one of the splits would get a gray background color.

bsod90:
Just tried building this branch on ubuntu. Here's what I get: http://d.pr/i/1fI1N right after tmux start (gray scree, already in copy mode).
tmux 2.0 with this patch https://gist.github.com/choppsv1/dd00858d4f7f356ce2cf works well. tmux 2.1 is incompatible with that patch.

jfelchner:
@sunaku how're we doin over there? 😋

sunaku:
:cold_sweat: Free time is scarce, I'll try this week. :hourglass_flowing_sand:

jfelchner:
@sunaku no rush man. We've waited this long. :grinning: I know how it is. I'll just ping you every once in a while so you don't forget. :wink:

sunaku:
@bergman @bsod90 This patch "did not work" for you because it has _evolved beyond_ the previous ones (upon which it builds) to properly require your terminal to advertise a `Tc` (24-bit / true color) capability before it attempts to emit those 24-bit RGB escape sequences.  :sweat_smile:  In contrast, its predecessors blindly assume that your terminal (regardless of what it actually is) supports 24-bit colors and blast away! :gun:

For the time being, you'll need to fake the `Tc` capability (because no such thing officially exists _yet_ :8ball:) by configuring tmux's terminal-overrides option as follows:

    tmux set-option -ga terminal-overrides ',xterm*:Tc'  

Don't give up on progress! :rocket:

sunaku:
@nicm I've been working through your suggestions but I've hit a roadblock regarding how the `Tc` capability is meant to be specified.  In particular, this little command seems like it ought to work:

    tmux set-option -ga terminal-overrides ',xterm*:Tc'

but the `tty_term_override()` function isn't being triggered by it (according to `-vvv` logs and my added `log_debug()` lines therein) :scream:.  As a result, tmux still thinks that `Tc` is not supported. :cry:

Any ideas? :cold_sweat:   Thanks for your consideration.

nicm:
After tmux starts does "show -s terminal-overrides" have your additions?

But "tmux info" shows Tc as missing?

It looks fine to me, you should try adding some debugging to
tty_term_override.

You should use tty_term_flag to check Tc, not tty_term_has. And the
tty_term_codes array should be sorted.

bergman:
Confirmed working after adding `set-option -ga terminal-overrides ",xterm*:Tc"` to ~/.tmux.conf

I do have some weird stuff happening, like below, where colors are messed up after the first colored output and even more messed up after first background colored output:

screen shot 2015-10-23 at 10 58 50

For anyone wanting to try it out with homebrew:
```patch
diff --git a/Library/Formula/tmux.rb b/Library/Formula/tmux.rb
index 44d0dd1..a2e3232 100644
--- a/Library/Formula/tmux.rb
+++ b/Library/Formula/tmux.rb
@@ -22,6 +22,12 @@ class Tmux < Formula
   depends_on "pkg-config" => :build
   depends_on "libevent"

+  option "with-truecolor", "Build with truecolor patch enabled"
+  patch do
+    url "https://github.com/tmux/tmux/pull/112.diff"
+    sha1 "eac7a6d755df770257267f131bffffa1dfd94f29"
+  end if build.with? "truecolor"
+
   def install
     system "sh", "autogen.sh" if build.head?

```

sunaku:
@nicm Thanks, this is truly weird. :fearful:  The overrides are taking effect only _after_ I restart tmux:

```sh
$ ./tmux -vvv -L test -f /dev/null

$ ./tmux show -s terminal-overrides
terminal-overrides "xterm*:XT:Ms=\E]52;%p1%s;%p2%s\007:Cs=\E]12;%p1%s\007:Cr=\E]112\007:Ss=\E[%p1%d q:Se=\E[2 q,screen*:XT"

$ ./tmux info | grep Tc
 201: Tc: [missing]

$ ./tmux set-option -ga terminal-overrides ',xterm*:Tc'

$ ./tmux info | grep Tc
 201: Tc: [missing]

$ exit
```

Now I restart tmux and this time it somehow remembers :ghost: the overrides I previously set:

```sh
$ ./tmux -vvv -L test -f /dev/null

$ ./tmux show -s terminal-overrides
terminal-overrides "xterm*:XT:Ms=\E]52;%p1%s;%p2%s\007:Cs=\E]12;%p1%s\007:Cr=\E]112\007:Ss=\E[%p1%d q:Se=\E[2 q,screen*:XT,xterm*:Tc"

$ ./tmux info | grep Tc
 201: Tc: (flag) true
```

Is this expected? :confused:

nicm:
Are you sure you are restarting tmux entirely?

sunaku:
@nicm Yes, I think so.  This is what I did to "restart" tmux:

1. Run `./tmux -vvv -L test -f /dev/null` to start tmux.
2. Quit the one and only pane containing the one and only shell inside the session by running `exit`.
3. Run `./tmux -vvv -L test -f /dev/null` to **re**start tmux.

jfelchner:
@sunaku that should do it but you can double check by running `ps -axf | grep tmux` in a shell that _isn't_ running inside tmux and make sure it's not running.

nicm:
Right, but are you sure that actually did quit tmux? tmux -Ltest kill-servr

alerque:
I'm having trouble getting this to work and I'm not sure what's going on. I've compiled tmux with the patch from this PR (against the 2.1 release tarball and against the current master). Setting the terminal-overrides value seems to work, either from my config file or from the command line inside a session. What doesn't work is anything about Tc showing up in `tmux info`, and I don't actually get 24 bit color. What else could I be missing here?

alerque:
@kylape I see you have rebased the commits from this PR on the 2.1 source. I've tried testing this on Linux and am still not getting much love out of it (e.g. no Tc value in `tmux info`). OSX seems to be treating this differently, but I can't get it to actually spin me a color wheel on Linux. Can I ask what platform you are testing on and what output you're able to get right now?

sghctoma:
@alerque Are you sure you're setting Tc for the correct terminal name (not the terminal name inside tmux, but from where you start tmux)? For example, my terminal name\* is:

```
[0x00 ~]$ echo $TERM
[0x00 ~]$ rxvt-unicode-256color
```

so I use this in my .tmux.conf:

```
set -ga terminal-overrides ',rxvt-unicode-256color:Tc'
```

Please note that for me this command does not work from inside tmux, only from the config file. Maybe this info is not updated when using the command from tmux, but I did not yet take a look at the code to confirm this.

\* it's not actually urxvt, since it can't handle 24bit colors, but I use this name nonetheless.

alerque:
@sghctoma What terminal are you actually using? I'm wondering if some other unrelated feature is being checked and tripping me up. I'm using Termite, which sets TERM to xterm-termite:

    $ echo $TERM
    xterm-termite

I was setting xterm*, but I just tried it with the full terminal name as well in my .tmux.conf:

    set -ga terminal-overrides ",xterm-termite:Tc"

This correctly shows up if I dump the value from inside a running tmux session:

    $ tmux show -s terminal-overrides
    terminal-overrides "xterm*:XT:Ms=\E]52;%p1%s;%p2%s\007:Cs=\E]12;%p1%s\007:Cr=\E]112\007:Ss=\E[%p1%d q:Se=\E[2 q,screen*:XT,xterm-termite:Tc"

But still no love from tmux info (run inside a session):

    $ tmux info | grep -c Tc
    0

Any other ideas for what could be affecting this? With the same patch on OS X I am able to get a working 24 bit color output inside tmux, although I have the same problem with artifacts that @bergman describes above. There `tmux info` includes info about Tc, either missing or true depending on whether I have the overrides above setup properly.

sghctoma:
@alerque I'm using [st](http://st.suckless.org/). I can see now that your `tmux info` does not show anything about Tc, not even the "`Tc: [missing]`" line. Based on this, a question: is there any chance that there is another, non-patched tmux running on your machine besides the patched one? If that's the case, kill it, or just simply use another socket for the patched one using the `-L` or `-S` options!

Btw, those artifacts are present here too. It seems that the normal fg color is not reset correctly after using some colors.

nicm:
This is a PR guys, can you please stop filling it up with irrelevant
discussion, this patch does not need testing in any case, it needs more
changes.

sunaku:
Indeed, sorry for the delays. :sweat: I'll get back to this PR this week hopefully. :sweat_smile:

jfelchner:
@nicm is right.  I know everyone is excited about this (as am I :grinning:), but it's not ready yet.  It's an incomplete PR.  It is not a "patch" until it's completed.  And we can't expect @sunaku to have everything working for everyone at every step of this process.

Once @nicm and @sunaku come up with a solution they're both happy with, at that point we can test it out before merging. :shipit:

nicm:
Suraj, with c5689a5a4031a43769b8b721cafa6d1eab6abc44 you can just expand fg and bg in grid_cell to u_int, and make sure the new colour flags are added to the condition for making an extended cell in grid_set_cell.

sunaku:
@nicm Thanks for the tips although I don't fully understand how the `GRID_FLAG_EXTENDED` flag works yet.  You suggested changing `fg` and `bg` to `u_int`, does that mean we're okay with storing colors as 32-bit integers?

I'm trying to wrap my head around the new changes which have presented themselves quite prominently as merge conflicts in 78a00c845cc4d4e80f36f7cda71e0ca9ea13dcf4 when in I tried rebasing this patch atop the current master branch. :cold_sweat:

nicm:
So now instead of being made up of arrays of grid_cell, the grid is made
up of arrays of grid_cell_entry which is a lot smaller. A
grid_cell_entry may be marked as extended (the GRID_FLAG_EXTENDED bit is
set). In that case it is an offset into the extddata array of grid_cell
structs. Any character which doesn't fit in grid_cell (currently, if it
is UTF-8) is marked as an extended cell.

If a cell is NOT extended, it is translated from a grid_cell_entry into
a grid_cell in grid_get_cell and vice versa in grid_set_cell. If it is
extended, it is just copied directly to and from the extddata array.

You can make the colours in grid_cell 32-bit, so long as RGB colour
cells are marked as extended - the same as UTF-8 cells are now. The
extra bytes are fine because there will be vastly fewer extended cells
than non-extended, and it is only affecting the few people who use RGB
colour or UTF-8 rather than punishing everyone.

I suggest that rather than doing u_int and using masks all over the
place to get the R, G and B out, it would still be nicer to do a union,
like this:

struct grid_cell_rgb {
	u_char	r;
	u_char	g;
	u_char	b;
};
struct grid_cell {
...
	union {
		u_char			fg;
		struct grid_cell_rgb	fg_rgb;
	};
	union {
		u_char			bg;
		struct grid_cell_rgb	fg_rgb;
	};
...
}

(Or there are other variations if they seem better.)

dasJ:
Sorry for being annoying, but is this PR dead?

ThomasAdam:
@dasJ no, it' s not.  But it seems people's patience are definitely dead.

Wait.  Just wait.

jfelchner:
@dasJ You're welcome to help out on the PR.  Otherwise, be patient. :smile:

sunaku:
:+1: Yup, definitely not dead. :sweat_smile: Thanks for your patience. :santa:

rhysd:
Is there any update for this PR?

sunaku:
Sorry to keep you all waiting. :santa: Free time has been scarce :runner: and motivation more so! :cold_sweat: :confounded: :scream: :boom: 

## Progress

I rebased the patch onto master, solving the merge conflicts I was stuck at in https://github.com/tmux/tmux/pull/112#issuecomment-158680798.

I have also implemented @nicm's suggestions for refactoring `tty_colours()` in https://github.com/tmux/tmux/pull/112#issuecomment-145831802.

## Obstacles

The `Tc` detection is _still_ not working for me under the **st** terminal, so I've been verifying the functionality using these compile-time overrides locally:
```diff
diff --git a/tty.c b/tty.c
index 6793a7e..9fe61bd 100644
--- a/tty.c
+++ b/tty.c
@@ -1498,7 +1498,7 @@ tty_check_fg(struct tty *tty, struct grid_cell *gc)
 		/*
 		 * And not a 24-bit terminal? Translate to 256-colour palette.
 		 */
-		if (!tty_term_has(tty->term, TTYC_TC)) {
+		if (0/* !tty_term_has(tty->term, TTYC_TC) */) {
 			gc->flags &= ~GRID_FLAG_FGRGB;
 			gc->flags |= ~GRID_FLAG_FG256;
 			gc->fg = colour_find_rgb(gc->fg_rgb.r, gc->fg_rgb.g, gc->fg_rgb.b);
@@ -1543,7 +1543,7 @@ tty_check_bg(struct tty *tty, struct grid_cell *gc)
 		/*
 		 * And not a 24-bit terminal? Translate to 256-colour palette.
 		 */
-		if (!tty_term_has(tty->term, TTYC_TC)) {
+		if (0/* !tty_term_has(tty->term, TTYC_TC) */) {
 			gc->flags &= ~GRID_FLAG_BGRGB;
 			gc->flags |= ~GRID_FLAG_BG256;
 			gc->bg = colour_find_rgb(gc->bg_rgb.r, gc->bg_rgb.g, gc->bg_rgb.b);
@@ -1707,7 +1707,7 @@ tty_try_rgb(struct tty *tty, struct grid_cell_rgb rgb, const char *type)
 {
 	char	s[32];
 
-	if (!tty_term_has(tty->term, TTYC_TC))
+	if (0/* !tty_term_has(tty->term, TTYC_TC) */)
 		return (-1);
 
 	xsnprintf(s, sizeof s, "\033[%s;2;%hhu;%hhu;%hhum", type, rgb.r, rgb.g, rgb.b);
```

## Pending

Need to update `grid_set_cell()` for RGB colors as described in https://github.com/tmux/tmux/pull/112#issuecomment-156408924.

Need to review this entire thread again and finish up any loose ends and TODO comments.

Feel free to jump in and carry the torch further if you want to speed this along. :santa: Cheers!

zchee:
@sunaku Thanks! awesome job!

jfelchner:
@sunaku amazing. 😄 If there's any simple stuff that I can help with please let me know.

rhysd:
@sunaku 

Thank you so much for the update.  It tried Neovim on iTerm2 & Retina display and the experience was excellent :smile:  I hope to implement this feature.

nicm:
Suraj, please look at [this diff](https://github.com/tmux/tmux/files/91718/rgb.diff.txt).

Based on your latest, some cleanup and:

- Add the EXTENDED flag dance to grid.c.
- Move the helper functions into tty.c and change the style.
- Use the same goto save_fg thing for RGB as 256 in tty_colours_fg.
- Pass a pointer into grid_try_rgb - this is not C++! :-).
- Fix the grid_default_cell definition.

Builds and runs but I didn't test too much, if it doesn't work, let me know... or fix it... :-).

sunaku:
Nicholas, thanks for reviewing and revising this patch! :joy_cat:

Some quantization (especially in green, on the third line) is happening with your patch, however:

nicm

Whereas it does not happen with mine:

sunaku

This is probably related to the EXTENDED flag dance. :crossed_flags: I'll dig more this weekend. :santa:

nicm:
I'd guess my patch is translating to 256 when it should not be, that would
cause this I think.

sunaku:
:+1: You're right, it was caused by a small bug in the `tty_same` functions:

```diff
diff --git a/tty.c b/tty.c
index 93e2e7d..9928909 100644
--- a/tty.c
+++ b/tty.c
@@ -73,8 +73,8 @@ tty_same_fg(const struct grid_cell *gc1, const struct grid_cell *gc2)
 {
 	int	flags1, flags2;
 
-	flags1 = (gc1->flags & (GRID_FLAG_FG256|GRID_FLAG_BG256));
-	flags2 = (gc2->flags & (GRID_FLAG_FG256|GRID_FLAG_BG256));
+	flags1 = (gc1->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
+	flags2 = (gc2->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
 
 	if (flags1 != flags2)
 	    return (0);
@@ -96,8 +96,8 @@ tty_same_bg(const struct grid_cell *gc1, const struct grid_cell *gc2)
 {
 	int	flags1, flags2;
 
-	flags1 = (gc1->flags & (GRID_FLAG_BG256|GRID_FLAG_BG256));
-	flags2 = (gc2->flags & (GRID_FLAG_BG256|GRID_FLAG_BG256));
+	flags1 = (gc1->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
+	flags2 = (gc2->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
 
 	if (flags1 != flags2)
 	    return (0);
```

zchee:
@sunaku @nicm Hi :)
Thanks for wonderful patch!!

I found a small bug.
`38` instead of `48`.

```diff
diff --git a/tty.c b/tty.c
index 77c7dfc..e805b40 100644
--- a/tty.c
+++ b/tty.c
@@ -1651,7 +1651,7 @@ tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
 
 	/* Is this a 24-bit colour? */
 	if (gc->flags & GRID_FLAG_FGRGB) {
-		if (tty_try_rgb(tty, &gc->fg_rgb, "48") == 0)
+		if (tty_try_rgb(tty, &gc->fg_rgb, "38") == 0)
 			goto save_fg;
 		/* Should not get here, already converted in tty_check_fg. */
 		return;
```

nicm:
Good spot, both of you. Here is the latest.



On Fri, Jan 15, 2016 at 01:44:38PM -0800, Koichi Shiraishi wrote:
>    [1]@sunaku [2]@nicm Hi :)
>    Thanks for wonderful patch!!
> 
>    I found a small bug.
>    38 instead of 48.
> 
>  diff --git a/tty.c b/tty.c
>  index 77c7dfc..e805b40 100644
>  --- a/tty.c
>  +++ b/tty.c
>  @@ -1651,7 +1651,7 @@ tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
> 
>      /* Is this a 24-bit colour? */
>      if (gc->flags & GRID_FLAG_FGRGB) {
>  -       if (tty_try_rgb(tty, &gc->fg_rgb, "48") == 0)
>  +       if (tty_try_rgb(tty, &gc->fg_rgb, "38") == 0)
>              goto save_fg;
>          /* Should not get here, already converted in tty_check_fg. */
>          return;
> 
>    --
>    Reply to this email directly or [3]view it on GitHub.
> 
>    Reverse link: [4]unknown
> 
> References
> 
>    Visible links
>    1. https://github.com/sunaku
>    2. https://github.com/nicm
>    3. https://github.com/tmux/tmux/pull/112#issuecomment-172102962
>    4. https://github.com/tmux/tmux/pull/112#issuecomment-172102962

diff --git a/grid.c b/grid.c
index 579eb96..b3ebde3 100644
--- a/grid.c
+++ b/grid.c
@@ -37,7 +37,7 @@
 
 /* Default grid cell data. */
 const struct grid_cell grid_default_cell = {
-	0, 0, 8, 8, { { ' ' }, 0, 1, 1 }
+	0, 0, { .fg = 8 }, { .bg = 8 }, { { ' ' }, 0, 1, 1 }
 };
 const struct grid_cell_entry grid_default_entry = {
 	0, { .data = { 0, 8, 8, ' ' } }
@@ -284,6 +284,7 @@ grid_set_cell(struct grid *gd, u_int px, u_int py, const struct grid_cell *gc)
 	struct grid_line	*gl;
 	struct grid_cell_entry	*gce;
 	struct grid_cell 	*gcp;
+	int			 extended;
 
 	if (grid_check_y(gd, py) != 0)
 		return;
@@ -293,8 +294,12 @@ grid_set_cell(struct grid *gd, u_int px, u_int py, const struct grid_cell *gc)
 	gl = &gd->linedata[py];
 	gce = &gl->celldata[px];
 
-	if ((gce->flags & GRID_FLAG_EXTENDED) || gc->data.size != 1 ||
-	    gc->data.width != 1) {
+	extended = (gce->flags & GRID_FLAG_EXTENDED);
+	if (!extended && (gc->data.size != 1 || gc->data.width != 1))
+		extended = 1;
+	if (!extended && (gce->flags & (GRID_FLAG_FGRGB|GRID_FLAG_BGRGB)))
+		extended = 1;
+	if (extended) {
 		if (~gce->flags & GRID_FLAG_EXTENDED) {
 			gl->extddata = xreallocarray(gl->extddata,
 			    gl->extdsize + 1, sizeof *gl->extddata);
diff --git a/input.c b/input.c
index ed1ebff..c1388cb 100644
--- a/input.c
+++ b/input.c
@@ -1629,18 +1629,20 @@ input_csi_dispatch_sgr_256(struct input_ctx *ictx, int fgbg, u_int *i)
 	c = input_get(ictx, *i, 0, -1);
 	if (c == -1) {
 		if (fgbg == 38) {
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = 8;
 		} else if (fgbg == 48) {
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = 8;
 		}
 	} else {
 		if (fgbg == 38) {
 			gc->flags |= GRID_FLAG_FG256;
+			gc->flags &= ~GRID_FLAG_FGRGB;
 			gc->fg = c;
 		} else if (fgbg == 48) {
 			gc->flags |= GRID_FLAG_BG256;
+			gc->flags &= ~GRID_FLAG_BGRGB;
 			gc->bg = c;
 		}
 	}
@@ -1666,13 +1668,18 @@ input_csi_dispatch_sgr_rgb(struct input_ctx *ictx, int fgbg, u_int *i)
 	if (b == -1 || b > 255)
 		return;
 
-	c = colour_find_rgb(r, g, b);
 	if (fgbg == 38) {
-		gc->flags |= GRID_FLAG_FG256;
-		gc->fg = c;
+		gc->flags &= ~GRID_FLAG_FG256;
+		gc->flags |= GRID_FLAG_FGRGB;
+		gc->fg_rgb.r = r;
+		gc->fg_rgb.g = g;
+		gc->fg_rgb.b = b;
 	} else if (fgbg == 48) {
-		gc->flags |= GRID_FLAG_BG256;
-		gc->bg = c;
+		gc->flags &= ~GRID_FLAG_BG256;
+		gc->flags |= GRID_FLAG_BGRGB;
+		gc->bg_rgb.r = r;
+		gc->bg_rgb.g = g;
+		gc->bg_rgb.b = b;
 	}
 }
 
@@ -1754,11 +1761,11 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 35:
 		case 36:
 		case 37:
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = n - 30;
 			break;
 		case 39:
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = 8;
 			break;
 		case 40:
@@ -1769,11 +1776,11 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 45:
 		case 46:
 		case 47:
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = n - 40;
 			break;
 		case 49:
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = 8;
 			break;
 		case 90:
@@ -1784,7 +1791,7 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 95:
 		case 96:
 		case 97:
-			gc->flags &= ~GRID_FLAG_FG256;
+			gc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			gc->fg = n;
 			break;
 		case 100:
@@ -1795,7 +1802,7 @@ input_csi_dispatch_sgr(struct input_ctx *ictx)
 		case 105:
 		case 106:
 		case 107:
-			gc->flags &= ~GRID_FLAG_BG256;
+			gc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			gc->bg = n - 10;
 			break;
 		}
diff --git a/tmux.h b/tmux.h
index ce40906..a773af6 100644
--- a/tmux.h
+++ b/tmux.h
@@ -387,6 +387,7 @@ enum tty_code_code {
 	TTYC_SMSO,	/* enter_standout_mode, so */
 	TTYC_SMUL,	/* enter_underline_mode, us */
 	TTYC_SS,	/* set cursor style, Ss */
+	TTYC_TC,	/* 24-bit "true" colour, Tc */
 	TTYC_TSL,	/* to_status_line, tsl */
 	TTYC_VPA,	/* row_address, cv */
 	TTYC_XENL,	/* eat_newline_glitch, xn */
@@ -643,16 +644,31 @@ enum utf8_state {
 #define GRID_FLAG_BG256 0x2
 #define GRID_FLAG_PADDING 0x4
 #define GRID_FLAG_EXTENDED 0x8
+#define GRID_FLAG_FGRGB 0x10
+#define GRID_FLAG_BGRGB 0x20
 
 /* Grid line flags. */
 #define GRID_LINE_WRAPPED 0x1
 
+/* Grid cell RGB colours. */
+struct grid_cell_rgb {
+	u_char	r;
+	u_char	g;
+	u_char	b;
+};
+
 /* Grid cell data. */
 struct grid_cell {
 	u_char			flags;
 	u_char			attr;
-	u_char			fg;
-	u_char			bg;
+	union {
+		u_char		fg;
+		struct grid_cell_rgb	fg_rgb;
+	};
+	union {
+		u_char		bg;
+		struct grid_cell_rgb	bg_rgb;
+	};
 	struct utf8_data	data;
 
 };
diff --git a/tty-term.c b/tty-term.c
index df69bd1..a673d04 100644
--- a/tty-term.c
+++ b/tty-term.c
@@ -254,6 +254,7 @@ const struct tty_term_code_entry tty_term_codes[] = {
 	[TTYC_SMSO] = { TTYCODE_STRING, "smso" },
 	[TTYC_SMUL] = { TTYCODE_STRING, "smul" },
 	[TTYC_SS] = { TTYCODE_STRING, "Ss" },
+	[TTYC_TC] = { TTYCODE_FLAG, "Tc" },
 	[TTYC_TSL] = { TTYCODE_STRING, "tsl" },
 	[TTYC_VPA] = { TTYCODE_STRING, "vpa" },
 	[TTYC_XENL] = { TTYCODE_FLAG, "xenl" },
diff --git a/tty.c b/tty.c
index da5eac4..059724c 100644
--- a/tty.c
+++ b/tty.c
@@ -36,8 +36,15 @@ static int tty_log_fd = -1;
 void	tty_read_callback(struct bufferevent *, void *);
 void	tty_error_callback(struct bufferevent *, short, void *);
 
+static int tty_same_fg(const struct grid_cell *, const struct grid_cell *);
+static int tty_same_bg(const struct grid_cell *, const struct grid_cell *);
+static int tty_same_colours(const struct grid_cell *, const struct grid_cell *);
+static int tty_is_fg(const struct grid_cell *, int);
+static int tty_is_bg(const struct grid_cell *, int);
+
 void	tty_set_italics(struct tty *);
 int	tty_try_256(struct tty *, u_char, const char *);
+int	tty_try_rgb(struct tty *, const struct grid_cell_rgb *, const char *);
 
 void	tty_colours(struct tty *, const struct grid_cell *);
 void	tty_check_fg(struct tty *, struct grid_cell *);
@@ -61,6 +68,74 @@ void	tty_default_colours(struct grid_cell *, const struct window_pane *);
 #define tty_pane_full_width(tty, ctx) \
 	((ctx)->xoff == 0 && screen_size_x((ctx)->wp->screen) >= (tty)->sx)
 
+static int
+tty_same_fg(const struct grid_cell *gc1, const struct grid_cell *gc2)
+{
+	int	flags1, flags2;
+
+	flags1 = (gc1->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
+	flags2 = (gc2->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
+
+	if (flags1 != flags2)
+	    return (0);
+
+	if (flags1 & GRID_FLAG_FGRGB) {
+		if (gc1->fg_rgb.r != gc2->fg_rgb.r)
+			return (0);
+		if (gc1->fg_rgb.g != gc2->fg_rgb.g)
+			return (0);
+		if (gc1->fg_rgb.b != gc2->fg_rgb.b)
+			return (0);
+		return (1);
+	}
+	return (gc1->fg == gc2->fg);
+}
+
+static int
+tty_same_bg(const struct grid_cell *gc1, const struct grid_cell *gc2)
+{
+	int	flags1, flags2;
+
+	flags1 = (gc1->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
+	flags2 = (gc2->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
+
+	if (flags1 != flags2)
+	    return (0);
+
+	if (flags1 & GRID_FLAG_BGRGB) {
+		if (gc1->bg_rgb.r != gc2->bg_rgb.r)
+			return (0);
+		if (gc1->bg_rgb.g != gc2->bg_rgb.g)
+			return (0);
+		if (gc1->bg_rgb.b != gc2->bg_rgb.b)
+			return (0);
+		return (1);
+	}
+	return (gc1->bg == gc2->bg);
+}
+
+static int
+tty_same_colours(const struct grid_cell *gc1, const struct grid_cell *gc2)
+{
+	return (tty_same_fg(gc1, gc2) && tty_same_bg(gc1, gc2));
+}
+
+static int
+tty_is_fg(const struct grid_cell *gc, int c)
+{
+	if (gc->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB))
+		return (0);
+	return (gc->fg == c);
+}
+
+static int
+tty_is_bg(const struct grid_cell *gc, int c)
+{
+	if (gc->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB))
+		return (0);
+	return (gc->bg == c);
+}
+
 void
 tty_create_log(void)
 {
@@ -1423,12 +1498,10 @@ void
 tty_colours(struct tty *tty, const struct grid_cell *gc)
 {
 	struct grid_cell	*tc = &tty->cell;
-	u_char			 fg = gc->fg, bg = gc->bg, flags = gc->flags;
 	int			 have_ax, fg_default, bg_default;
 
 	/* No changes? Nothing is necessary. */
-	if (fg == tc->fg && bg == tc->bg &&
-	    ((flags ^ tc->flags) & (GRID_FLAG_FG256|GRID_FLAG_BG256)) == 0)
+	if (tty_same_colours(gc, tc))
 		return;
 
 	/*
@@ -1437,8 +1510,8 @@ tty_colours(struct tty *tty, const struct grid_cell *gc)
 	 * case if only one is default need to fall onward to set the other
 	 * colour.
 	 */
-	fg_default = (fg == 8 && !(flags & GRID_FLAG_FG256));
-	bg_default = (bg == 8 && !(flags & GRID_FLAG_BG256));
+	fg_default = tty_is_fg(gc, 8);
+	bg_default = tty_is_bg(gc, 8);
 	if (fg_default || bg_default) {
 		/*
 		 * If don't have AX but do have op, send sgr0 (op can't
@@ -1451,48 +1524,52 @@ tty_colours(struct tty *tty, const struct grid_cell *gc)
 		if (!have_ax && tty_term_has(tty->term, TTYC_OP))
 			tty_reset(tty);
 		else {
-			if (fg_default &&
-			    (tc->fg != 8 || tc->flags & GRID_FLAG_FG256)) {
+			if (fg_default && !tty_is_fg(tc, 8)) {
 				if (have_ax)
 					tty_puts(tty, "\033[39m");
-				else if (tc->fg != 7 ||
-				    tc->flags & GRID_FLAG_FG256)
+				else if (!tty_is_fg(tc, 7))
 					tty_putcode1(tty, TTYC_SETAF, 7);
 				tc->fg = 8;
-				tc->flags &= ~GRID_FLAG_FG256;
+				tc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
 			}
-			if (bg_default &&
-			    (tc->bg != 8 || tc->flags & GRID_FLAG_BG256)) {
+			if (bg_default && !tty_is_bg(tc, 8)) {
 				if (have_ax)
 					tty_puts(tty, "\033[49m");
-				else if (tc->bg != 0 ||
-				    tc->flags & GRID_FLAG_BG256)
+				else if (!tty_is_bg(tc, 0))
 					tty_putcode1(tty, TTYC_SETAB, 0);
 				tc->bg = 8;
-				tc->flags &= ~GRID_FLAG_BG256;
+				tc->flags &= ~(GRID_FLAG_BG256|GRID_FLAG_BGRGB);
 			}
 		}
 	}
 
 	/* Set the foreground colour. */
-	if (!fg_default && (fg != tc->fg ||
-	    ((flags & GRID_FLAG_FG256) != (tc->flags & GRID_FLAG_FG256))))
+	if (!fg_default && !tty_same_fg(gc, tc))
 		tty_colours_fg(tty, gc);
 
 	/*
 	 * Set the background colour. This must come after the foreground as
 	 * tty_colour_fg() can call tty_reset().
 	 */
-	if (!bg_default && (bg != tc->bg ||
-	    ((flags & GRID_FLAG_BG256) != (tc->flags & GRID_FLAG_BG256))))
+	if (!bg_default && !tty_same_bg(gc, tc))
 		tty_colours_bg(tty, gc);
 }
 
 void
 tty_check_fg(struct tty *tty, struct grid_cell *gc)
 {
-	u_int	colours;
+	struct grid_cell_rgb	*rgb = &gc->fg_rgb;
+	u_int			 colours;
 
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_FGRGB) {
+		/* Not a 24-bit terminal? Translate to 256-colour palette. */
+		if (!tty_term_flag(tty->term, TTYC_TC)) {
+			gc->flags &= ~GRID_FLAG_FGRGB;
+			gc->flags |= ~GRID_FLAG_FG256;
+			gc->fg = colour_find_rgb(rgb->r, rgb->g, rgb->b);
+		}
+	}
 	colours = tty_term_number(tty->term, TTYC_COLORS);
 
 	/* Is this a 256-colour colour? */
@@ -1524,8 +1601,18 @@ tty_check_fg(struct tty *tty, struct grid_cell *gc)
 void
 tty_check_bg(struct tty *tty, struct grid_cell *gc)
 {
-	u_int	colours;
+	struct grid_cell_rgb	*rgb = &gc->bg_rgb;
+	u_int			 colours;
 
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_BGRGB) {
+		/* Not a 24-bit terminal? Translate to 256-colour palette. */
+		if (!tty_term_flag(tty->term, TTYC_TC)) {
+			gc->flags &= ~GRID_FLAG_BGRGB;
+			gc->flags |= ~GRID_FLAG_BG256;
+			gc->bg = colour_find_rgb(rgb->r, rgb->g, rgb->b);
+		}
+	}
 	colours = tty_term_number(tty->term, TTYC_COLORS);
 
 	/* Is this a 256-colour colour? */
@@ -1560,12 +1647,21 @@ tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
 	u_char			 fg = gc->fg;
 	char			 s[32];
 
+	tc->flags &= ~(GRID_FLAG_FG256|GRID_FLAG_FGRGB);
+
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_FGRGB) {
+		if (tty_try_rgb(tty, &gc->fg_rgb, "38") == 0)
+			goto save_fg;
+		/* Should not get here, already converted in tty_check_fg. */
+		return;
+	}
+
 	/* Is this a 256-colour colour? */
 	if (gc->flags & GRID_FLAG_FG256) {
-		/* Try as 256 colours. */
 		if (tty_try_256(tty, fg, "38") == 0)
 			goto save_fg;
-		/* Else already handled by tty_check_fg. */
+		/* Should not get here, already converted in tty_check_fg. */
 		return;
 	}
 
@@ -1581,9 +1677,12 @@ tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
 
 save_fg:
 	/* Save the new values in the terminal current cell. */
-	tc->fg = fg;
-	tc->flags &= ~GRID_FLAG_FG256;
-	tc->flags |= gc->flags & GRID_FLAG_FG256;
+	if (gc->flags & GRID_FLAG_FGRGB)
+		memcpy(&tc->fg_rgb, &gc->fg_rgb, sizeof tc->fg_rgb);
+	else
+		tc->fg = fg;
+	tc->flags &= ~(GRID_FLAG_FGRGB|GRID_FLAG_FG256);
+	tc->flags |= (gc->flags & (GRID_FLAG_FG256|GRID_FLAG_FGRGB));
 }
 
 void
@@ -1593,12 +1692,19 @@ tty_colours_bg(struct tty *tty, const struct grid_cell *gc)
 	u_char			 bg = gc->bg;
 	char			 s[32];
 
+	/* Is this a 24-bit colour? */
+	if (gc->flags & GRID_FLAG_BGRGB) {
+		if (tty_try_rgb(tty, &gc->bg_rgb, "48") == 0)
+			goto save_bg;
+		/* Should not get here, already converted in tty_check_bg. */
+		return;
+	}
+
 	/* Is this a 256-colour colour? */
 	if (gc->flags & GRID_FLAG_BG256) {
-		/* Try as 256 colours. */
 		if (tty_try_256(tty, bg, "48") == 0)
 			goto save_bg;
-		/* Else already handled by tty_check_bg. */
+		/* Should not get here, already converted in tty_check_bg. */
 		return;
 	}
 
@@ -1614,9 +1720,12 @@ tty_colours_bg(struct tty *tty, const struct grid_cell *gc)
 
 save_bg:
 	/* Save the new values in the terminal current cell. */
-	tc->bg = bg;
-	tc->flags &= ~GRID_FLAG_BG256;
-	tc->flags |= gc->flags & GRID_FLAG_BG256;
+	if (gc->flags & GRID_FLAG_BGRGB)
+		memcpy(&tc->bg_rgb, &gc->bg_rgb, sizeof tc->bg_rgb);
+	else
+		tc->bg = bg;
+	tc->flags &= ~(GRID_FLAG_BGRGB|GRID_FLAG_BG256);
+	tc->flags |= (gc->flags & (GRID_FLAG_BG256|GRID_FLAG_BGRGB));
 }
 
 int
@@ -1656,6 +1765,20 @@ fallback:
 	return (0);
 }
 
+int
+tty_try_rgb(struct tty *tty, const struct grid_cell_rgb *rgb, const char *type)
+{
+	char	s[32];
+
+	if (!tty_term_flag(tty->term, TTYC_TC))
+		return (-1);
+
+	xsnprintf(s, sizeof s, "\033[%s;2;%hhu;%hhu;%hhum", type, rgb->r,
+	    rgb->g, rgb->b);
+	tty_puts(tty, s);
+	return (0);
+}
+
 void
 tty_default_colours(struct grid_cell *gc, const struct window_pane *wp)
 {

zchee:
@nicm Thanks update :)

I patched your diff, 24bit color is fully working.
but, does not works save_bg(?) when switching window.
Caused by my settings? Or still under construction...?

I still do not understand all. Sorry :(

===

Here is simple `.tmux.conf`

```conf
set -ga terminal-overrides ',xterm*:Tc'
set -sg escape-time 1
set -sg focus-events on
set -g default-terminal "xterm-256color"

# Quick kill-session
bind C-q confirm-before 'kill-session'
```

===

and, screenshots.

Launched `tmux`, and run 24bit-color test script.

screen shot 2016-01-16 at 8 36 22 am

Ctrl-b - C, create a new-window

screen shot 2016-01-16 at 8 36 34 am

Last, Ctrl-b - n, switch window.
strange bg color.

screen shot 2016-01-16 at 8 36 40 am

nicm:
I found another bug, but I don't know if it is your problem 
[rgb3.diff.txt](https://github.com/tmux/tmux/files/92697/rgb3.diff.txt)

nicm:
This one probably does fix your bug
[rgb4.diff.txt](https://github.com/tmux/tmux/files/92701/rgb4.diff.txt)

zchee:
@nicm Thanks!
bg issue has been solved!

I am very grateful to you.
Thanks again.

sunaku:
:tada: Excellent! :triumph: PR updated with @nicm's latest version of the patch. :gift:

`Tc` detection is still broken for me in **st 0.6** so I'll try digging into that next. :fishing_pole_and_fish:

nicm:
I contacted Thomas Dickey and he doesn't have any better recommendation for tmux to detect this feature, so I think I am going to stick with just having a Tc flag. [Here is latest with man page.](https://github.com/tmux/tmux/files/94049/tmux-rgb.diff.txt)

jfelchner:
This is so exciting!! Been waiting for this for a long time. :grinning:

@sunaku thank you so very much for the hours and hours you spent on this!! :+1:

dasJ:
I don't know if I'm doing anything wrong, but it doesn't work for me. I use st as terminal emulator (supports True Colour of course) and if I just test, it's okay:

st

Now, when I run tmux (git version with the patch of this PR and with the configuration of @zchee), it looks like this: 

tmux

Is this a st problem? My config can be found here: https://github.com/dasJ/dotfiles/blob/master/st-config.h

sunaku:
@dasJ Yes, the `Tc` setting isn't working under **st** (version 0.6 here) for me either. :cold_sweat:

I've tried all of these overrides, but none trigger the `tty_term_override()` function:

```sh
./tmux set-option -ga terminal-overrides ',xterm*:Tc'
./tmux set-option -ga terminal-overrides ',st*:Tc'
./tmux set-option -ga terminal-overrides ',*:Tc'
./tmux set-option -ga terminal-overrides ',st-256color:Tc'
./tmux set-option -ga terminal-overrides ',screen:Tc'
```

So this is simply a matter of us being unable to register the `Tc` flag in that setup. :crossed_flags:

dasJ:
So if I'm getting this correctly, is st just too simple to support this?

nicm:
Do this:

tmux set -ag terminal-overrides ',*:Tc'
tmux detach
tmux attach

And show me the output of:

tmux show -sg terminal-overrides
tmux info|grep Tc

zchee:
@dasJ Yes,  I will work with the contents said by @nicm.
Or, I posted simple `tmux.conf`.
https://github.com/tmux/tmux/pull/112#issuecomment-172128418

```conf
set -ga terminal-overrides ',xterm*:Tc'
set -g default-terminal "xterm-256color"
```

Also will work on OS X, iTerm, zsh.

nicm:
This is wrong, default-terminal should never be anything except tmux*
or screen*. Maybe I should make tmux refuse to run if it is not.

zchee:
@nicm
Ah, sorry. 
Because it has been copy & paste from vim.
In fact, tmux* or screen*.
You're perfectly right.

dasJ:
@zchee Yes, I was previously using your configuration file.

@nicm terminal overrides: 
```
terminal-overrides "xterm*:XT:Ms=\E]52;%p1%s;%p2%s\007:Cs=\E]12;%p1%s\007:Cr=\E]112\007:Ss=\E[%p1%d q:Se=\E[2 q,screen*:XT,xterm*:Tc,*:Tc"
```
tmux info:
```
 197: Tc: (flag) true
```

After running: `tmux set -ag terminal-overrides ',*:Tc'`, **detaching and reattaching**, it works! It does not matter if I'm using xterm or screen in the configuration.

nicm:
*:Tc will match any TERM, and you always need to detach and reattach after changing terminal-overrides.

sunaku:
> you always need to detach and reattach after changing terminal-overrides

@nicm That did the trick! :bow: The man page _really_ ought to state this fact. :books:

dasJ:
Yes, it works perfectly:
colors

One thing I noticed (I don't know if it's related), my vim looks very bad:

vim1

It usually looks like that (Molokai theme):
vim2

sunaku:
@dasJ that looks like [a Vim BCE issue](http://sunaku.github.io/vim-256color-bce.html). :neckbeard: What does `:echo &t_ut` show?

dasJ:
@sunaku Yes, your blog post fixed it, thank you so much! :+1: I'm using tmux with this commit on all of my systems and I haven't found any problems. :tada:

sunaku:
@nicm Do you think this PR is ready to be merged? :shipit:  Is more testing necessary, perhaps? :fire:

sunaku:
I've added some usage instructions to the commit message now, in case this PR is ready to merge. :bowtie:

rhysd:
I tried this PR in my local machine (MacBook Pro Retina 2015, iTerm2) and it rendered Neovim with 24bit colors perfectly!  Thank you so much.

nicm:
Applied now.

zchee:
@nicm :tada: 
Thanks !!

dasJ:
Thank you so much! :+1:

rhysd:
:tada:

martin-svk:
Does it mean this was merged and tmux will have true colors support in next realease?

XVilka:
@martin-svk yes, see commit https://github.com/tmux/tmux/commit/427b8204268af5548d09b830e101c59daa095df9

martin-svk:
@XVilka thanks for clarification. 

That's awesome, finally I will be able to use all those fancy color schemes :+1: :tada:

jfelchner:
Thank you all!  Especially @sunaku.  I've been waiting for this for a long time.  What a fantastic job. :grinning:

sunaku:
Thanks everyone :relaxed: but we couldn't have done this without @nicm's tremendous assistance. :neckbeard: 

Hurray! :balloon: :birthday: :gift: :tada: :smile_cat: Let's enjoy our newfound 16 million colors! :rainbow: And long live tmux! :crown:

P.S. I wrote up [a little story](https://sunaku.github.io/tmux-24bit-color.html) (along with [usage instructions](https://sunaku.github.io/tmux-24bit-color.html#usage), of course) about all this [on my blog](https://sunaku.github.io). :memo:

jfelchner:
@sunaku absolutely right!  Tmux is a tool that I use every single day and it's work like this (and continued support by @nicm, @ThomasAdam and others) that makes it continue to stay relevant and immune to the temptation for people to create forks or other competitors touting "It's like Tmux but it does x".

I'm waiting for the actual stable release to use this, but I'm extremely excited. :)  @nicm any idea when the next release is going out?

ThomasAdam:
On 30 January 2016 at 21:23, Jeff Felchner <notifications@github.com> wrote:

> I'm waiting for the actual stable release to use this, but I'm extremely
> excited. :) @nicm <https://github.com/nicm> any idea when the next
> release is going out?
>
I release the portable version to coincide with OpenBSD releases.  The next
one is in May.

-- Thomas Adam

alexlafroscia:
Just a note for anyone that uses Homebrew and wants to install this patch; you can do so by uninstalling Tmux:

```bash
$ brew uninstall tmux
```

editing the formula for Tmux

```bash
$ brew edit tmux
```

and comment out references to the Bash completion files on lines `46` and `47`, since the `examples/` directory no longer exists.  Then you can install Tmux again from `master` using

```bash
$ brew install tmux --HEAD
```

which will grab the latest commit.  Of course, only do this if you're OK with using the development version of Tmux.

jfelchner:
@ThomasAdam thank you.
@alexlafroscia that'll work for me.  Gracias. :)

HeroCC:
Will this PR be included in 2.2? Or is it already in 2.1?

jfelchner:
@HeroCC it'll be released in the next release whatever number happens to be.

assaf758:
While trying 24bit-color support, I noticed there is some issue (when comparing to non-tmux terminal) - see http://ibin.co/2YAmCLcoBNRC.
Happens with two tests (see below) and two terminals (termite and gnome-terminal):
https://raw.githubusercontent.com/robertknight/konsole/master/tests/color-spaces.pl
https://github.com/gnachman/iTerm2/blob/master/tests/24-bit-color.sh

alerque:
@assaf758 You should open a separate issue report for this. This PR covered the initial implementation but if there are bugs now that it has been merged they should be reported separately. When you do please include more information about your setup including the exact commit you compiled TMUX from and what version of the terminals  you are using. I am unable to replicate this using those same scripts and Termite as my terminal, so something very specific to your setup is broken and we'll need more details to track it down.

neon64:
I'm not sure if this is the correct place to ask, so apologies in advance... but does this patch include support for 24-bit colour within tmux.conf as well? I tried using '00ff00' as a colour, for example, but that didn't work.

Issue #120, opened 2015-09-22
Using terminal overrides to force 256 colors on xterm stopped working around tmux 1.8


filbranden:
In order to reproduce it:

- Add this line to your `.tmux.conf`:
```
set -g terminal-overrides "xterm*:colors=256"
```
- Set `TERM=xterm` outside of tmux.
- Start tmux.
- Set `TERM=screen-256color` or something that supports 256 colors inside tmux.
- Run a 256color application, such as vim, and check for correct colors (e.g. `:set t_Co?` inside vim).

I traced this to commit f2e54e1e2ff5a1902cdab4db50a503541186638e, reverting this commit on top of HEAD seems to fix this issue, then the terminal overrides seem to work correctly as expected.

I imagine that the problem is because the terminal-overrides indicate that 256 colors are available, but when searching the terminfo for `setaf` and `setab` they are not present...

Not really sure what is the best fix for this issue. Is it possible to detect whether `setaf` and/or `setab` are present and functional? Or maybe fallback to the old command if the `colors=256` is coming from a `terminal-overrides` setting?

Or maybe that setting now needs to be updated to include a sample value of `setaf` and `setab` other than setting `colors`? Though that seems to be a fairly common recipe, being even (somewhat) included in the tmux manpage itself...
https://github.com/tmux/tmux/blob/master/tmux.1#L2477

For now, I'm running with that commit reverted... Let me know if you can think of a cleaner solution for this problem. I'd be happy to test and contribute a patch if you can guide me through how you think it should work.

Cheers!
Filipe

@nicm @ThomasAdam

filbranden:
This seems to work as a workaround:

```
set -g terminal-overrides "xterm:colors=256:\
setaf=\e[%?%p1%{8}%<%t3%p1%d%e%p1%{16}%<%t9%p1%{8}%-%d%e38;5;%p1%d%;m:\
setab=\e[%?%p1%{8}%<%t4%p1%d%e%p1%{16}%<%t10%p1%{8}%-%d%e48;5;%p1%d%;m"
```

I got the entries from `tput -T xterm-256color setaf` and `setab`.

But it's ugly... Suggestions?

nicm:
Try this please:

```
diff --git a/tty.c b/tty.c
index c4dfde1..7be952c 100644
--- a/tty.c
+++ b/tty.c
@@ -1648,14 +1648,19 @@ tty_try_256(struct tty *tty, u_char colour, const char *type)
 	char	s[32];
 
[...]

PR #121, opened 2015-09-23
Add new terminal-aliases server option

filbranden:
Fixes: #120

Well, in a way, it's still a workaround, but I think it's a more elegant way to enable 256 color support in `TERM=xterm`.

Thanks!
Filipe

@nicm @ThomasAdam

The original commit description follows:

The option adds a mapping between terminal settings in $TERM variable and the actual terminfo entry to use.

It can be more effective than terminal-overrides for tasks such as enabling 256 colour support, since that usually requires setting more than a single terminfo variable to the correct values.

For instance, until tmux 1.8 this recipe used to work to enable 256 colour support on xterm:

```
  set -g terminal-overrides "xterm:colors=256"
```

However, commit f2e54e1e2ff5a1 started using "setaf" and "setab" terminfo settings, which are insufficient in "xterm" to set 256 colours. It is possible to override those, but if the goal is to enable support for 256 colours, it is more effective to do so by instructing tmux to just use the full "xterm-256color" terminfo entry directly, which can now be achieved using:

```
  set -g terminal-aliases "xterm:xterm-256color"
```

Of course, ideally the $TERM setting should be correct to specify "xterm-256color" if that support is available, but unfortunately some terminal emulators are limited (for instance, xfce4-terminal uses termcap which has only a basic set of entries these days) and there are number of systems misconfigured to set TERM=xterm or similar from global configuration files such as /etc/profile, which often makes it more convenient to just override settings in tmux configuration (same as was already done, to quite an extent, by terminal-overrides, anyways.)

Tested on xfce4-terminal, using the "xterm:xterm-256color" mapping, confirming that vim syntax highlighting with 256 colour support was available inside the freshly started tmux session. Man page updated and tested locally.

### DIFF
--- a/examples/tmux.vim
+++ b/examples/tmux.vim
@@ -217,6 +217,7 @@ syn keyword tmuxOptsSet
 	\ status-right-length
 	\ status-utf8
 	\ staus-right-style
+	\ terminal-aliases
 	\ terminal-overrides
 	\ update-environment
 	\ visual-activity
--- a/options-table.c
+++ b/options-table.c
@@ -104,6 +104,11 @@ const struct options_table_entry server_options_table[] = {
 	  .default_num = 1
 	},
 
+	{ .name = "terminal-aliases",
+	  .type = OPTIONS_TABLE_STRING,
+	  .default_str = NULL
+	},
+
 	{ .name = "terminal-overrides",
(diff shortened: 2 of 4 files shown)


filbranden:
Superceded by #122

PR #123, opened 2015-09-24
Fix fill in code to store boolean flags in the .flag member.

filbranden:
@nicm I bumped into this while hacking the terminal-overrides code... It's doesn't really cause a bug but I think it should still be fixed. Thanks!


nicm:
Thanks, applied

PR #125, opened 2015-09-28
for 'choose-tree' add windows sorting by activity time

serge-name:
Tmux lacks a very convenient feature of GNU screen — the ability to sort windows by activity time (windowlist -m). So I made this tiny patch to implement such feature.


ThomasAdam:
Thanks; this idea has already been proposed once, and there's other changes happening in-the-works to make this sort of thing more generic.  I'm not too sure whether this change is worth it at this time.

serge-name:
I asked Nicholas to implement it about two years ago. Anyway, I'd be grateful if that feature will emerge in near future. Either using my patch or any generic and better code. Thank you.

nicm:
choose-tree is going to change and it should have this.

PR #130, opened 2015-10-03
Cache commands executed with #() and default to 1 for status-interval

yath:
As of 379400cfa69f8df9ac13b070c60d5f8a282ddf6e, commands have been run
every second when the status bar is being redrawn (on every event loop).

This commit introduces a ‘command-cache-time’ setting that controls the
caching time of #() commands and reduces status-interval to 1, so clocks
are redrawn in time and not only when the event loop happens to be
executed.

### DIFF
--- a/format.c
+++ b/format.c
@@ -219,6 +219,7 @@ format_job_get(struct format_tree *ft, const char *cmd)
 {
 	struct format_job	fj0, *fj;
 	time_t			t;
+	int			ct;
 
 	fj0.cmd = cmd;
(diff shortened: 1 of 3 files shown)


nicm:
I'm not clear what you are trying to fix here.

If you are saying that commands are running too often - sure, perhaps they shouldn't be run every time the status line is updated, even if it is at most once a second. But I don't want another option to control this, we need to think of some alternatives.

If you are saying that clocks are not being redrawn on time when status-interval is set to 1, then that is a bug and it should be fixed. Setting status-interval to 1 should force a status line redraw every second. However, it seems to work for me. The default for status-interval is not going to be changed.

Issue #161, opened 2015-10-23
Preference for more columns than rows in tiled layout


rkowalick:
I noticed tmux has a tiled window mode, but the layout seems to prefer having more rows over more columns: that is, If I had 6 windows, I would get the layout

```
---------
| 1 | 2 |
---------
| 3 | 4 |
---------
| 5 | 6 |
---------
```
Since most screens are wider than they are tall, I would expect most people would prefer:
```
-------------
| 1 | 2 | 3 |
-------------
| 4 | 5 | 6 |
-------------
```
It is a pretty straightforward change. Thoughts?

dvanallen:
Please reconsider accepting this change.  I would love to use the tmux tiling layout, but this is a deal breaker for me.  I believe most tiling window managers default to preferring vertical splits - I know for sure that dwm and i3 do.  With Retina displays, this is a huge improvement, especially when working with files that have an 80-character line length limit.

PR #162, opened 2015-10-23
Prefer more columns for tiled layouts

rkowalick:
Most screens are wider than they are taller, so we change the tiled
layouts to prefer having more columns that rows. For example, with 6
windows, this change gives 2 rows and 3 columns, while the old code
would have 3 rows and 2 columns.

Resolves https://github.com/tmux/tmux/issues/161

### DIFF
--- a/layout-set.c
+++ b/layout-set.c
@@ -480,9 +480,9 @@ layout_set_tiled(struct window *w)
 	/* How many rows and columns are wanted? */
 	rows = columns = 1;
 	while (rows * columns < n) {
-		rows++;
+		columns++;
 		if (rows * columns < n)
-			columns++;
+			rows++;
 	}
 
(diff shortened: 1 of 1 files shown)


nicm:
I'm not sure I'm convinced, columns are more useful than rows.

rkowalick:
It is certainly a personal preference thing. I'm just not sure what the right default is and was surprised when I used tiled mode and got 'more rows' over 'more columns'. Maybe the best solution is to allow a user to decide which they prefer?

To be honest, I didn't even know the tiled layout existed until accidentally stumbling upon it in the help screen.

nicm:
It has been years and you are the first to mention this, so I don't think an option is particularly desirable, we have too many options already. I think it is fine as it is, I think wider panes with fewer lines are more useful

PR #163, opened 2015-10-24
Add new format 'pane_current_path_basename'

ymattw:
This adds a new format option `pane_current_path_basename` to make `$(basename $(pwd))` available for use, typical usage is in `automatic-rename-format`.

The background: I wanted a way to put both cwd and current command in the auto window title but `pane_current_path` would be too long in most cases.  I know I can print a `\033k` escape sequence to set window title automatically (e.g via the `PROMPT_COMMAND` env for bash) but it does not support display current command.

Now this new format makes the work, for example I can auto rename window title to a fairly short and meaningful string:

```
set -g automatic-rename on
set -g automatic-rename-format \
    '#I#{?pane_in_mode,[tmux],[#{pane_current_path_basename}] #{pane_current_command}}#F'
```

image

### DIFF
--- a/format.c
+++ b/format.c
@@ -51,6 +51,8 @@ void	 format_cb_window_layout(struct format_tree *, struct format_entry *);
 void	 format_cb_start_command(struct format_tree *, struct format_entry *);
 void	 format_cb_current_command(struct format_tree *, struct format_entry *);
 void	 format_cb_current_path(struct format_tree *, struct format_entry *);
+void	 format_cb_current_path_basename(struct format_tree *,
+	     struct format_entry *);
 void	 format_cb_history_bytes(struct format_tree *, struct format_entry *);
 void	 format_cb_pane_tabs(struct format_tree *, struct format_entry *);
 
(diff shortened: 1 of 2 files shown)


ThomasAdam:
Can't you simply do:

```
if (cwd != NULL && *cwd == '/')
    fe->value = xstrdup(basename(cwd));
else
    fe->value = xstrdup(cwd)
```

This is via `<libgen.h>`, of course.

nicm:
I don't like adding special cases like this. Does #(basename #{pane_current_path}) not work?

If not then I think we should generalize this, so that say #{b:pane_current_path} is the basename. I'd also like to have #{t:...} for a time, so we could drop all the _string variants of the time. It should be easy to extend the length limit check to do this too.

ymattw:
@ThomasAdam Thanks for the tip of basename(), I've been away from C language too long :)

@nicm `#(basename #{pane_current_path}) ` gives me `<'basename /home/ymattw/ws/tmux' not ready>`.

I agree something like `#{b:pane_current_path}` is better, is that in your roadmap?  Please free feel to close this once you have answer for above question.

nicm:
http://pastebin.ca/raw/3217215

ymattw:
@nicm :+1: `#{b:pane_current_path}` works fine, thanks.

nicm:
Applied now, thanks.

PR #173, opened 2015-10-28
Stop window-active-style from fallbacking to the value of window-style

icetan:
This patch is to enable the case when you want to have the active pane's background or foreground "unstyled" and the non-active panes in some other style.

Currently setting `window-active-style` to `default` would make the active pane style fallback to whatever the `window-style` was set to.

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -1693,8 +1693,7 @@ tty_default_colours(struct grid_cell *gc, const struct window_pane *wp)
 		if (pgc->fg != 8 || (pgc->flags & GRID_FLAG_FG256)) {
 			gc->fg = pgc->fg;
 			gc->flags |= (pgc->flags & GRID_FLAG_FG256);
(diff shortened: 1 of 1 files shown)


nicm:
All styles in tmux work like this and this one is not going to be an exception.

icetan:
Is there an alternative solution to my problem then?

nicm:
Set the active style to an actual colour not to default?

nicm:
I think I would accept adding a new colour to mean "real default, no inheritance" (I can't think of a new name).

icetan:
Something like that, I was actually looking for that feature first. Maybe `clear` or `none`?

nicm:
None is already in use I think, try clear for now

-------- Original message --------

icetan:
Cool, I'll give it a shot.

PR #174, opened 2015-10-29
client_tty_short and pane_tty_short formats

tgyurci:
This change adds client_tty_short and pane_tty_short format. The /dev/tty prefix is redundant and takes valueable space if the tty is in the status line for example.
I'm not sure if the helper shorten_tty function is in the right place.


nicm:
You should be able to do this with the b: prefix

-------- Original message --------

tgyurci:
The b: prefix strips the /dev/ prefix, but does not strip /dev/tty.

nicm:
Can't you do it with #() and sed? 

-------- Original message --------

tgyurci:
It wors with ```#(...)``` and sed, but I thought it is cleaner to have a separate format variable and spare some process forks.

If you do not prefer to add two new variables, whould you consider to have a new prefix, for example ```l:```? I know that this can no be used widely, but the ```b:``` prefix is essentially a shortcut for ```$(basename)```.

nicm:
I don't think this is a common enough requirement to warrant extra code when sed works fine. If you want to generalize it and add support for substitution eg s/foo/bar:some_format then that would be much better.

-------- Original message --------

tgyurci:
Sorry for the late responding. The s/foo/bar would be a nice feature, and I see that you have made it. Thanks.

PR #185, opened 2015-11-09
Add option to use double-line characters for pane borders

ljonsson:
Single line pane borders sometime appear a bit thin.


nicm:
I'm not sure we want this at all but if we do, it will need to work or
fallback sanely on terminals without UTF-8. Also the option will need to
be CHOICE option, not a string.

PR #210, opened 2015-11-23
Update vim syntax with current commands and options

tgyurci:


nicm:
Why this?

-       \ confirm[-before]
+       \ confirm
+       \ confirm-before

tgyurci:
I thought that it's easier to maintain when all of the list items are expanded and there are short commands anyway which are not the exact prefix of their longer ones (swap-pane and swapp for example). If you think, that the old list style is better, then I willl change back it.

While writing this I had an idea: what about like this:
```
  \ bind[-key]
  \ break-pane breakp
  \ ...
```

That way the same commands are grouped together.

nicm:
Alright I suppose it is fine, applied, thanks.

PR #218, opened 2015-11-27
Add window status formats for activity and bell

ljonsson:


nicm:
No you can do this already with #{?}

PR #221, opened 2015-12-01
Fix segfault in log_debug call

mgraczyk:
When` window_get_active_at` returns `NULL`, `wp` is `NULL`, so `log_debug(... wp->id)` causes a segfault.
This crash started occurring when I began using a 4K monitor.


nicm:
Applied thanks.

PR #235, opened 2015-12-11
Looks for ncurses with pkg_config

dupgit:
On my fresh installed debian jessie 8.2 ncurses.h is not installed and configure script did said nothing. The compilation aborted with message "unable to found ncurses.h". That patch corrects this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,10 +155,16 @@ PKG_CHECK_MODULES(
         [
 		AC_SEARCH_LIBS(
 			setupterm,
-			[terminfo curses ncurses tinfo],
+			[terminfo curses tinfo],
 			found_curses=yes,
 			found_curses=no
 		)
+		AC_SEARCH_LIBS(
+                        addch,
+                        [ncurses],
+                        found_curses=yes,
+                        found_curses=no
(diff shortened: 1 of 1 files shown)


nicm:
Thanks, but this will fail if ncurses is present but not available through pkg-config, which is common. The check needs to fall back to the existing AC_SEARCH_LIBS (setupterm...) like the libevent check does.

dupgit:
This patch should now work better:
1) pkg-config search for ncurses
2) if it was not found by pkg-config it searchs with AC_SEARCH_LIBS 
2.1) setupterm function into terminfo curses tinfo libraries and
2.2) addch function into ncurses

On my system, without libncurses-dev installed this generates an error. Previously it didn't because AC_SEARCH_LIBS did found setupterm into tinfo and returned true.

nicm:
tmux does not use addch, does it not build with libtinfo?

nicm:
Oh it needs curses.h of course.

nicm:
Can you send me one patch with all your changes together please?

dupgit:
See Pull request #236. 
Thanks.

PR #236, opened 2015-12-11
Look for ncurses with pkg-config and falls back to AC_SEARCH_LIBS (total patch)

dupgit:
ncurses.h may be needed by tty-term.c.
This patch should do the following:
 1) pkg-config search for ncurses
 2) if it was not found by pkg-config it searchs functions into
    librairies with AC_SEARCH_LIBS to confirm that they really exists:
    2.1) it searches setupterm function into terminfo curses tinfo libraries and
    2.2) then it searches addch function into ncurses.

On my system, without libncurses-dev installed this generates an error.
Previously it didn't because AC_SEARCH_LIBS did found setupterm into tinfo
and returned true.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -143,15 +143,32 @@ if test "x$found_libevent" = xno; then
 	AC_MSG_ERROR("libevent not found")
 fi
 
-# Look for curses.
-AC_SEARCH_LIBS(
-	setupterm,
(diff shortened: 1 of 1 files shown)


nicm:
I have applied this but left out the addch check and just removed libtinfo because I don't know what it was for (libterminfo is for NetBSD).

PR #243, opened 2015-12-19
tmux.vim: sync with latest changes

mathstuf:


nicm:
Please take this to https://github.com/keith/tmux.vim

PR #246, opened 2015-12-24
Add bottom-border option

jibi:
This patch allows to specify the `bottom-border` option to display a (just esthetic) bottom border line between the status bar and the window panes.

This is without the `bottom-border` option:

2015-12-24-120703_884x604_scrot

and this is with:

2015-12-24-120723_884x604_scrot

### DIFF
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -197,6 +197,8 @@ cmd_new_session_exec(struct cmd *self, struct cmd_q *cmdq)
 	}
 	if (sy > 0 && options_get_number(global_s_options, "status"))
 		sy--;
+	if (sy > 0 && options_get_number(&global_s_options, "bottom-border"))
+		sy--;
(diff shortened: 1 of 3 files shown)


ThomasAdam:
Hi,

I really don't think this is needed. Does it also work with status position
set to top?

Thomas

jibi:
Hi,

> I really don't think this is needed.

Agree. This is not something that would be widely used, but I know some pepole who are actually using this now, and so I decided to open a PR instead of keeping rebasing my patch with each new release.

> Does it also work with status position set to top?

No, I wasn't even aware of this, but if you are interested in merging the patch I can fix this (maybe move from `bottom-border` to `statusline-border`)

nicm:
Thanks, but I don't see this as particularly useful, we have too many style options and whatnot already and I don't think this adds much value.

PR #247, opened 2015-12-25
Don't require space after ;.

eksoderstrom:
Create a new argv element whenever we parse a ';' preceded by a non-NULL string.


eksoderstrom:
Actually, this only correctly handles <command>;<command> and not <command ;<command>.
I'll close this pull request, and resubmit a more complete commit.

PR #248, opened 2015-12-25
Don't require space after ;

eksoderstrom:
When parsing a command string, treat ';' as a delimiter between commands rather than relying on whitespace as a delimiter. This allows commands of the form \<command\>;\<command\> and \<command\> ;\<command\>

### DIFF
--- a/TODO
+++ b/TODO
@@ -9,7 +9,6 @@
           config file as-is
 
 - make command sequences more usable
-	* don't require space after ;
 	* options for error handling: && and ||?
 
 - options bits and pieces:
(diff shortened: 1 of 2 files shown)


nicm:
Well, sort of:

$ ./tmux ls\;ls
unknown command: ls;ls

Someone tried to do this before but there was a problem with it... I can't find the emails so I don't know what it was. I think they tried to do it differently than you do though.

nicm:
I took a quick look at doing this for commands from outside tmux too, but it is kind of complicated to parse, so I don't think we should do this inside either. Thanks though!

PR #252, opened 2015-12-31
Status fix

nviennot:
Two commits:

> Fix message log appends

> The following has been improved:
> 1) Previous code had "u_int first" which is incorrect as
>    "first = c->message_next - limit" would result in negative values,
>    resulting in dropping the first `limit` messages.
> 2) Avoid to traverse the entire list of message to prune messages.
> 3) Extracted it into a global function as it can be useful for other part of
>    tmux (and tmate).

and

> Guard against NULL c->session in status_message_set()

### DIFF
--- a/status.c
+++ b/status.c
@@ -547,7 +547,7 @@ status_message_set(struct client *c, const char *fmt, ...)
 	struct message_entry	*msg, *msg1;
 	va_list			 ap;
 	int			 delay;
-	u_int			 first, limit;
+	u_int			 limit;
 
 	limit = options_get_number(global_options, "message-limit");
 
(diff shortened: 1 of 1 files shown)


nicm:
I don't think we need a helper function here at this point.

Why do you need a guard for c->session, where is this being called from with a NULL session?

nviennot:
I've amended the PR:
1) I removed the helper function, and just fixed whatever needed to be fixed
2) removed the c->session guard. Wasn't sure where it's really needed.

nicm:
The current code works correctly, "first" can never be negative, because it is unsigned. If limit > next, it will wrap and become very large, so msg_num will not be >= first.

However, I think your change is better, and we can break out of the loop early, so I will apply it.

nicm:
Thanks!

nviennot:
old code wasn't correct. try running an alert like "Activity in window 1", and then run tmux show-messages. You'll see nothing.
> it will wrap and become very large, so msg_num will not be >= first.

Yes, and so the messages get deleted

have a good one :)
happy 2016!

PR #259, opened 2016-01-03
Bash completion: dynamic list

przepompownia:
It allows to keep list of commands up to date (please compare it with the old static list). It would be good to add `list-commands` option `-s|--short` to display only commands without aliases and description column. Filtering by `cut` would be then redundant.

### DIFF
--- a/examples/bash_completion_tmux.sh
+++ b/examples/bash_completion_tmux.sh
@@ -2,104 +2,16 @@
 # This file is in the public domain
 # See: http://www.debian-administration.org/articles/317 for how to write more.
 # Usage: Put "source bash_completion_tmux.sh" into your .bashrc
-_tmux() 
+_tmux()
 {
-    local cur prev opts
-    COMPREPLY=()
-    cur="${COMP_WORDS[COMP_CWORD]}"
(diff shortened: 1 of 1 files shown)


ThomasAdam:
It'd be even better if this file was just removed, and we allowed upstream bash to handle this for us.

przepompownia:
I think that getting rid of control over something specific for tmux is not a good idea. There is a lot of packages that provide own completions without being depend on upstream (if such does exist) and distribution-wide centralized package. Example from my Debian:

     $ apt-file search /usr/share/bash-completion/completions/ | grep -v ^bash-completion: | cut -d: -f1 | sort -u | wc -l
    183

nicm:
Can you maintain this script yourself and we will link to it from README like we do for the vim syntax file?

przepompownia:
https://raw.githubusercontent.com/przepompownia/tmux-bash-completion/master/completions/tmux
https://raw.githubusercontent.com/przepompownia/tmux-bash-completion/tmux_version_up_to_2.1/completions/tmux

Shorten, completion-oriented `list-commands` output is still welcome.

nicm:
Great, thanks

PR #261, opened 2016-01-05
some warnings for Apple LLVM version 7.0.2 (clang-700.1.81) fixed

andrey-starodubtsev:


ThomasAdam:
Hi,

Thanks, but I'm not so sure we need any of this.  The compat/* changes can definitely be ignored; they'll have to be changed in OpenBSD, as I don't want to keep patches against them when syncing.

As with the printflike(), I'm not sure it's with it, and it's definitely not worth changing asprintf() to vasprintf()

andrey-starodubtsev:
I added `printflike` macros the same way as functions with variadic params have it. As https://gcc.gnu.org/onlinedocs/gcc-4.0.2/gcc/Function-Attributes.html says,

    For functions where the arguments are not available to be checked (such as vprintf), specify the third parameter as zero.

Warnings in compat are fixed the same way as they did in https://github.com/sspans/ladvd/commit/3cc2b92f4f57b959eac9b9c5d24ad43e4f0e2302

Anyway, fill free to cherry-pick anything from patch to reduce compiler clutter or let me know and I'll strip questionable changes.

nicm:
Hi

- I don't see much point in printflike for va_list functions so I am leaving it out. Does clang really warn about this?
- xmalloc.[ch] and imsg*.[ch] come from OpenBSD mostly unchanged so I'd like to leave them, I don't think these changes are appropriate to try and push upstream.
- fatalx is marked  __noreturn__ so a warning for session_group_index seems like a compiler bug, assuming it supports __noreturn__.
- I applied the osdep-darwin.c __unused check.
- I will apply the cfg_add_cause and input_reply printflike bits.
- I do not know what you are trying to fix in log.c?

Thanks

andrey-starodubtsev:
> I don't see much point in printflike for va_list functions so I am leaving it out. Does clang really warn about this?

yes, it does:

    xmalloc.c:102:21: warning: format string is not a string literal [-Wformat-nonliteral]
            i = vasprintf(ret, fmt, ap);
                              ^~~
    xmalloc.c:131:26: warning: format string is not a string literal [-Wformat-nonliteral]
            i = vsnprintf(str, len, fmt, ap);
                                    ^~~
    /usr/include/secure/_stdio.h:75:63: note: expanded from macro 'vsnprintf'
      __builtin___vsnprintf_chk (str, len, 0, __darwin_obsz(str), format, ap)
                                                                  ^
    2 warnings generated.

> I do not know what you are trying to fix in log.c?

It's the same warning about non-literal format for printf-like functions:

    log.c:99:22: warning: format string is not a string literal [-Wformat-nonliteral]
            if (vasprintf(&fmt, msg, ap) == -1)
                                ^~~
    1 warning generated.

static `log_vwrite` is used from `fatal` and `fatalx`, so if strerror(errno) called in `fatal` happens to have placeholder, stack overrun would happen. It's unlikely, I know, but it proves compiler warnings can be useful sometimes =)

Issue #147, opened 2015-10-19
2.1 set mouse to `copy-selection -x`.


NHDaly:
Hi, I cannot seem to get the `copy-selection -x` command to work for overriding the mouse's exit-copy-mode-on-selection behavior.

I tried this, but it wasn't able to change the behavior of click-drag selecting some text:

```
bind -n -tvi-copy  MouseUp1Pane copy-selection -x
bind -n -tvi-copy MouseDrag1Pane copy-selection -x
```

Just to be sure, i tried setting these to a different command:
```
bind -n -tvi-copy  MouseUp1Pane scroll-up
bind -n -tvi-copy MouseDrag1Pane scroll-up
```
and click-drag-release *still* performs the same behavior: copy and exit copy-mode. Simply clicking without dragging will do execute the `scroll-up`.

So I think there is a problem with not being able to override the MouseDrag1Pane inputs.

nicm:
Bindings in vi-copy and emacs-copy only work inside copy mode, you need to bind the keys outside copy mode. But there is no way to do what you want.

Issue #140, opened 2015-10-12
Option to disable "release mouse click to copy selection & exit copy-mode" behavior.


NHDaly:
I don't find this behavior desirable. Often, i am simply scrolling with my mouse and highlight something out of habit just to show someone, and I exit copy-mode. Also, I would like to use plugins like [tmux-yank](github.com/tmux-plugins/tmux-yank) to copy the highlighted text to my OS instead of just tmux's paste buffer. This is difficult with the current behavior:

https://github.com/tmux-plugins/tmux-yank/issues/24
https://github.com/tmux-plugins/tmux-yank/issues/28

It would be nice to have an option to disable this behavior. Ideally something like `set-option -g mouse-release-copies-selection off`, which could be set to `on` by default.

I'm not sure, but I *think* such an option would go here in the code?:
https://github.com/tmux/tmux/blob/4afe26fa82fd2701cbf686799668d685645b220e/window-copy.c#L2287

Thanks!

NHDaly:
Oh, I see that after commit bf635e7 the mouse-support options have changed. Is this already addressed by those changes? Thanks.

nicm:
No I like this behaviour.

NHDaly:
I'm sorry, no, I wasn't asking the default to change, just that there be *an option to change the behavior*. We don't *all* like that behavior.

NHDaly:
Will I be able to disable it by setting something like `bind -n   MouseUp1Pane copy-mode <space>` or something like that?

freitass:
I would use that option myself! Often times I end up in the following scenario:

1. Scroll up to analyse a huge log
2. Switch to another window (i.e. the browser)
3. (Accidentally) click on the terminal window to set it focused, leaving copy-mode and losing the point of the log where I was

twof:
+1 for this option.

spazm:
+1 for an alternate behavior/option.

  I use mouse select for highlighting as well as copying.   As a workaround I'll be disabling mouse mode.

sblask:
I have mouse mode disabled because of this behaviour and it would be great to have an option to disable it. I can see how it is useful if you want to copy something from the current screen, but if you have a huge log, try to find something and suddenly you click and lose the position, it's more of an annoyance. Also, not sure what I did, but for me dragging does not even copy, I just lose the position which totally looks like a bug...

ramn:
+1

maikelmeyers:
[...]

PR #273, opened 2016-01-20
Make mouse drag end behavior configurable

coderstephen:
This is a pull request to make the behavior of mouse dragging for copy mode configurable. As indicated by requests like #140 and #147, this would be a handy feature in the spirit of making everything in tmux scriptable. Currently, releasing a mouse button after dragging ends the drag event with no further action, and in the case of copy mode, copies the selected bytes into a buffer. This is a logical action, but not customizable. This patch allows this behavior to be customized by rebinding `MouseUp` keys.

**How it works**: A lot of users that have asked for this capability assumed that binding `MouseUpXPane` to something else would control what happens when dragging completes, when in reality such a key is not triggered on release. This patch changes dragging slightly, so that a corresponding `MouseUp` key is always triggered after ending a mouse drag.

**Regressions**: Probably some, but only one that I have noticed -- the behavior of binding to a `MouseUp` key is sort of dual and may or may not be unexpected behavior to users.

Note that default copy mode bindings are added so that `copy-selection` remains the default action when dragging, in line with the current behavior.

I'm not 100% confident in the technical side of this patch, but it is meant to show that it can be done and how. Constructive criticism is very welcome. Also, this is my first contribution to tmux, so everything I say above could be completely wrong. :wink:


nicm:
I like this. It would be possible to make it a new key (MOUSEDRAGEND1?) but I don't know if there is any advantage.

coderstephen:
@NHDaly I already did set `copy-selection` to the default behavior. :wink: See [mode-key.c#L350](https://github.com/coderstephen/tmux/blob/707b2357b552a26f86b3c895ae313c25b50edf18/mode-key.c#L350) and [mode-key.c#L499](https://github.com/coderstephen/tmux/blob/707b2357b552a26f86b3c895ae313c25b50edf18/mode-key.c#L499). By default `copy-selection` exits copy mode as well.

coderstephen:
@nicm I thought about making it a new key, something like `MouseDragEnd`, and that could easily be done, but as far as I know, `MouseUp` isn't being used all that much since `MouseDown` in general seems to be the one commonly used for handling "click" events.

I suppose, in a single click, it begins with a `Down` event, and ends with an `Up` event, whereas a drag begins by dragging the mouse around with the button down, and ends also whenever the mouse is `Up` as well. Either way seems reasonable though.

nicm:
I have applied this to my next branch with one change (window_copy_drag_release does nothing now, so it can be removed). It will be in master when the OpenBSD tree unlocks and I can apply it there. Thanks!

coderstephen:
Nice! Glad to contribute!

cmh:
@coderstephen Thanks for this contribution! It lets me finally replicate linux "middle mouse to copy" behaviour with `bind-key -t vi-copy -T root MouseUp1Pane copy-pipe "xclip -i -selection primary"`

nicm:
I think I am going to change this to MouseDragEndX because it is trivial to do and there is a use for binding Up and DragEnd differently in copy mode.

coderstephen:
Sounds good to me.

NHDaly:
Hi, what's the status on this? Was it released in `tmux 2.2`? :) Thanks!

+@jvanderpol who was asking me about it.

PR #277, opened 2016-01-22
Fix #188: sid's cwd overwrite pgrp's cwd

### LINKED ISSUE #188: split pane switch to pane_current_path
In 1.9 Version,
`bind - splitw -v -c "#{pane_current_path}"`
will split pane with path of current pane.

But in 2.1 version.
when run command in current pane, liake "man man". splitw will change to `/usr/share/man` path.

This is annoying. How to config it? or rollback?
Thanks!

liuyang1:


nicm:
What was the original issue for this?

liuyang1:
0) original code get path from "/proc/[PID]/cwd"; then change to "/proc/[PGRP]/cwd" as this commit.

Use pgrp of pty fd not pid of immediate child when recovering current working directory (like current process). From Marcel Partap.

1) then change to try "/proc/[SID]/cwd" when get "/proc/[PGRP/cwd" fail. as this commit.

If pgrp fails in osdep_get_cwd, try sid. Fixes eg cat foo|less. From Balazs Kezes.

2) I find cwd of PGRP not giving epxect path at  https://github.com/tmux/tmux/issues/188
but cwd os SID could. so make this change.

nicm:
Sorry, I think the current code is right, if I do:

(cd /; emacs)

pane_current_path should return /. Your change breaks this. It is unfortunate that man(1) changes the working directory but that is just the way it works.

PR #293, opened 2016-02-01
Incremental search

lukaspiatkowski:
The first diff is a refactor of existing search logic, it should not introduce any changes.
The second diff adds the incremental search.

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -47,8 +47,16 @@ int	window_copy_search_lr(struct grid *, struct grid *, u_int *, u_int,
 	    u_int, u_int, int);
 int	window_copy_search_rl(struct grid *, struct grid *, u_int *, u_int,
 	    u_int, u_int, int);
-void	window_copy_search_up(struct window_pane *, const char *);
-void	window_copy_search_down(struct window_pane *, const char *);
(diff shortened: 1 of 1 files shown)


netei:
Great feature !

nicm:
Ok I already made some changes to your first diff so rather than doing that again, I have just updated my changed diff to apply again. Please look at this [window-copy.c.diff.txt](https://github.com/tmux/tmux/files/242517/window-copy.c.diff.txt).

lukaspiatkowski:
@nicm Your changes looks good.

The only thing I would suggest is to leave some of the comments I wanted to add. I think that the "moveflag" or "direction" function parameters require reading through the code in order to understand them.

Other than that, I think you can merge this and I can rebase the "Incremental search in window copy mode" on top of your changes

nicm:
Ok, I have applied this, with a comment added back before
window_copy_search as well. It will be in GitHub next time it syncs, I
think it is every hour or so.

Can you please close one of the two PRs and post your new changes to the
other?

Cheers

nicm:
This is not moving forward and the code is going to change a bit soon anyway.

PR #297, opened 2016-02-06
Add xmalloc.h and new merged example file to EXTRA_DIST

orebokech:
make dist fails because it tries to include the old examples directory,
so use the merged example file instead.

Also include xmalloc.h which is required to build.


ThomasAdam:
OK.

PR #303, opened 2016-02-10
Added support for two (top and bottom) status bars...

tonyfischetti:
It works great for me. The only things that need to be done (unless I'm missing something) is configuration constraints (can't have the main bar and the aux bar in the same position) and documentation on the new feature.

Basically, the new bar, `aux-status`, is just like the main bar except all of it's configuration options are prefaced with `aux`. Also, it doesn't contain any of the information about the windows (like the main bar).

To make things more concrete, this is an excerpt from my `.tmux.conf`
```
set-option -g aux-status-left-length 100
set-option -g aux-status-right-length 100
set -g aux-status-style "bg=#989acc, fg=#fda878"

set-option -g status-position bottom
set-option -g aux-status-position top

set -g aux-status on
```

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -914,7 +914,7 @@ server_client_check_redraw(struct client *c)
 		else if (c->prompt_string != NULL)
 			redraw = status_prompt_redraw(c);
 		else{
-			aux_status_redraw(c);
+			redraw = aux_status_redraw(c);
 			redraw = status_redraw(c);
         }
 		if (!redraw)
--- a/server-client.c
+++ b/server-client.c
@@ -132,6 +132,7 @@ server_client_create(int fd)
 	c->tty.sy = 24;
 
 	screen_init(&c->status, c->tty.sx, 1, 0);
+	screen_init(&c->aux_status, c->tty.sx, 1, 0);
 
 	c->message_string = NULL;
 	TAILQ_INIT(&c->message_log);
@@ -331,7 +332,12 @@ server_client_check_mouse(struct client *c)
 	m->s = s->id;
 
 	/* Is this on the status line? */
+    /* it doesn't matter if it's on the aux status line */
+    /* but we need to store if there's anything on line 0 */
 	m->statusat = status_at_line(c);
+	m->auxstatusat = aux_status_at_line(c);
+    m->anyatzero = any_status_at_zero(c);
+    m->anyonlast = any_status_on_last(c);
 	if (m->statusat != -1 && y == (u_int)m->statusat) {
 		w = status_get_window_at(c, x);
 		if (w == NULL)
@@ -343,10 +349,10 @@ server_client_check_mouse(struct client *c)
 
 	/* Not on status line. Adjust position and check for border or pane. */
 	if (where == NOWHERE) {
-		if (m->statusat == 0 && y > 0)
+		if (m->anyatzero && y > 0)
 			y--;
-		else if (m->statusat > 0 && y >= (u_int)m->statusat)
-			y = m->statusat - 1;
+		if (m->anyonlast && y >= m->auxstatusat + m->statusat)
+			y = m->auxstatusat + m->statusat - 1;
 
 		TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
 			if ((wp->xoff + wp->sx == x &&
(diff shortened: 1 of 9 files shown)


ThomasAdam:
Hi,

Hmm.  Thanks for your patch, but I think this is going to need a lot more work.  I'm seeing a few warning during compilation you should address.  The indentation is screwed as well---please refer to OpenBSD's style(9) for cribbing; if nothing else, try and stick to the style of the surrounding code.  Stop mixing spaces and tabs.

As to the code itself, I do not think you need to have any of the aux-* settings.  Rather, the status lines should be numbered, or something, and then you can use the existing status-* options, without introducing everything again.  So for example:

`set -g status -1 ....`
`set -g statis -2 ...`

I'm sure you get the idea.  Just what shape those options should take, I'm not sure.

Setting:  `set -g aux-status-position bottom` having already had `aux-status-position` set to `top`, makes a mess of the grid_layout, and tmux is then confused as to the sx/sy properties of the grid.

I'd say this will need a lot of discussion before any more code is written.

nicm:
I am not convinced this is something we particularly need.

tonyfischetti:
Well, I spend a lot of time in terminals that are 80 columns wide. In order to fit all the information I want regarding my session name, which machine I'm on, and all my windows, 2 status bars is something I could definitely use. I guarantee there are many others who feel the same way

nicm:
I think I would prefer it if the existing status line could be multiple lines rather than a second status line entirely.

tonyfischetti:
That'd be nice, too. I mean it's up to you. All I can tell you is that there's an aesthetic benefit to having your work flanked by two bars on either side and that virtually everyone I talked to about two bars were happy to hear its possible now or excited because they "didn't know they needed it"

nicm:
It is simple enough to make the status line multiple lines; the question is how is it configured?

Do we limit it to two lines and just have an option with whatever the user wants to put in the second line? That would be simple. They can use #[] to get colours and whatnot.

If we want to make it an arbitrary number of lines then how do the options look?

There is no way we are copying all the existing status options even once more.

kylescottmcgill:
Because there is still no +1 button, I would like to +1 one it and provide a suggestion.

Both Dual Bars, and Multiline Bars could be done in a single swoop. Just allow the Secondary bar to place itself under Primary bar. Bam Multiline... kinda.

Opinion:
I am also one of the people who said I "needed" this, my reasoning/opinion is it gives me more space to work with, I currently run 2bwm tiling window manager with a bar at the top of my screen. If tmux could compensate by having two bars, I can dedicate a single bar to having OS Related stats/bars/etc and the secondary or bottom bar to having tmux related items, such as current window, session name etc.

Thomas, just wanted to say I have a hard time working/playing without tmux, thank you; and to all the contributors as well, you all make it a pleasure being a Sys Admin!

nicm:
This [status-text.diff.txt](https://github.com/tmux/tmux/files/140755/status-text.diff.txt) gives us a two line status line with the text on the second line specified by a single new option status-text

tonyfischetti:
Here are more +1's

hut8:
I would *really* like this feature. I want one on the top and one on the bottom. My `c` game is pretty rusty but I agree that simply duplicating the options for the existing line is not a great way to go about it (like @nicm said).

marcaddeo:
In the original patch, it seems that if the normal status line doesn't have any changes then the auxiliary line does not get updated. For instance, if I put the current time in the auxiliary line it won't be updated until something in the normal status line has updated.

nicm:
Here is updated diff against master, you need to set status-text to something: 
[tmux-two-status.diff.txt](https://github.com/tmux/tmux/files/242888/tmux-two-status.diff.txt)

YPetremann:
For me two statusbar or multiline status bar are both good idea :
- Multiple statusbar could be needed for example for peoples who want a main statusbar at top with windows, session name, user and domain, and the secondary one at the bottom, which could display some other informations like networking, battery ...

- Multiline statusbar work a little like more statusbar, one of the major advantage is that we could add a break between "applets" and so it could automatically wrap lines to put more applets. it could be also possible to force a new-line.

server-client.c
```diff
@@ -914,7 +914,7 @@ server_client_check_redraw(struct client *c)
 		else if (c->prompt_string != NULL)
 			redraw = status_prompt_redraw(c);
 		else{
-			aux_status_redraw(c);
+			redraw = aux_status_redraw(c);
```
hut8: How could this fix the issue? That variable is always overwritten immediately on the next line, right?

nicm:
This doesn't seem to be going anywhere so I am going to close this issue, it can be revisited if there is any movement.

PR #306, opened 2016-02-15
Improved utf8_width() function.

euske:
Hello, I use tmux in the CJK settings, and I've noticed that the widths of some characters aren't correctly handled. As you know, in a CJK setup, some characters are handled as a "wide" character that takes twice as much as other characters. tmux already supports this mechanism, but to make this work throughout the whole terminal, you need to have every program running on a terminal agree on their character widths. This means that the terminal emulator, fonts, tmux and a program running inside tmux (e.g. vi) all have to use the exactly same character width, for every character. There's an official width table for the Unicode characters now provided by the Unicode consortium, and I found that the width tables used in tmux is slightly off. This patch corrects it by using the table from EastAsianWidth.txt file which is now commonly used by xterm or glibc. In this process, I also found that the current implementation of utf8_width() which uses a balanced tree not very intuitive, I changed it to a simple binary search, eliminating the needs of "left" and "right" fields in the struct utf8_width_entry.


nicm:
There is no way for me to tell which characters you have changed so this
diff is useless. Please forget fiddling with the search algorithm and just
change the characters you think are wrong.

euske:
The width list is automatically generated from the Unicode database, so I can't exactly point out which character widths are wrong. Here's the Python script that generates the list (rename it to gencwidth.py)

[gencwidth.txt](https://github.com/tmux/tmux/files/133843/gencwidth.txt)

nicm:
The current list was generated from the tables in OpenBSD so I am not going
to just blindly change it all, you will need to figure out which ranges are
wrong and then I can investigate.

euske:
Here's the list of ranges that disagree with the UCD. 
[width_diff.txt](https://github.com/tmux/tmux/files/133994/width_diff.txt)

I was referring to the results of wcwidth() in glibc. Actually, their widths still don't agree 100% with the Unicode database, but I feel it's "more correct" than the current tmux widths (for example, U+3000 and U+FF21 is notably wrong there). Again, if you look at the Python code I posted, you'll see the logic behind each value (I made it as explicit as possible). Thanks for taking your time.

nicm:
Thanks.

0x0000ad...0x0000ad: ucd_width=0, tmux_width=1

This is soft hyphen and is deliberately 1.

0x0f0000...0x0ffffd: ucd_width=1, tmux_width=0
0x100000...0x10fffd: ucd_width=1, tmux_width=0

These are additional PUAs, probably better off as 1 like the BMP PUA.

I'll look at the rest another time.

euske:
FYI, here are the rules that I figured out and implemented in the generator program. (Again, I was referring to the glibc wcwidth() values.)

For each character:
- 1. Look up UnicodeData.txt. If the character is categorized either as "Mn" (Mark, Non-Spacing), "Me" (Mark, Enclosing) or "Cf" (Other, Format), the character has zero width.
- 1-a. However, some characters (listed in NONZERO_CHARS) are excluded from this rule and not assigned with zero width.
- 2. If the character is not assigned with a width, look up EastAsianWidth.txt. If the character is marked as "W" (Wide Width) or "F" (Full Width), the character has width 2.
- 2-a. However, some ranges marked as reserved (listed in RESERVED_CHARS) are excluded from this rule. (IMO, this shouldn't matter anyway.)
- 3. If the character is still not assigned with a width, the character has width 1.

Hope this helps.

nicm:
Thanks

nicm:
Ok I have talked to the OpenBSD developer who is working on UTF-8 and:

0x000cbf...0x000cbf: ucd_width=1, tmux_width=0
0x00302a...0x00302d: ucd_width=2, tmux_width=0
0x003099...0x00309a: ucd_width=2, tmux_width=0

These are Mn so width 0 is correct. We think glibc is wrong here, and
for 0CC6.

0x001160...0x0011ff: ucd_width=1, tmux_width=2
0x003000...0x003000: ucd_width=2, tmux_width=1
0x00ff01...0x00ff60: ucd_width=2, tmux_width=1
0x00ffe0...0x00ffe6: ucd_width=2, tmux_width=1

We think you are correct here, but I need to confirm with someone else.

0x00d800...0x00dfff: ucd_width=1, tmux_width=0

These are surrogates so they are not really valid. I expect 1 is
probably more useful than 0.

0x0111ca...0x0111cc: ucd_width=0, tmux_width=1
0x011300...0x011300: ucd_width=0, tmux_width=1
0x0115dc...0x0115dd: ucd_width=0, tmux_width=1
0x01171d...0x01171f: ucd_width=0, tmux_width=1
0x011722...0x011725: ucd_width=0, tmux_width=1
0x011727...0x01172b: ucd_width=0, tmux_width=1
0x01da00...0x01da36: ucd_width=0, tmux_width=1
0x01da3b...0x01da6c: ucd_width=0, tmux_width=1
0x01da75...0x01da75: ucd_width=0, tmux_width=1
0x01da84...0x01da84: ucd_width=0, tmux_width=1
0x01da9b...0x01da9f: ucd_width=0, tmux_width=1
0x01daa1...0x01daaf: ucd_width=0, tmux_width=1
0x02a6d7...0x02a6ff: ucd_width=2, tmux_width=1
0x02b735...0x02b73f: ucd_width=2, tmux_width=1
0x02b81e...0x02b81f: ucd_width=2, tmux_width=1
0x02fa1e...0x02fffd: ucd_width=2, tmux_width=1
0x030000...0x03fffd: ucd_width=2, tmux_width=1
0x02b820...0x02f7ff: ucd_width=2, tmux_width=1

We think you are correct for these; they are newer than OpenBSD's
Unicode tables (which come from the Perl version we use).

euske:
> We think you are correct here, but I need to confirm with someone else.

I was referring to the Unicode standard, which boils down to these two documents in this case:

- http://www.unicode.org/reports/tr11/
- http://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt

I don't know the reason why some glibc widths departed from this, but I'd be happy if each application agree with the widths.

nicm:
I looks like many of the differences are due to tmux using Unicode 7 and
these files are Unicode 8.

However, I think I'm going to change tmux to use the system wcwidth(),
probably shortly after the next release.

euske:
> I looks like many of the differences are due to tmux using Unicode 7 and these files are Unicode 8. 

Um, actually, no. I don't want to argue with you, but the character widths haven't been much changed since Unicode 3.0 back in 2001. So there should be some other reason for the discrepancy.

 - http://www.unicode.org/Public/3.0-Update/EastAsianWidth-3.txt
 - http://www.unicode.org/Public/7.0.0/ucd/EastAsianWidth.txt
 - http://www.unicode.org/Public/8.0.0/ucd/EastAsianWidth.txt

> However, I think I'm going to change tmux to use the system wcwidth(),

That would be a better solution. (Originally I was going to suggest it, but I thought you avoided using wcwidth for some reason.)

nicm:
Widths haven't changed but they have added new codepoints...

nicm:
You can try this if you like, but it won't be applied for a while (until after the next release): [tmux.wcwidth.diff.txt](https://github.com/tmux/tmux/files/136794/tmux.wcwidth.diff.txt)

nicm:
We will change to s

PR #336, opened 2016-03-17
New window option 'utmp-inhibit'.

koniu:
This allows the user to opt out of having utmp entries written for
each pane. Can help prevent clobbering of utmp records on busy shell
servers.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -710,6 +710,12 @@ const struct options_table_entry options_table[] = {
 	  .default_num = 0
 	},
 
+	{ .name = "utmp-inhibit",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_WINDOW,
+	  .default_num = 0
+	},
+
 	{ .name = "window-active-style",
 	  .type = OPTIONS_TABLE_STYLE,
 	  .scope = OPTIONS_TABLE_WINDOW,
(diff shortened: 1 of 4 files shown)


nicm:
I would prefer a configure option to build without utempter.

koniu:
That's an option of course but a config option would be nice for the lazy folk who, like me, like to get their software from distro packages. 

Eg. Debian now has tmux built with utempter (tho it's nowhere to be seen in build-deps) and since I pulled v2.1 from jessie-backports to replace v1.9 no-one knows anymore who's actually logged in because 'who' output is 3 screens long!

nicm:
I don't think we need a runtime option for this.

koniu:
So you're suggesting that because a feature was added now we have to build tmux from source to retain sanity? I understand there's reasons for keeping a number of runtime options down but you had made quite an impactful change and giving users the choice that doesn't rely on custom builds is surely a better way to go.

nicm:
If we had a configure flag then the few people who don't want utempter could build without it.

Support for utempter was added more than two years ago and so far nobody else has requested even a configure flag. We have far too many options already and we are not adding another that hardly anyone will use.

Even if packagers do not provide a package without utempter, tmux is simple to build, it is not asking a huge burden.

nicm:
I would be happy for the default to be to build without utempter.

nicm:
Try this [x.txt](https://github.com/tmux/tmux/files/178782/x.txt).

koniu:
It could be that there's been no such request because the version in question has not yet made it to stable distros which are used in multi-user environment, as opposed to rolling distros which people mostly use on their personal machines.

Configure flag is a half-measure because sometimes a user might want to have both hanging sessions for different tasks and a general session/window where they might like to be able to receive 'write' messages or talk requests.

Fair play about the amount of options anyhow even if you're making heavy assumptions about what users might want/need. EOT as far as I'm concerned, cheerio!

PR #348, opened 2016-03-25
Minor change to fix building tmux on Oracle Solaris 11.x

khanyooh:
The ncurses headers are located in a nested dir:
$ pkg contents library/ncurses | egrep "(ncurses\.h|term\.h)"
usr/include/ncurses/ncurses.h
usr/include/ncurses/term.h
(It's required to build libevent2 prior to building tmux2, as official repos provide libevent1.4 only)

### DIFF
--- a/tty-term.c
+++ b/tty-term.c
@@ -21,12 +21,19 @@
 #if defined(HAVE_CURSES_H)
 #include <curses.h>
 #elif defined(HAVE_NCURSES_H)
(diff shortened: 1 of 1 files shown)


nicm:
Don't do this with an ifdef in the C file. Better to end up passing
-I/usr/include/ncurses, either by using pkg-config or by adding a test in
configure.ac.

PR #355, opened 2016-03-29
Add set-clipboard-cmd option

matthiasvegh:
This configuration parameter allows a user to automatically pipe the selection to xclip or whatever tool they fancy for managing their clipboard.

https://github.com/matthiasvegh/dotfiles/blob/master/home/.tmux.conf#L6-L7 shows example usage of this. This works well even through Putty if X forwarding is on. (ie.: windows clipboard is kept in sync with the selected text).

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -113,6 +113,12 @@ const struct options_table_entry options_table[] = {
 	  .default_num = 1
 	},
 
+	{ .name = "set-clipboard-cmd",
+	  .type = OPTIONS_TABLE_STRING,
+	  .scope = OPTIONS_TABLE_SERVER,
+	  .default_str = ""
+	},
+
 	{ .name = "terminal-overrides",
 	  .type = OPTIONS_TABLE_STRING,
 	  .scope = OPTIONS_TABLE_SERVER,
--- a/tmux.1
+++ b/tmux.1
@@ -2444,6 +2444,8 @@ disallowedWindowOps: 20,21,SetXprop
 Or changing this property from the
 .Xr xterm 1
 interactive menu when required.
+.It Ic set-clipboard-cmd Ar string
+Command to pipe the selection into when copying
 .It Ic terminal-overrides Ar string
 Contains a list of entries which override terminal descriptions read using
 .Xr terminfo 5 .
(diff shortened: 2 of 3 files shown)


nicm:
Can't you already do this with copy-pipe?

matthiasvegh:
Yes and no. Copy-pipe is used by https://github.com/tmux-plugins/tmux-yank for instance, and a myriad of other tmux.confs floating around, all of them however require a key to be pressed to copy from the selection to the system clipboard (See: https://github.com/tmux-plugins/tmux-yank#notes).

I've yet to come up with a configuration-only way of getting mouse release to copy to X's clipboard. The fact that tmux-yank still has the notice (despite something interesting going on with iTerm tmux-plugins/tmux-yank#51) seems to confirm this.

nicm:
You should be able to change the binding for MouseDragEnd1Pane in copy
mode to use copy-pipe.

matthiasvegh:
That would certainly be easier, I'll try in just a moment and get back to you. (If it _were_ possible however, what is the purpose of `set-clipboard`?)

nicm:
set-clipboard means to use the terminal's own escape sequences to set
the clipboard, which means you don't need any external tool at all, it
just works. I think only xterm and possibly iTerm2 support this.

matthiasvegh:
```bind-key -t vi-copy MouseUp1Pane copy-pipe "xclip -in -selection clipboard"``` Seems to work as you said. I imagined set-clipboard-cmd as an extension of set-clipboard but a config-only approach seems to make more sense than this commit.

matthiasvegh:
I've been using your suggestion and have noticed that newlines are not being handled correctly (selections with newlines then pasted elsewhere have the newlines removed). I'm still checking whether this is a tmux issue or an xclip one.
Using set-clipboard-cmd however the result is correct (which also uses xclip).
This sounds a lot like tmux-plugins/tmux-yank#42

matthiasvegh:
Nevermind my last, this seems to be a question of using xclip -in or xclip -in -selection clipboard.
In any event, copy-pipe seems does as it should, so there is  no need for this option.
Thanks for your help!

PR #367, opened 2016-04-11
Add next-paragraph and previous-paragraph commands

casey:
Credit for this patch goes to jxraynor from the tmux-users list:

https://groups.google.com/forum/#!topic/tmux-users/UFM75p-v-pE


Morantron:
Omgosh, this is a very quick way to move around in copy-mode. Nice :clap:

PR #373, opened 2016-04-15
Fix default socket on noacl cygwin root

graywolf:
With cygwin noacl root, permissions are just weird and tmux
fails to create default socket even when it can (and stuff
works if it does).


nicm:
I don't like adding more ifdefs for obscure Cygwin configurations. This check is here for a reason.

graywolf:
I understand, I'll just maintain own fork then.

PR #382, opened 2016-04-25
Update build instruction

letientai299:
I could not build tmux success due to missing dependencies, until I check the .travis.yml file.


nicm:
This is obviously not portable so I don't think we should include it.

PR #390, opened 2016-04-29
New window and split window in the current folder by default

ershov:
Hi,

I'm missing the tmux feature to open new windows and panes in the current directory.
Here's a patch to bring it back.

Many OS distros are still shipped with that old version 1.8 when tmux did that.
As I can see, when people try newer tmux, they get surprized about not having new windows in the current folder.

I know I can just add `-c` to the key bindings but still, currently there's no solution how to run a custom command in the current directory.

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -93,6 +93,9 @@ cmd_split_window_exec(struct cmd *self, struct cmd_q *cmdq)
 		format_defaults(ft, cmdq->state.c, s, NULL, NULL);
 		to_free = cwd = format_expand(ft, args_get(args, 'c'));
 		format_free(ft);
+        } else if ((cwd = osdep_get_cwd(wp->fd)) ||
+		   (cwd = osdep_get_cwd(s->curw->window->active->fd))) {
+		/* */
 	} else if (cmdq->client != NULL && cmdq->client->session == NULL)
 		cwd = cmdq->client->cwd;
 	else


nicm:
I do not think we should do this and certainly not make it the default.

ershov:
Maybe a flag then?
This used to be the default and many users get upset when they lose this feature after upgrade.

nicm:
The line that users are upset is not convincing for a change that happened more than two years ago with few complaints since. You may be upset but that does not make everyone.

Saying that, what I would like to see is a way to set default flags on a command, so you could set the default flags to new-window to "-c '#{pane_current_path}'" and -c would always be applied. Perhaps this would be best done by having a command that worked from a hook to modify the arguments.

ershov:
Looks like a common mechanism for a single purpose.
I can't figure neither any other use of this nor any other 'default' value for this flag.

Anyway, I got your point.

Thank you for your consideration.

nicm:
Well yes initially for a single purpose but I am constantly surprised by
the uses people think up for generalised mechanisms. Whereas if we make a
special case then it can't be used for anything else even if we want to.

ershov:
It's a good point, too.
Since you are into generalization, you might like an idea I have.
It can solve many problems in tmux including multiple commands binding in 'mode', command aliases, named command combinations, variables, loops and many others.
Can be done quite easily, without heavy code rewrite.
Let me know if you are interested.

PR #392, opened 2016-04-29
Fix if-shell not expanding all current formats

coderstephen:
Title says it all. `if-shell` wasn't expanding all formats, such as one pertaining to panes. Passing the format state to `format_defaults()` fixes the problem. Nothing fancy here. :grinning:


nicm:
Yes, changed run-shell but not if-shell before... applied, thanks.

PR #396, opened 2016-05-05
evbuffer_readline() for libevent < 2.0

bnjf:
libevent 2.0 has deprecated `evbuffer_readline()` in favour of `evbuffer_readln()`.  Use it only with older versions.

### DIFF
--- a/control.c
+++ b/control.c
@@ -61,7 +61,11 @@ control_callback(struct client *c, int closed, __unused void *data)
 		c->flags |= CLIENT_EXIT;
 
 	for (;;) {
+#if defined(_EVENT_NUMERIC_VERSION) && _EVENT_NUMERIC_VERSION >= 0x2000000
 		line = evbuffer_readln(c->stdin_data, NULL, EVBUFFER_EOL_LF);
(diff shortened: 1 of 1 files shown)


nicm:
This issue implies that tmux uses evbuffer_readline, which it doesn't and shouldn't because it is irredeemably broken. Which I reported and fixed and is the reason for libevent 1.4.13 :-).

bnjf:
Phew!

bnjf:
A-ha.  I recall being able to build against 1.4, but I thought I'd submit this pull request for chumps stuck with it (RHEL 6, for instance).

nicm:
No, that is not the point. We are not using evbuffer_readln because evbuffer_readline is deprecated, we are using it because evbuffer_readline is broken.

If you don't use control mode, then commenting out the evbuffer_readln call or replacing it with evbuffer_readline and building against a libevent older than 1.4.13 is fine. But we are not going to let tmux build with it silently broken.

You can either comment the line out yourself or build against a newer libevent (it is easy to build yourself).

PR #399, opened 2016-05-08
cmd-bind-key.c: fix argument count check for `bind -t` case

intelfx:
`bind-key` may accept multiple arguments as commands like this:

    bind-key -t vi-copy WheelUpPane scroll-up \; scroll-up \; scroll-up

Fixes issue #400.


nicm:
This patch does not change anything, arguments beyond the second will be ignored.

PR #401, opened 2016-05-09
Add binary clock

cosmo-ray:
Add the possibility to choose 'binary' as style in clock-mode-style.

Now the question is why ?

Let's imagine this patch is upstream,
peoples will get the possibility to read binary far more often than they usually do,
they will become familiar with binary algorithm,
Developers will get habit to use bit-mask, once they have learn how to manipulate binary,
they will get curious about low level stuff,
they will use programming language that allow to manipulate bit-mask easily, like C.

But if they don't learn how to manipulate bit-mask, developers will stop learning C because it's too complex for them, they will go java, OS will be create in java, kernel will be in Java,
the world will use java, and java logo is coffee, so the world will drink coffee.
What will happen with then ?
Tea market will drop, coffee grow up, and do you think nations like Brazil produce enough coffee for all china ?
No china will obviously invade Brazil to have more coffee,
and from this a new world war will happen.

So voilà, as a humanist I can't let the world be destroy and decide to save it,
let's introduce a binary clock in Tmux and save the world with it !

Thanks.

Signed-off-by: Matthias Gatto <uso.cosmo.ray@gmail.com>

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -36,7 +36,7 @@ const char *options_table_mode_keys_list[] = {
 	"emacs", "vi", NULL
 };
 const char *options_table_clock_mode_style_list[] = {
-	"12", "24", NULL
+	"12", "24", "binary", NULL
 };
 const char *options_table_status_keys_list[] = {
 	"emacs", "vi", NULL
--- a/window-clock.c
+++ b/window-clock.c
@@ -21,6 +21,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <endian.h>
 
 #include "tmux.h"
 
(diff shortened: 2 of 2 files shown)


nicm:
I don't think we need this in tmux, thanks.

cosmo-ray:
I'm sorry If you didn't like the commit message,
I know this patch is useless, and so I've try to make a funny message because I had no other argument than, "I'd like to have binary clock."
Now even if it's useless, it's still something funny to have, and I don't think this patch add a lot of complexity to tmux code.

nicm:
I don't dislike the commit message, we just do not need this feature. But
anyone who does want it can now apply your patch.

ThomasAdam:
Plus, external binary clocks exist...

PR #412, opened 2016-05-17
Change session directory

andjscott:
What is the thinking on changing the working directory of a session? I didn't seen any other issues about it. To do this currently, you must detach from the session and reattach with an argument.  `tmux attach-session -t $session -c $directory`

This method would allow changing of session's working directory without detaching.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -70,6 +70,7 @@ dist_tmux_SOURCES = \
 	cmd-bind-key.c \
 	cmd-break-pane.c \
 	cmd-capture-pane.c \
+	cmd-change-session-directory.c \
 	cmd-choose-buffer.c \
 	cmd-choose-client.c \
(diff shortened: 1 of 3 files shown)


krader1961:
What problem does this solve? And why didn't you document the new commands?

nicm:
A new command is silly, add a flag to attach-session so that it does not attach.

nicm:
Something like this maybe
[x.txt](https://github.com/tmux/tmux/files/267612/x.txt)

andjscott:
You are right that the command is unneeded in tmux. I have also found a way of doing what I wanted:
```
cwd() {
	cd "$@"
	local session=$(tmux display-message -p '#S')
	(unset TMUX; tmux attach -t $session -c $PWD\; new-window exit \; detach-client) > /dev/null
}

```

PR #421, opened 2016-05-23
window-copy: remove unused ox variable

mathstuf:
---
Doesn't look like it's supposed to be used.


nicm:
Yes, applied, thanks!

PR #434, opened 2016-06-02
Allow control mode clients to connect without resizing windows

dlobraico:
@msaffer and I were experimenting with control mode and were surprised to find that connecting to an existing session in control mode caused a window resize to the default terminal size of 80x24. This patch skips resizing for clients with the `CLIENT_CONTROL` flag set.


nicm:
We can't do this, because then there is no way for a control mode user
to limit the size.

I think it would be OK to skip control clients which have not called
refresh-client -C, but you will need a CLIENT_CONTROLRESIZED flag or
something for that.

PR #444, opened 2016-06-08
spelling fixes

jsoref:


nicm:
Applied, except for ansicode.txt which came from somewhere else. Thanks

Issue #312, opened 2016-02-18
Badly positioned split after closing top-bottom split


netei:
This issue comes from here : https://github.com/junegunn/fzf/issues/201 , but is caused by tmux.

Here's a way to reproduce the problem

1. Open a tmux session
2. Run command `echo foo`
3. Clear your screen with `C-l`
4. Open a top-bottom split with `C-b` `"`
5. Close the bottom split with `C-d`

Now the window contains 

```
$ echo "foo"
foo
```

Even tough the screen was "cleared" before. 

They is a workaround by binding `C-l` to `clear; tmux clear-history` but this makes us loose the tmux history feature.

What do you think ?

nicm:
When you resize a pane, tmux will fill with lines from the history if
possible, it is not a bug.

netei:
Is they an option to disable this behaviour (eg do not fill lines when removing a split), or would you accept to add an option for this case ?

ThomasAdam:
Why would you not want this?

netei:
Because I find it disturbing because when I clear my screen, It would be better if it would stay cleared whenever I use fzf (which is basically a fuzzy finder that will open a top-bottom split, let you select a line, and close that split, and enter it into the command line)

nicm:
We do not need an option for this. I wouldn't mind altering the default
behaviour if the change is small.

kennykaye:
@nicm I would also support the addition/modification of this feature. Do you have plans to implement this?

PR #451, opened 2016-06-15
Don't pull cleared (not scrolled) history back into view.

chaoren:
- It's intuitive to see the cursor position as being relative to the
  top of the pane. Pulling lines down unexpectedly changes the cursor
  position.
- Certain third party tools will use a temporary pane for the user to
  make a selection. The rapid opening and closing of a helper pane
  unexpectedly changes the state of the original pane.
  See demonstration video here:
      https://github.com/junegunn/fzf/issues/201

### DIFF
--- a/grid-view.c
+++ b/grid-view.c
@@ -67,6 +67,8 @@ grid_view_clear_history(struct grid *gd)
 		grid_collect_history(gd);
 		grid_scroll_history(gd);
 	}
+
+	gd->hscrolled = 0;
 }
 
 /* Clear area. */
--- a/grid.c
+++ b/grid.c
@@ -99,7 +99,7 @@ grid_create(u_int sx, u_int sy, u_int hlimit)
 
 	gd->flags = GRID_HISTORY;
 
-	gd->hsize = 0;
+	gd->hscrolled = gd->hsize = 0;
 	gd->hlimit = hlimit;
 
 	gd->linedata = xcalloc(gd->sy, sizeof *gd->linedata);
@@ -170,6 +170,8 @@ grid_collect_history(struct grid *gd)
 
 	grid_move_lines(gd, 0, yy, gd->hsize + gd->sy - yy);
 	gd->hsize -= yy;
+	if (gd->hscrolled > gd->hsize)
+		gd->hscrolled = gd->hsize;
 }
 
 /*
@@ -186,6 +188,7 @@ grid_scroll_history(struct grid *gd)
 	    sizeof *gd->linedata);
 	memset(&gd->linedata[yy], 0, sizeof gd->linedata[yy]);
 
+	gd->hscrolled++;
 	gd->hsize++;
 }
 
@@ -196,7 +199,7 @@ grid_clear_history(struct grid *gd)
 	grid_clear_lines(gd, 0, gd->hsize);
 	grid_move_lines(gd, 0, gd->hsize, gd->sy);
 
-	gd->hsize = 0;
+	gd->hscrolled = gd->hsize = 0;
 	gd->linedata = xreallocarray(gd->linedata, gd->sy,
 	    sizeof *gd->linedata);
 }
@@ -231,6 +234,7 @@ grid_scroll_history_region(struct grid *gd, u_int upper, u_int lower)
 	memset(gl_lower, 0, sizeof *gl_lower);
 
 	/* Move the history offset down over the line. */
+	gd->hscrolled++;
 	gd->hsize++;
 }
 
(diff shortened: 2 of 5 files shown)


chaoren:
Related issue: https://github.com/tmux/tmux/issues/312.

nicm:
I like the way it is now and it is how xterm works, we are not going to
revert it.

chaoren:
Can we at least leave this pull request open to get some more opinions?

nicm:
We are not going to just revert this change under any circumstances, and since it has been in tmux since 2009 I think there has been plenty of time for opinions. But I would consider a different solution.

chaoren:
Perhaps we can consider different behavior depending on whether the height is increased at the top, or increased at the bottom?

I agree that if the bottom edge of the pane is fixed, and the top grows upward, then it does make sense to pull lines out of history to fill the new gap, but if it's the top edge of the pane that stays fixed, and the bottom is growing, then pulling lines out of the top won't make any sense. It should be the bottom getting filled (with blank lines) in this case. I think it would be nice if we could keep the cursor fixed relative to the entire window if possible. Don't you agree?

nicm:
I don't think this makes more sense, we should try to show as much text
as we can from the history no matter how the pane is resized. Doing
something different depending on what direction we think it is changing
sounds like a mess.

What is the actual problem here? If you don't like that a temporary pane
changes the layout, why not use a temporary window instead?

chaoren:
Imagine issuing a `clear` command, then opening a new pane underneath to do some side stuff, then closing it. The original pane is no longer `clear`ed, since some history just got pulled out. It's only a minor annoyance, but tmux is doing something unintuitive here: filling extra space at the bottom by pushing down text from the top.

chaoren:
> we should try to show as much text as we can from the history

I don't think this is an universally agreed upon idea. This is why the `clear` commands exists. Some people use it to clear the view after finishing one thing and starting another.

chaoren:
The workaround I gave the people on that issue is to just use:
```
alias clear='clear; tmux clear-history'
```
And it works fine for me at least, but the tmux scrollback buffer would still be very nice to have if you realize much later that you needed an output from before, but most of the time you want it out of your way.

chaoren:
> Doing something different depending on what direction we think it is changing sounds like a mess.

I didn't think this would be difficult. Does screen have access to absolute `x` and `y` coordinates relative to the parent window, in addition to its height and width?

chaoren:
`clear` is definitely the culprit here.

If you never clear the screen, there should only be text in the scrollback buffer if the current screen is already filled. In that case, opening and closing a pane below would cause you to lose half a screen worth of text from view if you don't pull from history.

If you have cleared your screen, and your screen is only a quarter filled, opening and closing a pane below only affects the empty part of the original screen, so pulling history down would be weird in this case.

But there's no way tmux could know which part of the scrollback buffer is before or after the most recent `clear` or `reset` command.

chaoren:
Could we do something like this?

1. At the end of `grid_view_clear_history`, we mark the current position in the grid.
2. When we're pulling lines out of history in `screen_resize_y`, we check if we reach that mark.
3. Once we reach that mark, we stop pulling history, and just put blank lines at the bottom.

I think the mark can be stored as an offset from `grid->hsize`. When we do `grid_collect_history`, we decrease the value of the mark by `yy`. When the mark becomes negative or when `grid_clear_history`, we just invalidate the mark.

If I'm understanding the purpose of `grid_view_clear_history` correctly, the cleared history is undesirable to keep in view, but still desirable to keep in the buffer. I think it makes sense for us to keep only those cleared history out of view when resizing. Is there any other event that can cause out-of-view history to come back into view?

Only `clear` and `reset` triggers `grid_view_clear_history`, correct?

ThomasAdam:
I really don't think this is worth the effort.

chaoren:
> I really don't think this is worth the effort.

Yeah, I guess only people who clear the screen very often would run into this mild annoyance. Though, I find this mild annoyance to be worth *my* effort.

I'm offering to provide all of the implementation effort, but first I need confirmation that
1. the behavior of pulling cleared (not scrolled) history back into view is wrong
2. the solution outlined in https://github.com/tmux/tmux/pull/451#issuecomment-226394333 makes sense

I haven't used xterm, but when you clear screen in xterm, it doesn't go into a scrollback buffer, right? I'm not sure how you would compare this behavior.

chaoren:
@nicm @ThomasAdam please have a look at https://github.com/tmux/tmux/pull/451/commits/67006aed8bd128b9ddf63f2ef7653837457bc8bc.

nicm:
The xterm I currently have in front of me doesn't appear to put what is on screen into scrollback when you clear but that directly contradicts the log message for ecc22c521d83f8ce480c0e2866fdf3ed14bd6131, so I expect that some xterm does :-). I seem to remember something about terminals or applications which clear the screen by sending a load of CRs.

But I think it is fine not to use anything that has been cleared when resizing.

chaoren:
@nicm, thanks for confirming the expected behavior. Could you please review https://github.com/tmux/tmux/pull/451/commits/67006aed8bd128b9ddf63f2ef7653837457bc8bc and accept it if it looks fine?

aykamko:
@nicm I've always been waiting on this. Can you please take a look and stamp it?

nicm:
Applied, thanks.

PR #453, opened 2016-06-15
Add info about #{buffer_name} to man page

awalGarg:
Thanks to user Nei in irc for informing me about this variable. It was missing from the man page, though.


nicm:
Applied thanks

PR #464, opened 2016-06-23
Fix the UTF-8 detection on non-GNU systems

7heo:
Not all systems have locale environment variables set; so the `CLIENT_UTF8` flag can be unset on a such UTF-8 configured system.

This commit fixes that by, instead calling of `getenv()`, calling `nl_langinfo(CODESET)`; which, according to the man,
> returns a string which is - *the name of the character encoding used in the selected locale, such as "UTF-8"* - in the program's current global locale.

after the locale has been set at the top of the `main()`.

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -271,15 +271,9 @@ main(int argc, char **argv)
 	if (getenv("TMUX") != NULL)
 		flags |= CLIENT_UTF8;
 	else {
-		s = getenv("LC_ALL");
-		if (s == NULL || *s == '\0')
-			s = getenv("LC_CTYPE");
-		if (s == NULL || *s == '\0')
-			s = getenv("LANG");
-		if (s == NULL || *s == '\0')
-			s = "";
-		if (strcasestr(s, "UTF-8") != NULL ||
-		    strcasestr(s, "UTF8") != NULL)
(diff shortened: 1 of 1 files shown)


nicm:
nl_langinfo is not reliable, so we can do it as well as getenv, but not instead.

nicm:
Note that LANG and LC_ALL are specified by POSIX, they are not GNU extensions.

nicm:
Actually, we shouldn't do this.

The reason we are doing this check is a convenience - if the user set the locale to UTF-8, it is a sensible default to assume a UTF-8 terminal.

There are two problems with using nl_langinfo() here:

Firstly, the return value from nl_langinfo() doesn't reflect the actual locale if the setlocale() stuff at the top of main() succeeds - it will /always/ return UTF-8. So we would have to call it before that, and probably swap the "" and "en_US.UTF-8" calls.

Secondly, it would mean there is no way to use tmux on a non-UTF-8 terminal if the platform defaults to UTF-8. Assuming a UTF-8 terminal is not necessarily correct and we need a way for the user to override it.

7heo:
@nicm Agreed. All of what you said makes sense. I will update this PR with that in mind.

PR #471, opened 2016-07-05
Add no resize argument (-n) for tmux attach

mie00:
`tmux attach -n` for attaching a client that has no effect on the screen resize.

It's mainly useful in cases where the client is a read-only client and it shouldn't affect the main user.

I thought it is useful when I wrote this https://github.com/mie00/tmux-guest to find that it's annoying when someone with a smaller screen try to attach to your sessions in read-only mode.


nicm:
But the larger clients will be mangled and not display properly.

PR #473, opened 2016-07-06
Enable support for 24-bit tmux color options and clean up color code

townba:
* Enable support for 24-bit colors for `tmux` options. Before this change, `tmux` options are limited to 256 colors even when running in a terminal that supports 24-bit color.
* Clean up color code. Before this change, color information for a cell was divided between its flags (indicating what color model was used: regular, 256 color, or 24-bit color) and its foreground/background value (which itself was a union of an 8-bit value or a struct with separate red/green/blue values). With this change, all color information is stored in a single 32-bit value, with the higher 8 bits indicating the color model and the lower 24 bits indicating value.

### DIFF
--- a/colour.c
+++ b/colour.c
@@ -54,7 +54,7 @@ colour_to_6cube(int v)
  * evenly spread (8, 18, 28 ... 238).
  */
 int
-colour_find_rgb(u_char r, u_char g, u_char b)
+colour_rgbto256(u_char r, u_char g, u_char b)
 {
 	static const int	q2c[6] = { 0x00, 0x5f, 0x87, 0xaf, 0xd7, 0xff };
 	int			qr, qg, qb, cr, cg, cb, d, idx;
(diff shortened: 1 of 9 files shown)


nicm:
Looks mostly good, thanks, but:

- I do not want to increase the size of grid_cell_entry by 6 bytes. You
will need to use extended cells for both 256 and RGB colours, or
(preferably) keep the flags for 256 and continue to use extended cells
for RGB.

- The flag names don't need "COLOUR" in them twice:

+#define COLOUR_FLAG_256COLOURS 0x01000000
+#define COLOUR_FLAG_24BITCOLOUR 0x02000000

And I would prefer to stick with "RGB colour" instead of "24-bit colour"
throughout. How about these:

+#define COLOUR_FLAG_256 0x01000000
+#define COLOUR_FLAG_RGB 0x02000000

- In colour.c, I think we do not need to add this, we have enough forms to
specify colours already:

+               else if (len == 4) {
+                       n = sscanf(s + 1, "%1hhx%1hhx%1hhx", &r, &g, &b);
+                       r *= 0x11;
+                       g *= 0x11;
+                       b *= 0x11;

- This should use colour_24bittorgb:

+       if (c & COLOUR_FLAG_24BITCOLOUR) {
+               xsnprintf(s, sizeof s, "#%02X%02X%02X", (c >> 16) & 0xFF,
+                   (c >> 8) & 0xFF, c & 0xFF);
+               return (s);

townba:
Thank you for the thorough review. PTAL. I've applied your recommended changes.

townba:
Ugh. Hold off on this. I found a bug.

townba:
Fixed. PTAL.

nicm:
Applied with fixes for a few style nits. Thanks!

PR #480, opened 2016-07-14
Support for the escape sequences of the Terminology shell

lordwolfchild:
These changes make sure the escape sequences used by Terminology (https://github.com/billiob/terminology) are passed through tmux in raw form, so the terminal emulators graphical extensions can be used. It does not address the problem that Terminology does not know about the separate scrollback buffer of tmux, though, the demo programs issued by Terminology (tyls, etc.) will produce strange effects, but if the sequences are used in own programs/scripts, it works like a charm...

### DIFF
--- a/input.c
+++ b/input.c
@@ -113,6 +113,7 @@ void	input_enter_apc(struct input_ctx *);
 void	input_exit_apc(struct input_ctx *);
 void	input_enter_rename(struct input_ctx *);
 void	input_exit_rename(struct input_ctx *);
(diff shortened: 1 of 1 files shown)


nicm:
I don't think we want to support this, thanks. We already have a general purpose escape sequence to pass through to the terminal unaltered.

lordwolfchild:
You're probably right; thought this through by myself over night, tmux is not the right place for this... Thanks for your time anyway!

PR #501, opened 2016-08-10
Add wrap-select-pane option

alexdavid:
I've found select-pane wrapping around edges of the terminal to be error prone for my workflow. It looks like others have the same pain (hence https://github.com/dalejung/tmux-select-pane-no-wrap).

Would you consider merging this option to disable it, so that those who don't want this functionality don't have to patch it with a janky shell script?

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -892,6 +892,12 @@ const struct options_table_entry options_table[] = {
 	  .default_str = "default"
 	},
 
+	{ .name = "wrap-select-pane",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_SERVER,
+	  .default_num = 1
+	},
+
 	{ .name = "wrap-search",
 	  .type = OPTIONS_TABLE_FLAG,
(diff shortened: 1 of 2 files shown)


ThomasAdam:
This has come up before.  I really don't think we need this.

alexdavid:
Ok, I'll just use my fork then. Thanks anyway

kevintraver:
@alexdavid thanks for your branch! It works great, but there is one issue:
When trying to switch panes at the edge, it doesn't switch like it's supposed to, but it does unzoom the other pane.

Not sure if it's an easy fix, but would be glad to help.

alexdavid:
Just to be clear, you don't want to unzoom a pane if you move in a direction that would normally result in a wrap, correct?

kevintraver:
Yes. It seems to make sense to me that when you arrive at an edge, and then try to keep moving in that same direction the action should just be nothing.

NVolcz:
Can we revisit this issue? I would like to have the option for select-pane not to wrap around (zoomed in or not).

PR #505, opened 2016-08-14
Add full-size window split feature

smkent:
This adds a new argument (`-f`) to the `split-window` command which causes
the new pane to be created spanning the full window height (with `-h`) or
full window width (with `-v`) instead of splitting the active pane.

When creating a new full-size split, all existing panes are resized
proportionally to fit within the reduced window space. The command fails
if there is not sufficient space available in the window.

To illustrate, `split-window -h` on the following layout (with pane 1
active) produces:

```
    +---------+      +----+----+
    | 1       |      | 1  | 3  |
    +---------+  ->  +----+----+
    | 2       |      | 2       |
    +---------+      +---------+
```

With this feature, `split-window -h -f` on the same layout produces:

```
    +---------+      +----+----+
    | 1       |      | 1  | 3  |
    +---------+  ->  +----+    |
    | 2       |      | 2  |    |
    +---------+      +---------+
```

----

These types of full-size splits can be created in vim (using `:topleft` or `:botright`), and I wrote this to add the same type of feature to tmux. I can create full-size splits using \<Prefix\>\[hjkl\] (vim-style direction keys) with the following tmux config:

```
bind h split-window -h -f -b
bind j split-window -v -f
bind k split-window -v -f -b
bind l split-window -h -f
```

This is my first time writing any code for tmux. I'm happy to answer questions or receive feedback.

### DIFF
--- a/cmd-join-pane.c
+++ b/cmd-join-pane.c
@@ -124,7 +124,7 @@ join_pane(struct cmd *self, struct cmd_q *cmdq, int not_same_window)
 		else
 			size = (dst_wp->sx * percentage) / 100;
 	}
-	lc = layout_split_pane(dst_wp, type, size, args_has(args, 'b'));
+	lc = layout_split_pane(dst_wp, type, size, args_has(args, 'b'), 0);
 	if (lc == NULL) {
 		cmdq_error(cmdq, "create pane failed: pane too small");
 		return (CMD_RETURN_ERROR);
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -38,8 +38,8 @@ const struct cmd_entry cmd_split_window_entry = {
 	.name = "split-window",
 	.alias = "splitw",
 
-	.args = { "bc:dF:l:hp:Pt:v", 0, -1 },
-	.usage = "[-bdhvP] [-c start-directory] [-F format] "
+	.args = { "bc:dfF:l:hp:Pt:v", 0, -1 },
+	.usage = "[-bdfhvP] [-c start-directory] [-F format] "
 		 "[-p percentage|-l size] " CMD_TARGET_PANE_USAGE " [command]",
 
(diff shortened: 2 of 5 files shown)


nicm:
Thanks, I will take a look at this, probably in a couple of weeks.

nicm:
Please take a look at this [tmux-full-split.diff.txt](https://github.com/tmux/tmux/files/453288/tmux-full-split.diff.txt) which is mostly style changes (int -> u_int, less use of ?:, renaming some overly long names) and moving a bit of code about. I've also added a few comments as well.

smkent:
I took a look at your diff, everything seems reasonable to me. I did some cursory testing and the feature still seems to work as expected. Thanks for simplifying the loop in `layout_set_size_check`, I didn't think about that the first time around. Also, the variable rename in `layout_count_cells` (`n` -> `count`) wasn't part of my original PR; should that be moved to a separate commit? I guess I don't have a strong opinion either way.

How should we integrate your feedback with my change? Should I squash your diff into my commit or should we leave them separate? Since your feedback is reasonably substantial, it might make sense to make it a separate commit so it's clear who authored what.

Thanks again for taking the time to look at my change. I'm looking forward to getting this merged!

nicm:
I will apply it and credit you in the commit message, I suggest because you made relatively large code additions we add you to the copyright in the layout.c file. Let me know your name and I will add a copyright line as well. If I do not do it today it will be when I get home on Wed.

smkent:
Sounds great. I'm happy to be credited in layout.c; my name is Stephen Kent.

nicm:
Applied now, thanks.

smkent:
Thanks!

For anyone who finds this thread later, this was fed1e384ad7eb88cc6203e49d4efa8e5ed4edb59 (merged in f68a908c8ae4cd013e1a0ca5c5643381b0acb725).

PR #523, opened 2016-08-29
Default key binding for RectangleToggle: v -> Ctrl-v

GHPS:
Many aspects of vi-copy mimic VIM's visual mode. To select a rectangular section of the text in VIM (visual-block-mode) one uses Ctrl-v. It therefore makes sense to use the same key binding in tmux.


nicm:
I'm not going to do this right now but I'll add C-v as part of an upcoming diff, thanks.

Issue #515, opened 2016-08-23
Mac utf8 Rendering Issue


joshuarubin:
![screen shot 2016-08-23 at 11 03 37 am](https://cloud.githubusercontent.com/assets/194275/17901614/610ee6b4-6921-11e6-8922-4b6f5b239b18.png)

Scrolling through a vertical vim split that has a lot of multibyte utf8 characters causes horizontal overflows.

Mac OS X 10.11.6

Tried using both tmux 2.2 and HEAD (installed with homebrew)

```sh
$ locale
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL="en_US.UTF-8"
```

```sh
$ echo $TERM
tmux-256color
```

Had same results with non-italics `$TERM` `screen-256color`.

Issue occurs in both `Terminal.app` and `iTerm2.app` and in both `nvim` and `vim`.

Running `:redraw!` (in vim) fixes the issue. Running `refresh-client` (in tmux) never fixes it.

The issue does not occur outside of tmux.

nicm:
OS X has many missing widths for Unicode characters, tmux will assume width 1 for those. I guess you are using some of them that should actually be 2, but without a log it is hard to tell.

joshuarubin:
I'm happy to provide data for you to reproduce. What do you need from me? What kind of logs?

nicm:
tmux logs with -vvvv
Make sure you kill tmux entirety and start with -vvvv to get the server log

joshuarubin:
What was the issue?

nicm:
No logs

joshuarubin:
Huh, I thought I posted them, not sure what happened, here:

There were also issues with this char in my status line (unicode 2705): ✅
That caused some `wcwidth() == -1` issues in the log. I removed that and still had drawing issues.

Also, you might want to warn people that the logs contain their env vars. I had things like my github api key there. I had to jump through some hoops to sanitize the logs to be safe to post publicly,

Here are the logs.
[tmux-server-23676.log.txt](https://github.com/tmux/tmux/files/442528/tmux-server-23676.log.txt)

nicm:
[...]

PR #524, opened 2016-08-30
use utf8proc on platforms with broken wcwidth

joshuarubin:
As discussed in #515, this fixes rendering issues on platforms with broken `wcwidth` implementations (primarily mac) by using utf8proc.

The version of `utf8proc_data.c` differs slightly from the one in the utf8proc project as described in JuliaLang/utf8proc#83.

The difference only ensures that certain characters have width 1 (instead of 2), which means that it is no less correct than the current case of setting a width of 1 whenever `wcwidth` returns `-1`. In most cases, it should be correct much more of the time.

As/if the upstream data file is updated, I will update this pull request, or make a new one as necessary.


nicm:
No no no, don't copy and paste code from utf8proc - what's the point in that? Link against the library and implement wcwidth() and mbtowc() using it.

There should be no changes to utf8.c.

joshuarubin:
Linking against utf8proc will require changes to anything that packages tmux (like homebrew). Further, there are some problems in utf8proc, as I described, that would make using the upstream version worse than the current hack. Once upstream gets a few issues sorted out, it would be possible, but still seems unnecessary to introduce a new external dependency.

I can move things around with compat to minimize changes to `utf8.c`, but since it imports `<wchar.h>`, it will have to be updated or else there will be multiple definitions of `wcwidth`. Also, not sure how I can redefine `mbtowc` since it is included by `<stdlib.h>`.

And how about giving it a shot before knocking it down. This was not a small amount of work. And it is a whole lot better than the existing hack.

nicm:
Sorry, we are not going to import 1200 lines of code from another
project, we deliberated moved to system wcwidth() to avoid having to
carry around our own Unicode stuff. It is an external dependency or
status quo. Probably I should have been clear on this when I originally
mentioned utf8proc.

If there are currently issues with utf8proc, this should probably just
wait until they are resolved.

There are several options for how we use it, but ideally it should be
entirely contained within compat/ and compat.h with minimal changes to
the main *.c files. It is important not to pollute the main codebase
with #ifdef goo (one or two lines is OK, 10 is not).

We should be able to just link an alternative wcwidth() (it should be a
weak symbol in libc), but if that doesn't work then we can call it
compat_wcwidth() or something and use some small #ifdef to pick the
right function (or do the osdep_* thing we do for event_init()).

joshuarubin:
I updated the PR to use an externally linked utf8proc. It is an opt-in parameter at configure time, not platform dependent.

joshuarubin:
BTW, `utf8.c` is unchanged from `master` and utf8proc is good enough IMO with the minor changes I made (so that emoji and "private use characters" are considered width 1). There is still a corner case of some characters returning width 2 when they should return 1, but that is now in the hands of the utf8proc team and my PR there.

nicm:
This looks good, although I don't like having the code in a header. Doesn't it work if the code goes in a .c file?

nicm:
Also please add a copyright and licence statement to the top of any new file (like the existing files, but obviously with your name). Thanks

joshuarubin:
> Doesn't it work if the code goes in a .c file?

When I put the code in `charwidth.c`, this is the error I get:

```
duplicate symbol _wcwidth in:
    utf8.o
    compat/charwidth.o
duplicate symbol _mbtowc in:
    utf8.o
    compat/charwidth.o
duplicate symbol _wctomb in:
    utf8.o
    compat/charwidth.o
```

Maybe I need to define the functions differently in the header or provide some linker option?

This is the header I used:

```c
#ifndef _CHARWIDTH_H_
#define _CHARWIDTH_H_

#include <wchar.h>

int wcwidth(wchar_t wc);
int mbtowc(wchar_t *pwc, const char *s, size_t n);
int wctomb(char *s, wchar_t wchar);

#endif
```

joshuarubin:
Header file updated with copyright.

nicm:
OK clearly the linker doesn't like it, I think messing with linker options
is not going to be portable. I'm not sure what the tidiest thing to do is,
I'll take a look but I'm out tomorrow so it'll be either Thu or Sat.

nicm:
Here is an updated diff: [utf8proc.diff.txt](https://github.com/tmux/tmux/files/449392/utf8proc.diff.txt)

- Rename the functions to utf8proc_* and put them in compat/utf8proc.c.
- #ifdef them in utf8.c. I thought of calling them osdep_wcwidth or whatever with some ifdef stuff in compat.h but I think this is fine, it is only one line for each and still makes it obvious something is going on without cluttering up the whole show. If we have to do any other funny stuff on other platforms we can reconsider.
- Various style fixes.

Obviously I can only test on Linux, so please test on OS X.

joshuarubin:
OK, changes merged in. Looks good on my side.

nicm:
Applied now, thanks!

joshuarubin:
Thanks a bunch 👍  💯 !

PR #526, opened 2016-09-07
Support compilation on Solaris 11.3

paisleyrob:
Note: Still fails with default build order (autoreconf -if && ./configure
&& make) due to linking against -lncurses. Instead linking against
-lcurses works properly.

I haven't figured out how to convince configure to use -lcurses instead of -lncurses

### DIFF
--- a/osdep-sunos.c
+++ b/osdep-sunos.c
@@ -17,6 +17,7 @@
  */
 
 #include <sys/types.h>
+#include <sys/param.h>
 #include <sys/stat.h>
 
 #include <event.h>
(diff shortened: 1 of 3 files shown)


nicm:
It is probably not correct to include sys/param.h and sys/types.h, one or the other.

ACCESSPERMS must go in compat.h.

I think the filio.h bit should go in compat.h too.

It should use curses unless setupterm() is found in ncurses, but without a config.log I can't tell you any more.

paisleyrob:
Request migrated to https://github.com/tmux/tmux/pull/527

PR #529, opened 2016-09-10
Fixed osdep_get_name for Darwin

gpakosz:
On Darwin, `osdep_get_name` is broken since tmux `1.9`.

In tmux `1.8`, `osdep_get_name` uses `sysctl` to retrieve the command.

Then, in between tmux `1.8` and `1.9`, `sysctl` got dropped in favor of `proc_pidinfo` + `PROC_PIDT_SHORTBSDINFO`:

```
commit 7360ff4496bfd159ac5935a18de675f067c375db
Author: Nicholas Marriott <nicholas.marriott@gmail.com>
Date:   Tue Feb 12 09:40:22 2013 +0000

    Use proc_pidinfo on Darwin for process name too, from OZAKI Kiichi.
```

However, `PROC_PIDT_SHORTBSDINFO` is not available on older Mac OS versions. As such, `PROC_PIDT_SHORTBSDINFO` got ultimately changed to `PROC_PIDTBSDINFO` in commit:
```
commit 46b3c1a02564268ab810d0f54190f1e8fc1f4f6d
Author: Nicholas Marriott <nicholas.marriott@gmail.com>
Date:   Wed Apr 10 12:46:29 2013 +0100

    Use proc_bsdinfo which works on older OS X versions, from OZAKI Kiichi.
```

However, `PROC_PIDTBSDINFO` requires privileges and will only check for processes from the same user. In practice, this breaks `#{pane_current_command}`: in tmux `1.9+`, on Darwin: when you e.g. `sudo -i`, window name won't update anymore (provided you have `automatic-rename` set to `on`).

### DIFF
--- a/osdep-darwin.c
+++ b/osdep-darwin.c
@@ -17,7 +17,9 @@
  */
 
 #include <sys/types.h>
+#include <sys/sysctl.h>
 
+#include <Availability.h>
 #include <event.h>
 #include <libproc.h>
 #include <stdlib.h>
@@ -33,18 +35,34 @@ struct event_base	*osdep_event_init(void);
 char *
 osdep_get_name(int fd, __unused char *tty)
 {
-	struct proc_bsdinfo		bsdinfo;
+#ifdef __MAC_10_7
+	struct proc_bsdshortinfo	bsdinfo;
 	pid_t				pgrp;
 	int				ret;
 
 	if ((pgrp = tcgetpgrp(fd)) == -1)
 		return (NULL);
 
-	ret = proc_pidinfo(pgrp, PROC_PIDTBSDINFO, 0,
-	    &bsdinfo, sizeof bsdinfo);
-	if (ret == sizeof bsdinfo && *bsdinfo.pbi_comm != '\0')
-		return (strdup(bsdinfo.pbi_comm));
-	return (NULL);
+	ret = proc_pidinfo(pgrp, PROC_PIDT_SHORTBSDINFO, 0,
+			&bsdinfo, sizeof bsdinfo);
+	if (ret == sizeof bsdinfo && *bsdinfo.pbsi_comm != '\0')
+		return (strdup(bsdinfo.pbsi_comm));
+#else
+	int	mib[4] = { CTL_KERN, KERN_PROC, KERN_PROC_PID, 0 };
+	size_t	size;
+	struct kinfo_proc kp;
+
+	if ((mib[3] = tcgetpgrp(fd)) == -1)
+		return (NULL);
+
+	size = sizeof kp;
+	if (sysctl(mib, 4, &kp, &size, NULL, 0) == -1)
+		return (NULL);
+	if (*kp.kp_proc.p_comm == '\0')
+		return (NULL);
+
+	return (strdup(kp.kp_proc.p_comm));
+#endif
 }
 
 char *


nicm:
What does your change do?

gpakosz:
I'm running El Capitan version `10.11.16`. In tmux `1.9+` (i.e. `1.9`, `1.9a`, `2.0`, `2.1`, `2.2` and `HEAD`), once I `sudo -i` automatic window titles stop working:
  - launch `$tmux -f /dev/null` --> window title gets renamed to "bash"
  - launch `vim` --> window title gets renamed to "vim"
  - quit `vim` --> window title gets renamed back to "bash"
  - sudo -i
  - *[enter password]* --> window title **doesn't get renamed** to "sudo" while entering the password nor "sh" (default system shell) once logged as root
  - launch `vim` --> window title  **doesn't get renamed** to "bash"

Executing the same steps with tmux `1.8` gets the window renamed at every step, even when after `sudo -i`.

This is because `PROC_PIDTBSDINFO` requires privileges and will only check for processes from the same user and won't see processes owned by `root` user.

This PR reintroduces conditional compilation and uses `sysctl` when compiling for Mac OS < 10.7 or uses `PROC_PIDT_SHORTBSDINFO` instead of `PROC_PIDTBSDINFO` when compiling for Mac OS >= 10.7.

Once the PR applied, I'm getting the following behavior:

tmux-516 mov

You can see that window title renaming works again, but doesn't update until next status interval or when I press the Enter key. In tmux `1.8`, renaming was instantaneous.

nicm:
Applied now, thanks.

gpakosz:
Thanks.

BTW should I open another issue for the slight delay I'm experiencing in the gif above? In tmux `1.8` window title is refreshed as soon as the root prompt gets displayed. In the gif above, I have to press Enter to somehow trigger a status refresh, otherwise tmux seems to wait for next `status-interval` timeout.

nicm:
name or title? tmux will change title when the application does, so if
you are doing it in the prompt it should happen and redraw immediately.

gpakosz:
Sorry, I'm talking about **window name** and `automatic-rename`.

In tmux `1.8`, if I do what I do in the gif:

  1. launch `$tmux -f /dev/null`
  2. `$ sudo -i`
  3. *[enter your password]*
  4. *press return key*

As soon as I press the return/enter key in step 4., tmux refreshes the status line and window gets renamed to `0:sh*`.

But in tmux `HEAD` (with this PR applied), as illustrated by the gif above, status refresh doesn't happen until `status-interval` timeout kicks in **OR** until I press the return/enter key one more time. The gif shows the case where I press the return/enter key one more time to force a refresh. As you can see, when the pane displays `tatooine:~ root#` for the first time, tmux *hasn't caught up yet* and window name is still `0:sudo*`.

nicm:
tmux won't change name unless the pane changes, so if it gets the pane
change before the kernel is returning the right name, it'll get the old
name.

gpakosz:
How to explain behavior is different from tmux `1.8` and tmux `HEAD` in that respect? I'm too new to the codebase to circumvent what changed in between 4 versions of tmux :)

nicm:
It used to just update it on a timer.

ThomasAdam:
tmux syncs from openbsd. Not from github.

PR #550, opened 2016-09-26
Added #include <stdarg.h> without which these files will not compile.

orenwatson:
I don't know what version of compiler tmux is typically compiled with, but on my system it complains about va_args, etc, not being defined. This patch allows it to compile by including the relevant header.

### DIFF
--- a/arguments.c
+++ b/arguments.c
@@ -18,6 +18,7 @@
 
 #include <sys/types.h>
 
+#include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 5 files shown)


ThomasAdam:
This really isn't necessary.  What system are you compiling this on, and which compiler?

nicm:
For va_list you need stdarg.h or stdio.h so this is probably right.

orenwatson:
I was using GCC 4.8 on Ubuntu 14.04

nicm:
This is not required because stdarg.h is included from tmux.h. If it
wasn't, you would need to add it in a number of other files. I suspect
something is wrong on your system.

PR #551, opened 2016-09-27
cmd-find.c: follow ppids for current window

bmalehorn:
Before, a command like

    tmux rename-window foo

would rename *current* window, not the window where `tmux rename-window
foo` originated from.

This behavior can be easily reproduced: run a delayed `rename-window`:

    +-----------------------------+         +-----------------------------+
    |$ (sleep 1 &&                |         |$                            |
    |   tmux rename-window xxx) & |         |                             |
    |  tmux next-window           |  ---->  |                             |
    |                             |         |                             |
    |[0] 0:bash* 1:bash-          |         |[0] 0:bash- 1:xxx*           |
    +-----------------------------+         +-----------------------------+

Even though `rename-window` was issued from window 0,
window 1 was renamed.

This is especially obvious if you run a command that took a while, then
renamed the window:

    tmux rename-window foo
    ssh foo
    tmux set-window-option automatic-rename on

When the `ssh` command exits hours later, it sets automatic renaming on
whichever window I happened to be selected on. Annoying!

The fix is to walk up the process tree until we find the client's
pane. Then, we rename that window (or do whatever we were going to do).
Just like commands within a session will use that session by default,
all commands from within a pane will use that pane by default.

Now, that same command does this:

    +-----------------------------+         +-----------------------------+
    |$ (sleep 1 &&                |         |$                            |
    |   tmux rename-window xxx) & |         |                             |
    |  tmux next-window           |  ---->  |                             |
    |                             |         |                             |
    |[0] 0:bash* 1:bash-          |         |[0] 0:xxx- 1:bash*           |
    +-----------------------------+         +-----------------------------+

### DIFF
--- a/cmd-find.c
+++ b/cmd-find.c
@@ -973,9 +973,35 @@ cmd_find_target(struct cmd_find_state *fs, struct cmd_find_state *current,
 	else
 		fs->current = current;
 
-	/* An empty or NULL target is the current. */
-	if (target == NULL || *target == '\0')
+	/* An empty or NULL target is default. */
+	if (target == NULL || *target == '\0') {
+		/*
+		 * Follow parent pids from client until reaching a window_pane.
+		 * If this window_pane is in current session, use it.
+		 * Otherwise, default to whatever window the user is looking at.
+		 */
+		if (cmdq->client) {
+			pid_t pid = cmdq->client->pid;
+			struct window_pane *wp = NULL;
+			struct winlink *wl;
+			while (pid != 0 &&
+			       (wp = window_pane_find_by_pid(pid)) == NULL)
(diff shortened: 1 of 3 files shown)


nicm:
Your change is not portable. It is better to do this by looking at the
pty name.

But we already reverted this in 6f417ec9438de4440522483be02c51803d628313.

Scripts should avoid relying on the current session, window or pane and
instead make use of display -p and new/neww/splitw -P:

	P=$(tmux display -p '#{pane_id}'); sleep 1; tmux renamew -t$P xxx

bmalehorn:
Ah, I didn't know you could get / specify the pane id through those commands. This gets me what I want without having to apply a patch. Thanks!

PR #574, opened 2016-10-10
Do not use placeholders for job output

gpakosz:
When status line contains long commands inside `#()`, `"<'cmd' not ready>"`
or `"<'cmd' didn't start>"` placeholders can occupy a lot of space.
By now, when a job output isn't available it's just replaced by the
empty string and `log_debug()` is called instead.

### DIFF
--- a/format.c
+++ b/format.c
@@ -237,7 +237,7 @@ format_job_get(struct format_tree *ft, const char *cmd)
 		fj = xcalloc(1, sizeof *fj);
 		fj->cmd = xstrdup(cmd);
 
-		xasprintf(&fj->out, "<'%s' not ready>", fj->cmd);
+		log_debug("%s: <'%s' not ready>", __func__, fj->cmd);
 
 		RB_INSERT(format_job_tree, &format_jobs, fj);
 	}
@@ -247,8 +247,7 @@ format_job_get(struct format_tree *ft, const char *cmd)
 		fj->job = job_run(fj->cmd, NULL, NULL, format_job_callback,
 		    NULL, fj);
 		if (fj->job == NULL) {
-			free(fj->out);
-			xasprintf(&fj->out, "<'%s' didn't start>", fj->cmd);
+		    log_debug("%s: <'%s' didn't start>", __func__, fj->cmd);
 		}
 		fj->last = t;
 	}


nicm:
I like the placeholders.

gpakosz:
Though, when `status-left` or `status-right` contain long commands, it kind of causes flicker. And I assume people would never launch commands that run for long from their status line on purpose?

nicm:
Blank would cause flicker too though, best is to avoid commands where the
actual command changes - then tmux can use the cached output until it
finishes.

gpakosz:
I would argue it flickers less.

The reason I can't rely on tmux caching output is because I'm doing e.g. `#(cmd #{pane_tty} #D)` to e.g. to detect whether the current pane runs an ssh session to a remote server.

nicm:
Get the pane_tty etc inside the script instead.

On 10 Oct 2016 9:11 pm, "Gregory Pakosz" <notifications@github.com> wrote:

I would argue it flickers less.

The reason I can't rely on tmux caching output is because I'm doing e.g. #(cmd
#{pane_tty} #D) to e.g. to detect whether the current pane runs an ssh
session to a remote server.

—

<https://github.com/tmux/tmux/pull/574#issuecomment-252733269>, or mute the
thread

gpakosz:
> Get the pane_tty etc inside the script instead.

That doesn't fly. When doing so status line doesn't redraw right when I'm changing pane. Having `#{pane_tty}` and `#D` in the command invocation has this nice side effect.

Sure I could iterate through bindings and dynamically add `refresh-client -S` to any binding that does `select-pane` but  `list-keys`' output doesn't make it easy to dynamically amend bindings because it's hard to get quoting right when trying to feed output of `list-keys` to `bind`

nicm:
Ah yes because it won't run the same command more than once a second, which
is intentional. Perhaps we should provide a way to force that as part of
the command ($(!cmd...) or something), or just reduce the time. Although it
should still have the cached output from the last run since the pane tty
and ID should be the same as when you were last in that pane. Anyway I like
the the placeholders and removing them is not the right solution here.

gpakosz:
Alright.

Well I would welcome extending `#()` but I don't really know where to aim. Like I would like to have a way to have conditionals: `#(?cmd,a,b)`, or substitution or whatnot. But maybe that's too fragile to parse cmd correctly in such circumstances and also maybe you don't want more "fanciness" in tmux?

nicm:
You could try this, which adds #(!...) to force a command to run anytime the status line is updated, assuming it is not already running (that is, it bypasses the 1/second limit): [x.diff.txt](https://github.com/tmux/tmux/files/520414/x.diff.txt)

It also stops it redrawing the status line when the command finishes, otherwise the command runs infinitely. But status line should be redrawn enough that this shouldn't matter.

I am not entirely certain I like this, however. I might change it to be #(F:...) or something so we can add other extensions later.

gpakosz:
Nicholas, I'm sorry I didn't follow up earlier.

Well, it's unfortunate that the status line stops redrawing when the command finishes. The original intent was just to avoid long placeholder strings "blinking" (even more distracting when status line uses colors) until the result is available.

Do you think it's a good idea to have `#(user defined place holder:cmd)` syntax ?

nicm:
I don't understand your comment.

I don't think we want user defined placeholder strings.

gpakosz:
Well in my case, the user defined placeholder would be an empty string.

Here is my current `status-right`:
```
$ tmux show -gv status-right
#(cut -c3- ~/.tmux.conf | sh -s _loadavg)#(cut -c3- ~/.tmux.conf | sh -s _uptime)#(cut -c3- ~/.tmux.conf | sh -s _battery)#[fg=#080808,bg=default,none]#[fg=#8a8a8a,bg=#080808,none]#[fg=none]#[bg=none]#[none]#{?client_prefix,⌨,}#[fg=none]#[bg=none]#[none]#{?session_many_attached,👓,}#{?@battery_status, #{@battery_status},}#{?@battery_hbar, #{@battery_hbar},}#{?@battery_percentage, #{@battery_percentage},} #[fg=#8a8a8a,bg=#080808,none]| %R #[fg=#8a8a8a,bg=#080808,none]| %d %b #[fg=#d70000,bg=#080808,none]#[fg=#e4e4e4,bg=#d70000,none] #(cut -c3- ~/.tmux.conf | sh -s _username #{pane_tty} false #D)#[fg=none]#[bg=none]#[bold,blink]#(cut -c3- ~/.tmux.conf | sh -s _root #{pane_tty} #D)#[default]#[fg=#e4e4e4,bg=#d70000,none] #[fg=#e4e4e4,bg=#d70000,none]#[fg=#000000,bg=#e4e4e4,bold] #(cut -c3- ~/.tmux.conf | sh -s _hostname #{pane_tty} false #D)
```

You may argue my [`tmux` configuration](https://github.com/gpakosz/.tmux) does too much but `#(cut -c3- ~/.tmux.conf | sh -s _username #{pane_tty} false #D)` and `#(cut -c3- ~/.tmux.conf | sh -s _hostname #{pane_tty} false #D)` give me nice SSH aware username and hostname display.

Screenshot

And as much as I try to optimize those, from time to time, I see my status line flicker with `<'#(cut -c3- ~/.tmux.conf | sh -s _username #{pane_tty} false #D)' not ready>` because those commands are long strings to being with :(

nicm:
Right, but we already went over this, the problem is that you are firing different commands all the time, so tmux hasn't got a cached copy of the last output. The solution is to always fire the same command and look  up the pty etc inside the script.

gpakosz:
While I understand you're not in favor of having a way to not have placeholders, the solution certainly isn't in firing the same command and lookup pty inside the script: current pane may have changed already and this results in a less snappy sync between effective current pane and status line update.

The initial PR was a way to avoid adding yet more options and complexity to tmux while removing a imho rare behavior: seeing the placeholder strings. It looks like I under appreciated the importance and frequency of place holder strings display.

nicm:
They aren't particularly important but I like them, we have had them for
years, and they are a lot clearer than just printing nothing. You are
asking to remove them or to add code to change them to solve a rare
problem which comes from using #() in a way that is not recommended.

If you have a command that changes with every iteration, tmux cannot
find its previous output so it will have to use something - making that
an empty string is as likely to cause problems for someone else as the
placeholder string does for you now. It is not a general solution, and I
am not convinced there is a good one.

You could probably deal with the pane changing while the script is
running by making tmux pass TMUX_PANE into the #() command like we do
for panes themselves.

gpakosz:
I understand. All I was trying to say is that while reducing "flickering" would be nice, I didn't feel like being in a position to suggest a new, and more complicated to maintain, syntax for commands.

I tried looking for `TMUX_PANE` while being in the script, but this ends up feeling "laggy" when quickly changing panes since status line refresh is bounded. Another solution for my case would be injecting `refresh-client -S` calls in bindings but so far I wasn't able to reliably parse the output of `list-keys` to amend/transform and re-inject bindings.

nicm:
One possibility actually would be able to name commands, so you could do like:

$(mycmdname:sh foobarbaz.sh #{pane_tty})

And then tmux could use that name to cache the output instead of the whole command. For your case you would also need a way to be able to force execution though.

nicm:
Named commands would be a good step if we ever want to handle commands that run permanently in the background, although we've managed without that for years so it is probably not a priority...

nicm:
Try this please: [y.diff.txt](https://github.com/tmux/tmux/files/589317/y.diff.txt)

So you can prefix a command with some stuff separated by colons: #(flags:name:cmd)

Only flag is F for force, so you can try #(F:mycmd:/path/to/mycmd #{whatever})

gpakosz:
I understand the patch you're offering allows me to have a shorter placeholder string. Like `<'hostname' not ready>` instead of my overly verbose `<'cut -c3- ~/.tmux.conf | sh -s _hostname #D' not ready>`.

I tried with the following script:

```
#!/bin/sh

pane_id="$(tmux display -p '#D')"
echo $1 -- $pane_id >> /tmp/log.txt
echo "$pane_id"
```

And `./tmux-2.4 -S /tmp/tmux24 set status-right '#(F:paneid:/tmp/script.sh #D)'`.

First feedback:
  - upon switching pane, `/tmp/log.txt` gets appended immediately once, and then again 1 second later and the status line is refreshed at the same time as the second log entry
  - a space in the command name doesn't seem to be parsed correctly
  - I can't reassign a different command to the same name

I'll play with the code more.

nicm:
No the idea is to have a way to allow tmux to know what the previous
command so it can reuse it's output.

nicm:
You can't have spaces in the name. It won't let you use the same name for a
different command but I will change that.

nicm:
OK please forget the previous diff, here is a much better idea.

Rather than having an extra command name, lookup by the /unexpanded/ command. If the expanded form has changed, run the command again.

So "sh /tmp/x.sh #D" will be executed every time #D changes, and at most once a second if it has not changed.

I've left the force flag in there too as it may be useful, but you shouldn't need it. I might take it out again.

Please try this:  [xyz.diff.txt](https://github.com/tmux/tmux/files/594564/xyz.diff.txt)

nicm:
Here is the version without the force flag: [xyz2.diff.txt](https://github.com/tmux/tmux/files/594628/xyz2.diff.txt)

gpakosz:
Tried the latest patch. Seems to work well.

Of course, the first time a command runs, if it's taking too much time, one can get the `<'cmd' not ready>` message. But I'm confident that would mitigate about all the "flickering" I have experienced by bringing stability to colored status line sections/segments' width. See what I mean?

Thanks Nicholas!

nicm:
Applied now.

PR #581, opened 2016-10-13
Add -f flag to list of flags split-window takes.

frosencrantz:
Recent pull request (https://github.com/tmux/tmux/pull/505) added this flag, but the flag was not listed in the man page as one the command takes. The flag is documented, just missing from the list.


nicm:
Fixed thanks.

PR #587, opened 2016-10-15
Support multiplexing SSH-agent sessions

apmasell:
Add a feature to multiplex other systems through the same connectivity as tmux
windows and clients. This creates a SSH agent for each window and forwards
requests to the SSH agents discovered by clients, then synthesises the
responses. There is a unique agent for each window and the current
window-client connections determine which forwarding is valid.

This is meant to be extensible to other protocols, including GPG agent and
DBus.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -163,6 +163,7 @@ dist_tmux_SOURCES = \
 	notify.c \
 	options-table.c \
 	options.c \
+	othermux.c \
 	paste.c \
 	proc.c \
 	resize.c \
--- a/environ.c
+++ b/environ.c
@@ -101,7 +101,7 @@ environ_find(struct environ *env, const char *name)
 }
 
 /* Set an environment variable. */
-void
+struct environ_entry *
 environ_set(struct environ *env, const char *name, const char *fmt, ...)
 {
 	struct environ_entry	*envent;
@@ -118,6 +118,7 @@ environ_set(struct environ *env, const char *name, const char *fmt, ...)
 		RB_INSERT(environ, env, envent);
 	}
 	va_end(ap);
+	return envent;
 }
 
 /* Clear an environment variable. */
@@ -138,21 +139,23 @@ environ_clear(struct environ *env, const char *name)
 }
 
 /* Set an environment variable from a NAME=VALUE string. */
-void
+struct environ_entry *
 environ_put(struct environ *env, const char *var)
 {
 	char	*name, *value;
+	struct environ_entry *result;
 
 	value = strchr(var, '=');
 	if (value == NULL)
-		return;
+		return NULL;
 	value++;
 
 	name = xstrdup(var);
 	name[strcspn(name, "=")] = '\0';
 
-	environ_set(env, name, "%s", value);
+	result = environ_set(env, name, "%s", value);
 	free(name);
+	return result;
 }
 
 /* Unset an environment variable. */
--- a/othermux.c
+++ b/othermux.c
@@ -1,0 +1,1097 @@
+/* $OpenBSD$ */
+
+/*
+ * Copyright (c) 2007 Nicholas Marriott <nicholas.marriott@gmail.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF MIND, USE, DATA OR PROFITS, WHETHER
+ * IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
+ * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <arpa/inet.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/time.h>
+#include <sys/un.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <string.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <time.h>
+
+#include "tmux.h"
+
+/*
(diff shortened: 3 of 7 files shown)


nicm:
Thanks, but I don't think we want this in tmux, at least not in this form. What problem is this intended to solve?

apmasell:
I never expected you to hit the merge button. ;-) I wanted to start a discussion and felt it was better to do so with a working prototype than just nice ideas.

I've seen a bunch of different attempts to solve this problem: when a tmux instance is created, it inherits the environment variables for the SSH agent and those become stale when a new client attaches. There's currently the `update-environment` option, but this can only affect new windows. There's also methods to set the environment variable to a symlink and modify the symlink before a user starts as session. That works so long as only a single attachment takes place (i.e., if you attach from client 1, then client 2, then client from system 2, making the environment flip back to client 1 is difficult). If there is long-running simultaneous attachment, then there's no guarantee to be connected to the environment (something I do nearly always). If windows are transferred between sessions, then the shell environment has no way to access the correct information.

The necessary state information exists with in tmux itself: what the provided SSH agents are and what the windows are connected to those clients. If each window gets a separate SSH agent, it can correctly determine which clients' SSH agents are candidates for the request.

I'm happy to discuss other forms, but I created this as a demonstration for how such a thing would work. What are you concerns for direct addition to tmux? Would using a plugin system alleviate it (the protocol specific modules get `dlopen`ed by tmux)? It could be put in a separate daemon, but tmux would need to be able to 1) request environment variables for new sessions, 2) provide the environment variables of new clients, 3) provided updated window-client links. That would involve adding an information socket of some kind to tmux for this purpose. I want this to be as minimally invasive as possible and don't want the maintenance burden of these other protocols to fall to the tmux developers.

It would also be possible to use this as a framework for other protocols. SSH agent was my first choice because of how frequently I use it and the simplicity of the protocol. GPG agent would probably be of similar complexity. DBus would be possible, though much more complex to ensure unique paths and ids. X11 is also possible, but rather than attempting true multiplexing (as I have with this SSH agent design), it could simply forward to the X11 session of the most-recently-used client.

nicm:
I don't want to carry around a bunch of external protocol information in tmux, even as relatively simple as ssh-agent. And definitely not DBus, GPG or X11.

And dlopen() is never going to fly, loading external SOs into our address space is a disaster.

This needs to live outside tmux and talk to it somehow. But I don't think you need an information socket, I think this could be done as a separate daemon that attached to tmux with control mode. If control mode doesn't give you the information you need, it could be extended to do so.

nicm:
By the look of it, all you need is to add a flag to some command to report the client environment and possibly a notification when a window is moved. You can already get and set the environment variables of new sessions.

We should probably also make a change so that a control client only limits the attached session size if it has called refresh-client -C.

apmasell:
Okay. I'm looking at the control mode commands:

It ideal to have a `set-environment` that could do substitution on its argument to include the window ID since I wouldn't want to have to reset it constantly.

The `%unlinked-window-add` notification doesn't convey which session the window is being added to.

Also, with respect to the `refresh-client -C`, this client wouldn't want to be associated with any session; it should really care about all of them. So, perhaps call this a “global client” and make it disallowed to call refresh (and not be associated with a session, so not get `%output`)?

So, proposal 1 is looking like:
- modify a `%window-add` and `%unlinked-window-add` to include a session ID
- create `%window-moved` notification
- create `%client-attached`, `%client-detached`, `%client-switched`, `%client-variable` notifications (using the pointer value as the ID)
- create `set-environment-template` that can do substitution on value using the same patterns as else where
- either keep a state bit as to whether `refresh-client` has been called or create a slightly different control mode so that this session doesn't participate in resizing (I prefer the latter; thoughts?)
- Ideally, create a `%client-active` message to indicate when the last actively used client changes. (The client version of `monitor-activity`.)

Problems:
There are going to be two races of the same flavour:
- when tmux initially starts up, this demon will have to connect in parallel with the real client and do so before the first shell is spawned to set the environment variable correctly. I have no idea how to solve this.
- for every new window that is created, the demon will have to create a socket, but the initial shell environment is already being created, so command's in the user's login script (or SSH itself if that's the command being run) might start executing with the environment variable pointing to a non-existent socket. I don't know how to get around this without some kind of interlock.

nicm:
On Mon, Oct 17, 2016 at 03:48:48AM -0700, Andre Masella wrote:
>    Okay. I'm looking at the control mode commands:
> 
>    It ideal to have a set-environment that could do substitution on its
>    argument to include the window ID since I wouldn't want to have to reset
>    it constantly.
> 
>    The %unlinked-window-add notification doesn't convey which session the
>    window is being added to.
> 
>    Also, with respect to the refresh-client -C, this client wouldn't want to
>    be associated with any session; it should really care about all of them.
>    So, perhaps call this a ***global client*** and make it disallowed to call
>    refresh (and not be associated with a session, so not get %output)?
> 
>    So, proposal 1 is looking like:
> 
>      * modify a %window-add and %unlinked-window-add to include a session ID

Sounds good, but the session ID should go after the existing window ID
please.

>      * create %window-moved notification
>      * create %client-attached, %client-detached, %client-switched,
>        %client-variable notifications (using the pointer value as the ID)

These are both fine. I would suggest that instead of using pointers, we
just give clients a unique id like sessions, windows etc. I don't like
the idea of addressing commands to clients by the pointer and an ID
could be useful elsewhere (at the moment we mostly use the tty name
which I never liked much).

>      * create set-environment-template that can do substitution on value
>        using the same patterns as else where

Please don't add any new commands. Add a flag or - better - just make it
do it by default.

>      * either keep a state bit as to whether refresh-client has been called
>        or create a slightly different control mode so that this session
>        doesn't participate in resizing (I prefer the latter; thoughts?)

I would add a bit, CLIENT_CONTROLRESIZED or something for the resizing.

If you need to make it so that control client can live without a session
then I think that would be OK, but it would probably need more changes,
and I think it would be better done separately from the resize change
(which is sensible regardless).

>      * Ideally, create a %client-active message to indicate when the last
>        actively used client changes. (The client version of
>        monitor-activity.)

This is fine too.

> 
>    Problems:
>    There are going to be two races of the same flavour:
> 
>      * when tmux initially starts up, this demon will have to connect in
>        parallel with the real client and do so before the first shell is
>        spawned to set the environment variable correctly. I have no idea how
>        to solve this.

I suggest that (initially at least) you just create the first tmux
session via the daemon, that is the daemon creates the first session to
start the server and then users just attach to that server. So the
daemon is always the first client.

>      * for every new window that is created, the demon will have to create a
>        socket, but the initial shell environment is already being created, so
>        command's in the user's login script (or SSH itself if that's the
>        command being run) might start executing with the environment variable
>        pointing to a non-existent socket. I don't know how to get around this
>        without some kind of interlock.

I think that perhaps this is something users will just have to be aware of.

Possibly you could use "tmux wait", like make the daemon issue "wait -S"
for a channel named predictably per window, then users can just do "tmux
wait whatever" to make sure they are in sync.

apmasell:
Okay. That sounds good to me. The `wait` command is probably all I need.

I'll make the changes in separate commits in this PR and include my demon for testing purposes. Once you're happy with the changes, I'll remove the demon.

Thanks for your design insights. :-)

nicm:
OK, that is fine.

nicm:
Closing this, please continue to comment or open a new issue when you have something to look at.

PR #593, opened 2016-10-18
Add missing GRID_FLAG_{BG,FG}RGB so that RGB style strings behave properly

tombuzbee:
This fixes an issue I noticed where #rrggbb colors are downgrading to 256 colors, even on a true color terminal. Adding these two flags seems to work just fine for me. Thanks!

Test case:
tmux set -g status-left '#[bg=#000000] #[bg=#100000] #[bg=#200000] #[bg=#300000] #[bg=#400000] #[bg=#500000] #[bg=#600000] #[bg=#700000] #[bg=#800000] #[bg=#900000]'

Good result: 10 shades of red
Bad result: 3 shades

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -305,14 +305,19 @@ grid_get_cell(struct grid *gd, u_int px, u_int py, struct grid_cell *gc)
 		return;
 	}
 
-	gc->flags = gce->flags & ~(GRID_FLAG_FG256|GRID_FLAG_BG256);
+	gc->flags = gce->flags &
+		~(GRID_FLAG_FG256|GRID_FLAG_BG256|GRID_FLAG_FGRGB|GRID_FLAG_BGRGB);
 	gc->attr = gce->data.attr;
 	gc->fg = gce->data.fg;
 	if (gce->flags & GRID_FLAG_FG256)
 		gc->fg |= COLOUR_FLAG_256;
+	if (gce->flags & GRID_FLAG_FGRGB)
+		gc->fg |= COLOUR_FLAG_RGB;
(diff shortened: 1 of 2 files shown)


nicm:
This is wrong, cells with RGB colours are always extended cells with the
full 32 bits for the colours. GRID_FLAG_FG256 and GRID_FLAG_BG256 are
only needed to differentiate 256 from ANSI colours in non-extended cells
where the fg and bg members are only 8 bits.

tombuzbee:
Hmm, I guess it wasn't as simple as it seemed (although it fixed my test case). Dropping for now. Thanks for taking a look.

nicm:
As far as I can tell tmux is working fine here, the RGB colours are making it through to the terminal in the right form. I guess that either your tmux is too old, or your terminal (like xterm) doesn't fully support RGB colour, or you have hit another bug, but without any more information it is impossible to tell.

tombuzbee:
No, something is definitely not quite right. I built at HEAD and am using iTerm2 with a custom terminfo which includes Tc. RGB colors work from commands within a pane, but not in tmux itself.

nicm:
Logs? All three *.log files please, and using just a config with just your status-left example above.

nicm:
That is:

echo 'set -g status-left ...' >/tmp/foo
tmux -Ltest kill-server
tmux -Ltest -vvv -f/tmp/foo new
exit

tombuzbee:
Well, this is embarrassing. I must have had a stale server running in the background, because I can't reproduce my issue after running the kill-server command.

Apologies for the noise!

PR #615, opened 2016-10-28
Fix powerline on Unicode 9-compliant terminals

yan12125:
Powerline uses [the WATCH character](https://github.com/powerline/powerline/blob/fd54f63/powerline/config_files/themes/powerline.json#L41) (U+231A) when displaying current time. In [Unicode 8.0.0](http://www.unicode.org/Public/8.0.0/ucd/EastAsianWidth.txt), it's NEUTRAL:
```
2313..231F;N     # So    [13] SEGMENT..BOTTOM RIGHT CORNER
```
However, [Unicode 9.0.0](http://www.unicode.org/Public/UNIDATA/EastAsianWidth.txt) changed it to WIDE:
```
2313..2319;N     # So     [7] SEGMENT..TURNED NOT SIGN
231A..231B;W     # So     [2] WATCH..HOURGLASS
231C..231F;N     # So     [4] TOP LEFT CORNER..BOTTOM RIGHT CORNER
```
This breaks Unicode 9-compliant terminal emulators, including [Gnome/VTE3](https://bugzilla.gnome.org/show_bug.cgi?id=771591) (the core component of gnome-terminal and xfce4-terminal) and [QTerminal](https://github.com/lxde/qterminal/issues/260). If an user is on Fedora Rawhide, which has [patched glibc for Unicode 9.0.0 support](http://pkgs.fedoraproject.org/cgit/rpms/glibc.git/commit/?id=5c031d86de2393daa5181ad5ccdbca52c5eb3554), terminals use ```wcwidth()``` will be affected, too. I suggest to remove the assumption "Symbols, like emoji, should always use width 1." so that tmux works fine with those terminal emulators.

/cc @joshuarubin I didn't find any description about the assumption regarding UTF8PROC_CATEGORY_SO. Does values from ```utf8proc_charwidth()``` break something? (#524)

### DIFF
--- a/compat/utf8proc.c
+++ b/compat/utf8proc.c
@@ -33,10 +33,6 @@ utf8proc_wcwidth(wchar_t wc)
 		 */
 		return (1);
 	}
-	if (cat == UTF8PROC_CATEGORY_SO) {
-		/* Symbols, like emoji, should always use width 1. */
-		return (1);
-	}
 	return (utf8proc_charwidth(wc));
 }
 


yan12125:
Forgot to paste steps to reproduce:
* ```pip3 install powerline-status --user```
* ```~/.tmux.conf```: ```source ~/.local/lib/python3.5/site-packages/powerline/bindings/tmux/powerline.conf```
* xfce4-terminal 0.8.0 & tmux 2.3
* Arch Linux x86_64
xfce4-terminal-tmux

nicm:
Aha, I should have checked to see if it had changed width in Unicode 9, well spotted. Applied your change, thanks.

choco:
This change actually messes up things on macOS :/ If Fedora has a working wcwidth, why use utf8proc? This change was specifically made to mitigate a non-working implementation.

yan12125:
The motivation is that the original codes break at least xfce4-terminal on Arch Linux, and some people reported QTerminal on Fedora is also broken.

@choco Could you paste an screenshot?

nicm:
The problem here appears to be that Fedora wcwidth() is Unicode 8 but some terminals have their own internal implementation that is Unicode 9.

Because wcwidth() is in libc, it is very often going to lag behind, and if it has errors it may be a long time before fixes reach distributions. And of course on OS X, it is just rubbish.

The intention of allowing tmux to use utf8proc is to provide an alternative for exactly this case - a library that should be easier to update than libc, while still keeping the burden of UTF-8 details away from tmux.

My inclination is to keep this change, we should not be overriding the information utf8proc gives us - if utf8proc is still wrong on OS X, you should take this up with the utf8proc developers.

joshuarubin:
@yan12125:
> I didn't find any description about the assumption regarding UTF8PROC_CATEGORY_SO. Does values from utf8proc_charwidth() break something? (#524)

It was primarily for powerline fonts if I remember correctly.

It would be nice if there was a way to configure unicode 8/9 widths at runtime (like iTerm2 can now do in nightlies).

yan12125:
> It would be nice if there was a way to configure unicode 8/9 widths at runtime

Indeed. Unfortunately I know little about tmux's configuration language.

joshuarubin:
While I don't know if such a runtime flag would even be considered, that is not such a big challenge. utf8proc does not seem to have a way to differentiate between unicode 8 and 9 widths AFAIK (http://julialang.org/utf8proc/doc/utf8proc_8h.html).

Also, even if we could coerce utf8proc to do this, the built in wcwidth couldn't, so the runtime flag would only be useful if tmux was built with utf8proc.

It seems less likely that this kind of change (a runtime flag that either does nothing based on build configuration or only exists if built a certain way) would be included -- and that's *if* we can find a simple way to calculate widths differently by unicode version.

When monospace isn't monospace... what a mess.

yan12125:
yeah nothing is changed in utf8proc. Instead, widths are changed in Unicode standards.

> _if_ we can find a simple way to calculate widths differently by unicode version

I guess that would be much complicated. Ideally all applications should follow the latest Unicode standards. Nevertheless, that won't be a real world, never, especially on Mac, _a buggy Unix_.

PR #648, opened 2016-11-15
Prevent reporting error when name does not change.

ShikChen:
Allowing rename-session with the same current name is a more user friendly behavior.
It's frustrating to see the error like `duplicate session: <current_session_name>`.


nicm:
Applied a slightly different version of this, thanks!

Issue #642, opened 2016-11-12
Keep selection when scrolling in copy-mode


choco:
**Platform:** OS X
**Version:** tmux git afdef6d122ab667a48b843af74541e9aafcf98e6 (last commit)
**Terminal emulator:** iTerm2
**$TERM:**
- outside: `screen-256color`
- inside: `screen-256color`

I have the following configuration in my tmux.conf:
```
bind -T copy-mode-vi MouseDrag1Pane      send -X begin-selection
unbind  -T copy-mode-vi MouseDragEnd1Pane
```
This allows me to keep a selection without clearing it immediately.
When I create a selection using the bind above and then try to scroll, the selection will get expanded from where the scroll is initiated. I use the following slightly modified bind but the default also triggers the issue
```
bind -T copy-mode-vi WheelUpPane         send -X -N 3 scroll-up
bind -T copy-mode-vi WheelDownPane       send -X -N 3 scroll-down
```
Basically i'd like to be able to scroll and not mess with current selection if possible.

wincent:
@choco: I noticed the same. I'm avoiding the expansion [like this](https://github.com/wincent/wincent/blob/7537a6db15d5b8e62560be5ffab31908191e8ac3/roles/dotfiles/files/.tmux.conf#L69-L71):

```
bind-key -T copy-mode-vi WheelUpPane send-keys -X clear-selection\; send-keys -X -N 3 scroll-up
bind-key -T copy-mode-vi WheelDownPane send-keys -X clear-selection\; send-keys -X -N 3 scroll-down
```

But keeping the selection would be nice if possible.

nicm:
It would be fine if scroll-up/scroll-down commands did not move the selection.

mlgh:
Hi!
Try launching tmux with this patch: #652 
and add execute something like this
`bind-key -T copy-mode-vi b send -X stop-dragging-selection`
Then in copy mode you can try pressing b and selection will live indepndently from cursor. If you want to modify the selection, press "o" ("other-end" command) and you'll be able to modify selection.

mlgh:
@choco @wincent I believe you are interested in this, could you please try the #652 PR and report if this is the desired behavior for you? Also this PR allows you to move cursor independently from selection, do you see any need in such functionality?
Thanks!

choco:
I was trying this new feature and it works perfectly for a selection created by dragging, but I can't make it work when the selection was created in other ways.
Basically I'm trying to get this to work
```
[...]

PR #650, opened 2016-11-15
Don't stop dragging selection if mouse wheel is scrolled.

mlgh:
How to reproduce: enable mouse mode, click somehwere and start dragging a selection. Then scroll up with a wheel. Now selection doesn't react on mouse move, though in rxvt/termite you would expect it to follow the mouse.

The m->valid part is about pressing page up while dragging mouse.

I'm not quite sure about the m->valid. It seems that no one uses m->key when m->valid is 0,  maybe you know the actual meaning of this variable?

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -421,7 +421,7 @@ have_event:
 		m->wp = -1;
 
 	/* Stop dragging if needed. */
-	if (type != DRAG && c->tty.mouse_drag_flag) {
+	if (type != DRAG && type != WHEEL && c->tty.mouse_drag_flag) {
 		if (c->tty.mouse_drag_release != NULL)
 			c->tty.mouse_drag_release(c, m);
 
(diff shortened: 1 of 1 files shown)


mlgh:
It seems like there is a bunch of artifacts connected with this behaviour: you can select a block of text with a mouse, then release mouse button, then move it to another place and scroll wheel up two times - and you will see that some of the lines look like they are not selected (those lines were not redrawn).

mlgh:
Perhaps we could implement #642 here - in the following way:
If mouse is held, drag the selection, otherwise scroll-up scroll-down scrolls the screen without touching the selection. Perhaps modifying the selection when it is invisible should bring us back to selection.

nicm:
scroll-up and scroll-down commands should change to scroll the view but not
move the selection. You may have to to settle what happens when the
selection goes off screen though.

nicm:
Applied the wheel bit, thanks (I had the same diff in a tree here because someone else reported it, but they clearly never tested it so I dropped it).

It is not correct to set m->valid when it may not have been filled in by server_client_check_mouse. In fact we should probably be clearing the valid flag if that function fails.

mlgh:
Thanks!
I'll try to find a correct way to restore m->valid when dragging mouse and pressing non-mouse button.

PR #651, opened 2016-11-15
Initialize modeprefix with 1, otherwise first send-keys will be a no-op

mlgh:
reproduce: launch tmux, select something with mouse and press page up - the first one will be a no-op because modeprefix is 0 (and modeprefix is a synonym to sendkeys -N argument)


nicm:
Applied thanks.

PR #654, opened 2016-11-17
Add MOUSEDRAG*UPDATE key. Make mouse not valid if key is not set.

mlgh:
reproduce: copy mode, select anything with mouse, press page up, now mouse doesn't drag selection.
Please tell if you like the following approach, then we could reduce copypaste somehow, I believe.

Also, maybe MouseDrag* should be renamed into MouseDragBegin* ? But this will break peoples tmux configs.


nicm:
Your explanation is not very good, why does it no longer drag the selection after page up? A page up key press should not cancel mouse drag.

mlgh:
Whenever you start dragging the mouse, `server_client_check_mouse` returns `MOUSEDRAG1` event and 
https://github.com/tmux/tmux/pull/654/files#diff-bf88952e84cb9bf8e86922316b3d1eafR771
it hits the `m->valid = 1` line.
Then the m->tty.mouse_drag_update is set to window_copy_drag_update.

So you keep dragging the mouse, and `DRAG` event comes. it hits this line 
https://github.com/tmux/tmux/pull/654/files#diff-bf88952e84cb9bf8e86922316b3d1eafR475
and calls `mouse_drag_update` and then returns `KEYC_UNKNOWN`
This leads us to situation where `m->key = MOUSEDRAG1` and `m->valid = 1` but the event actually is not MOUSEDRAG1 anymore. So m->valid = 1 is just because of some old event.

Now I press page up, it hits
https://github.com/tmux/tmux/pull/654/files#diff-bf88952e84cb9bf8e86922316b3d1eafR777
this line `m->valid = 0` and after you drag mouse a little bit more, the event is just ignored since `server_client_check_mouse` will return `KEYC_UNKNOWN`.

I propose to fix this by adding a new type of event `MOUSEDRAG*UPDATE` - in this case after page up is pressed and you continue dragging the mouse, `server_client_check_mouse` will return not `KEYC_UNKNOWN`, but `MOUSEDRAG1UPDATE`, and `m->valid` will be turned to `1`.
Also this pull request sets `m->valid` to `0` whenever `server_client_check_mouse` fails, that is - we no longer have some old event make it look like `m->valid` is `1`.

I have been using this patch since I've made the pull request and hadn't met any issues with it.

nicm:
I don't think we need actual keys for this, just a marker. Does this work? [x.diff.txt](https://github.com/tmux/tmux/files/611464/x.diff.txt)

mlgh:
Yeah, I was not sure if `KEYC_MOUSE` was ok, so your patch is better! Everything works like a charm.
Also, don't we want to set `m->valid = 0` for `m->key = KEYC_UNKNOWN`? That's missing in x.diff.txt

nicm:
Applied, but you are right about m->valid and in fact there is another bug. It should be set when we call the drag callback.

So I think it is better to use a new KEYC_DRAGGING and fire the callback later: 
[y.diff.txt](https://github.com/tmux/tmux/files/611574/y.diff.txt)

mlgh:
Are you still going to commit y.diff.txt? Since right now m->valid may still be 1 for invalid events. https://github.com/tmux/tmux/commit/84319aa8f013238ccc1bbd6ea44b9e6be7c58db2

nicm:
Yes I will.

nicm:
Applied now, thanks.

mlgh:
Looks great, thanks!

PR #658, opened 2016-11-20
Autoconfigure pledge().

cgull:
On older versions of OpenBSD, `pledge()` does not exist.  Use Autoconf to detect it.


ThomasAdam:
Hi,

This isn't necessary.  Although tmux-portable should be buildable by OpenBSD, there's no need for it to, and we make no guarantees that it's possible.  These changes would make it awkward with respect to the code in OpenBSD.  I won't be applying this.

PR #661, opened 2016-11-21
Regular expression support for capture-pane

ericpruitt:
I have modified capture-pane to implement regular expression support. This is useful when searching for lines of text in scrollback buffers that have a lot of data in them but only a small subset of the data is of interest to the user. I'm not comfortable with autconf, so I have not made any changes to it. If you would like to merge this change, I would be happy to update the build system (if necessary) if you can provide some general guidance.

### DIFF
--- a/cmd-capture-pane.c
+++ b/cmd-capture-pane.c
@@ -100,7 +100,7 @@ cmd_capture_pane_history(struct args *args, struct cmdq_item *item,
 	const struct grid_line	*gl;
 	struct grid_cell	*gc = NULL;
 	int			 n, with_codes, error, escape_c0, join_lines;
-	u_int			 i, k, sx, top, bottom, tmp;
+	u_int			 i, sx, top, bottom, tmp;
 	char			*cause, *buf, *line, *subset;
 	const char		*Sflag, *Eflag, *regex;
 	size_t			 linelen;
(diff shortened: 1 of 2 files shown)


nicm:
I am not convinced by this, what are some use cases where you can't use
grep instead?

ericpruitt:
Nothing in my original message implies I cannot use grep; this is purely a performance optimization. If I dump a large buffer that's thousands of lines long, there's often a very significant amount of lag because tmux has to write out a bunch of data. On top of that, I typically end up ignoring 99% of the output when I grep the scrollback buffer.

nicm:
OK, but you didn't imply that you can either.

How big is your history and how long is the delay?

ericpruitt:
> OK, but you didn't imply that you can either.

Sorry; I thought "This is useful when searching for lines of text in scrollback buffers that have a lot of data in them but only a small subset of the data is of interest to the user" made my use case clear.

> How big is your history and how long is the delay?

The largest scrollback buffer limit I regularly use is 25,000 lines.

I just ran a test with a scrollback buffer populated with `for _ in {1..25000}; echo $LONG_LINE; done`, and this patch was slower than piping to both GNU grep (~2x faster) and whatever grep OpenBSD (only slightly faster) provides by default. As the number of lines is reduced, using the internal regular expressions becomes faster which is what I observed in my earlier tests. I didn't narrow it down to an exact number of lines, but I assume it depends on processor speed; whether or not the grep binary is cached in memory; and if it's not cached, how fast the filesystem its stored on is. I will close this pull request since its efficacy is dubious.

PR #662, opened 2016-11-21
try again to fix unicode 9 character widths

joshuarubin:
I know you're really sick of hearing about this issue and I'm not sure if it is reasonable for me to keep bringing it up, but I just wanted to give things one more try before giving up.

This patch removes utf8proc support which was a great idea but problematic for most users (myself included) in practice.

In its place a new library is included that calculates char widths strictly according to the unicode spec (which is the way the unicode 9 terminals I am aware of have built their own width tables). The library is called [`wcwidth9`](https://github.com/joshuarubin/wcwidth9) and it is _only_ for unicode 9. I pulled it together from neovim sources and have done comprehensive testing against utf8proc. `wcwidth9` supports identification of ambiguous width characters (which should be width 1 except in some east asian language contexts). It also separately identifies characters from the "private use area" which are also ambiguous width, but do not necessarily change with the east asian language context.

The width tables were taken from neovim and vim which use a script to generate them from the official unicode 9 documents for east asian character widths and emoji data.

This change is implemented as an opt-in build-time option (--enable-wcwidth9), but as it does not introduce another third party dependency. It could (and perhaps should) be implemented as a runtime (window level) option. I tried seeing what it would take to plumb it through as a runtime option, but it affected too much code for me to be comfortable even suggesting. If implemented as a runtime option, other options for how to handle ambiguous width and private use area characters could also be implemented.

Again, I understand if you don't want to reopen this can of worms, but I appreciate the opportunity to try to fix this one more time.


nicm:
The point of using wcwidth() and utf8proc is to pull the Unicode stuff
out of tmux and we are not adding it back.

Does utf8proc not plan to support Unicode 9?

In any case, I am not convinced we should worry about it unduly, it is
still relatively new, and systems will catch up.

joshuarubin:
> The point of using wcwidth() and utf8proc is to pull the Unicode stuff out of tmux and we are not adding it back.

Yeah, I figured.

> Does utf8proc not plan to support Unicode 9?

Well, it actually only supports unicode 9. But the widths it chooses are based on gnu unifont which often makes different choices than the unicode 9 spec. As terminals like iTerm2 support unicode 9 widths (optionally), they have implemented the widths according to the spec. As a result of the disagreements between utf8proc and the terminal width tables, there are still many rendering issues.

> In any case, I am not convinced we should worry about it unduly, it is still relatively new, and systems will catch up.

Roger that. I'm going to maintain my own branch for a while because this is a problem for me and others may find it useful. We'll see where things are down the road. Thanks for your time and for the great tool!

oblitum:
Thanks @joshuarubin! Your patch fixed all my issues on #1057, at last I can use Unicode with tmux! I've rebased your patch on top of master locally before applying. This really got messy trying to get solved, and the least worse option was to grab [`libutf8proc-julia`](https://aur.archlinux.org/packages/libutf8proc-julia), the most updated one, but still broken.

oblitum:
> The point of using wcwidth() and utf8proc is to pull the Unicode stuff out of tmux and we are not adding it back.

Despite being header only, wcwidth9 is still a dependency, just like the others, if all of them are managed outside the tmux project, and tmux links dynamically or statically with any of those, it seems there isn't much a difference.

oblitum:
False alarm, it didn't solve all my issues there.

PR #672, opened 2016-11-25
Remove lineflag as it seems to be broken and nobody uses it.

mlgh:
I'm removing it because it was implemented through window_copy_other_end and
now this function has different meaning.


andrey-starodubtsev:
fixes #701 as well

nicm:
Apparently people do use it, but it does need to be improved. Still let's defer removing it for now.

Issue #671, opened 2016-11-25
`-N` option for send-keys for non-copymode commands


choco:
I'm not sure if it is supposed to work or it is restricted to copy mode commands, but it would be nice if it was possible to use the `-N` option of `send-key` for everything.
Basically I'd like to be able to do:
```
send -N 3 Up
```
but right now this returns `not in mode` message

mlgh:
Don't we want to get rid of `modeprefix` and just issue the command `N` times in `cmd-send-keys.c`?
This will break `select-line`, (perhaps this is fixable somehow), other than that I don't immediately see any problems with such approach. This will make send-keys more uniform.

nicm:
I don't mind if -N changes to work without -X, but I suggest just adding a loop rather than removing modeprefix.

PR #675, opened 2016-11-28
Make -N work as repeat count in non-mode

mlgh:
Tries to solve #671 
e.g.
`bind-key -n C-o send -N 50 Up`
will result in Up being sent 50 times


nicm:
Applied with some trivial changes.

nicm:
Thanks

mlgh:
Great, thanks!

PR #705, opened 2016-12-30
Fix the cygwin tmux always creates new windows in HOME directory

dindinw:
Verify the issue by using the command
```shell
$tmux new-session -s test -n test -c "/usr/bin"
```
in the cygwin environment, the '-c' option never work, the new window always go to the HOME directory. 

The issue is caused by the extra `-` added to the `argv0` in lines:
```c
xasprintf(&argv0, "-%s", ptr + 1);
```
and
```c
xasprintf(&argv0, "-%s", wp->shell);
```
The issue is only happen in the cygwin tmux, the macos and linux is ok for using the wrong `argv0`, Just wondering why the extra `-` is needed, is it a typo?

### DIFF
--- a/window.c
+++ b/window.c
@@ -943,9 +943,9 @@ window_pane_spawn(struct window_pane *wp, int argc, char **argv,
 			fatal("execl failed");
 		}
 		if (ptr != NULL && *(ptr + 1) != '\0')
-			xasprintf(&argv0, "-%s", ptr + 1);
+			xasprintf(&argv0, "%s", ptr + 1);
 		else
-			xasprintf(&argv0, "-%s", wp->shell);
+			xasprintf(&argv0, "%s", wp->shell);
 		execl(wp->shell, argv0, (char *)NULL);
 		fatal("execl failed");
 	}


dindinw:
I think I know why the extra '-' is here, from the `man bash`
> A login shell is one whose first character of argument zero is a -,  or
one started with the --login option.

The problem is from the cygwin bash, which `bash --login` command always go to the $HOME directory. Please ignore the PR.
Thanks

nicm:
The - is correct, it means to create a login shell. Are you sure you don't
change to home directory in your shell profile?


On 30 Dec 2016 5:56 am, "Alex Wu" <notifications@github.com> wrote:

Verify the issue by using the command

$tmux new-session -s test -n test -c "/usr/bin"

in the cygwin environment, the '-c' option never work, the new window
always go to the HOME directory.

The issue is caused by the extra - added to the argv0 in lines:

xasprintf(&argv0, "-%s", ptr + 1);

and

xasprintf(&argv0, "-%s", wp->shell);

The issue is only happen in the cygwin tmux, the macos and linux is ok for
using the wrong argv0, Just wondering why the extra - is needed, is it a
typo?
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/tmux/tmux/pull/705
Commit Summary

   - Fix the cygwin tmux always creates new windows in HOME directory,
   remove the extra '-' before the argv0

File Changes

   - *M* window.c <https://github.com/tmux/tmux/pull/705/files#diff-0> (4)

Patch Links:

   - https://github.com/tmux/tmux/pull/705.patch
   - https://github.com/tmux/tmux/pull/705.diff

—
You are receiving this because you are subscribed to this thread.
<https://github.com/tmux/tmux/pull/705>, or mute the thread

dindinw:
@nicm thanks for your reply, it's sure not the profile script problem.
```shell
$ mv .profile .profile_test
$ mv .bash_profile .bash_profile_test
$ cd /usr/bin
$ bash
$ pwd
/usr/bin
$ exit
exit
$ bash --login
$ pwd
/home/Alex
$ exit
logout
$ pwd
/usr/bin
```

nicm:
Then this is a bug in Cygwin.

dindinw:
@nicm sorry again, you are correct , my fault. do `profile` thing. see `/etc/profile` of cygwin
```shell
# Make sure we start in home unless invoked by CHERE
if [ ! -z "${CHERE_INVOKING}" ]; then
  unset CHERE_INVOKING
else
  cd "${HOME}" || echo "WARNING: Failed attempt to cd into ${HOME}!"
fi
```
So the problem is fixed, I can set `CHERE_INVOKING` to make the cheat for cygwin.  
Thank you again for your help.

PR #707, opened 2017-01-03
Implement ccc on a per-pane basis

s-gilles:
Each pane has a 256color palette.  Sending the OSC sequence

    \x1b]4;d;rgb:r/g/b\x1b\\

sets the dth palette entry to the RGB color denoted by r, g, b and redraws
the pane. In tty_colours_{bg,fg}, this palette is used for substitution
before the actual color is set, so that palette changing is implemented
via 24-bit color support.

The special-case logic in tty_colours is slightly altered to allow
fg/bg values of 8 to match the pane-specific palette instead of the host
terminal's defaults.

### DIFF
--- a/input.c
+++ b/input.c
@@ -162,6 +162,7 @@ enum input_esc_type {
 	INPUT_ESC_SCSG0_ON,
 	INPUT_ESC_SCSG1_OFF,
 	INPUT_ESC_SCSG1_ON,
+	INPUT_ESC_ST,
 };
 
 /* Escape command table. */
@@ -179,6 +180,7 @@ static const struct input_table_entry input_esc_table[] = {
 	{ 'E', "",  INPUT_ESC_NEL },
 	{ 'H', "",  INPUT_ESC_HTS },
 	{ 'M', "",  INPUT_ESC_RI },
+	{ '\\', "", INPUT_ESC_ST },
 	{ 'c', "",  INPUT_ESC_RIS },
 };
 
@@ -1188,6 +1190,9 @@ input_esc_dispatch(struct input_ctx *ictx)
 	case INPUT_ESC_SCSG1_OFF:
 		ictx->cell.g1set = 0;
 		break;
+	case INPUT_ESC_ST:
+		/* ST terminates OSC, but the state transition already did that */
+		break;
 	}
 
 	return (0);
@@ -1849,6 +1854,9 @@ input_exit_osc(struct input_ctx *ictx)
 	case 2:
 		screen_set_title(ictx->ctx.s, p);
 		server_status_window(ictx->wp->window);
+		break;
+	case 4:
+		window_pane_initc(ictx->wp, p);
 		break;
 	case 12:
 		if (*p != '?') /* ? is colour request */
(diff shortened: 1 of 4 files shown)


s-gilles:
I feel I made enough design decisions during this patch that I should
justify them.

 * Implementing initc via 24bit color:

   I'm only interested in this feature if I can have two programs in two
   separate panes, each with their own palette, and with ncurses reporting
   to each program a modifiable palette of 256 colors.  Since panes
   can be shuffled around, this implies that if tmux is to support `ccc`,
   it must be able to show a window that uses more than 256 colors at
   the same time. Therefore, tmux can only implement this feature by
   showing 256+ different colors concurrently, each specified by RGB
   triples, which is exactly what 24bit color is for.

 * Not using `tty_check_{bg,fg}()`

   My first approach to this patch was to have these functions perform
   the palette substitution.  I later backtracked on that because I found
   I had to create flags for whether substitution had occurred in order
   to support changing colors that had already been printed. Because
   `tty_colours()` already did similar work by special-casing 8, I put the
   logic there.

 * Adding `INPUT_ESC_ST`

   As I was tracking down a bug which turned out to be unrelated, I
   saw that the final `\x1b\\` of the palette changing sequence caused
   `input_esc_dispatch()` to log a warning that `\\` was unknown.
   I figured I should avoid submitting code that adds spurious warnings
   in response to supported sequences, and `INPUT_ESC_ST` seemed to be
   a decent way of handling that.

I believe the ugliest part of this patch is the use of `sscanf` to decode
the sequence.  If desired, I have no problem replacing it with a mass of
`if ('a' <= *p && *p <= 'f')...` blocks.

I do have a question going forward: What is
the correct path to change the terminfo entry for
tmux-256color? Changing it should be a simple as adding `ccc` and
`initc=\E]4;%p1%d;rgb\:%p2%{255}%*%{1000}%/%2.2X/%p3%{255}%*%{1000}%/%2.2X/%p4%{255}%*%{1000}%/%2.2X\E\\,`,
but I don't know if I should take it to the terminfo/ncurses maintainers
or if an actual tmux dev should do that.

nicm:
I think this is mostly fine. I don't see any particular need for this
because nothing really uses it, but doing it like this is OK.

Comments:

- You are adding 1 KB to every pane for something few will use, please
  allocate palette[] on demand instead.

- I am not convinced by the last_wp dance, particularly since it will
  apply even if the palette can't change because it hasn't been set. But
  it is OK for now.

- Please don't mix declarations and code in tty_colours.

- sscanf is fine, but I don't like that it is happening in
  window.c. Either do the whole lot in a function in tty.c, or parse the
  string in tty.c and pass a colour number into a
  window_pane_set_palette function or something.

- I think we will need to have OSC 104 as well.

nicm:
You also only handle one colour definition, but there could be
multiple. I think you will probably have to parse it differently.

s-gilles:
Thanks for the comments, I'll see if I can address them without making too much of a mess for a minor feature and reopen when it's ready.

nicm:
Great thanks

PR #710, opened 2017-01-05
Allow changing 256 colour palettes per-pane

s-gilles:
Each pane has a 256color palette.  OSC sequences 4 and 104 are
implemented: 4 sets the dth palette entry to the RGB color denoted by r,
g, b and redraws the pane, while 104 resets entries.  The RIS sequence
also resets the entire palette. In tty_colours_{bg,fg}, and tty_reset,
this palette is used for substitution before the actual color is set,
so that palette changing is implemented via 24-bit color support.

The special-case logic in tty_colours is slightly altered to allow
fg/bg values of 8 to match the pane-specific palette instead of the
host terminal's defaults. tty_reset and style_equal are also slightly
augmented.


s-gilles:
This is the update to #707.  Your comment about `last_wp` was completely correct, ncim, and I now think the correct way of handling that is in `tty_reset()`.  The history started getting a bit disjointed, with `tty_reset` changes happening around OSC 104 implementations, so I rebuilt the history and forced the branch.  Apparently github doesn't like that, so here's a brand new PR.

nicm:
Thanks, this looks good, but a few more comments:

- I know I said to do the parsing in tty.c, but that was a mistake, I
  actually meant input.c, sorry... that is where this should go - only
  input.c needs to care about the %d;rgb:%x/%x/%x form. So please move
  the functions, tty_osc_4 to input_osc_4 etc.

- I don't think your implementation of tty_osc_4 is an improvement on
  sscanf, I was thinking more of a loop of sscanf, something like:

  s = copy = xstrdup(p);
  for (;;) {
      cp = strsep(&s, ";");
      if (cp == NULL)
      	 break;
      if (sscanf(cp, "%u", &c) != 1) /* or strtoul/strtonum if you prefer */
      	 break;

      cp = strsep(&s, ";");
      if (cp == NULL)
      	 break;
      if (sscanf(cp, "rgb:%x/%x/%x", &r, &g, &b) != 3)
      	 continue;

      /* and so on */
  }
  free(copy);

- tty_osc_104 is all numbers, so it can use input_split and walk the
  param list like everything else?

- Is the < 0x100 check in style_default correct, don't you still want to
  check in the palette for colours with COLOUR_FLAG_256 set?

- Please use NULL not 0 for pointers.

s-gilles:
Thanks for looking at it (and sorry about tty.c - I was a bit confused myself).  I flipped back and forth internally on `sscanf` myself, and  chose to do the explicit `char`-checking because I was worried about lax handling of leading whitespace, etc. but it should be easy to use something like what you've provided, and I don't suppose it would be bad to be lenient in the OSC sequences.

Hopefully I can address your comments without mangling history this time; I'll re-open when complete.

s-gilles:
I believe I've addressed your comments.  I agree with checking in `style_default` for `COLOUR_FLAG_256`, and I used your `sscanf` loop close to verbatim for the OSC 4 sequences.  I wasn't immediately able to have `input_split` do what I wanted, so I used a very dumb `strtol` loop instead for the 104, but if you'd prefer I can spend a bit more time figuring out how to make it work.

As an idle comment, I'm a little unsure of why the `style_equal(&grid_default_cell, &wp->colgc);` is not inverted when checked in `window_redraw_active_switch()`, but I preserved the original meaning and it seems to work.

nicm:
OK just a couple more things:

- You see to be looking up default colours (colour 8) in the palette as
  colours 0 and 7, but that doesn't seem right. In xterm at least,
  redefining colours 0 and 7 does not change the defaults. I don't think
  we should either, colour 8 should be left alone.

- I think style_switch_implies_redraw would be better in window.c as
  something like window_pane_is_default_style, since it is only used by
  one function in there. Then it could also be static.

- I think your return value from style_switch_implies_redraw is inverted
  in the default palette == NULL case. Also there is no point in
  checking if wp != NULL - it has to be non-NULL or &wp->colgc will core
  too.

- It seems like this:

if (gc->bg & COLOUR_FLAG_256 || gc->bg < 0x100) && wp->palette[gc->bg & 0xff] != 0)

Would be better as a macro of some kind, but not sure what it would look
like. Maybe something like:

#define COLOUR_PALETTE(wp, c) \
	(((wp)->palette != NULL && (~(c) & COLOUR_FLAG_RGB) && (c) != 8) ? (wp)->palette[(c) & 0xff] : -1)

Or maybe just:

#define COLOUR_IN_PALETTE(c) ((~(c) & COLOUR_FLAG_RGB) && (c) != 8)

nicm:
I mean your function returns 0 for the palette cases meaning no switch
needed (ie the palette is default) but the style_equal will return 1 if it
is default not 0. No?


On 5 Jan 2017 8:03 pm, "s-gilles" <notifications@github.com> wrote:

As to 8, I'd been using st as my reference, which seems to treat 8 as
context-dependent 0/7. Given the casing in tty_colours() (or at least, how
I read it) I thought that was the intended behavior. xterm is certainly
more authoritative though, I'll see if I can follow that instead. Hopefully
it should only be a matter of removing code, not adding it.

I have no problem with the changes to style_switch_implies_redraw, but I'm
pretty sure that, in the palette == NULL case, I'm following the current
master behavior of window_redraw_active_switch(). Is it possible that it's
backwards in the current code?

I'll look into how often I need to check the palettes. There are certainly
a few places where the conditionals have gotten a bit long, but I'd prefer
to define as few macros as possible, so I'll see what cleans up the code
the best.

—

<https://github.com/tmux/tmux/pull/710#issuecomment-270742917>, or mute the
thread

s-gilles:
The function should definitely return 1 in the `palette != NULL` cases, because the switch is definitely needed - I  did make that mistake when I first wrote it, but 7d8dfb90 should correct it.  In the `palette == NULL` case, it's supposed to behave exactly like `style_equal(&grid_default_cell, &wp->colgc)`, since it replaces that.  I don't trust myself to not be missing a ! mentally somewhere, though - what would be a good test?  Setting `window-active-style` to something different than `window-style` and switching panes?

nicm:
Probably I misunderstood it I'll take another look when I get home.

On 5 Jan 2017 8:18 pm, "s-gilles" <notifications@github.com> wrote:

The function should definitely return 1 in the palette != NULL cases,
because the switch is definitely needed - I did make that mistake when I
first wrote it, but 7d8dfb9
<https://github.com/tmux/tmux/commit/7d8dfb908989d398a20f158c4a12e8e73a68a595>
should correct it. In the palette == NULL case, it's supposed to behave
exactly like style_equal(&grid_default_cell, &wp->colgc), since it replaces
that. I don't trust myself to not be missing a ! mentally somewhere, though
- what would be a good test? Setting window-active-style to something
different than window-style and switching panes?

—

<https://github.com/tmux/tmux/pull/710#issuecomment-270746139>, or mute the
thread

s-gilles:
By the way, it looks like exempting the default 8 from the lookup makes the changes much less intrusive.  `tty_reset` no longer needs a changed prototype, special cases of 8 that I had put in all throughout tty.c can be removed, etc.  Assuming I can satisfactorily address other comments, would you like me to redo the history?  As it is, there's a large patch adding some extra logic, and it looks like there will be a large patch which backs out half that logic.

nicm:
It doesn't matter I only use the final diff.



On 5 Jan 2017 8:59 pm, "s-gilles" <notifications@github.com> wrote:

By the way, it looks like exempting the default 8 from the lookup makes the
changes much less intrusive. tty_reset no longer needs a changed prototype,
special cases of 8 that I had put in all throughout tty.c can be removed,
etc. Assuming I can satisfactorily address other comments, would you like
me to redo the history? As it is, there's a large patch adding some extra
logic, and it looks like there will be a large patch which backs out half
that logic.

—

<https://github.com/tmux/tmux/pull/710#issuecomment-270755770>, or mute the
thread

s-gilles:
I believe I've addressed the latest round of comments.  Using xterm-style handling of 8 makes the diff a lot cleaner.  I've convinced myself through testing that the logic for `window_redraw_active_switch` is what it has to be, but given that, the name of `window_pane_is_default_style` seems a little off.  I may rename it after some thought, but I'll gladly take any suggestions.

nicm:
OK great, this is much tidier without the default colours and works very
well. Please take a look at this diff, mostly style tweaks but also:

- I don't really remember why window_redraw_active_switch() is the way
  it is, it seems weird, but let's assume it is correct :-). Anyway, I
  think we can just eliminate the new function altogether and do the
  check inline.

- Added errno checks for strtol in input_osc_4 and input_osc_104.

- I think we should continue rather than break if sscanf fails, it could
  be an X colour name, if we don't support it we should just ignore it.

- Make send-keys -R also reset the palette.

- Instead of the palette swapping bit you are doing for identify mode, I
  have added GRID_FLAG_NOPALETTE and use it for identify and the various
  modes.

? 710.diff
? tmux
Index: cmd-send-keys.c
RCS file: /cvs/src/usr.bin/tmux/cmd-send-keys.c,v
retrieving revision 1.36
diff -u -p -r1.36 cmd-send-keys.c
--- cmd-send-keys.c	11 Dec 2016 22:55:25 -0000	1.36
+++ cmd-send-keys.c	6 Jan 2017 08:41:58 -0000
@@ -112,8 +112,10 @@ cmd_send_keys_exec(struct cmd *self, str
 		return (CMD_RETURN_NORMAL);
 	}
 
-	if (args_has(args, 'R'))
+	if (args_has(args, 'R')) {
+		window_pane_reset_palette(wp);
 		input_reset(wp, 1);
+	}
 
 	for (; np != 0; np--) {
 		for (i = 0; i < args->argc; i++) {
@@ -128,8 +130,9 @@ cmd_send_keys_exec(struct cmd *self, str
 			if (literal) {
 				ud = utf8_fromcstr(args->argv[i]);
 				for (uc = ud; uc->size != 0; uc++) {
-					if (utf8_combine(uc, &wc) == UTF8_DONE)
-						window_pane_key(wp, NULL, s, wc, NULL);
+					if (utf8_combine(uc, &wc) != UTF8_DONE)
+						continue;
+					window_pane_key(wp, NULL, s, wc, NULL);
 				}
 				free(ud);
 			}
Index: input.c
retrieving revision 1.106
diff -u -p -r1.106 input.c
--- input.c	9 Dec 2016 21:39:27 -0000	1.106
+++ input.c	6 Jan 2017 08:41:59 -0000
@@ -18,6 +18,7 @@
 
 #include <sys/types.h>
 
+#include <errno.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
@@ -105,6 +106,9 @@ static void	input_set_state(struct windo
 		    const struct input_transition *);
 static void	input_reset_cell(struct input_ctx *);
 
+static void	input_osc_4(struct window_pane *, const char *);
+static void	input_osc_104(struct window_pane *, const char *);
+
 /* Transition entry/exit handlers. */
 static void	input_clear(struct input_ctx *);
 static void	input_ground(struct input_ctx *);
@@ -162,6 +166,7 @@ enum input_esc_type {
 	INPUT_ESC_SCSG0_ON,
 	INPUT_ESC_SCSG1_OFF,
 	INPUT_ESC_SCSG1_ON,
+	INPUT_ESC_ST,
 };
 
 /* Escape command table. */
@@ -179,6 +184,7 @@ static const struct input_table_entry in
 	{ 'E', "",  INPUT_ESC_NEL },
 	{ 'H', "",  INPUT_ESC_HTS },
 	{ 'M', "",  INPUT_ESC_RI },
+	{ '\\', "", INPUT_ESC_ST },
 	{ 'c', "",  INPUT_ESC_RIS },
 };
 
@@ -1141,6 +1147,7 @@ input_esc_dispatch(struct input_ctx *ict
 
 	switch (entry->type) {
 	case INPUT_ESC_RIS:
+		window_pane_reset_palette(ictx->wp);
 		input_reset_cell(ictx);
 		screen_write_reset(sctx);
 		break;
@@ -1188,6 +1195,9 @@ input_esc_dispatch(struct input_ctx *ict
 	case INPUT_ESC_SCSG1_OFF:
 		ictx->cell.g1set = 0;
 		break;
+	case INPUT_ESC_ST:
+		/* ST terminates OSC but the state transition already did it. */
+		break;
 	}
 
 	return (0);
@@ -1850,10 +1860,16 @@ input_exit_osc(struct input_ctx *ictx)
 		screen_set_title(ictx->ctx.s, p);
 		server_status_window(ictx->wp->window);
 		break;
+	case 4:
+		input_osc_4(ictx->wp, p);
+		break;
 	case 12:
 		if (*p != '?') /* ? is colour request */
 			screen_set_cursor_colour(ictx->ctx.s, p);
 		break;
+	case 104:
+		input_osc_104(ictx->wp, p);
+		break;
 	case 112:
 		if (*p == '\0') /* no arguments allowed */
 			screen_set_cursor_colour(ictx->ctx.s, "");
@@ -1960,4 +1976,76 @@ input_utf8_close(struct input_ctx *ictx)
 	screen_write_cell(&ictx->ctx, &ictx->cell.cell);
 
 	return (0);
+}
+
+/* Handle the OSC 4 sequence for setting (multiple) entries. */
+static void
+input_osc_4(struct window_pane *wp, const char *p)
+{
+	char	*copy, *s, *next = NULL;
+	long	 idx;
+	u_int	 r, g, b;
+
+	copy = s = xstrdup(p);
+	while (s != NULL && *s != '\0') {
+		errno = 0;
+		idx = strtol(s, &next, 10);
+		if (*next++ != ';')
+			goto bad;
+		if (errno == ERANGE && (idx == LONG_MAX || idx == LONG_MIN))
+			goto bad;
+		if (idx < 0 || idx >= 0x100)
+			goto bad;
+
+		s = strsep(&next, ";");
+		if (sscanf(s, "rgb:%2x/%2x/%2x", &r, &g, &b) != 3) {
+			s = next;
+			continue;
+		}
+
+		window_pane_set_palette(wp, idx, colour_join_rgb(r, g, b));
+		s = next;
+	}
+
+	free(copy);
+	return;
+
+bad:
+	log_debug("bad OSC 4: %s", p);
+	free(copy);
+}
+
+/* Handle the OSC 104 sequence for unsetting (multiple) entries. */
+static void
+input_osc_104(struct window_pane *wp, const char *p)
+{
+	char	*copy, *s;
+	long	idx;
+
+	if (*p == '\0') {
+		window_pane_reset_palette(wp);
+		return;
+	}
+
+	copy = s = xstrdup(p);
+	while (*s != '\0') {
+		errno = 0;
+		idx = strtol(s, &s, 10);
+		if (*s != '\0' && *s != ';')
+			goto bad;
+		if (errno == ERANGE && (idx == LONG_MAX || idx == LONG_MIN))
+			goto bad;
+		if (idx < 0 || idx >= 0x100)
+			goto bad;
+
+		window_pane_unset_palette(wp, idx);
+		if (*s == ';')
+			s++;
+	}
+	free(copy);
+	return;
+
+bad:
+	log_debug("bad OSC 104: %s", p);
+	free(copy);
 }
Index: screen-redraw.c
RCS file: /cvs/src/usr.bin/tmux/screen-redraw.c,v
retrieving revision 1.41
diff -u -p -r1.41 screen-redraw.c
--- screen-redraw.c	13 Oct 2016 20:27:27 -0000	1.41
+++ screen-redraw.c	6 Jan 2017 08:41:59 -0000
@@ -588,6 +588,8 @@ screen_redraw_draw_number(struct client 
 		gc.bg = active_colour;
 	else
 		gc.bg = colour;
+	gc.flags |= GRID_FLAG_NOPALETTE;
+
 	tty_attributes(tty, &gc, wp);
 	for (ptr = buf; *ptr != '\0'; ptr++) {
 		if (*ptr < '0' || *ptr > '9')
@@ -615,6 +617,8 @@ draw_text:
 		gc.fg = active_colour;
 	else
 		gc.fg = colour;
+	gc.flags |= GRID_FLAG_NOPALETTE;
+
 	tty_attributes(tty, &gc, wp);
 	tty_puts(tty, buf);
 
Index: tmux.h
RCS file: /cvs/src/usr.bin/tmux/tmux.h,v
retrieving revision 1.683
diff -u -p -r1.683 tmux.h
--- tmux.h	5 Jan 2017 09:07:16 -0000	1.683
+++ tmux.h	6 Jan 2017 08:41:59 -0000
@@ -576,6 +576,7 @@ enum utf8_state {
 #define GRID_FLAG_PADDING 0x4
 #define GRID_FLAG_EXTENDED 0x8
 #define GRID_FLAG_SELECTED 0x10
+#define GRID_FLAG_NOPALETTE 0x20
 
 /* Grid line flags. */
 #define GRID_LINE_WRAPPED 0x1
@@ -829,6 +830,8 @@ struct window_pane {
 
 	struct grid_cell colgc;
 
+	int		*palette;
+
 	int		 pipe_fd;
 	struct bufferevent *pipe_event;
 	size_t		 pipe_off;
@@ -857,6 +860,11 @@ struct window_pane {
 TAILQ_HEAD(window_panes, window_pane);
 RB_HEAD(window_pane_tree, window_pane);
 
+#define WINDOW_PANE_PALETTE_HAS(wp, c)			\
+	((wp) != NULL && (wp)->palette != NULL &&	\
+	    ((c) < 0x100 || (c) & COLOUR_FLAG_256) &&	\
+	    (wp)->palette[(c) & 0xff] != 0)
+
 /* Window structure. */
 struct window {
 	u_int		 id;
@@ -2125,6 +2133,9 @@ void		 window_pane_alternate_on(struct w
 		     struct grid_cell *, int);
 void		 window_pane_alternate_off(struct window_pane *,
 		     struct grid_cell *, int);
+void		 window_pane_set_palette(struct window_pane *, u_int, int);
+void		 window_pane_unset_palette(struct window_pane *, u_int);
+void		 window_pane_reset_palette(struct window_pane *);
 int		 window_pane_set_mode(struct window_pane *,
 		     const struct window_mode *);
 void		 window_pane_reset_mode(struct window_pane *);
Index: tty.c
RCS file: /cvs/src/usr.bin/tmux/tty.c,v
retrieving revision 1.218
diff -u -p -r1.218 tty.c
--- tty.c	7 Dec 2016 09:16:55 -0000	1.218
+++ tty.c	6 Jan 2017 08:42:00 -0000
@@ -47,11 +47,14 @@ static void	tty_force_cursor_colour(stru
 static void	tty_cursor_pane(struct tty *, const struct tty_ctx *, u_int,
 		    u_int);
 
-static void	tty_colours(struct tty *, const struct grid_cell *);
+static void	tty_colours(struct tty *, const struct window_pane *,
+		    const struct grid_cell *);
 static void	tty_check_fg(struct tty *, struct grid_cell *);
 static void	tty_check_bg(struct tty *, struct grid_cell *);
-static void	tty_colours_fg(struct tty *, const struct grid_cell *);
-static void	tty_colours_bg(struct tty *, const struct grid_cell *);
+static void	tty_colours_fg(struct tty *, const struct window_pane *wp,
+		    const struct grid_cell *);
+static void	tty_colours_bg(struct tty *, const struct window_pane *wp,
+		    const struct grid_cell *);
 
 static void	tty_region_pane(struct tty *, const struct tty_ctx *, u_int,
 		    u_int);
@@ -1518,7 +1521,7 @@ tty_attributes(struct tty *tty, const st
 	 * Set the colours. This may call tty_reset() (so it comes next) and
 	 * may add to (NOT remove) the desired attributes by changing new_attr.
 	 */
-	tty_colours(tty, &gc2);
+	tty_colours(tty, wp, &gc2);
 
 	/* Filter out attribute bits already set. */
 	changed = gc2.attr & ~tc->attr;
@@ -1548,7 +1551,8 @@ tty_attributes(struct tty *tty, const st
 }
 
 static void
-tty_colours(struct tty *tty, const struct grid_cell *gc)
+tty_colours(struct tty *tty, const struct window_pane *wp,
+    const struct grid_cell *gc)
 {
 	struct grid_cell	*tc = &tty->cell;
 	int			 have_ax;
@@ -1594,14 +1598,14 @@ tty_colours(struct tty *tty, const struc
 
 	/* Set the foreground colour. */
 	if (gc->fg != 8 && gc->fg != tc->fg)
-		tty_colours_fg(tty, gc);
+		tty_colours_fg(tty, wp, gc);
 
 	/*
 	 * Set the background colour. This must come after the foreground as
 	 * tty_colour_fg() can call tty_reset().
 	 */
 	if (gc->bg != 8 && gc->bg != tc->bg)
-		tty_colours_bg(tty, gc);
+		tty_colours_bg(tty, wp, gc);
 }
 
 void
@@ -1688,29 +1692,38 @@ tty_check_bg(struct tty *tty, struct gri
 }
 
 static void
-tty_colours_fg(struct tty *tty, const struct grid_cell *gc)
+tty_colours_fg(struct tty *tty, const struct window_pane *wp,
+    const struct grid_cell *gc)
 {
 	struct grid_cell	*tc = &tty->cell;
 	char			 s[32];
+	int			 colour = 0;
+
+	/* If this colour is changed in the palette, substitute. */
+	if ((~gc->flags & GRID_FLAG_NOPALETTE) &&
+	    WINDOW_PANE_PALETTE_HAS(wp, gc->fg))
+		colour = wp->palette[gc->fg & 0xff];
+	else
+		colour = gc->fg;
 
 	/* Is this a 24-bit or 256-colour colour? */
-	if (gc->fg & COLOUR_FLAG_RGB ||
-	    gc->fg & COLOUR_FLAG_256) {
-		if (tty_try_colour(tty, gc->fg, "38") == 0)
+	if (colour & COLOUR_FLAG_RGB ||
+	    colour & COLOUR_FLAG_256) {
+		if (tty_try_colour(tty, colour, "38") == 0)
 			goto save_fg;
 		/* Should not get here, already converted in tty_check_fg. */
 		return;
 	}
 
 	/* Is this an aixterm bright colour? */
-	if (gc->fg >= 90 && gc->fg <= 97) {
-		xsnprintf(s, sizeof s, "\033[%dm", gc->fg);
+	if (colour >= 90 && colour <= 97) {
+		xsnprintf(s, sizeof s, "\033[%dm", colour);
 		tty_puts(tty, s);
 		goto save_fg;
 	}
 
 	/* Otherwise set the foreground colour. */
-	tty_putcode1(tty, TTYC_SETAF, gc->fg);
+	tty_putcode1(tty, TTYC_SETAF, colour);
 
 save_fg:
 	/* Save the new values in the terminal current cell. */
@@ -1718,29 +1731,38 @@ save_fg:
 }
 
 static void
-tty_colours_bg(struct tty *tty, const struct grid_cell *gc)
+tty_colours_bg(struct tty *tty, const struct window_pane *wp,
+    const struct grid_cell *gc)
 {
 	struct grid_cell	*tc = &tty->cell;
 	char			 s[32];
+	int			 colour = 0;
+
+	/* If this colour is changed in the palette, substitute. */
+	if ((~gc->flags & GRID_FLAG_NOPALETTE) &&
+	    WINDOW_PANE_PALETTE_HAS(wp, gc->bg))
+		colour = wp->palette[gc->bg & 0xff];
+	else
+		colour = gc->bg;
 
 	/* Is this a 24-bit or 256-colour colour? */
-	if (gc->bg & COLOUR_FLAG_RGB ||
-	    gc->bg & COLOUR_FLAG_256) {
-		if (tty_try_colour(tty, gc->bg, "48") == 0)
+	if (colour & COLOUR_FLAG_RGB ||
+	    colour & COLOUR_FLAG_256) {
+		if (tty_try_colour(tty, colour, "48") == 0)
 			goto save_bg;
 		/* Should not get here, already converted in tty_check_bg. */
 		return;
 	}
 
 	/* Is this an aixterm bright colour? */
-	if (gc->bg >= 90 && gc->bg <= 97) {
-		xsnprintf(s, sizeof s, "\033[%dm", gc->bg + 10);
+	if (colour >= 90 && colour <= 97) {
+		xsnprintf(s, sizeof s, "\033[%dm", colour + 10);
 		tty_puts(tty, s);
 		goto save_bg;
 	}
 
 	/* Otherwise set the background colour. */
-	tty_putcode1(tty, TTYC_SETAB, gc->bg);
+	tty_putcode1(tty, TTYC_SETAB, colour);
 
 save_bg:
 	/* Save the new values in the terminal current cell. */
@@ -1826,6 +1848,9 @@ tty_default_colours(struct grid_cell *gc
 			gc->fg = agc->fg;
 		else
 			gc->fg = wgc->fg;
+
+		if (gc->fg != 8 && WINDOW_PANE_PALETTE_HAS(wp, gc->fg))
+			gc->fg = wp->palette[gc->fg & 0xff];
 	}
 
 	if (gc->bg == 8) {
@@ -1835,6 +1860,9 @@ tty_default_colours(struct grid_cell *gc
 			gc->bg = agc->bg;
 		else
 			gc->bg = wgc->bg;
+
+		if (gc->bg != 8 && WINDOW_PANE_PALETTE_HAS(wp, gc->bg))
+			gc->bg = wp->palette[gc->bg & 0xff];
 	}
 }
 
Index: window-choose.c
RCS file: /cvs/src/usr.bin/tmux/window-choose.c,v
retrieving revision 1.82
diff -u -p -r1.82 window-choose.c
--- window-choose.c	16 Oct 2016 19:04:05 -0000	1.82
+++ window-choose.c	6 Jan 2017 08:42:00 -0000
@@ -796,6 +796,7 @@ window_choose_write_line(struct window_p
 
 	last = screen_size_y(s) - 1;
 	memcpy(&gc, &grid_default_cell, sizeof gc);
+	gc.flags |= GRID_FLAG_NOPALETTE;
 	if (data->selected == data->top + py)
 		style_apply(&gc, oo, "mode-style");
 
Index: window-clock.c
RCS file: /cvs/src/usr.bin/tmux/window-clock.c,v
retrieving revision 1.21
diff -u -p -r1.21 window-clock.c
--- window-clock.c	13 Oct 2016 20:27:27 -0000	1.21
+++ window-clock.c	6 Jan 2017 08:42:00 -0000
@@ -230,6 +230,7 @@ window_clock_draw_screen(struct window_p
 			screen_write_cursormove(&ctx, x, y);
 
 			memcpy(&gc, &grid_default_cell, sizeof gc);
+			gc.flags |= GRID_FLAG_NOPALETTE;
 			gc.fg = colour;
 			screen_write_puts(&ctx, &gc, "%s", tim);
 		}
Index: window-copy.c
RCS file: /cvs/src/usr.bin/tmux/window-copy.c,v
retrieving revision 1.162
diff -u -p -r1.162 window-copy.c
--- window-copy.c	5 Jan 2017 09:07:16 -0000	1.162
+++ window-copy.c	6 Jan 2017 08:42:01 -0000
@@ -1158,6 +1158,7 @@ window_copy_write_line(struct window_pan
 	size_t				 size = 0;
 
 	style_apply(&gc, oo, "mode-style");
+	gc.flags |= GRID_FLAG_NOPALETTE;
 
 	if (py == 0) {
 		size = xsnprintf(hdr, sizeof hdr,
@@ -1360,6 +1361,7 @@ window_copy_update_selection(struct wind
 
 	/* Set colours and selection. */
 	style_apply(&gc, oo, "mode-style");
+	gc.flags |= GRID_FLAG_NOPALETTE;
 	screen_set_selection(s, sx, sy, endsx, endsy, data->rectflag, &gc);
 
 	if (data->rectflag && may_redraw) {
Index: window.c
RCS file: /cvs/src/usr.bin/tmux/window.c,v
retrieving revision 1.176
diff -u -p -r1.176 window.c
--- window.c	16 Nov 2016 09:22:16 -0000	1.176
+++ window.c	6 Jan 2017 08:42:01 -0000
@@ -447,24 +447,30 @@ window_set_active_pane(struct window *w,
 void
 window_redraw_active_switch(struct window *w, struct window_pane *wp)
 {
-	const struct grid_cell	*agc, *wgc;
+	const struct grid_cell	*gc;
 
 	if (wp == w->active)
 		return;
 
 	/*
 	 * If window-style and window-active-style are the same, we don't need
-	 * to redraw panes when switching active panes. Otherwise, if the
-	 * active or inactive pane do not have a custom style, they will need
-	 * to be redrawn.
+	 * to redraw panes when switching active panes.
 	 */
-	agc = options_get_style(w->options, "window-active-style");
-	wgc = options_get_style(w->options, "window-style");
-	if (style_equal(agc, wgc))
+	gc = options_get_style(w->options, "window-active-style");
+	if (style_equal(gc, options_get_style(w->options, "window-style")))
 		return;
-	if (style_equal(&grid_default_cell, &w->active->colgc))
+
+	/*
+	 * If the now active or inactive pane do not have a custom style or if
+	 * the palette is different, they need to be redrawn.
+	 */
+	if (WINDOW_PANE_PALETTE_HAS(w->active, w->active->colgc.fg) ||
+	    WINDOW_PANE_PALETTE_HAS(w->active, w->active->colgc.bg) ||
+	    style_equal(&grid_default_cell, &w->active->colgc))
 		w->active->flags |= PANE_REDRAW;
-	if (style_equal(&grid_default_cell, &wp->colgc))
+	if (WINDOW_PANE_PALETTE_HAS(wp, wp->colgc.fg) ||
+	    WINDOW_PANE_PALETTE_HAS(wp, wp->colgc.bg) ||
+	    style_equal(&grid_default_cell, &wp->colgc))
 		wp->flags |= PANE_REDRAW;
 }
 
@@ -829,6 +835,7 @@ window_pane_destroy(struct window_pane *
 	free((void *)wp->cwd);
 	free(wp->shell);
 	cmd_free_argv(wp->argc, wp->argv);
+	free(wp->palette);
 	free(wp);
 }
 
@@ -1089,6 +1096,37 @@ window_pane_alternate_off(struct window_
 	grid_destroy(wp->saved_grid);
 	wp->saved_grid = NULL;
 
+	wp->flags |= PANE_REDRAW;
+}
+
+void
+window_pane_set_palette(struct window_pane *wp, u_int n, int colour)
+{
+	if (n > 0xff)
+		return;
+
+	if (wp->palette == NULL)
+		wp->palette = xcalloc(0x100, sizeof *wp->palette);
+
+	wp->palette[n] = colour;
+	wp->flags |= PANE_REDRAW;
+}
+
+void
+window_pane_unset_palette(struct window_pane *wp, u_int n)
+{
+	if (n > 0xff || wp->palette == NULL)
+		return;
+
+	wp->palette[n] = 0;
+	wp->flags |= PANE_REDRAW;
+}
+
+void
+window_pane_reset_palette(struct window_pane *wp)
+{
+	free(wp->palette);
+	wp->palette = NULL;
 	wp->flags |= PANE_REDRAW;
 }

nicm:
GitHub is a bit poo and doesn't understand attachments, so here it is: 
[x.diff.txt](https://github.com/tmux/tmux/files/689402/x.diff.txt)

nicm:
Actually you don't need the errno check because of the range check anyway, so ignore that bit.

nicm:
[y.diff.txt](https://github.com/tmux/tmux/files/689421/y.diff.txt)

nicm:
OK, there is one problem remaining. If you try to change the palette in a non-Tc terminal, it does not fallback to 256 colours properly, because you are doing the mapping in tty_colours_fg not in tty_check_fg.

nicm:
Is there a reason you can't do it in tty_check_fg? Because the wrong colour would be saved in tc->fg?

s-gilles:
Let me look into why tty_check_fg isn't used.  It might be another thing which can be changed because of 8's special treatment. 

One thing though: `GRID_FLAG_NOPALETTE` is missing from window-clock.c around line 245 (the clock numbers are still being paletted).

s-gilles:
Whatever my original reasoning for not using `tty_check_{bg,fg}` was, it no longer applies.  xterm (which I haven't forced to have Tc) with `TERM=xterm-256color` also seems much happier, so I think it's probably a good idea to move the logic.

I'd rebase against latest master, but github's recommendations for that are to force push, and last time I force pushed it made the PR worthless, so whatever: here's the [latest diff](http://sprunge.us/MMGf?diff) (and [raw](http://sprunge.us/MMGf)), based on y.diff.txt.

nicm:
Looks great to me, seems to work fine.

nicm:
I am not entirely certain we should try to get this into the tmux terminfo entry, because it is only really works  as intended on full RGB colour terminals. Is anyone going to need this who wouldn't be able to enable it themselves?

s-gilles:
I suppose not - it will only work anyway if someone sets Tc anyway.  But perhaps it should be mentioned alongside the documentation for Tc?  Something like

    Tc    Indicate that the terminal supports the ‘direct colour’ RGB escape sequence
          (for example, \e[38;2;255;255;255m).  This will be used to implement the
          initialize_colo[u]r capability, used by e.g. ncurses if ‘ccc’ is added to tmux's
          terminfo file.

?

nicm:
I have applied this now, it will appear in GitHub next time the repository updates from OpenBSD (hourly I think). Many thanks for your work!

I've added a note similar to what you proposed to tmux(1), it isn't particularly helpful but at least it mentions the feature. Perhaps a more extensive FAQ entry might be appropriate, although I am a bit off the FAQ as it is just a bit of a dumping ground.

s-gilles:
Great!  Thank you for helping me through the revisions!

nicm:
There is a bug with aixterm colours (colours 90-97), they are currently mapped to palette entries 90-97 but they should in fact map to 8-15.

nicm:
Fixed this by replacing WINDOW_PANE_PALETTE_HAS with a function window_pane_get_palette.

s-gilles:
Sorry about that, you fixed it before I even saw the notification.

PR #714, opened 2017-01-11
tmux.1: Document pane border colour options.

simono:
These were added in 1.2 https://github.com/tmux/tmux/blob/23254f59daea7415c96bc323d3b48d35c1b21327/CHANGES#L626

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -3070,6 +3070,12 @@ see the
 option.
 Attributes are ignored.
 .Pp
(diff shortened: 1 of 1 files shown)


nicm:
We do not document the -fg, -bg, -attr variants because they are
deprecated in favour of -style. It is probably time to delete at least
some of them.

simono:
Ok, thanks.

PR #717, opened 2017-01-12
Add exec-client Command

jennamagius:
This patch adds a new command, "exec-client", that causes a running, attached client to detach and morph into a different process, specified by command. This is useful for when you connect to a remote system via ssh with tmux as your ssh command. If you then want to connect to a tmux session owned by a different user, without nesting, it is currently necessary to disconnect and reauthenticate to SSH. By doing exec-client /bin/bash, a client can "detach", even if its parent process is not an interactive shell.

NOTE: This implementation is a little slap-dash, I would appreciate feedback from people more familiar with the code base on better ways for implementing this. (Maybe without bumping PROTOCOL_VERSION? Maybe a more elegant way of acquiring the shell option?)


nicm:
Hi

Some comments:

This should not be a new command, please make it a flag (-E?) to
switch-client.

You can add message numbers at the end (after MSG_WAKEUP) without
bumping PROTOCOL_VERSION, please do that.

Instead of looking up default-shell in global_s_options, use the
client's session if it has one.

Don't use alloca() under any circumstances. This is a particularly bad
place to use it because shell_size and cmd_size come from the
use. malloc() it instead.

Otherwise it looks fine, although I'm not particularly convinced this is
needed.

nicm:
Or actually detach client not switch client

jennamagius:
Hello,

Thank you for your feedback. I have amended per your recommendations.

nicm:
In the MSG_EXEC case in client_dispatch_attached I think you should make the fatalx() check also that strlen(data) != datalen, as it is now a message with only one string (one \0) would run off the end of the buffer.

Otherwise this looks fine apart from minor style nits.

jennamagius:
Good call, I made a mental note that I needed to do that, then forgot to actually do it. Done now.

nicm:
OK applied with some mostly trivial changes, will be in GitHub next time
it syncs from OpenBSD. Thanks!

jennamagius:
Thank you! :)

PR #721, opened 2017-01-13
Display plural or singular of 'window(s)' and 'pane(s)' accordingly

santigl:


nicm:
I am happy with it as it is, I don't think we need the extra code to worry about this.

santigl:
Thanks for the quick reply, @nicm. 

But the added code is minimal, and it is not modifying any functionality code. All modifications are encapsulated inside the output formatting functions (which seem to be structured that way to allow for these kind of extensions, aren't they?).

PR #727, opened 2017-01-19
configure: check for ncursesw as well

The-42:
This makes tmux build on systems which only have _ncursesw_, which is ncurses with wide character support (built with `--enable-widec`). There may be room for improvement, but for now both ncurses and ncursesw are detected via pkgconfig reliably.


nicm:
Applied a slightly modified version. Thanks

PR #734, opened 2017-01-22
Simple handling of OSC 52 clipboard sharing

fpqc:
A much simpler and stripped-down no-nonsense implementation of OSC 52 clipboard sharing.  

Please note that in order to get this to actually work, I did have to recompile my TERMINFO db and add the line:
`Ms=\E]52;%p1%s;%p2%s\007,`

I was unable to get it to work with terminal overrides in the tmux configuration.

### DIFF
--- a/input.c
+++ b/input.c
@@ -22,6 +22,8 @@
 #include <string.h>
 #include <time.h>
 
+#include <resolv.h> 
+
 #include "tmux.h"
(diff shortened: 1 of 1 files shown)


nicm:
Applied with some trivial adjustments, should be in GitHub next time it syncs.

PR #746, opened 2017-01-29
add window name to break-pane

arwild01:
Surprised that the break-pane command didn't give a way to name the window that gets created.  Not 100% positive this is the "right" way to do things, but it appears to work in my limited testing.


nicm:
Applied to OpenBSD now, will be in GitHub soon. Thanks!

PR #752, opened 2017-01-31
Add missing wchar.h header when enable-utf8proc(HAVE_UTF8PROC)

zchee:
Fix compile error when `--enable-utf8proc`.
The `utf8proc_{wcwidth,mbtowc,wctomb}` functions uses `wchar_t` type as its argument, but `wchar.h` header was not included.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -22,6 +22,7 @@
 
 #include <limits.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


nicm:
Great, applied, thanks!

zchee:
@nicm Thanks :D

PR #756, opened 2017-02-02
Just use Terminfo instead of ncurses (if it exists)

arwild01:
I realized I don't need to overrule or manipulate pkg-config in any way... in fact we can use it more.

It turns out none of the functions used by tmux are actually in libncurses.so... they all fall on the libtinfo.so side of the split library.  Also, "tinfo" is supported in pkg-config just like ncurses is.

So if the platform has the tinfo packge, use it, else try ncurses, ncursesw, and finally curses.

I've updated the manual search to follow similar logic. 

As a plus this actually eliminates an unnecessary library from the tmux executable.  Obviously if tmux expands to use any of the functions from the "ncurses" side of the library then this logic would have to be revisited, but there's quite a bit of code in tmux today and none of it actually needs libcurses.so on RHEL/CentOS (and possibly other platforms) where the Terminfo library is separate from the ncurses library.

### DIFF
--- a/TODO
+++ b/TODO
@@ -87,9 +87,10 @@
 	  options list with each option having a type (window or session), then
 	  options on window, on session, and global. for window options we look
 	  window->session->global, and for session we look session->global.
-          problem: what about windows in multiple sessions? there are contexts
-          where we do not know which session, or where multiple choices makes
-          no sense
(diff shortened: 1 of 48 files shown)


arwild01:
I can certainly live with that.

nicm:
Applied, thanks!

PR #757, opened 2017-02-04
reverting key table changes.

amosbird:
Without clearing key tables for session switch, some configuration may lead to
undesired behavior after switching -- prefix indicator is on while not behaving
like prefix has been pressed, and one key event will be consumed silently.

Configuration that can reproduce such problematic behavior is in https://github.com/amosbird/tmuxconfig

### DIFF
--- a/cmd-attach-session.c
+++ b/cmd-attach-session.c
@@ -98,6 +98,7 @@ cmd_attach_session(struct cmdq_item *item, int dflag, int rflag,
 			environ_update(s->options, c->environ, s->environ);
 
 		c->session = s;
+		server_client_set_key_table(c, NULL);
 		status_timer_start(c);
 		notify_client("client-session-changed", c);
 		session_update_activity(s, NULL);
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -277,6 +277,7 @@ cmd_new_session_exec(struct cmd *self, struct cmdq_item *item)
 		} else if (c->session != NULL)
 			c->last_session = c->session;
 		c->session = s;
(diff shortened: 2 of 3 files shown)


nicm:
Try this instead please: [tmux-repeat.diff.txt](https://github.com/tmux/tmux/files/754741/tmux-repeat.diff.txt)

nicm:
Note i only changed switch-client,  attach-session and new-session need the same if you use those:

~~~~
diff --git a/cmd-attach-session.c b/cmd-attach-session.c
index d29c4dd1..de10193e 100644
--- a/cmd-attach-session.c
+++ b/cmd-attach-session.c
@@ -98,6 +98,8 @@ cmd_attach_session(struct cmdq_item *item, int dflag, int rflag,
                        environ_update(s->options, c->environ, s->environ);

                c->session = s;
+               if (!repeat)
+                       server_client_set_key_table(c, NULL);
                status_timer_start(c);
                notify_client("client-session-changed", c);
                session_update_activity(s, NULL);
diff --git a/cmd-new-session.c b/cmd-new-session.c
index 4e7b413b..deda88f2 100644
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -277,6 +277,8 @@ cmd_new_session_exec(struct cmd *self, struct cmdq_item *item)
                } else if (c->session != NULL)
                        c->last_session = c->session;
                c->session = s;
+               if (!item->repeat)
+                       server_client_set_key_table(c, NULL);
                status_timer_start(c);
                notify_client("client-session-changed", c);
                session_update_activity(s, NULL);

amosbird:
works fine, thanks.

PR #765, opened 2017-02-08
Update CONTRIBUTING

timotheecour:
might as tell users what to run (even better would be smthg like `brew doctor` or `brew config`)

### DIFF
--- a/CONTRIBUTING
+++ b/CONTRIBUTING
@@ -13,6 +13,11 @@ Please also include:
 - logs from tmux (see below);
 - at most one or two screenshots, if helpful.
 
+On ubuntu show output of:
+```
+tmux -V && ecoh $TERM && lsb_release -a
+```
+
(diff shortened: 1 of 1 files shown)


nicm:
Applied, but uname instead of lsb_release, thanks.

timotheecour:
ok thx. for reference: https://github.com/tmux/tmux/commit/9af5f3ff812044a157a120ba7e2c70283b3b36dd

timotheecour:
@nicm how is `uname -sp` better than `lsb_release -a` ?
http://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name quotes:
> lsb_release -a is likely going to be your best option for finding this information out, and being able to do so in a consistent way.

whereas:
```
uname -sp
Linux unknown
```

unless you don't care about which version of linux (i would assume it would matter in some cases)

PR #777, opened 2017-02-14
Interpret `tmux source-file` path relative to $PWD

sfiera:
Previously, they were interpreted relative to the server's current
directory, which was surprising when using tmux from the command line:

    $ cd ~ && tmux  # following commands within tmux session
    $ cd ~/.ssh
    $ file ../.tmux.conf
    /home/$USER/.tmux.conf: ASCII text
    $ tmux source-file ../.tmux.conf
    ../.tmux.conf: No such file or directory

This change considers the current directory when calling tmux from the
command-line. As a bonus, it lets you mark scripts for tmux with a
shebang:

    #!/usr/bin/env tmux source-file

Previously, you could do that and the file would work if called with an
absolute path or from the tmux server's $PWD, but not in general.

Implementation notes:

  * The code to determine the appropriate $PWD is copied from
    cmd-load-buffer.c. I'm not sure what the difference between c->cwd
    and c->session->cwd is, but I assume that load-buffer is correct on
    it.

  * I've chosen to escape all non-alphanumeric characters in $PWD when
    constructing the glob. This is over-cautious, but I think a
    conservative whitelist is safer than a blacklist.

  * I've preserved the existing behavior of -q, which interprets
    globbing failure differently from sourcing failure. For example:

        $ tmux source-file -q /etc/sudoers && echo yes || echo no
        yes
        $ tmux source-file -q ~root/.tmux.conf && echo yes || echo no
        no

    This happens because sudoers can be globbed but not read, whereas
    glob can't determine if ~root/.tmux.conf exists or not.

### DIFF
--- a/cmd-source-file.c
+++ b/cmd-source-file.c
@@ -49,21 +49,44 @@ cmd_source_file_exec(struct cmd *self, struct cmdq_item *item)
 {
 	struct args		*args = self->args;
 	struct client		*c = item->client;
+	struct session  	*s;
 	int			 quiet;
 	struct cmdq_item	*new_item;
 	enum cmd_retval		 retval;
+	char			*glob_pattern, *tmp;
+	const char		*cwd;
 	glob_t			 g;
 	u_int			 i;
 
-	quiet = args_has(args, 'q');
-	if (glob(args->argv[0], 0, NULL, &g) != 0) {
-		if (quiet && errno == ENOENT)
-			return (CMD_RETURN_NORMAL);
-		cmdq_error(item, "%s: %s", args->argv[0], strerror(errno));
-		return (CMD_RETURN_ERROR);
+	if (*args->argv[0] == '/')
+		glob_pattern = xstrdup(args->argv[0]);
+	else {
+		if (c != NULL && c->session == NULL && c->cwd != NULL)
+			cwd = c->cwd;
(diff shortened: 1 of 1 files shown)


nicm:
Good idea, but don't escape manually, use stravis() with VIS_GLOB, and
don't use strcpy (ever). Something like this would be better I think:

if (stravis(&tmp, cwd, VIS_GLOB) != 0)
	fatal("stravis failed");
xasprintf(&pattern, "%s/%s", tmp, args->argv[0]);
free(tmp);

nicm:
You will actually need to use utf8_stravis() not stravis() or it will
break paths containing UTF-8 characters.

sfiera:
Yes, thanks, that looks better. Since cmd-source-file.c is the only file that uses glob(3), I assumed there wasn't any glob-escaping function already in tmux, but I'd much rather use one if it's there.

It looks like utf8_stravis() doesn't need to be checked; no other caller is checking the result, so I haven't either. I tested with `cd ~/テスト`.

nicm:
Applied with some additional changes to tidy up loadb and saveb by adding some helper functions. Should be in GitHub next time it syncs up. Thanks!

sfiera:
Great, thanks for maintaining great software.

PR #778, opened 2017-02-15
Change "status bar" occurrences to "status line"

ben4ever:
This is to consistently use "status line" when talking about it which
simplifies searching through the man page for this term.


nicm:
Good, applied, thanks.

PR #785, opened 2017-02-19
Enhance osdep_get_name() on NetBSD

krytarowski:
Follow the cannonical solution to reset mib[5] for KERN_PROC2 calls to 0 to
retrieve the number of bytes required to allocate.

Don't overallocate buf, len is already in bytes.


nicm:
Applied thank.s

PR #808, opened 2017-03-06
clarify FAQ regarding OS X reattach-to-user-namespace

krader1961:
Mention a couple more OS X commands that may not work if
`reattach-to-user-namespace` isn't used.


nicm:
Applied with some changes, thanks!

PR #829, opened 2017-03-17
choose-tree: Mapped h / l keys to tree-collapse / tree-expand

gpakosz:


nicm:

PR #830, opened 2017-03-18
Update README file for installing from version control

JESii:
Instructions to get/build the latest from version control don't actually install tmux, but just build it. I ran the autogen and then ./configure && make and tmux wasn't installed.

So changed line to `./configure && make install` so that it's parallel to the instructions for building from a release tarball.

### DIFF
--- a/README
+++ b/README
@@ -27,7 +27,7 @@ To get and build the latest from version control:
 	$ git clone https://github.com/tmux/tmux.git
 	$ cd tmux
 	$ sh autogen.sh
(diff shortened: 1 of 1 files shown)


nicm:
I don't think this is necessary. Many people want to build from Git master without installing, and the release tarball instructions just above cover installing (I have clarified this in the text) and it is reasonable to expect people to infer that that the same "make install" command will do for master if they require.

JESii:
I like your clarification; thanks!

PR #831, opened 2017-03-18
Fix changing directions in linewise copy mode

osandov:
Since commit 7e6c2cb23868, if I enter copy mode with select-line, move
the cursor up a few lines, then move the cursor down, the selection
grows from the bottom instead of shrinking from the top. This fixes
that condition.


nicm:
Applied, thanks!

PR #834, opened 2017-03-21
tmux.c: TMPDIR for tmux sockets

thac0:
After consulting the documentation, I saw that TMUX_TMPDIR was used to
prefix tmux sockets. This was a surprise, since I have TMPDIR set in my
environment. Curious, I consulted the git history. Quoting 3034a71488ce:

    Over the years, I have seen $TMPDIR set up worse than /tmp many
    times, and don't know how this practice infected other parts of the
    system.

As it is standard, return to using TMPDIR to store tmux sockets. Per the
standard, TMPDIR is "a directory made available for programs that need a
place to create temporary files". [IEEE Std 1003.1-2008, Sec 8.3]

Fixes: 3034a71488ce ("Let's see if anyone screams...")
Signed-off-by: Joe Konno <joe.konno@intel.com>

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -143,7 +143,7 @@ session created, and continues to process the rest of the configuration file.
 .It Fl L Ar socket-name
 .Nm
 stores the server socket in a directory under
-.Ev TMUX_TMPDIR
+.Ev TMPDIR
 or
 .Pa /tmp
 if it is unset.
--- a/tmux.c
+++ b/tmux.c
@@ -118,7 +118,7 @@ make_label(const char *label)
 
 	uid = getuid();
 
-	if ((s = getenv("TMUX_TMPDIR")) != NULL && *s != '\0')
+	if ((s = getenv("TMPDIR")) != NULL && *s != '\0')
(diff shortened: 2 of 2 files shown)


thac0:
Also send via git-send-email to tmux-users@googlegroups.com

nicm:
I think we have been without TMPDIR long enough now that there is no need to restore it at this point.

Jesin:
It looks like `TMUX_TMPDIR` is primarily used for sockets. Most of the programs on my system that use per-user sockets put them in `XDG_RUNTIME_DIR`, which for me is `/run/user/$UID`. If we're going to move anything out of `TMUX_TMPDIR`, why not put it in `XDG_RUNTIME_DIR`, following the example of gpg-agent, gvfs, i3, pulse, [...]

nicm:
We are not going to move the sockets and we are not going to use XDG_RUNTIME_DIR. If you want the sockets created in a different places you can use one of the two existing mechanisms: TMUX_TMPDIR, the -f flag.

PR #850, opened 2017-04-04
handle pty devices from different namespaces

brauner:
Various programs that deal with namespaces will use pty devices that exist in
another namespace. One obvious candiate are containers. So far ttyname() was
incorrectly handling this case because the pts device stems from the
host and thus cannot be found amongst the current namespace's /dev/pts/<n>
entries. Serge Hallyn and I recently upstreamed patches to glibc that allow
ttyname{_r}() to correctly handle this case. At a minimum, ttyname{_r}() will
set errno to ENODEV in case it finds that the /dev/pts/<n> device that the
symlink points to exists in another namespace. This commit will allow tmux to
handle this case and behave correctly in a container.

Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/client.c
+++ b/client.c
@@ -220,11 +220,11 @@ client_main(struct event_base *base, int argc, char **argv, int flags,
 	struct cmd		*cmd;
 	struct cmd_list		*cmdlist;
 	struct msg_command_data	*data;
-	int			 cmdflags, fd, i;
+	int			 cmdflags, fd, i, ret;
 	const char		*ttynam, *cwd;
 	pid_t			 ppid;
 	enum msgtype		 msg;
-	char			*cause, path[PATH_MAX];
+	char			*cause, path[PATH_MAX], ttynam_in_ns[PATH_MAX];
 	struct termios		 tio, saved_tio;
 	size_t			 size;
 
@@ -283,7 +283,22 @@ client_main(struct event_base *base, int argc, char **argv, int flags,
 		if ((cwd = find_home()) == NULL)
 			cwd = "/";
 	}
-	if ((ttynam = ttyname(STDIN_FILENO)) == NULL)
+
+	/* In case STDIN_FILENO refers to a /dev/pts/<n> device existing in
+	 * another namespace or refers to a /dev/pts/<n> device that exists in
+	 * the current namespace but its st_dev and st_ino fields do not match
+	 * then ttyname{_r}() will set errno to ENODEV. In this case we have
+	 * tmux operate directly on the symlink itself.
+	 */
+	errno = 0;
+	ttynam = ttyname(STDIN_FILENO);
+	if (errno == ENODEV) {
+		ret = snprintf(ttynam_in_ns, PATH_MAX, "/proc/%d/fd/%d", getpid(), STDIN_FILENO);
+		if (ret < 0 || ret >= PATH_MAX)
+			return (1);
+		errno = ENODEV;
+		ttynam = ttynam_in_ns;
+	} else if (ttynam == NULL)
 		ttynam = "";
 
 	/*
--- a/tty.c
+++ b/tty.c
@@ -95,6 +95,10 @@ int
 tty_init(struct tty *tty, struct client *c, int fd, char *term)
 {
 	char	*path;
+	/* len(/proc/self/fd/) = 14 + len(64-bit-int-in-chars) = 21 */
+	#define MAXPROC_SYMLINKLEN (14 + 21)
+	char	tmppath[MAXPROC_SYMLINKLEN];
+	int	ret, saved_errno;
 
 	if (!isatty(fd))
 		return (-1);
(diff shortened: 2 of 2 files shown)


brauner:
This require a version of glibc that includes this patch (This commit will not
change tmux behavior wrt to an unpatched glib version.):


	Author: Christian Brauner <christian.brauner@canonical.com>
	Date:   Fri Jan 27 15:59:59 2017 +0100

	    linux ttyname and ttyname_r: do not return wrong results

	    If a link (say /proc/self/fd/0) pointing to a device, say /dev/pts/2, in a
	    parent mount namespace is passed to ttyname, and a /dev/pts/2 exists (in a
	    different devpts) in the current namespace, then it returns /dev/pts/2.
	    But /dev/pts/2 is NOT the current tty, it is a different file and device.

	    Detect this case and return ENODEV.  Userspace can choose to take this as a hint
	    that the fd points to a tty device but to act on the fd rather than the link.

	    Signed-off-by: Serge Hallyn <serge@hallyn.com>
	    Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

To test whether you are running a patched glibc version you can do:

        sudo  unshare --fork --pid --mount-proc

and then to mount a new devpts instance:

        mount -t devpts devpts /dev/pts

If you call

        tty

you should see "not a tty". Now you can compile this test program:

```
/*
 *
 * Copyright © 2017 Christian Brauner <christian.brauner@ubuntu.com>.
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License version 2, as
 * published by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License along
 * with this program; if not, write to the Free Software Foundation, Inc.,
 * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */

#define _GNU_SOURCE
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

/* Test whether this is a correctly patched version of ttyname(). */
int main(int argc, char *argv[])
{
        int fd;
        char buf[4096];

        fd = open("/proc/self/fd/0", O_RDONLY);
        if (fd < 0) {
                fprintf(stderr, "Could not open \"/proc/self/fd/0\": %s.\n",
                        strerror(errno));
                exit(EXIT_FAILURE);
        }

        if (!ttyname(fd)) {
                /* COMMENT(brauner): ENODEV will only be set by a patched
                 * glibc.
                 */
                if (errno == ENODEV) {
                        printf("ttyname(): The pty device might exist in a "
                               "different "
                               "namespace: %s\n",
                               strerror(errno));
                } else {
                        exit(EXIT_FAILURE);
                }
        }

        if (ttyname_r(fd, buf, sizeof(buf))) {
                /* COMMENT(brauner): ENODEV will only be set by a patched
                 * glibc.
                 */
                if (errno == ENODEV) {
                        printf("ttyname_r(): The pty device might exist in a "
                               "different "
                               "namespace: %s\n",
                               strerror(errno));
                } else {
                        exit(EXIT_FAILURE);
                }
        }

        exit(EXIT_SUCCESS);
}
```

Now, in this setting you can also test a patched tmux version. A non-patched
tmux version should refuse to run, a patched tmux version should run just fine.

brauner:
To see the combined `glibc` and `tmux` patch in action in a `LXD` container see:

[asciicast](https://asciinema.org/a/110681)

nicm:
/proc/PID is not portable, so it is a poor idea to use it like this. But we don't need it anyway: on all platforms except Cygwin, tmux only uses the tty as a label to identify the client, and to check if we are running inside tmux itself by comparing with the pty name.

If we add a separate member for the client name, then instead of using the ttyname, we can use a portable, fake name while still keeping the actual tty (if available) for comparing with the pty and for the client_tty format.

We can also remove tty->path and the ttyname() call in tty_init entirely, it should be the same as the one the client gave us.

Please try this instead: [x.diff.txt](https://github.com/tmux/tmux/files/893125/x.diff.txt)

brauner:
The whole patch would only affect Linux anyway I guess but yeah, if we can avoid `/proc/PID` sure, let's do it. I wasn't sure where else tmux relies on the actual tty name.

brauner:
@nicm, patch works fine according to my testing. Are you closing this one and pushing your proposed fix now?

nicm:
I will apply it at some point in the next week probably.

nicm:
Thanks

brauner:
Cool, can I close this one then?

nicm:
No leave it open or I will forget.

brauner:
Thanks!

nicm:
I have applied this to OpenBSD now, it will be in GitHub next time it syncs up. Thanks!

brauner:
@nicm, cool thanks! we've pushed a version of this patch to the Ubuntu tmux package. It already hit `17.04` proposed. Finally, tmux in containers. :)

PR #854, opened 2017-04-05
New 'monitor-activity-ignore-delay' option added

bocadillodeatun:
This is a pull request for a feature I found I needed for a particular use case, but I also understand that almost no one else might find it useful, so please feel free to ignore this pull request if you find this new functionality to be too obscure :)

A new parameter ("**monitor-activity-ignore-delay**") has been created that accepts a value (in milliseconds) and activity that takes place in a background window **before** the specified time (since the window changed from being the active one to its current background state) is ignored.

For example, let's say we set this value to "2000" (ie. "2 seconds") and we have a window that prints "hello" every 5 seconds. Under these conditions, if we change to another window on "second 1" after a "hello" is printed, the original window will be marked with the "activity flag" in 4 seconds (as expected). **However**, if we change to another window on "second 4" after "hello" is printed, the original window will be marked with the "activity flag" in **6 seconds** (1+5) instead of in 1 second (because there is a 2 seconds "ignore window")

Now let me explain why I needed this (in case it helps me convince you to accept this PR, hehe): I am using a vim mapping that calls a vimscript function which ends up invoking "tmux new-window". The thing is that it looks like vim, **after** executing this command (ie. after tmux changes to a new window) still writes a few extra characters to the terminal (they seem to be control sequences to reposition the cursor on the screen) which tmux rightfully detects as "new activity".
This means that **every time** I create a new tmux window from vim, the vim window is immediately marked with the "new activity" flag.
Thanks to the new parameter that this PR introduces, if I set it to, say, 100 ms, the problem disappears.

### DIFF
--- a/alerts.c
+++ b/alerts.c
@@ -222,6 +222,9 @@ alerts_check_activity(struct window *w)
 {
 	struct winlink	*wl;
 	struct session	*s;
+	struct timeval	now;
+	struct timeval	tv;
+	u_int			delayms;
 
 	if (~w->flags & WINDOW_ACTIVITY)
 		return (0);
@@ -231,11 +234,19 @@ alerts_check_activity(struct window *w)
 	TAILQ_FOREACH(wl, &w->winlinks, wentry)
 		wl->session->flags &= ~SESSION_ALERTED;
 
+	gettimeofday(&now, NULL);
+
 	TAILQ_FOREACH(wl, &w->winlinks, wentry) {
 		if (wl->flags & WINLINK_ACTIVITY)
 			continue;
 		s = wl->session;
 		if (s->curw == wl)
+			continue;
+
+		timersub(&now, &wl->window->background_time, &tv);
+		delayms = options_get_number(wl->window->options, "monitor-activity-ignore-delay");
+
+		if ((tv.tv_sec * 1000) + (tv.tv_usec / 1000) < delayms)
 			continue;
 
 		wl->flags |= WINLINK_ACTIVITY;
--- a/options-table.c
+++ b/options-table.c
@@ -627,6 +627,14 @@ const struct options_table_entry options_table[] = {
 	  .default_num = 0
 	},
 
+	{ .name = "monitor-activity-ignore-delay",
+	  .type = OPTIONS_TABLE_NUMBER,
+	  .scope = OPTIONS_TABLE_WINDOW,
+	  .minimum = 0,
+	  .maximum = INT_MAX,
+	  .default_num = 0
+	},
+
 	{ .name = "monitor-silence",
 	  .type = OPTIONS_TABLE_NUMBER,
 	  .scope = OPTIONS_TABLE_WINDOW,
--- a/session.c
+++ b/session.c
@@ -539,6 +539,9 @@ session_set_current(struct session *s, struct winlink *wl)
 	if (wl == s->curw)
 		return (1);
 
+	if ((s->curw != NULL) && (s->curw->window != NULL))
+		gettimeofday(&s->curw->window->background_time, NULL);
+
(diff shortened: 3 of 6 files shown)


nicm:
I don't think we want an additional option for this. A fully generic solution would be to add a new option type to express a range but i don't think it is worth it just for this.

Why not start with monitor activity off and turn it on after a delay using run-shell -b or in the script itself?

bocadillodeatun:
I see what you mean. After all this is a very specific use case and there might be a workaround (I'm still trying to find one) using scripting.

In case any one can help, this is a simple way to reproduce my "problem":

1. Add the following lines to your `~/.vimrc`:
```
function! TmuxPreviousWindow()
    silent! call system("tmux previous-window;")
endfunction

map <silent> <A-e> :call TmuxPreviousWindow()<cr>
```
(obviously this is a silly example, because if you just wanted to change tmux window, you would directly call a tmux mapping, not a vim one, but bare with me...)

2. Now create two tmux windows and, in the second one, open "vim" and press "Alt+e". You will be "moved" to the first tmux window *and* the original (second) vim window will be marked with the "activity flag" (indicated by a dark background in the tmux status bar).

So... in order to avoid this we could just do this:
```
silent! call system("tmux setw monitor-activity off; tmux previous-window;")
```
...but then "monitor-activity" detection would remain disabled for the vim window.

I have tried all these alternatives:
```
1) silent! call system("tmux setw monitor-activity off; tmux previous-window; tmux setw -t 2 monitor-activity on")
2) silent! call system("tmux setw monitor-activity off; tmux previous-window; (sleep 3 && tmux setw -t 2 monitor-activity on)")
3) silent! call system("tmux setw monitor-activity off; tmux previous-window; (sleep 3 && tmux setw -t 2 monitor-activity on 2>/dev/null)")
4) silent! call system("tmux setw monitor-activity off; tmux previous-window; nohup 'sleep 3 && tmux setw -t 2 monitor-activity on 2>/dev/null'")
5) silent! call system("tmux setw monitor-activity off; tmux previous-window; tmux run-shell -b 'sleep 3 && tmux setw -t 2 monitor-activity on 2>/dev/null'")
```

...but all of them fail for different reasons:
```
1) tmux window 2 is immediately marked with the "activity" flag
2) tmux window 2 is marked with the "activity" flag after 3 seconds (even if there is no activity)
3) Same as 2)
4) Same as 1), seems like "nohup" prevents "tmux" from "sending orders" to a session (makes sense)
5) Same as 2)
```

As I said on the original message, please feel free to discard this pull request. I'll try to update this thread with the workaround in case I find one :)

bocadillodeatun:
Let me fix something I said in the previous comment: when using "alternative 5" (which seems to be the more reasonable one)...
```
silent! call system("tmux setw monitor-activity off; tmux previous-window; tmux run-shell -b 'sleep 3 && tmux setw -t 2 monitor-activity on 2>/dev/null'")
```
... I said the vim window is marked with the activity flag after three seconds, but this is not true. The actual behaviour goes like this:
* After changing window, if (in the new window) I type anything (for example, simply press the "space" key) **before the sleep pause** (3 seconds), then the vim window is never flagged (which is what we want!)
* However, if I don't type anything for those 3 seconds and **then** I type anything (and either press [ENTER] or wait 1-2 seconds), the vim window is flagged :(

I originally didn't spot this because I was using a pause of just one second instead of 3.

There is something fishy going on...

In any case, as you said, it is probably not worth wasting more time investigating this. I guess I can live with simply disabling alerts when leaving a vim windows :)

nicm:
If you think you have found another problem, please try to reproduce without using vim (you can make a window that does delayed output with something like 'tmux new "sleep 0.1;echo"') then open an issue and I'll take a look.

PR #865, opened 2017-04-20
Remove getopt.h include

ericvw:
'getopt()' is provided by `<unistd.h>`.  Including `<getop.h>` is not
necessary unless `getopt_long()` or `getopt_long_only()` are needed,
which are GNU extensions.

This allows compilation for non-GNU platforms (e.g., AIX).


nicm:
Applied, and from compat.h too, thanks.

PR #875, opened 2017-04-22
freeing out variable before its reassignment.

devnexen:
controlling eventually 'optimistic' utf8_append return.


nicm:
Applied status.c, I think the other change is unnecessary. Thanks

PR #884, opened 2017-04-25
Fixed the way ANSI sequences are constructed

kibos1er:
ANSI sequences are now OK when changing attributes/colors.

Before
======

Server:
server-master
Captured buffer:
client-master

After
=====

Server:
server-kibos1er
Captured buffer:
client-kibos1er

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -678,6 +678,7 @@ grid_string_cells_code(const struct grid_cell *lastgc,
 	u_int	attr = gc->attr;
 	u_int	lastattr = lastgc->attr;
 	char	tmp[64];
+	int	attrchanged = 0;
 
 	struct {
 		u_int	mask;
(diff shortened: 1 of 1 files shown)


nicm:
Please give me a simple printf that shows this problem.

nicm:
Are you sure this is a problem with tmux? It should not be necessary to reapply SGR colours when an attribute changes.

kibos1er:
1. create a tmux session
2. type inside the tmux session
```
printf '\033[1;31mbold_red \033[0;31mnormal_red\n'; tmux capture-pane; tmux show-buffer | head -n 5
```
You should see that the second 'normal_red' text is not red (it is white in my terminal).

Jesin:
It shows up as red in my terminal, using the current tmux master accaff8ac0c73ed20ee01a98d863f4e63d7d40d4 and using termite as my terminal.

kibos1er:
Well, it is not working on my setup. I have checked with xfce4-terminal, terminology, xterm, gnome-terminal and the latest tmux-master. I will check this evening with termite and let you know.

nicm:
OK it is only when we do SGR 0 - it clears the colours too, so we need to add them back. Please try this instead: [xx.diff.txt](https://github.com/tmux/tmux/files/955311/xx.diff.txt)

nicm:
Hmm I don't really like how grid_string_cells works at all and it won't work correctly with RGB colours.

nicm:
Please try this one instead actually: [yy.diff.txt](https://github.com/tmux/tmux/files/955399/yy.diff.txt)

kibos1er:
Your patch works perfectly on my setup. Thanks.

nicm:

PR #891, opened 2017-04-27
Make man page to show the right value

delphinus:
The [CHANGES][] says the default value has been changed for `xterm-keys`. This fixes the man page for this.

[CHANGES]: https://github.com/tmux/tmux/blob/e4b4125310b27ef8e33f7652eb5f96fe7e362515/CHANGES#L39-L39


nicm:
Fixed thanks, but I just removed the default from the man page. It is too much of a pain to document all the defaults when it is trivial to see them just by running tmux.

delphinus:
OK! I think it is better too. 👍

PR #916, opened 2017-05-09
Replace Red-black tree with memory-efficient version

jstimpfle:
Hi,

tmux built from source on my amd64 machine currently weighs about 370K of object code

```
$ nm --print-size --size-sort --radix=decimal tmux|awk '{sum += $2} END{print sum}'
368812
```
26K thereof come from the RB_* functions (generated by BSD tree.h macros) alone.
```
$ nm --print-size --size-sort --radix=decimal tmux| grep RB_ | awk '{sum += $2} END{print sum}'
25811
```
That is 7% of the tmux code, or 40% of most CPU's instruction cache size!

The problem with BSD tree.h is that it doesn't share code between the macro-generated tree instances.

This pull request is to announce that I'm currently working on porting tmux over to the Red-black tree that I've been working on for the last couple of weeks.

My version is mostly generic machine code (about 2-3K depending on compilation flags); type-safety equivalent to BSD tree.h's is provided by macro-generated wrappers that weigh almost nothing. (Optionally comparison functions can be inlined for optimized comparison-directed operations like find, insert, delete. But the gains are marginal and this costs about 300-400 bytes per tree instance).

My version is also about 10% faster for all common operations and uses 8 bytes less of memory per node. But that might not matter for tmux - are there places where the Red-black tree is heavily used?

This pull request contains an example of the port for 3 of the 14 BSD tree.h instances currently used in tmux; When done for all 14 instances about 22K of object code will be saved. That is 6% of total tmux code or 34% of instruction cache size.

As can be seen, the port can mostly be automated with sed substitutions. In the next few days I will still improve the usability of my tree code to iron out the few places where more lines of source code are added than deleted.

Is there a chance of getting this into tmux? Are there any problems that need to be solved? I can adapt to further directions and can also design the interface to be even more similar to the BSD tree's interface.

Kind greetings
Jens Stimpfle


nicm:
No we are not going to do this in tmux, sorry. I'm not going to drop the existing, maintained, reviewed, tested OpenBSD trees for something new. If you can show some numbers indicating anywhere that the RB trees are a significant, real world performance hit, then I'll take a look, but a solution is unlikely to involve replacing all the RB trees in tmux.

You might want to try and get it into OpenBSD first, then tmux will use it, but I would recommend keeping the API unchanged as far as possible, probably trying to make the changes incrementally rather than replacing the whole lot at once, and getting some measured performance numbers rather than just raw sizes.

jstimpfle:
Late reply - thanks for looking at it. I'm not aware of any performance problems caused by the current tree implementation. Judging from the names it doesn't seem like the trees in tmux are used for anything performance-sensitive, but I don't know the tmux source well - and keeping a low cache footprint seems worthwhile if it doesn't come with other significant costs. (I fully understand that switching to an unknown implementation is very risky).

I will follow your suggestion and try to get the OpenBSD people interested. But first I'm going to see if I can make a shim for the existing Macro interface. Except for RB_LEFT, RB_RIGHT, RB_PARENT I think it should work.

Issue #11, opened 2015-06-04
Allow interpolating format variables before shell invocation


ezyang:
http://unix.stackexchange.com/questions/193943/passing-tmux-variable-into-shell-command-in-tmux-conf has a pretty decent discussion of the request. Allowing this would mean you could completely reimplement any format string via shell and would offer a lot more flexibility. Additionally, the behavior of shell commands format strings would be more like if-shell, which does interpolate format commands before running the shell.

nicm:
Try git please.

ezyang:
Yup, it's fixed. Looks like this commit:

{{{
commit 66edb3392b234ccd9a940039936edb34258c2102
Author: Nicholas Marriott <nicm@openbsd.org>
Date:   Sun Mar 24 09:33:35 2013 +0000

    Expand format variables in the run-shell and if-shell shell commands,
    from Thiago Padilha.
}}}

PR #927, opened 2017-05-13
add -a option to force ACS

yanma:
I added -a option to force ACS.

I'm Japanese and the line border of the Japanese font has double width.
When I split the terminal vertically, it cause overlap between the border and right terminal.
This option is implemented to fix this issue.

https://github.com/yanma/tmux/commit/aac7c65529bc361c08546b5fabe4ff1b5120d11e

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -1651,8 +1651,6 @@ server_client_dispatch_identify(struct client *c, struct imsg *imsg)
 		c->tty.flags |= TTY_UTF8;
 	if (c->flags & CLIENT_256COLOURS)
 		c->tty.term_flags |= TERM_256COLOURS;
-	if (c->flags & CLIENT_FORCEACS)
-		c->tty.flags |= TTY_FORCEACS;
 
 	tty_resize(&c->tty);
 
--- a/server-client.c
+++ b/server-client.c
@@ -1651,6 +1651,8 @@ server_client_dispatch_identify(struct client *c, struct imsg *imsg)
 		c->tty.flags |= TTY_UTF8;
 	if (c->flags & CLIENT_256COLOURS)
 		c->tty.term_flags |= TERM_256COLOURS;
+	if (c->flags & CLIENT_FORCEACS)
+		c->tty.flags |= TTY_FORCEACS;
 
 	tty_resize(&c->tty);
 
--- a/tmux.c
+++ b/tmux.c
@@ -210,13 +210,10 @@ main(int argc, char **argv)
 		flags = 0;
 
 	label = path = NULL;
-	while ((opt = getopt(argc, argv, "2ac:Cdf:lL:qS:uUVv")) != -1) {
+	while ((opt = getopt(argc, argv, "2c:Cdf:lL:qS:uUVv")) != -1) {
 		switch (opt) {
 		case '2':
 			flags |= CLIENT_256COLOURS;
-			break;
-		case 'a':
-			flags |= CLIENT_FORCEACS;
 			break;
 		case 'c':
 			free(shellcmd);
--- a/tmux.c
+++ b/tmux.c
@@ -210,10 +210,13 @@ main(int argc, char **argv)
 		flags = 0;
 
 	label = path = NULL;
-	while ((opt = getopt(argc, argv, "2c:Cdf:lL:qS:uUVv")) != -1) {
+	while ((opt = getopt(argc, argv, "2ac:Cdf:lL:qS:uUVv")) != -1) {
 		switch (opt) {
 		case '2':
 			flags |= CLIENT_256COLOURS;
+			break;
+		case 'a':
+			flags |= CLIENT_FORCEACS;
 			break;
 		case 'c':
 			free(shellcmd);
--- a/tmux.h
+++ b/tmux.h
@@ -1077,7 +1077,6 @@ struct tty {
 #define TTY_OPENED 0x20
 #define TTY_FOCUS 0x40
 #define TTY_BLOCK 0x80
-#define TTY_FORCEACS 0x100
 	int		 flags;
 
 	struct tty_term	*term;
@@ -1353,7 +1352,6 @@ struct client {
 #define CLIENT_DOUBLECLICK 0x100000
 #define CLIENT_TRIPLECLICK 0x200000
 #define CLIENT_SIZECHANGED 0x400000
-#define CLIENT_FORCEACS 0x800000
 	int		 flags;
 	struct key_table *keytable;
 
--- a/tmux.h
+++ b/tmux.h
@@ -405,6 +405,7 @@ enum tty_code_code {
 	TTYC_SS,	/* set cursor style, Ss */
 	TTYC_TC,	/* 24-bit "true" colour, Tc */
 	TTYC_TSL,	/* to_status_line, tsl */
+	TTYC_U8,
 	TTYC_VPA,	/* row_address, cv */
 	TTYC_XENL,	/* eat_newline_glitch, xn */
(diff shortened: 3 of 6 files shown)


nicm:
I'm not wild about this. Why are these symbols double-width in this font? Unicode codepoint widths are the same no matter what other symbols you are using.

yanma:
OK, I understood it is font matter. Then, how about this explanation?

Some fonts doesn't have Unicode border glyph. When tmux works with UTF-8 mode, it refers unsupported glyph and substitutes it with corresponding glyph. As a result, the border becomes discontinuous(looks like "dashed border").

On the other hand, NCURSES has extended characters graphic and it includes border(See: [tty-acs.c](https://github.com/tmux/tmux/blob/master/tty-acs.c#L32-L65)). So, when this option is enabled, continuous border can be used instead of unsupported dashed border. It helps using incomplete fonts which supports limited glyphs.

krader1961:
@yanma, I do not understand your explanation. Whether or not a font has a glyph for a specific Unicode code point is not relevant. As @nicm explained the character cell width (i.e., terminal column width) of a Unicode code point is defined by the Unicode standard. If your terminal, using your preferred font, does not do the right thing then you should be talking to the people maintaining your operating system or terminal.

yanma:
@nicm, @krader1961, Thank you for your comments.

I'm sorry to make you confuse. I talk about not only character width. I also understood the character width may be font-set, terminal, or OS issue. Basically I should talk with their maintainer.

But I still want to fix dashed border issue too. Some fonts don't support [box-drawing characters](https://en.wikipedia.org/wiki/Box_Drawing). When we use this kind of font with UTF-8 environment, it causes dashed border like below.

withoutacs

If tmux has the option to force ACS border set, we can use continuous border without changing base font as below.

withacs

I understand the dashed border issue still should be fixed by font creators because it is resolved when they support box-drawing characters. But as you may know, not all font creators have enough time to support so many glyph. Some of them voluntarily make fonts. So, I believe the option is still helpful to provide alternative way to draw the border properly.

nicm:
Yes, thinking again something like this isn't a bad idea because there are several broken terminals and fonts out there. But I don't know if a command line argument is the best way.

The reason we use UTF-8 line drawing characters is because putty doesn't support ACS and UTF-8 at the same time.

How do you force ncurses programs not to use UTF-8 line drawing?

yanma:
Yes, you are right. I think there are some alternative ways to realize it.

1. Command line argument(my implementation).
2. tmux.conf option.
3. Always use ACS for line drawing.
4. Fix font itself (e.g. replace unsupported UTF-8 line with other correct font by the soft such as fontforge).

Currently we have no choice except [4]. [3] is easy solution but it spoils the designer's effort if they support UTF-8 line border with their original design. So, I think [1] or [2] is the practical solution. If you don't prefer [1], how about [2]?

hattya:
I think this is related to [UAX #11: East Asian Width](http://unicode.org/reports/tr11/).

In that specification, BOX DRAWING characters are marked as East Asian Ambiguous (A), so that Japanese fonts are designed them as the fullwidth character.

@nicm
```sh
$ export NCURSES_NO_UTF8_ACS=0
```

hattya:
@yanma I have a patch for [[2]](https://gist.github.com/hattya/4768f15e84814eda644b59f30c414780).

yanma:
@hattya Great! It is logically OK and perfectly worked in my environment.

@nicm Could you check [solution [2]](https://gist.github.com/hattya/4768f15e84814eda644b59f30c414780) by @hattya? He realized this feature by another approach. Maybe you prefer this way because it doesn't use command line argument.

In fact, there are many local solutions are reported in Japan. I guess the situation is similar with other countries which have multi-width character set. To avoid this confusion, I hope tmux officially supports this feature. Thank you very much.

- http://emonkak.hatenablog.com/entry/2011/12/30/131055
- http://qiita.com/milu0_0_0/items/3aa45b9ff40df1bcc06c
- https://gist.github.com/silenvx/3895468

nicm:
I like that idea. The problem with it is you can only specify it globally, not per client. But maybe that is OK.

nicm:
An option is neat, but I'm not sure this is worth it. It may be better just to look at the NCURSES_NO_UTF8_ACS environment variable which people will have to set anyway.

nicm:
Try this please: [tmux-acs.diff.txt](https://github.com/tmux/tmux/files/999580/tmux-acs.diff.txt)

yanma:
Workable but opposite. According to the man, setting NCURSES_NO_UTF8_ACS tells ncurses to use Unicode values which correspond to the VT100 line-drawing glyphs. So if it is nonzero value, tmux shouldn't force ACS. Maybe the correct implementation is below.
```C
if (envent != NULL && atoi(envent->value) == 0)
```
Anyway, I think this solution is the best because NCURSES_NO_UTF8_ACS exactly has the meaning that I mentioned. I think tmux also should consider this environment variable to draw line characters.

nicm:
Is that how ncurses actually behaves though? It seems opposite to the name
of the variable. I'll take a look at the ncurses code when I get home.

On 14 May 2017 7:00 pm, "Hiroaki Yamazoe" <notifications@github.com> wrote:

Workable but opposite. According to the man, setting NCURSES_NO_UTF8_ACS
tells ncurses to use Unicode values which correspond to the VT100
line-drawing glyphs. So if it is nonzero value, tmux shouldn't force ACS.
Maybe the correct implementation is below.

if (envent != NULL && atoi(envent->value) == 0)

Anyway, I think this solution is the best because NCURSES_NO_UTF8_ACS
exactly has the meaning that I mentioned. I think tmux also should consider
this environment variable to draw line characters.

—
You are receiving this because you were mentioned.

<https://github.com/tmux/tmux/pull/927#issuecomment-301329002>, or mute the
thread

nicm:
You are right: I think the name of the variable is meant to be read as "no ACS if UTF-8". Anyway, please try this instead: [tmux-acs2.diff.txt](https://github.com/tmux/tmux/files/1000308/tmux-acs2.diff.txt)

nicm:
Wait wait wait, now I have actually read the man page and I see there is already a terminfo entry, so we can use that instead of mucking about with the environment.

Try this please: [tmux-acs3.diff.txt](https://github.com/tmux/tmux/files/1000342/tmux-acs3.diff.txt)

You will need to do:
~~~~
set -ag terminal-overrides ',*:U8=1'

nicm:
OK there were some bugs in that one, sorry. Please try this instead, hopefully it is the final version, it also makes the check for ACS a bit less confusing by doing it in one place: [tmux-acs4.diff.txt](https://github.com/tmux/tmux/files/1000370/tmux-acs4.diff.txt)

hattya:
https://github.com/mirror/ncurses/blob/v6.0/ncurses/tinfo/lib_setup.c#L584

It seems that `U8=0` use ACS, and `U8=1` use UTF-8.

nicm:
Ugh did I get it the wrong way round? This is confusing

nicm:
Yes, you are right: please try this one instead [tmux-acsX.diff.txt](https://github.com/tmux/tmux/files/1000996/tmux-acsX.diff.txt)

yanma:
I confirmed it works well!

So I replaced my pull request with tmux-acsX.diff.txt. It helps us to resolve many confusion about drawing border with UTF-8 environment. I also learned a lot from this discussion. Thank you very much.

nicm:
Good stuff, I have applied this to OpenBSD now, it will be in GitHub soon (next time it syncs up). Please let me know if you see any further problems after that happens.

PR #942, opened 2017-05-22
Add OSC 11 support for changing background color

bertnp:
Add support for OSC 11 escape sequences to allow the background color of
the current pane to be changed. For example, the following changes the
background color to red:
   printf "\033]11;rgb:ff/00/00\033\\"

More documentation on OSC 11 codes can be found here in the Operating System Controls section:
https://www.xfree86.org/4.8.0/ctlseqs.html

This allows shell color scheme scripts such as base16-shell to set
themes with correct background colors for individual panes. Note, in
order to set the color scheme only for the current pane, base16-shell
scripts should be called with "env TMUX= <script path>" to avoid sending
the escape sequence which tells tmux to pass the escape sequence through
to the terminal.

### DIFF
--- a/input.c
+++ b/input.c
@@ -110,6 +110,7 @@ static void	input_set_state(struct window_pane *,
 static void	input_reset_cell(struct input_ctx *);
 
 static void	input_osc_4(struct window_pane *, const char *);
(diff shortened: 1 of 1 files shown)


nicm:
Applied, thanks!

PR #949, opened 2017-05-28
Update man page to reflect default window size changes.

cgull:


nicm:
I have changed this to mention the initial 80x24 size, but I didn't mention control mode because I don't think many people use it and I don't want to spread mentions of it all over the manual. People who do use it will figure out what is going on quickly enough, of they can ask.

Thanks

PR #957, opened 2017-06-04
subtle mem leak

devnexen:


nicm:
free(NULL) is fine so there is no need for the extra conditional, otherwise fixed, thanks!

PR #965, opened 2017-06-08
Add "pane-set-clipboard" hook for when the terminal clipboard content is set using the OSC 52 escape sequences.

townba:
Add **pane-set-clipboard** hook for when the terminal clipboard content is set using the OSC 52 `xterm(1)` escape sequences.

For example, `set-hook -g pane-set-clipboard "run-shell '( tmux show-buffer | xsel --input --clipboard )'"` allows `tmux` to set the clipboard on X terminals that don't support OSC 52 much like it does on terminals that *do* support it.

Context: I've been using `bind-key C-c run-shell "( tmux show-buffer | xsel --input --clipboard )"` for a while for terminals that don't support OSC 52, but sometimes I'm using a terminal that does support it. Having such a hook allows me to either always call `xsel` or write a script that can determine whether or not to bother.

(I'm aware that it's not really a pane notification, but that's the context in which the clipboard operations are done.)


townba:
Ugh. Didn't want it to clutter things up with all those intermediate commits. I don't know if that's a problem; if so, I can redo this pull request.

nicm:
Applied, thanks. I forgot to credit you in the commit message though, sorry :-/.

Issue #478, opened 2016-07-14
Seemingly no differentiation between key and shift-key


izemmouri:
Hi,

As I tried to bind M-key and M-Shift-key to 2 different commands, I realized tmux didn't seem to differentiate between keys and shifted keys. I don't know if I'm missing something or if this was the intended behaviour. If it was indeed the intended behaviour, would you mind indicating me which file in the source code is responsible for it?

Best regards

krader1961:
What key are you trying to apply the [shift] modifier to? Because things like [shift]+[enter] don't normally make any sense unless you've done something unusual to your terminal configuration. Try removing the tmux bindings then run `od -tx1z` or `xxd` and press the keys you're trying to bind. I'm betting those commands will show the same sequence of characters with and without the [shift] modifier. In which case there is no way for tmux to distinguish the two cases.

izemmouri:
You opened my eyes. Thank you.
I guess the workaround would be to make my terminal send some escape sequence known to tmux for the [shift]+[enter] combination (like the sequence for a function key, say F3) and then bind F3 to a command in tmux.
Would there be any way to make tmux recognize an arbitrary key sequence that wouldn't conflict with the ones it already knows about? For example, if I made urxvt send "\e0pq" for shift+enter, would there be a way to tell tmux that "\e0pq" is, say, key "Senter", and the bind M-Senter to a command?

krader1961:
Unfortunately tmux only allows single character sequences or a few special key names defined in the terminfo definition. Possibly prefixed with `M-` which simply means an extra escape character is the first char of the sequence. So you can't bind an arbitrary sequence like you can in shells like zsh or fish. So you'll need to reuse the sequence sent by another key (e.g., [F3]).

PR #981, opened 2017-06-21
Allow user-defined escape sequences for keys

da-x:
These days in other terminal programs such as Zsh and Vim, it is possible to bind arbitrary escape sequences to actions. While tmux allows to pass right throught from the terminal program,
bindings them to tmux itself requires editing terminfo entries.

For example, consider an addition of a key binding for Ctrl-Shift-[, which is by default indistinguishable from Ctrl-[. one can modify terminal program settings such as the one of rxvt via the `Xdefaults` file:

    URxvt.keysym.C-braceleft      : \033[5;30012~

And then add the same bindings in `.tmux.conf`:

    define-key "\e[5;30012~" C-braceleft
    bind-key -n C-braceleft resize-pane -L 3

### DIFF
--- a/cmd-set-option.c
+++ b/cmd-set-option.c
@@ -240,6 +240,12 @@ cmd_set_option_exec(struct cmd *self, struct cmdq_item *item)
 		TAILQ_FOREACH(loop, &clients, entry)
 			server_client_set_key_table(loop, NULL);
 	}
+	if (strcmp(name, "user-keys") == 0) {
+		TAILQ_FOREACH(loop, &clients, entry) {
+			if (loop->tty.flags & TTY_OPENED)
+				tty_keys_build(&loop->tty);
+		}
+	}
 	if (strcmp(name, "status") == 0 ||
 	    strcmp(name, "status-interval") == 0)
 		status_timer_start_all();
--- a/key-string.c
+++ b/key-string.c
@@ -110,12 +110,17 @@ static const struct {
 static key_code
 key_string_search_table(const char *string)
 {
-	u_int	i;
+	u_int	i, user_idx;
 
 	for (i = 0; i < nitems(key_string_table); i++) {
 		if (strcasecmp(string, key_string_table[i].string) == 0)
 			return (key_string_table[i].key);
 	}
+
+	if (sscanf(string, "User%d", &user_idx) == 1)
+		if (user_idx < KEYC_NUSER)
+			return (KEYC_USER + user_idx);
+
 	return (KEYC_UNKNOWN);
 }
 
@@ -265,6 +270,10 @@ key_string_lookup_key(key_code key)
 		return ("MouseMoveStatus");
 	if (key == KEYC_MOUSEMOVE_BORDER)
 		return ("MouseMoveBorder");
+	if (key >= KEYC_USER && key < KEYC_USER + KEYC_NUSER) {
+		snprintf(out, sizeof(out), "User%d", (u_int)(key - KEYC_USER));
+		return (out);
+	}
 
 	/*
 	 * Special case: display C-@ as C-Space. Could do this below in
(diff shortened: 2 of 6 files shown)


da-x:
This PR addresses issue #478.

We would probably need to think a bit more on how to implement this feature, and also add the proper documentation for it.

nicm:
What is this useful for?

Don't add a new command please, this can be a flag to bind-key.

da-x:
It's useful for binding tmux commands to key combinations that don't have an associated terminfo entry.

A new command may make sense, because the mapping between the key sequence and its key combination name is singular, where there could be several 'bind-key' commands for the same key combination, for example a single key combo may be configured to do different actions under various tmux modes.

But I understand your concern. Originally I thought about extending the parsing of `terminal-overrides` instead of adding the 'define-key' command. Do you think that may be better?

nicm:
You can already do this by binding one of the existing function keys:
~~~~
set -as terminal-overrides ',*:kf48=\E[5;30012~'
bind -n C-S-F12 resize-pane -L 3
~~~~
It is annoying that the terminfo(5) capability names don't specify the modifiers but that's how it is.

So there are a couple of ways I think we could go.

Firstly - the nicest way to do this would be to allow escape sequences to be bound directly, so you could just do:
~~~~
bind '\e[5;30012~' resize-pane -L3
~~~~
This might be pretty complicated to implement and TBH I really doubt it is worth it.

Secondly - would be to try and make it easier to do, without using the function keys. In this case I would add say 10 new user defined keys (KEYC_USER0-KEYC_USER9, "User0" to "User9") and just have an array option to specify them, so you would do something like:
~~~~
set -s user-keys[0] '\e[5;30012~'
bind User0 resize-pane -L 3
~~~~
This would be much easier to implement.

da-x:
Yeah, overriding the terminfo for innocent entries seemed too hackish for me, to the extent that it drove me to implement this feature. 

I think that being able to name the new user codes makes a more readable configuration, so how about: 

```
set -s user-keys.anyname '\e[5;30012~'
bind User.anyname resize-pane -L 3
```
?

nicm:
No I don't want a new type of option or option name.

da-x:
I've re-implemented according to what you suggested.

nicm:
This looks good, however you need to detach and reattach all clients using that terminfo entry to make the option take effect (tty_keys_build is only called when a terminfo is first created). It would be possible to walk the option in tty_keys_next and strncmp(), sort of like we do for mouse and xterm keys. Perhaps that would be better?

It would be possible to drop the 20 user keys limit by starting them off a KEYC_USER base like you did before and special casing them in key_string_lookup_string and key_string_lookup_key. Then they would just be limited by the array size limit (1000). I'm not sure it is worth it though, is anyone really going to need more than 20?

Also you will need to add this to tmux.1.

da-x:
See this new implementation,

- Followed your suggestion to bring back the keycode handling of User<index> to be dynamic, allowing for 1000 user defined key codes. Supporting a large number may be useful considering a large number of combinations that can be made with the modifier keys Alt-Ctrl-Shift-Meta along with various keyboard keys.
- Acting on the setting of `user-keys` from `cmd-set-option` seems to fix the problem regarding existing terminals. However removing escape sequences from the parse tree is not currently supported. Perhaps we are fine with that, for the meanwhile.
- Described the new option array in the man page.

nicm:
If you do this you will need to clear the old keys out when the option
changes or rebuild the entire tree from nothing, it is not OK to just
leave them.

Please don't move OPTIONS_ARRAY_LIMIT into tmux.h. I'd just define the
user bits outside the enum:

#define KEYC_USER 0x000020000000ULL
#define KEYC_NUSER 1000

da-x:
I've made the requested changes.

nicm:
Great, applied to OpenBSD now with some minor tweaks. Should be in GitHub soon. Thanks!

Issue #983, opened 2017-06-26
defining macros


eri451:
tl;dr: I would be nice to be able to define some sort of macros.

I like to keep a session with interactive programs as background session ("apps"). And even get some into specific panes of an attached session.
It's super nice that tmux does that already like so:
     $ tmux swap-pane -d -s apps:2.0

but since I would like to grab multiple applications that way, wondering which window holds what app again it is not very nice. But macros would easily solve this.
So someone could go:
     $ tmux macro app1 swap-pane -d -s apps:1:0
    $ tmux macro app2 swap-pane -d -s apps:5:0
and so
    $ tmux app1
would apply the command.

I'm running:
$ uname -srv && tmux -V && echo $TERM
Linux 4.9.0-3-amd64 #1 SMP Debian 4.9.30-1 (2017-06-04)
tmux 2.3
xterm-256color

nicm:
You can write a script and then set it as a command alias.

eri451:
then I have to call 
    :so myscript<cr>
this would just work if the script is in my current working directory

PR #989, opened 2017-06-29
defining Macros

eri451:
in relation to feature request #983 I added in some functionalty to make
macros a thing.  macros shall give you the option do save some more
complex calls to tmux (e.g swap-pane -d -s foo:1.3 or new-window -c ./ "make")
as abbreviations or alias or macros depends on how you call it.
Keybindings can do that so I copied a lot of code from that component
and started from there.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,6 +90,7 @@ dist_tmux_SOURCES = \
 	cmd-list.c \
 	cmd-load-buffer.c \
 	cmd-lock-server.c \
+	cmd-macro.c \
 	cmd-move-window.c \
 	cmd-new-session.c \
 	cmd-new-window.c \
@@ -144,6 +145,7 @@ dist_tmux_SOURCES = \
 	layout.c \
 	log.c \
 	mode-tree.c \
+	macros.c \
 	names.c \
 	notify.c \
 	options-table.c \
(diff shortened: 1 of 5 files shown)


nicm:
You can do this already with command-alias.

eri451:
there is no command alias afaik:
    $ tmux alias
     unknown command: alias

nicm:
It's an option.

eri451:
tmux also supports user options which are prefixed with a ‘@’.  User
     options may have any name, so long as they are prefixed with ‘@’, and be
     set to any string.  For example:

           $ tmux setw -q @foo "abc123"
           $ tmux showw -v @foo
           abc123


like this?

nicm:
No:
~~~~
             command-alias[] name=value
                     This is an array of custom aliases for commands.  If an
                     unknown command matches name, it is replaced with value.
                     For example, after:

                           set -s command-alias[2] zoom='resize-pane -Z'

                     Using:

                           zoom -t:.1

                     Is equivalent to:

                           resize-pane -Z -t:.1

                     Note that aliases are expanded when a command is parsed
                     rather than when it is executed, so binding an alias with
                     bind-key will bind the expanded form.

eri451:
cool

nicm:
But I suggest you start at command-alias[100] or something, because the default ones start at 0.

eri451:
ohh man all this trouble ... but at least I had fun with C.

PR #994, opened 2017-07-02
Try C.UTF-8 locale if en_US.UTF-8 isn't usable

rfrancoise:
The en_US.UTF-8 locale may not always be usable on Linux, either because
the locales package isn't installed or because the user uses
another locale. However, modern distributions (at least Debian and
derivatives, and Fedora and derivatives) provide a fallback C.UTF-8
locale that is guaranteed to always be present and work, so try that if
en_US.UTF-8 isn't available.


nicm:
Applied, thanks!

PR #999, opened 2017-07-07
correct the shortcut of window splitting

jingpengw:
fix the issue: https://github.com/tmux/tmux/pull/999
the `"` should split vertically, and `%` should split horizontally!
Not sure whether this is a correct fix or not.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -254,13 +254,13 @@ Suspend the
 client.
 .It !
 Break the current pane out of the window.
-.It \&"
+.It %
 Split the current pane into two, top and bottom.
 .It #
 List all paste buffers.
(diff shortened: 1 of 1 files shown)


nicm:
It works correctly as it is.

nicm:
Man page looks correct to me? " is top-bottom like it says

jingpengw:
oh, I see. I was looking at https://gist.github.com/andreyvit/2921703
but isn't  `"` represent vertical split more resonable? it looks like vertical split.
`%` looks more like horizontal split.

nicm:
The time to change it was 10 years ago, we are not changing it now.

PR #1014, opened 2017-07-21
Fix failure to update old content at end of line.

Jesin:
In the latest tmux commit 26db50d6df220afc4530d379e81acf9010ae02d8, when a short line should replace a long line on the screen, old content from the end of the long line often sticks around on the right side of the screen, instead of getting erased like it should.

Actions affected by this bug include:
- scrolling with page-up and page-down in `copy-mode-vi`
- switching between tmux windows using the shortcuts for `previous-window` and `next-window`
- quitting vim

Affected terminal emulators and platforms include:
- thestinger/termite@d87e7c8269718f8408b80e31a7d5655b259cbf7a with thestinger/vte-ng@d5149e91e8670b06716cdbd7ba425297f4454723 on Arch Linux
- [ConnectBot v1.9.2](https://play.google.com/store/apps/details?id=org.connectbot) on Android 6.0, served by [OpenSSH 7.5p1](https://www.openssh.com/portable.html) on Arch Linux

It appears this bug was caused by commit 8c6ad5532076254ec138f9b598b41d035bdf68e7. Reverting that commit has fixed the bug on my systems.

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -885,7 +885,7 @@ tty_draw_line(struct tty *tty, const struct window_pane *wp,
 	u_int			 i, j, sx, nx, width;
 	int			 flags, cleared = 0;
 	char			 buf[512];
-	size_t			 len, old_len;
+	size_t			 len;
 
 	flags = (tty->flags & TTY_NOCURSOR);
 	tty->flags |= TTY_NOCURSOR;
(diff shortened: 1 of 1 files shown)


nicm:
Does this fix it? [x.diff.txt](https://github.com/tmux/tmux/files/1166892/x.diff.txt)

nicm:
Actually it won't but I just applied another fix to OpenBSD that should fix it, let me know it works once it reaches GitHub.

Jesin:
@nicm it looks like 26db50d6df220afc4530d379e81acf9010ae02d8 patched with https://github.com/tmux/tmux/files/1166892/x.diff.txt still leaves about 68 characters unerased. Is there some way you can speed up the fix propagation from OpenBSD to GitHub, or is there somewhere I can go to see the fix before it propagates?

Jesin:
Found it: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/tmux/tty.c.diff?r1=1.291&r2=1.292

Jesin:
Fixed by 358df1019153c473bea4f7ab8415c63131cf9b1a. Thanks.

PR #1016, opened 2017-07-24
Make select pane return error when encountering window edges.

amosbird:
The whole point of this is to execute custom shell scripts when pane navigation fails.
Specificly, one can move out the terminal emulator and active the next window right to it using window manager scripts. This is also consistent with the behavior of `swap-pane`.


nicm:
I don't like this, I don't want error messages when I navigate panes and it
is also even more contrary to how tmux usually works (it should probably
wrap at edges to be consistent). It should be fairly simple to write a
script to tell if the current pane is at an edge before you try and move
(use the pane_x window_x and so on formats), so you could do that. Or
maintain this patch privately.

nicm:
I would be open to formats to tell you explicitly if a pane is at the left
right top or bottom which would make things simpler for you.

amosbird:
ok, just add new formats to check if it's at the edge of current window.

amosbird:
Um, may I ask why this is closed?

nicm:
Is anything happening?

amosbird:
Huh? I was thinking you're ok with the new formats.

nicm:
I'm OK with them but someone needs to add them :-). Did I miss that?

amosbird:
Oh, I've added them in this pr.

nicm:
OK you updated the pr I see now I'll look when I get home

amosbird:
cool, thanks!

nicm:

PR #1033, opened 2017-08-13
Add 'delay' option to 'display-panes'

Crote:
This PR adds a `-d` option to the `display-panes` command, allowing you to override `display-panes-time`. This gives you the ability to use `display-panes` with different timeout values. One possible use case is to have a long delay for picking panes (to mark them or to switch to them), and a very short delay for identifying them, which could be combined with movement commands to quickly identify the pane you have moved to.

Additionally, it changes the meaning of a delay of 0. Previously, the indicators would simply never be shown. This changes it to mean that the indicators should be displayed until further input.

### DIFF
--- a/cmd-display-panes.c
+++ b/cmd-display-panes.c
@@ -37,8 +37,8 @@ const struct cmd_entry cmd_display_panes_entry = {
 	.name = "display-panes",
 	.alias = "displayp",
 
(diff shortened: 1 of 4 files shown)


nicm:
Applied with some minor changes, thanks. Will appear in GitHub later.

PR #1034, opened 2017-08-15
Run alert-* hooks based on *-action settings; add monitor-bell window option

townba:
* Change `alert-[activity|bell|silence]` hooks to run based on `[activity|bell|silence]-action` settings rather than only for non-current windows.
* Add `monitor-bell` window option to allow disabling bells for individual windows (nice when one window is particularly noisy; also matches `monitor-activity` and `monitor-silence`).

### DIFF
--- a/alerts.c
+++ b/alerts.c
@@ -30,12 +30,12 @@ static int	alerts_enabled(struct window *, int);
 static void	alerts_callback(int, short, void *);
 static void	alerts_reset(struct window *);
 
+static int	alerts_action_applies(struct winlink *, int);
 static int	alerts_check_all(struct window *);
 static int	alerts_check_bell(struct window *);
 static int	alerts_check_activity(struct window *);
 static int	alerts_check_silence(struct window *);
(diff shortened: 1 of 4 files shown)


nicm:
Thanks.

I have applied the BELL_* rename bits of this. I'm not going to rename the option structs though.

Can you please split the remainder into two diffs, one to add monitor-bell and one to change the alert behaviour?

One comment on the code: in alerts_set_message, the loop does nothing if !applies, but applies is not modified inside the function, so it could be checked outside the loop. Personally I would hoist both the action and applies checks into the callers and move the options lookup inside, so they all looked like:
~~~~
                if (action != ALERT_NONE && applies != 0)
                        alerts_set_message(wl, "Silence", "visual-silence");
~~~~

nicm:
Well, I actually didn't apply as-is but changed ALERT_ACTION_* to ALERT_*.

townba:
Acknowledged. Closing.

PR #1037, opened 2017-08-16
Add monitor-bell window option.

townba:
Add `monitor-bell` window option.


nicm:
Applied to OpenBSD now, thanks!

PR #1039, opened 2017-08-16
Allow multiple bells in non-current windows.

townba:
Allow multiple bells in non-current windows. Without this, a bell can sound only once in non-current windows.


ThomasAdam:
Is this necessary?  I can see that getting annoying.

townba:
I think the current behavior (allowing one bell in another window to come through but not subsequent bells) is at least surprising. (One-and-done makes sense for "activity" and "silence".) Pull #1037 adds a per-window `monitor-bell` option that could be used to mitigate annoyances.

nicm:
Applied, thanks!

PR #1041, opened 2017-08-17
Fix search when the match is at the end of the line.

townba:
Fix search when the match is at the end of the line. Fixes `search-forward` (finding and highlighting) and `search-backward` (highlighting).


nicm:
Good spot, fixed, thanks!

PR #1050, opened 2017-08-25
CHANGES: fix search-backward-incremental example

aspiers:
`search-backward-incremental` doesn't work with `command-prompt` unless the `-i` option is given in order to re-run the `search-backward-incremental` command each time a key is pressed.


nicm:
Applied, thanks.

PR #1051, opened 2017-08-30
Fix memory leaks in history clearing with BCE

da-x:
When history contains lines that utilize BCE, their line data is not being properly freed when history lines are being moved. This change fixes this memory leak.

The problem is reproduced with current tmux master branch. To reproduce the issue, I have a large file containing ANSI codes that I can`cat` multiple times while observing tmux server's memory usage growing indefinitely, even when the history limit is small.

### DIFF
--- a/grid-view.c
+++ b/grid-view.c
@@ -75,7 +75,7 @@ grid_view_clear_history(struct grid *gd, u_int bg)
 
 	/* Scroll the lines into the history. */
 	for (yy = 0; yy < last; yy++) {
-		grid_collect_history(gd, bg);
+		grid_collect_history(gd);
 		grid_scroll_history(gd, bg);
 	}
 	if (last < gd->sy)
@@ -100,7 +100,7 @@ grid_view_scroll_region_up(struct grid *gd, u_int rupper, u_int rlower,
     u_int bg)
 {
 	if (gd->flags & GRID_HISTORY) {
-		grid_collect_history(gd, bg);
+		grid_collect_history(gd);
 		if (rupper == 0 && rlower == gd->sy - 1)
 			grid_scroll_history(gd, bg);
 		else {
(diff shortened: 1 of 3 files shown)


nicm:
I don't get it, tell me how this fix works please.

nicm:
I think this is an error in grid_move_lines, not in grid_collect_history. The lines that are being overwritten should be freed, no matter the background colour for the new lines. Try this instead please:
~~~~
diff --git a/grid.c b/grid.c
index 83ad89ed..52231257 100644
--- a/grid.c
+++ b/grid.c
@@ -522,7 +522,7 @@ grid_move_lines(struct grid *gd, u_int dy, u_int py, u_int ny, u_int bg)
 	for (yy = dy; yy < dy + ny; yy++) {
 		if (yy >= py && yy < py + ny)
 			continue;
-		grid_clear_lines(gd, yy, 1, bg);
+		grid_clear_lines(gd, yy, 1, 8); /* 8 so they are freed */
 	}
 
 	memmove(&gd->linedata[dy], &gd->linedata[py],

nicm:
I think you may be right that grid_collect_history doesn't need to worry about the background colour as well.

da-x:
Yeah, I already tried the smaller version of the fix that you suggested,
but it didn't completely resolve the issue.

nicm:
Yes... try this please instead: [x.diff.txt](https://github.com/tmux/tmux/files/1263035/x.diff.txt)

nicm:
I want to change so if something is operating on lines that are or will be gone it should free, clear should be reserved for stuff that will stay on screen. It is confusing that clear with bg == 8 means free.

Likewise move_lines should only be used if it can be arbitrary line regions which are staying around, when we are just doing a simple move of the history I think it is simpler to memmove it directly (I'll change grid_clear_history too).

da-x:
Thanks, the patch you sent works.

nicm:
Applied now, thanks!

Issue #10, opened 2015-06-04
Clarify pull request stance in Readme


af:
Hi,

Now that Github is the primary repo for tmux, my initial assumption was that pull requests would be the preferred way to submit patches. Since [it is not](https://groups.google.com/forum/#!topic/tmux-users/rySuwnHcJJo), it would be nice to mention this in the readme. 

Even though the Readme does explicitly say "Patches should be sent by email to
the mailing list at tmux-users@googlegroups.com.", I can't be the only person that would assume PRs would also be welcome. If you don't want to receive them, a mention of this in the Readme would probably save you some time.

bruno-:
Also an idea: github has a nice feature where "guidelines for contributing" will be visibly prompted to a user before he/she opens a new issue or pull request as described in [this blog post](https://github.com/blog/1184-contributing-guidelines).

All that's needed it to create a `CONTRIBUTING` file in the project root. This could probably filter out some percentage of impulsive issue or PR openings and reduce the maintenance burden on the maintainers.

nicm:
Let's see what happens before jumping to add to the documentation.

And no more documentation files, we have too many already. README or not at all.

-------- Original message --------

fatih:
Checkout https://github.com/golang/go which doesn't not accept Pull Requests. They have a `CONTRIBUTING.md` and don't use Pull Requests at all.

aleksandrs-ledovskis:
Another option - just edit project description. Check [git/git](https://github.com/git/git) for example

PR #1062, opened 2017-09-09
Compact line's grid extended data when it enters history

da-x:
When working with a 50,000 line history, heavy with 24-bit color ANSI, I grew suspicious noticing the memory resident size of the tmux being 1.6GB. Some debugging discovered that the `extdsize` of many history lines was much greater than their `cellsize`. Regardless of the cause of that, this commit makes sure that the history's memory usage is kept closer to the optimum. After this fix, the memory resident size
of tmux under the same 50,000 line history scenario dropped down to 170MB.

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -94,6 +94,49 @@ grid_need_extended_cell(const struct grid_cell_entry *gce,
 	return (0);
 }
 
+/* Free up unused grid_cells. */
+static void
+grid_compact_line(struct grid_line *gl)
+{
+	int			new_extdsize = 0;
+	struct grid_cell	*new_extddata;
+	struct grid_cell_entry	*gce;
+	struct grid_cell	*gc;
+	u_int			px, extd_i;
+
+	if (gl->extdsize == 0)
+		return;
+
+	for (px = 0; px < gl->cellsize; px++) {
+		gce = &gl->celldata[px];
+		if (gce->flags & GRID_FLAG_EXTENDED)
+			new_extdsize += 1;
+	}
+
+	if (new_extdsize == 0) {
+		free(gl->extddata);
+		gl->extddata = NULL;
+		gl->extdsize = 0;
+		return;
+	}
+
+	new_extddata = xreallocarray(NULL, new_extdsize, sizeof *gl->extddata);
+
+	extd_i = 0;
+	for (px = 0; px < gl->cellsize; px++) {
+		gce = &gl->celldata[px];
+		if (gce->flags & GRID_FLAG_EXTENDED) {
+			gc = &gl->extddata[gce->offset];
+			memcpy(&new_extddata[extd_i], gc, sizeof *gc);
+			gce->offset = extd_i++;
+		}
+	}
+
+	free(gl->extddata);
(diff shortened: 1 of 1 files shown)


nicm:
Yes... I see the problem and this looks mostly reasonable, thanks.

Some style nits:

- all local variable declarations go at the top of the function not in inner blocks;
- no {}s around compound statement bodies when they are a single line;
- u_int not unsigned int please, and line up the variable names to the nearest tab;
- do not use struct copies please, use memcpy.

There is a slight performance hit (5% or so) for jot/seq 10000000 but since this has no extended cells, it must be from the first loop over cellsize. Adding this to the start of grid_compact_line fixes it:
~~~~
        if (gl->extdsize == 0)
                return;
~~~~
Also I don't think you need to use calloc when you are going to write over the whole thing, use xreallocarray:
~~~~
       new_extddata = xreallocarray(NULL, new_extdsize, sizeof *gl->extddata);
~~~~

da-x:
Fixed according to suggests

nicm:

frosencrantz:
The fix applied appears to be segfaulting in tree mode.

This is the stack trace I'm seeing:
Program received signal SIGSEGV, Segmentation fault.
grid_scroll_history (gd=gd@entry=0x1da06b0, bg=bg@entry=8) at grid.c:330
330             grid_compact_line(&gd->linedata[gd->hsize]);
(gdb) where
#0  grid_scroll_history (gd=gd@entry=0x1da06b0, bg=bg@entry=8) at grid.c:330
#1  0x000000000041d886 in grid_view_clear_history (gd=0x1da06b0, bg=bg@entry=8) at grid-view.c:79
#2  0x0000000000431758 in screen_write_clearscreen (ctx=ctx@entry=0x7ffc28a73830, bg=bg@entry=8) at screen-write.c:1128
#3  0x0000000000429459 in mode_tree_draw (mtd=0x1d9ec10) at mode-tree.c:489
#4  0x000000000044c475 in window_tree_key (wp=0x1ddcf90, c=0x1dcef00, s=<optimized out>, key=118, m=<optimized out>) at window-tree.c:988
#5  0x0000000000434d76 in server_client_handle_key (c=0x1dcef00, key=118) at server-client.c:1003
#6  0x000000000043f3db in tty_keys_next (tty=tty@entry=0x1dcf000) at tty-keys.c:704
#7  0x00000000004409b8 in tty_read_callback (fd=<optimized out>, events=<optimized out>, data=0x1dcf000) at tty.c:169
#8  0x00007fe0faaa2f53 in event_base_loop () from /usr/lib/libevent-1.4.so.2
#9  0x000000000042e33f in proc_loop (tp=0x1d8d530, loopcb=loopcb@entry=0x4377b0 <server_loop>) at proc.c:199
#10 0x0000000000437d2c in server_start (client=0x1d8cd30, base=base@entry=0x1d8c590, lockfd=lockfd@entry=5, lockfile=0x1d8cbc0 "") at server.c:201

nicm:
This will be fixed soon, thanks.

PR #1071, opened 2017-09-12
Add "%elif" and "%else" directives for configuration file; allow nesting "%if" directives

townba:
Add `%elif` and `%else` directives for configuration file. Allow nesting `%if` directives.

### DIFF
--- a/TODO
+++ b/TODO
@@ -22,7 +22,6 @@
 	  multiple attached sessions, and is the active window in multiple
 	  attached sessions?
 	* comparison operators like < and > (for #{version}?)
-	* %else statement in config file
 
 - improve monitor-*:
 	* straighten out rules for multiple clients
--- a/cfg.c
+++ b/cfg.c
@@ -26,6 +26,16 @@
 
 #include "tmux.h"
 
+struct cfg_cond {
+	size_t			line;
+	int			may_meet;
+	int			met;
+	int			in_else;
+	int			meets;
+	SLIST_ENTRY(cfg_cond)	entry;
+};
+static SLIST_HEAD(, cfg_cond) conds = SLIST_HEAD_INITIALIZER(conds);
+
 static char		 *cfg_file;
 int			  cfg_finished;
 static char		**cfg_causes;
@@ -111,7 +121,8 @@ load_cfg(const char *path, struct client *c, struct cmdq_item *item, int quiet)
 	char			*buf, *cause1, *p, *q, *s;
 	struct cmd_list		*cmdlist;
 	struct cmdq_item	*new_item;
-	int			 condition = 0;
+	struct cfg_cond		*condition = NULL, *tcondition;
+	enum { NA, IF, ELIF }	 if_type;
 	struct format_tree	*ft;
 
 	log_debug("loading %s", path);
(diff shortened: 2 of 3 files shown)


ThomasAdam:
I've not looked at this in much detail -- and although I appreciate `queue.h` has a lot of different list types, we've standardised on RB tree or TAILQs.  I would suggest the use of a TAILQ instead of a SLIST.

I can't help but think that this augmentation of the simple %if syntax which was already in place rather overlaps with formats in a way which doesn't appreciably make a lot more sense to me.

nicm:
I think SLIST is OK so long as you remember not to use SLIST_REMOVE. But I would have used TAILQ and there isn't much benefit to using SLIST in code like this.

I think at this point the %if/%endif/etc code should be broken out of load_cfg into one or more helper functions; it might be better to do this first before adding %else.

townba:
I'll change it to TAILQ – I was considering it before anyway. I'll also break `load_cfg` up for readability.

townba:
Here's the latest. (I considered moving the `cfg_dir_*` functions into another file such as `cfg-directive.c` and maybe using "directive" rather than the abbreviation "dir", but I decided to see how this goes first.)

Attached is an example [`.tmux.conf`](https://github.com/tmux/tmux/files/1323015/tmux.conf.txt) file that exercises nested `%if` directives and error handling.

nicm:
OK looks good. Please try this, with the following changes:

- I don't see why you need "met" and "meets" and "may_meet", it seems just "met" and "may_meet" (I renamed this to "skipped") would be enough. A condition is skipped if its parent condition was not met, and if it is skipped it and any else/elif/else also can't be met. Perhaps I am missing something though.
- I don't see the need for a function pointer table here, so just use if/else.
- Move more of the code out of load_cfg.
- Rename some of the variables and functions and add some comments. Personal taste probably but anyway.
- You should not need to be trimming spaces and handling comments so much: fparseln will remove comments, and load_cfg will trim spaces from start and end. So it is only necessary to trim trailing spaces before calling format_expand. And if we check for "%if " (with a space) etc then empty ones will be an error.
- I don't think we need %error and if so we can do it separately.

Anyway please try this: [x.diff.txt](https://github.com/tmux/tmux/files/1323864/x.diff.txt)

nicm:
Oh also you can't do TAILQ_REMOVE after free()... :-)

townba:
Thank you very much for the review, @nicm. Notes:
* You were right about only needing two variables to track conditions, though it needed more work to get nested `%if`s working properly. Fixed.
* Went with the `if`/`else if` approach to testing directives.
* Moved code out of `load_cfg` à la your approach.
* Used your names (and code comments) with just a few tweaks.
* Removed most of the whitespace and comment handling I'd added.
  * Note: `fparseln` is being called without a comment delimiter ([code](https://github.com/tmux/tmux/blob/ff526e43de84cff3770d32c0b6ea215a7359f521/cfg.c#L108)), so it doesn't remove `#` comments. Instead, I'm simply ignoring any text on a line after `%else` and `%endif`.
* I'll try to remember to remove `%error`. (I'm using it in my testing.)
* Moved the `free` after the `TAILQ_REMOVE`.
* Switched to use `TAILQ_FOREACH_REVERSE_SAFE` to dump out unterminated `%if`s so they're now from earliest to latest.

nicm:
Oh well if fparseln isn't stopping comments I suggest they not be allowed
after %else etc either, so anything except spaces should be invalid.

nicm:
Looks good, thanks. If you are going to skip spaces and still it is OK to make p char * instead of const char * and modify it, might be tidier. Anyway will have another look after 2.6 is released.

townba:
Text on a line after `%else` or `%endif` is now an error.

nicm:

PR #1072, opened 2017-09-13
Reset window name or pane title when attempting to rename/retitle using invalid UTF-8

townba:
Since the user can change the window name and pane title using escape sequences, it'd be nice to have a way to go back to the defaults using a similar method. This gives the user a way to change the window name and pane title back to their defaults using escape sequences (any non–UTF-8 name/title). For example, this would allow a terminfo entry to specify `dsl=\E]0;\377\E\\` to reset it.

### DIFF
--- a/input.c
+++ b/input.c
@@ -1956,8 +1956,10 @@ input_exit_osc(struct input_ctx *ictx)
 	case 2:
 		if (utf8_isvalid(p)) {
 			screen_set_title(ictx->ctx.s, p);
-			server_status_window(ictx->wp->window);
+		} else {
+			screen_set_title(ictx->ctx.s, NULL);
 		}
+		server_status_window(ictx->wp->window);
 		break;
 	case 4:
 		input_osc_4(ictx->wp, p);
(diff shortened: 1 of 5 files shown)


nicm:
I'm not happy inventing a new way to do this, how do you do it in other
terminals?

nicm:
xterm has a title stack, it might be nice to add that instead.

townba:
Acknowledged. What would you think of:
1. Interpreting setting an empty window name as setting the name to automatic. (This could either be universal or only for the escape sequence.)
2. Adding an xterm-like stack for the pane title.
3. Adding at least partial support for DECSSDT (Select Status Display Type) if not already there for the pane title. That'd give us a standard way of handling setting and resetting the pane title.

nicm:
I don't think we should invent new behaviour for this. So for the pane
title setting stuff (OSC 0,1,2) we should do what xterm does, and for
screen's stupid rename sequence (\033k) we should do what screen does.

I don't think setting an empty name should do anything magic.

Adding xterm pane title stack would be good.

What does xterm do for DECSSDT?




On 13 Sep 2017 8:58 am, "Brad Town" <notifications@github.com> wrote:

Acknowledged. What would you think of:

   1. Interpreting setting an empty window name as setting the name to
   automatic. (This could either be universal or only for the escape sequence.)
   2. Adding an xterm-like stack for the pane title.
   3. Adding at least partial support for DECSSDT (Select Status Display
   Type) if not already there for the pane title. That'd give us a standard
   way of handling setting and resetting the pane title.

—

<https://github.com/tmux/tmux/pull/1072#issuecomment-329089437>, or mute
the thread

PR #1073, opened 2017-09-13
Remove unused parameters from various functions

d-e-s-o:


nicm:

PR #1075, opened 2017-09-14
Add xterm-like stack for pane titles

townba:
Add xterm-like stack for pane titles.

* `\033[22;0t` pushes the current pane title onto the pane's stack. The title is unchanged.
* `\033[23;0t` pops the current pane title from the pane's stack and changes the pane title.

(References: http://invisible-island.net/xterm/ctlseqs/ctlseqs.pdf)

### DIFF
--- a/screen.c
+++ b/screen.c
@@ -24,6 +24,13 @@
 
 #include "tmux.h"
 
+struct screen_title_entry {
+	char				*text;
+
+	TAILQ_ENTRY(screen_title_entry)	 entry;
+};
+TAILQ_HEAD(screen_titles, screen_title_entry);
+
 static void	screen_resize_x(struct screen *, u_int);
 static void	screen_resize_y(struct screen *, u_int);
 
@@ -39,6 +46,8 @@ screen_init(struct screen *s, u_int sx, u_int sy, u_int hlimit)
 	s->cstyle = 0;
 	s->ccolour = xstrdup("");
 	s->tabs = NULL;
+	s->titles = xmalloc(sizeof *s->titles);
+	TAILQ_INIT(s->titles);
 
 	screen_reinit(s);
 }
@@ -47,6 +56,8 @@ screen_init(struct screen *s, u_int sx, u_int sy, u_int hlimit)
 void
 screen_reinit(struct screen *s)
 {
+	struct screen_title_entry *title_entry;
+
 	s->cx = 0;
 	s->cy = 0;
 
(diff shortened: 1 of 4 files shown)


nicm:
This looks good, but I think the code would be better in screen.c to match screen_set_title(). I think you should put struct title_entry in screen.c instead of tmux.h (use a pointer in struct screen and malloc it on first use).

Then have a small API that input.c can use that matches screen_set_title, something like:

~~~~
void screen_push_title(struct screen *);
void screen_pop_title(struct screen *);
~~~~

townba:
Thank you for the review and suggestions. PTAL.

screen.c
```diff
@@ -60,16 +71,30 @@ screen_reinit(struct screen *s)
 	grid_clear_lines(s->grid, s->grid->hsize, s->grid->sy, 8);
 
 	screen_clear_selection(s);
+
+	while ((title_entry = TAILQ_FIRST(s->titles)) != NULL) {
+		TAILQ_REMOVE(s->titles, title_entry, entry);
```
ThomasAdam: I know this is based on the original code, but can you check if `TAILQ_FOREACH_SAFE` and hence `TAILQ_REMOVE` does the same thing?  If so, I'd suggest switching over to that.

ThomasAdam: Also, move this line *after* you'e free()d the other entries below, please.

screen.c
```diff
@@ -60,16 +71,30 @@ screen_reinit(struct screen *s)
...
 {
+	struct screen_title_entry *title_entry;
+
 	free(s->tabs);
 	free(s->title);
 	free(s->ccolour);
 	grid_destroy(s->grid);
+
+	while ((title_entry = TAILQ_FIRST(s->titles)) != NULL) {
+		TAILQ_REMOVE(s->titles, title_entry, entry);
```
ThomasAdam: Ibid.

screen.c
```diff
@@ -110,6 +135,35 @@ screen_set_title(struct screen *s, const char *title)
...
+ * empty, do nothing.
+ */
+void
+screen_pop_title(struct screen *s)
+{
+	struct screen_title_entry *title_entry;
+
+	title_entry = TAILQ_FIRST(s->titles);
+	if (title_entry) {
+		TAILQ_REMOVE(s->titles, title_entry, entry);
```
ThomasAdam: Ibid.

nicm:
I think you are missing a free(s->titles) somewhere but otherwise this looks OK.

What is this for in input_csi_dispatch_winops:

~~~~
 		case 9:
 		case 10:
+			m++;
+			if (input_get(ictx, m, 0, -1) == -1)
+				return;

nicm:
Oh you are just missing a break there.

Here is a version with some minor tweaks. If this looks OK I will apply it after release: [y.diff.txt](https://github.com/tmux/tmux/files/1347190/y.diff.txt)

Oh if you want to change to use TAILQ_FOREACH_SAFE go for it, I don't think it matters too much.

nicm:
Oh my diff also needs this bit too. Also I didn't test it yet :-). Updated: 
[z.diff.txt](https://github.com/tmux/tmux/files/1347203/z.diff.txt)

~~~~
--- tmux-title-stack2/screen.c  Sun Oct  1 15:35:57 2017
+++ tmux-title-stack/screen.c   Sun Oct  1 15:36:13 2017
@@ -62,6 +62,7 @@ screen_init(struct screen *s, u_int sx, u_int sy, u_in
 {
        s->grid = grid_create(sx, sy, hlimit);
        s->title = xstrdup("");
+       s->titles = NULL;

        s->cstyle = 0;
        s->ccolour = xstrdup("");

screen.c
townba: `TAILQ_FOREACH_SAFE` and `TAILQ_REMOVE` should work fine.

townba: I'm sorry, I don't understand. Move which line below where? (I can't move the `TAILQ_REMOVE` line below `free(title_entry)`.)

townba:
@nicm, your latest diff looks fine to me. You and @ThomasAdam can battle over `TAILQ_FOREACH_SAFE` vs. the current approach. :smile:

nicm:
Applied to OpenBSD now, will be in GitHub later, thanks!

PR #1096, opened 2017-10-04
Nice termination message if remain-on-exit is used

NetSpida:
This adds screen-like termination messages for panes if remain-on-exit is
used. The three settings status-pane-died, status-pane-exitcode and
status-pane-exitsignal are available to be used with format strings.
Their message is formatted and printed in case the program died
normally, the program terminated with non-zero exitcode and the program
terminated with a signal, respectively. In either case, the templates
are first passed through strftime with the time of death. In cases of
abnormal termination, the format option #{pane_dead_status} will contain
either the exitcode or the signal, depending on the cause of
termination.

Test case:
tmux new-window "false"

Test Configuration:
set -g remain-on-exit on
setw -g status-pane-died '#[fg=green]#[bg=black]Pane >#{pane_current_command}< completed successfully at %H:%M:%S on %d.%m.%Y#[default]'
setw -g status-pane-exitcode '#[fg=yellow]#[bg=black]Pane >#{pane_current_command}< died with exit code #{pane_dead_status} at %H:%M:%S on %d.%m.%Y#[default]'
setw -g status-pane-exitsignal #[fg=red]#[bg=black]Pane >#{pane_current_command}< terminated with signal #{pane_dead_status} at %H:%M:%S on %d.%m.%Y#[default]'

### DIFF
--- a/cmd-pipe-pane.c
+++ b/cmd-pipe-pane.c
@@ -18,6 +18,7 @@
 
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/wait.h>
 
 #include <errno.h>
 #include <fcntl.h>
@@ -63,6 +64,7 @@ cmd_pipe_pane_exec(struct cmd *self, struct cmdq_item *item)
 	int			 old_fd, pipe_fd[2], null_fd;
 	struct format_tree	*ft;
 	sigset_t		 set, oldset;
+	int	 status;
 
 	/* Destroy the old pipe. */
 	old_fd = wp->pipe_fd;
@@ -71,6 +73,10 @@ cmd_pipe_pane_exec(struct cmd *self, struct cmdq_item *item)
 		close(wp->pipe_fd);
 		wp->pipe_fd = -1;
 
+		if (gettimeofday(&wp->death_time, NULL) != 0)
+			fatal("gettimeofday failed");
+		if (waitpid(wp->pid, &status, WNOHANG|WUNTRACED) > 0)
+			wp->status = status;
 		if (window_pane_destroy_ready(wp)) {
 			server_destroy_pane(wp, 1);
 			return (CMD_RETURN_NORMAL);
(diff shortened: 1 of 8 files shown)


ThomasAdam:
Hook support should be used for this, IMO.

NetSpida:
Adding log of discussion on IRC for reference (shortened for readability):

nicm: ok firstly we aren't adding three new options for this
nicm: so as far as options go my inclination would be just to choose the text you want and not give people a choice at all
Spida: I used three separate options to allow different formatting (I made some examples of green highlight on clean termination, yellow on non-zero exitcode, red on signal)
nicm: well, i think we can do that with one option
nicm: if you have one option and several formats i think static text would be better
nicm: you can change the colour if you like 
nicm: although i wouldn't change the colour of all the text, i would make it say something like "Pane exited with status X" or "Pane exited on signal X" and make just the X coloured maybe
nicm: dunno, would need to see what it looks like

nicm: i think pane_death_time would be better as pane_dead_time to match the others, even if it is not as grammatical

nicm: you can't just scatter calls to waitpid all over the place
nicm: it should only happen when we get SIGCHLD
nicm: you may need changes to make sure we don't draw the "pane is dead" stuff until the signal comes in
nicm: like you set a flag the first time server_destroy_pane is called if remain-on-exit is set and only draw the text when it is called from server_child_exited
nicm: it will only work if the pane exited on its own or because of a signal and tmux got the signal before the pty was closed
nicm: but you can't call waitpid anywhere else, because the child might not have exited yet
nicm: so what you will need to do, is not draw the text when server_destroy_pane is first called but draw it later when it is called from server_child_exited
nicm: instead of if (old_fd == -1) return; do something like if (wp->flags & PANE_STATUSDRAWN) return; if (!WIFEXITED(status) && !WIFSIGNALED(status)) return; wp->flags |= PANE_STATUSDRAWN; ... draw the status...
nicm: so that you only draw the text once the exit status has been updated

NetSpida:
I will rewrite the patch and do a new pull request.

PR #1100, opened 2017-10-07
Correct man page 'history_size'

ideasman42:


nicm:

PR #1103, opened 2017-10-07
Add termination messags if remain-on-exit is used

NetSpida:
This adds screen-like termination messages for panes if remain-on-exit is used. Third version of the patch, as discussed with nicm on IRC, without configuration options.


nicm:
OK I have made some minor changes, please take a look at this: [x.diff.txt](https://github.com/tmux/tmux/files/1367028/x.diff.txt)

- I don't think you need the formats changes and there is no use in making pane_dead_status the signal without some way to tell it is a signal and not a status code. I suggest adding them separately if you want them.
- There is no guarantee -1 is not a valid value for wp->status. I think 0 is better.
- Do not use pane_current_command or a format for the message, I think a static message is going to be more reliable.

nicm:
Applied now with some further minor changes. Will be in GitHub a bit later. Thanks!

PR #1123, opened 2017-10-19
for tree mode session nodes, restored keybindings 'h' and 'l'

electriceel1985:
I'm so used to 'h' and 'l' for vi-mode session tree navigation, and I know I can't be the only one. These bindings worked prior to the 2.6 tree code rewrite.


nicm:
OK, applied, thanks.

PR #1124, opened 2017-10-19
Fix description about vim(1) support

rhysd:
Vim now bundles the syntax file in official package and supports syntax highlight for `.tmux.conf`. Vim user no longer needs to install syntax file manually for tmux.

https://github.com/vim/vim/blob/036986f1507d223549d110af300144468bd3a1f7/runtime/syntax/tmux.vim

It was added on Mar 6 and in v8.0.420. I updatebbd README file following the update.

### DIFF
--- a/README
+++ b/README
@@ -43,7 +43,8 @@ the source tree with:
 
 A small example configuration in example_tmux.conf.
 
-A vim(1) syntax file is available at:
+vim(1) officially supports syntax highlight in v8.0.420 or later.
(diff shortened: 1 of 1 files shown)


ThomasAdam:
It's probably easier just to remove references to this completely.

Will fix.

nicm:
Changed, thanks.

rhysd:
ok, thank you for your change.

PR #1139, opened 2017-10-26
Fixed choose-tree command line bug

ForTheReallys:
Small fix for the choose-tree command line.  Without this, when the user opens the choose-tree command line with ':', and then closes it with ^C, the server would crash.


nicm:
Fixed, thanks!

PR #1140, opened 2017-10-28
Fix East Asian Ambiguous (Unity half-width)

atton:
UTF-8 has [East Asian Ambiguous characters](http://unicode.org/reports/tr11/).
This characters fit to special characters fit to half-width main language(e.g. English) or full-width main language(e.g. Japanese)
So this characters has ambiguous width depend on language environment(LC_ALL, LC_CTYPE).
However, [utf8proc](https://github.com/JuliaLang/utf8proc/blob/master/data/charwidths.jl#L116) cannot solve this problem(statically fix to 1).
I resolve this problem in my environment with ad-hoc [patch](https://github.com/atton/utf8proc/blob/ambiguous/data/charwidths.jl#L114) (statically Ambiguous fix to 2)
I created this PR modifies tty-acs table to half-width UTF-8 characters.

Screen shots before this Pull Request:
before-1
image

Screen shot after this Pull Request:
after-1
after-2

My Environments
- OS: macOS 10.12.6
- terminal: Terminal.app(2.7.3)
- Font: Ricty 3.2.2(Japanese font. ambiguous width = full(2))

Reference : [EasitAsianAmbiguous Character Lists](http://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt)

### DIFF
--- a/tty-acs.c
+++ b/tty-acs.c
@@ -48,7 +48,7 @@ static const struct tty_acs_entry tty_acs_table[] = {
 	{ 'n', "\342\224\274" },	/* large plus or crossover */
 	{ 'o', "\342\216\272" },	/* scan line 1 */
 	{ 'p', "\342\216\273" },	/* scan line 3 */
-	{ 'q', "\342\224\200" },	/* horizontal line */
+	{ 'q', "\357\277\232" },	/* halfwidth hangul letter eu */
 	{ 'r', "\342\216\274" },	/* scan line 7 */
 	{ 's', "\342\216\275" },	/* scan line 9 */
 	{ 't', "\342\224\234" },	/* tee pointing right */
--- a/tty-acs.c
+++ b/tty-acs.c
@@ -48,14 +48,14 @@ static const struct tty_acs_entry tty_acs_table[] = {
 	{ 'n', "\342\224\274" },	/* large plus or crossover */
 	{ 'o', "\342\216\272" },	/* scan line 1 */
 	{ 'p', "\342\216\273" },	/* scan line 3 */
-	{ 'q', "\357\277\232" },	/* halfwidth hangul letter eu */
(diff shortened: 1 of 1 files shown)


nicm:
Don't  change the comments please they are the ACS name not the Unicode
character.

atton:
OK. I will fix it.

atton:
I fixed it.

nicm:
Unfortunately these characters do not work for me, we can't change characters to fix a problem for a few users that will cause bigger problems for many others.

Did you try using the U8 flag instead? Try:
~~~~
set -as terminal-overrides ',*:U8=0'

atton:
OK. I will try it.

By the way, why can my patch work your environment?
So please show me your environment (include screen shots)

nicm:
It would probably work if I used a different font or installed some East
Asian fonts and used a terminal which supported substituting fonts, but I'm
not going to do that when the existing characters work without.

atton:
> set -as terminal-overrides ',*:U8=0'
this option cannot work in my environment like it.

screen shot 2017-10-30 at 15 54 01

nicm:
Are you sure you restarted tmux entirely? If so, perhaps your tmux is too old or your terminal does not support ACS correctly. tmux needs a terminal with either correct UTF-8 line drawing characters (no U8) or working ACS (U8=0). Does the same happen in xterm or iterm2?

atton:
> set -as terminal-overrides ',*:U8=0'

works in iterm2 and xterm using latest tmux.
So Terminal.app does not support ACS...

screen shot 2017-10-30 at 19 22 48

screen shot 2017-10-30 at 19 24 54

nicm:
That's weird, but Terminal.app is known for having a lot of bugs and missing features. Maybe it doesn't handle ACS properly at all or maybe not in your locale.

PR #1144, opened 2017-10-31
Add support for xdg config home

fekir:
Hello,

this is a small pull request for adding support to the XDG Base Directory Specification (https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

I supposed that a breaking change would not have been welcome, so I've mplemented the following logic:
if $HOME/tmux.conf is not present (could not be loaded), then try to load $XDG_CONFIG_HOME/tmux.conf (the default value is therefore $HOME/.config/tmux.conf).

Let me know if there is something that I should change in order to make you accept this pull request

Kind regards

Federico

### DIFF
--- a/cfg.c
+++ b/cfg.c
@@ -79,7 +79,8 @@ set_cfg_file(const char *path)
 void
 start_cfg(void)
 {
-	const char	*home;
+	const char	*home = NULL;
+	const char	*config_dir;
 	int		 quiet = 0;
 	struct client	*c;
 
@@ -106,8 +107,17 @@ start_cfg(void)
 		xasprintf(&cfg_file, "%s/.tmux.conf", home);
 		quiet = 1;
 	}
-	if (cfg_file != NULL)
+	if (cfg_file == NULL || load_cfg(cfg_file, NULL, NULL, quiet) <= 0){
+		cfg_file = NULL;
+	}
+
+	if (cfg_file == NULL && (config_dir = find_config_dir(home)) != NULL) {
+		xasprintf(&cfg_file, "%s/tmux.conf", config_dir);
+		quiet = 1;
+	}
+	if (cfg_file != NULL){
 		load_cfg(cfg_file, NULL, NULL, quiet);
+	}
 
 	cmdq_append(NULL, cmdq_get_callback(cfg_done, NULL));
 }
--- a/tmux.c
+++ b/tmux.c
@@ -185,6 +185,34 @@ find_home(void)
 	return (home);
 }
 
+const char *
+find_config_dir(const char* home)
+{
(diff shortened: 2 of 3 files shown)


nicm:
This has come up a few times before and we aren't going to add it, there
are already several ways to specify where the config is, we don't need
more.

fekir:
Oh, I'm sorry. 
I should have searched in the closed bugs if there was already a similar issue.

I've currently made an alias in my .bashrc (alias tmux="tmux -f $HOME/.config/tmux.conf"), but it has several drawbacks.
For example I need to duplicate the alias when using a different shell.
And when executing directly the binary file (like I did now for testing what I built), the configuration won't get loaded (AFAIK I can't define a variable like TMUX_CONFIG_FILE).
Same happens if tmux is executed from another program.


Is there any chance that $XDG_CONFIG_HOME will be the default directory instead of $HOME?
Or any chance that this pull request is going to be reconsidered?

I know you can specify any file you want through the command line, but it does not cover every use case, and I know that it may be a personal preference, but it has the big advantage that all configuration files are easier to manage, for example when putting them under revision control or when copying them from one computer to another.

Even on windows (with cygwin) it would be nice, since dot-files are not hidden automatically and they clutter you home directory.

(neovim did the same change for example, I do not know if the reasoning where the same)

nicm:
Yes, there is absolutely no chance we will ignore decades of Unix
practice (not to mention years of tmux versions) and make
XDG_CONFIG_HOME the default directory :-).

As far as supporting it at all, well, there are a lot of opinions on
where config files should go and what they should be called and I don't
want to try and support more than one.

To avoid the drawbacks of an alias, you could use a symlink, or a
~/.tmux.conf containing just "source ~/.config/tmux.conf".

(If you really want an alias, wouldn't it be better to ask bash to
support XDG_CONFIG_HOME, then you can put the alias in .config/bashrc
and it'll still be there when you copy that between computers?)

If Cygwin doesn't understand hidden files, you should probably take that
up with them.

fekir:
Thank you for the hint, writing "source ~/.config/tmux.conf" in my ~/.tmux.conf file is surely a better approach! I did not think about it.

Unfortunately symlinks do not work well on windows, that's why I made the alias.

It's not a problem of cygwin, but on Windows/NT Platforms the visibility of a file does not depend on the filename.
Of course one could go through all ".*" files and mark them as hidden (it's what I do sometimes), or use another graphical file manager instead of explorer.exe.


I'm sure you have already considered it more than once if it was already requested, so I'm not going to insist.
There are of course different opinions, but AFAIK the XDG is the only "standard" apart from the Microsoft standard for Windows, and the UNIX practice (which is not a standard, but just a common practice), so there are not really a lot of alternatives that one would need to support.

PR #1147, opened 2017-11-02
Faster grid reflow for vertical resize with many split lines

da-x:
For large 200,000 lines pane histories which contain split lines, the latency in vertical pane resize is noticable due to joins and re-splits in the grid reflow process.

If we take into account that the pane's horizontal size has not changed, we can just move those lines from the old grid instead of spending time joining and re-spliting them.

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -52,7 +52,8 @@ static void	grid_reflow_join(struct grid *, u_int *, struct grid_line *,
 		    u_int);
 static void	grid_reflow_split(struct grid *, u_int *, struct grid_line *,
 		    u_int, u_int);
-static void	grid_reflow_move(struct grid *, u_int *, struct grid_line *);
+static void	grid_reflow_move(struct grid *, u_int *, struct grid_line *,
+				 u_int);
 
 static size_t	grid_string_cells_fg(const struct grid_cell *, int *);
 static size_t	grid_string_cells_bg(const struct grid_cell *, int *);
@@ -1055,7 +1056,8 @@ grid_reflow_split(struct grid *dst, u_int *py, struct grid_line *src_gl,
 
 /* Move line data. */
 static void
-grid_reflow_move(struct grid *dst, u_int *py, struct grid_line *src_gl)
+grid_reflow_move(struct grid *dst, u_int *py, struct grid_line *src_gl,
+		 u_int flags_to_remove)
 {
 	struct grid_line	*dst_gl;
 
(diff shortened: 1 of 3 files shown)


da-x:
I've found a bug in my implementation - re-pushed with the fix.

nicm:
Hmm... is there any case where would we need to reflow at all if the
width hasn't changed?

Index: screen.c
RCS file: /cvs/src/usr.bin/tmux/screen.c,v
retrieving revision 1.48
diff -u -p -r1.48 screen.c
--- screen.c	5 Oct 2017 13:29:18 -0000	1.48
+++ screen.c	2 Nov 2017 15:51:23 -0000
@@ -197,7 +197,8 @@ screen_resize(struct screen *s, u_int sx
 		 * is simpler and more reliable so let's do that.
 		 */
 		screen_reset_tabs(s);
-	}
+	} else
+		reflow = 0;
 
 	if (sy != screen_size_y(s))
 		screen_resize_y(s, sy);

da-x:
True, that change works too

nicm:
I have applied my change now, thanks! We'll see if any problems show up...

da-x:
BTW I wonder whether as a run-time option, being able to disable line splitting for a pane would a useful feature - some cases may benefit (but we may also need to add scroll-to-left/right).

nicm:
I don't think we need an option, unless you have a case in mind that will affect many users. If it is just for you with 200k lines of history, we don't really need any more options that only one person will use. Of course it is not nice that the history size is limited by the speed of reflow, but it would be better to see if the reflow code can be made faster, or look at other possibilities.

Perhaps we could only reflowing the first 100k lines or something (that is easy but a large hammer), or reflowing panes on demand, eg after a resize only reflow the visible panes or the visible *parts* of the visible panes, but do the rest later either when idle or when the pane/window is next selected. Or even have the history limit work across panes, so we keep N lines of *total* history, and collect from different panes by age of history - meaning a single pane could grow to 200k but to do so it would have to free lines from other panes in the same session, leaving fewer total lines to reflow when the session resizes.

But reflow was added about 5 years ago and so far you are the first or maybe second person to have any issue with its performance, and you are capable of patching tmux the way you want yourself, so I'm not sure how much effort this warrants.

nicm:
It would be relatively easy to only reflow visible panes (those that are in the current window of an attached session), the remainder could flagged and reflowed when needed (when a command is run with them as target, or when the become visible).

PR #1161, opened 2017-11-17
Correctly unset user options

CHAZICLE:
Server died when trying to unset user options because it was trying to set them to defaults that don't exist for user options.


nicm:

PR #1165, opened 2017-11-21
Add SIXEL integration

dvn0:
This is pretty old, but it's also rather small. How close is this to be able to merge? It would be great to be able to have sixel graphics while using tmux.


nicm:
This is missing any attempt to track it so tmux can redraw it when the pane needs to be redrawn.

dvn0:
Thanks for looking at it. I didn't write the patch, but maybe @saitoha would be interested in getting it ready for upstream, or someone else who is also interested in seeing sixel support added to tmux.

PR #1184, opened 2017-12-08
Fix doubled WheelUp and WheelDown.

koie:


nicm:

PR #1247, opened 2018-01-28
Making cmd-split-window -v/-h  works as expected

hihihippp:
The current actual action of  :split-window -v and :split-window -h are not as the description in  the split-window part of the manual https://github.com/tmux/tmux/blob/19afd842bf63b28f4ea087b171a6be36a404d8e2/tmux.1#L1236

I have test on cygwin tmux 2.6 and the tmux 2.6 in Ubuntu 18.10.

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -94,9 +94,9 @@ cmd_split_window_exec(struct cmd *self, struct cmdq_item *item)
 	else
 		cwd = s->cwd;
 
-	type = LAYOUT_TOPBOTTOM;
+	type = LAYOUT_LEFTRIGHT;
 	if (args_has(args, 'h'))
-		type = LAYOUT_LEFTRIGHT;
+		type = LAYOUT_TOPBOTTOM;
 
 	size = -1;
 	if (args_has(args, 'l')) {


nicm:
split-window is fine as it is, thanks.

hihihippp:
The following are two popular tmux config file in github which are aware of the problem of the split-window command, my [suggestion](https://github.com/tmux/tmux/pull/1247/commits/3b32b2fae5f7d03b814e2c4acf3b58eedd5d7e41) can fix this problem:
1.  :split-window **-v** is **horisiontal** split window actually instead of vertal split window.
2. :split-window **-h** is **vertical** split window actually instead of horisiontal split window.
these action are not as we expected intutively and not consistent with the manual descritipion in the split-window part.  
- [gf3/dotfiles(the line 42, and 45) 782 stars on github](https://github.com/gf3/dotfiles/blob/a40853f440341d548aa11b16643de4c7d6014792/.tmux.conf#L42)
- [tony/tmux-config(line 7) 1135 stars on github](https://github.com/tony/tmux-config/blob/5b348ee402913fae-67ab51e725a2b4971ab65f3d/.tmux.conf#L7)

and the following is a vidwo on youtube realted the problem of split-window command(time-line 1:35):
[Create Custom Layout - Linux TMUX](https://youtu.be/sxw-n5Du600?list=PL5BE1545D8486D66D&t=95)
and related [tmux config file](https://github.com/gotbletu/shownotes/blob/b686e2e9c01dfa5061262847be2e8ea82ce0a447/tmux_layout.txt#L6)

Issue #512, opened 2016-08-20
tmux server with no sessions automatically closes itself


BrainDamage:
tmux server with no sessions seems to automatically close itself, i'm not sure if this is intentional or a bug, if it's the first, could an option to provide permanence even with no sessions?
~❯ tmux start-server \; set-option -g exit-unattached off 
~❯ tmux ls 
no server running on /tmp/tmux-1000/default

ThomasAdam:
That's correct.  There's no sessions created, and the command exited.

Issue #182, opened 2015-11-05
tmux start-server doesn't start the server in HEAD


paulhybryant:
Installed tmux through brew install --HEAD
tmux -V shows 2.2

tmux start-server finished with exit code 0
tmux show-options -g failed with no server running on  /tmp/tmux-...

It works in my previously installed version (2.1 I think).

ThomasAdam:
No, see:

`tmux -Lx start \; show -g`

paulhybryant:
Thanks for the reply. It is still not clear to me why the following command returns error?

tmux start-server
tmux list-sessions
no server running on /private/tmp/tmux-180029/default

The only thing in ~/.tmux.conf is 
set-option -sg exit-unattached off

ThomasAdam:
For that reason alone---start-server doesn't create any sessions.

paulhybryant:
And thus the server exits? If so is there a flag / option that prevent the server from exiting in this case? I didn't find such a flag / option in the manual. Thanks you.

nicm:
A tmux server will exit without a session. Create a session and it will
not exit.

kchr:
Sorry for reviving an old thread, but what is the purpose of `start-server` if you can't use it to start a server?

kjell:
I agree this feels bogus. I got around it in my setup script by spawning a ghost window with `sleep 1` to keep the server alive while I spawn a few other sessions. When the `sleep`ing session ends it exits and the other sessions keep the server running.

```
tmux has || tmux start \; new-session -d 'sleep 1'
<spawn your tmux sessions here>
```

And a bit easier way to do this is just starting the desired sessions within the same set of instructions as `start`:

```
tmux start \; new-session -d -s session1 \; new-session -d -s session2
```

nicm:
You do not need to use tmux start if you are using tmux new. new also
starts the server.



On 20 Feb 2017 10:17 pm, "Kjell Olsen" <notifications@github.com> wrote:

I agree this feels bogus. I got around it in my setup script by spawning a
ghost window with sleep 1 to keep the server alive while I spawn a few
other sessions. When the sleeping session ends it exits and the other
sessions keep the server running.

tmux has || tmux start \; new-session -d 'sleep 1'
<spawn your tmux sessions here>

And a bit easier way to do this is just starting the desired sessions
within the same set of instructions as start:

tmux start \; new-session -d -s session1 \; new-session -d -s session2

—
You are receiving this because you commented.
[...]

PR #1262, opened 2018-02-20
Add persist-without-sessions server option

Shados:
Defaults to `off`, if set to `on` it causes the tmux server process to keep running even without any active sessions.

## Use Case
My motivation for this change is that I'd like to run a tmux server as a systemd user service, and while I can currently do so by having it spawn an empty or default session, this is kind of a hack.
I typically start a new session per terminal rather than defaulting to re-connecting to an existing session, and having an extra session hanging around purely to keep the server alive is... somewhere between annoying and potentially confusing. 

More generally, the benefits of running the server under some form of service management are:
- If using systemd, can persist across logouts (if lingering is enabled for the user, and generally only relevant if logind is configured with `KillUserProcesses=yes`)
- Can use a different configuration file location than the default, without having to pass it in the command line directly or use an alias
- Can have the tmux server ready and waiting before any terminal or ssh sessions need to use it, effectively eliminating server startup time (in my experience this is trivial, but I know it affects some people on older hardware more noticeably)

There are at least a few other people who want this behaviour, based on a couple of previous issues where people were confused as to the behaviour of the `exit-unattached` option (namely #182 and #512).

## Notes

- I've only tested this out under Linux (NixOS, specifically)
- The unmodified `list-sessions` code called against an empty server simply results in no lines being printed, which seems suitable to me

----------

As an aside, is there anything in the way of development or high-level design documentation for tmux? I noticed the presentations but haven't done more than glance through them yet. 

I'm asking because I'm interested in contributing more generally; I haven't written much C in a long time, and am feeling oddly nostalgic. What I've seen of tmux's code looks pretty clean to me, so it's an attractive option as something to work on.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -96,6 +96,12 @@ const struct options_table_entry options_table[] = {
 	  .default_num = 500
 	},
 
+	{ .name = "exit-empty",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_SERVER,
+	  .default_num = 1
+	},
+
 	{ .name = "exit-unattached",
 	  .type = OPTIONS_TABLE_FLAG,
 	  .scope = OPTIONS_TABLE_SERVER,
--- a/server.c
+++ b/server.c
@@ -254,6 +254,10 @@ server_loop(void)
 
 	server_client_loop();
 
+	if (!options_get_number(global_options, "exit-empty") && !server_exit) {
+		return (0);
+	}
+
 	if (!options_get_number(global_options, "exit-unattached")) {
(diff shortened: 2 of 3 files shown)


nicm:
I would invert the sense of this and call it exit-no-sessions (default to on) to match exit-unattached.

Shados:
Grouping it with exit-unattached makes sense to me, but I think exit-no-sessions is pretty unintuitive as far as option names go. How about maybe 'exit-empty'?

nicm:
OK

Shados:
OK, I've changed it.

nicm:

PR #1274, opened 2018-02-25
Add C-g key to exit tree mode.

spudlyo:
### Summary
Tree mode has a number of hardcoded emacs and vi key bindings. One more couldn't hurt right?

### Use Case
We emacs users are pretty used to hitting C-g to abort out of things, this just makes tree-mode that much more comfortable for us.


nicm:

PR #1275, opened 2018-02-28
ISSUE_TEMPLATE: make audience explicit

waldyrious:
The instructions in CONTRIBUTING are useful for people who are reporting problems with tmux (bugs, unexpected behavior, etc.) but they don't apply to feature requests or other types of issues. This change makes that clearer.


nicm:
I think it is fine as it is. All issues of any type should include at least a tmux version and people should use their common sense with the rest.

waldyrious:
Fair enough. Thanks for the quick response.

PR #1276, opened 2018-02-28
COPYING: format text

waldyrious:
Use standard quote format to make the quoted text evident
and add .md extension to allow rendering as markdown.

Also fix a run-on sentence (replaced comma with semicolon).

ps - I suspect that git(hub) fails to detect the file rename due to the total size of the file being small in comparison to the number of lines changed by the addition of quotation marks. In any case, that's just a UI failure; the actual diff is correct as far as I can tell.


nicm:
No markdown please.

waldyrious:
@nicm would you accept this change without the addition of the .md extension tp the filename? The `>` quoting format is not exclusive to markdown.

nicm:
No I think it is fine as it is, I don't want formatting in text files.

PR #1284, opened 2018-03-15
quickfix for breaking pane_index order by "split-window -f"

koie:
## description

"split-window -f" breaks pane_index order.

## example

    tmux split-window -h -b \; split-window -h -f \; display-pane -d 10000
    => 0 2 1

    tmux split-window -h \; split-window -h -f -b \; display-pane -d 10000
    => 1 0 2

I expect 0 1 2 in both cases.


nicm:
Applied with some minor changes, thanks!

Issue #1288, opened 2018-03-19
libevent epoll is disabled inside tmux sessions


hq6:
Here is my tmux and OS version information.

```
Linux x86_64
tmux 2.1
screen
```

Applications running inside a `tmux` session cannot use libevent with epoll, because the `EVENT_NOEPOLL` environmental variable is set by tmux and continues to be set for applications running inside it..

PR #1289, opened 2018-03-19
Remove EVENT_NOEPOLL from environment after initializing libevent

hq6:
This pull request fixes [Issue 1288](https://github.com/tmux/tmux/issues/1288).


nicm:
I think this is OK for recent libevent where event_reinit does not check the variable again, but you are missing the same change for FreeBSD and Darwin.

hq6:
I have updated the pull request.

nicm:
Great, applied now, thanks!

PR #1290, opened 2018-03-21
Add format spec for the PID of the current process

pedroarthur:


nicm:
This is the pgrp not the current process, it isn't enough on some platforms. I don't think we can really be reliable enough to make this worth it, sorry.

pedroarthur:
Could you elaborate some more on why we can't be reliable enough? I saw osdep-* and every platform that implements `osdep_get_name` has a mechanism through which we could have the PID of the process group. I could implement this as an osdep, but I can only test it on Linux (in amd64, i386, ARM, and MIPS architectures).

PR #1294, opened 2018-03-26
Add --enable-jemalloc to `configure`

da-x:
I noticed that long-running tmux sometimes spends a lot of time in glibc's realloc, even when not processing much terminal data. A few months ago, I began linking in with jemalloc, and my long-running tmux processes seems much more responsive.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -224,6 +224,29 @@ else
 	fi
 fi
 
+# jemalloc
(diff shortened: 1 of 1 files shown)


nicm:
I don't want to add this kind of flag, can't you do it with LD_PRELOAD anyway?

Or talk to the glibc developers about why their malloc is crap.

PR #1298, opened 2018-03-28
Add KRB5CCNAME to the default list of variables handled by update-environment.

georgyo:
In environments with kerberos authentication for ssh, each connection creates a unique ticket store and exports the KRB5CCNAME environment variable for that session. This is similar to other environment variables already in the default, such as SSH_AUTH_SOCK.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -513,8 +513,8 @@ const struct options_table_entry options_table[] = {
 	{ .name = "update-environment",
 	  .type = OPTIONS_TABLE_ARRAY,
 	  .scope = OPTIONS_TABLE_SESSION,
-	  .default_str = "DISPLAY SSH_ASKPASS SSH_AUTH_SOCK SSH_AGENT_PID "
-			 "SSH_CONNECTION WINDOWID XAUTHORITY"
+	  .default_str = "DISPLAY KRB5CCNAME SSH_ASKPASS SSH_AUTH_SOCK "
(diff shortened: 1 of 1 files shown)


nicm:
I think we have enough in the default, and I'd prefer just to keep it to variables useful in the OpenBSD base system.

georgyo:
This is part of the openbsd systems as well

https://man.openbsd.org/ssh.1#K

nicm:
I don't see KRB5CCNAME anywhere in that page?

georgyo:
You are correct about that, it KRB5CCNAME is not explicitly listed there because it technically set by the kerberos library that OpenSSH is linked against. However if KerberosAuthentication is set on SSHD and a user uses kerberos to authenticate, KRB5CCNAME is **always** set. All gssapi implementations that I am aware of (MIT, Heimdal, and shishi) all behave the same way in this regard.

It is also technically possible to have PAM do the kerberos authentication (both with tickets and passwords) with ssh, and this will still set KRB5CCNAME for the user.

To me, this is an officially supported OpenSSH authentication method that allows forwarding/delegation of the method used to authenticate, very similar to ssh keys and agents.  It seems odd to include SSH_AUTH_SOCK but not include KRB5CCNAME in the defaults for tmux.

nicm:
Kerberos was removed from OpenBSD a few releases ago and OpenSSH doesn't
link against it. There are lots of environment variables various programs
want and I don't want to end up adding many more. Maybe this one is OK
though, I'll think about it.

nicm:
I have added this now.

PR #1306, opened 2018-04-03
Make README easier to read

kenden:
The README looks old, when viewed on github. This is a first step in making it more readable.

Renamed to README.md so it displays better on github.
Add markdown formatting.
Add sections (Installation / Dependencies / Usage...).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,54 +1,81 @@
+# tmux
+
 Welcome to tmux!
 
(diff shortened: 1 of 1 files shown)


nicm:
This has come up several times before. We are not going to clutter the text files that are shipped with tmux with GitHub's markup.

Issue #516, opened 2016-08-24
Pane resizing breaks bash, zsh, csh prompts.


ruipgpinheiro:
Hello,

For some reason, the resizing feature of tmux 2.2 seems to break the bash and zsh prompt. Csh breaks as well if you type something while the prompt is line-wrapping.

I'm using xterm, and can reproduce it every time even when running the shell with "-f", by changing the prompt to something simple (it happens no matter what prompt I use), splitting the window into 4 panes, and shrinking one of the panes down to 1x1. Upon resizing that pane back to a normal size, the whole prompt is corrupted.

For example, if I set the prompt to "1234567", follow the above steps, and then type "a" to find out where the cursor is:

csh:
> 1234567a

bash:
> 1234567a
> 1234567
> 567
> 345671212
> 2123456

zsh:
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567
>1234567a

As can be seen, the only prompt that seems to be unaffected by the issue is csh. I am unable to reproduce this issue outside tmux when using xterm directly.

When you have large prompts like I do, or even a right-sided prompt, this issue occurs *anytime* you resize and/or split a pane, and is very very annoying as it corrupts your whole buffer. I spent the last 2 weeks searching the web for solutions, trying work-arounds, but none have fixed it.

From what I've been able to gather, the issue seems to be that when zsh and bash redraw the prompt, they are not starting from the beginning of the "buffer line", but instead from the beginning of the current terminal line, which means if the line reflows part of the prompt will stay in previous lines. When resizing back to a larger size, tmux reflows the text meaning that part of the redrawn prompt will go back into previous lines, polluting the buffer even more. Interestingly, bash seems to be able to recover to the correct cursor position, while zsh does not.

Csh however "cheats", in that it does not redraw the prompt until you start typing again (both zsh and bash seem to redraw automatically on sigwinch instead), which means that by then it can clear the whole line correctly again, as it's not longer wrapping. Csh also breaks the prompt if you type something while the prompt has 1x1 size, then delete it before enlarging it back to the normal size:

>11234567a
>12
>12
>12
>12
>12341234567

[...]

PR #1323, opened 2018-04-18
Improve reflow by taking subprocess reflow on SIGWINCH into account

jiixyj:
Some programs like shells already reflow their output on SIGWINCH, at least the current line. The reflow algorithm of `tmux` can interfere with this, resulting in glitches (issue #516).

I've tried to mitigate this with the `winch_mod_y` variable. On a resize this is set to the height of the screen. When the subprocess tries to draw something in reaction to the SIGWINCH this variable is updated with the line number. The `tmux` reflow algorithm then just reflows up to this line. So if the subprocess is not SIGWINCH aware this is equivalent to the current behavior.

Here is another comment: https://github.com/tmux/tmux/issues/516#issuecomment-347769975
Since then I've readded SIGWINCH rate limiting (one SIGWINCH per 20ms) to give the application always some time to redraw itself.

Sadly, the SIGWINCH mechanism is inherently racy. So there is still the possibility of glitches.


nicm:
There is no reliable way for tmux to know if something is in response to SIGWINCH or not, or where the extent of any changes due to SIGWINCH starts or ends, I don't think trying to guess this is a good idea.

Applications that want to handle SIGWINCH and not have tmux (or indeed any other terminal that reflows) interfere can enter cursor positioning mode.

The only application here that knows what it wants the terminal to look like after SIGWINCH is zsh, it should be dealing with this, not tmux.

nicm:
What do the zsh developers say about this problem? How do other terminals that support reflow (like vte) deal with it?

jiixyj:
VTE doesn't seem to handle this well. It also produces junk:
https://github.com/jiixyj/rewrap-screenshots/blob/master/vte.png

I agree, tmux cannot know if the changes came in response to SIGWINCH or not. It is only a heuristic that seems to do well in practice. It is also attempting to handle this case:
 - Tmux wraps a line, two "new" lines are created
 - The shell does not redraw itself below both new lines as it should, but replaces the second line.
 - However, the wrapped flag of the first "new" line is set. Now, when resizing the window, tmux constantly tries to unwrap, but it uses data that is constantly redraw again by the shell. This results in "junk" that seemingly appears out of thin air.

But even without the `winch_mod_y` this patch improves the reflow by giving the application some time to redraw itself before tmux reflowing kicks in. I can take out the `winch_mod_y` stuff for now. My primary use case is a program I write that uses ncurses filter mode (i.e. only one line is in curses mode). It currently looks like this:
https://github.com/jiixyj/rewrap-screenshots/blob/master/ncurses-filter-rewrap-unpatched.png
With this branch:
https://github.com/jiixyj/rewrap-screenshots/blob/master/ncurses-filter-rewrap-patched.png

I'm writing to the screen every 40ms. It looks like the current reflow code doesn't trigger at all because the terminal never 'settles'.

nicm:
If you take out the bits that rely on it knowing what the application is
doing then I'll take a look, I don't mind changing the timing.

jiixyj:
I've taken out the code that looked at where the application wrote after receiving SIGWINCH.

Instead, I've implemented another idea: What about removing the 'wrapped' flag of lines where some specific escape sequences are executed? For example "clear line to beginning": Here, a 'wrapped' flag of the previous line wouldn't make much sense because it would try to join something that isn't there anymore. Conceptually this approach would be similar to the "line feed" command that clears the 'wrapped' flag of the current line.

This works great for bash and zsh prompts so far.

nicm:
One change at a time.

So what you are trying to do with the timer changes is resize immediately without reflow and then reflow later on the timer?

What is force_wrap_disabled for?

I don't think it is necessary to move mode resize into the timer, it can happen immediately.

jiixyj:
> So what you are trying to do with the timer changes is resize immediately without reflow and then reflow later on the timer?

Yes, that's the plan. This gives the application a bit of time to react to SIGWINCH. I think I've tried to call screen_resize in the timer and not in window_pane_resize, but that didn't work. Ideally what I'd like to do is to present a screen with a fixed size to the application for those 20ms. This would also fix the race below. Do you think this is doable? Can window_pane and screen size be decoupled?


> What is force_wrap_disabled for?

This temporarily turns off wrapping to mitigate the following race:
 1) Screen resized to width 60 from 61
 2) SIGWINCH sent to subprocess, 20ms timer started
 3) Screen resized to width 59 (no SIGWINCH sent, timer reset to 20ms)
 4) Subprocess writes to screen reacting to the SIGWINCH. It writes 60 cells, but this will introduce an unwanted newline.
This variable is only in play during the second or subsequent resizes. For the first resize this is set to 1 but then immediately set to 0 before the SIGWINCH is sent. But I can take it out for now as it doesn't really fix this race properly and might break other things.


> I don't think it is necessary to move mode resize into the timer, it can happen immediately.

Sounds good.

nicm:
I don't think force_wrap_disabled is a good idea, the screen needs to be the right size before sending SIGWINCH. I don't think we can just turn wrapping off, you can't rely on what the application will send.

You could copy the screen and use the copy for display until the application has finished updating, but that is likely to be very expensive.

But I don't know about the idea of delaying reflow. Why do we want to reflow after the application has redrawn? We want to reflow what is already there, what it is draws after SIGWINCH should already take the new size into account, we should not need to reflow that.

nicm:
OK so, the problem is not that tmux reflows the existing text, but that the cursor position after reflow may not be on the expected line. So if you have this with the X being the cursor

    1 | AAA AAA AAA X |

You may end up with:

    1 | AAA AAA |
    2 | AAA X   |

Then the application will draw on line 2, leaving line 1 as a leftover.

I don't know that delaying reflow is a good idea, after SIGWINCH the application has drawn what it wants at the new size, we should not be reflowing it.

nicm:
Well, in theory that is how it should look, but there is something funny about where tmux puts the cursor on resize compared with eg VTE.

nicm:
I have been trying to fix the cursor positioning but have not managed to do it completely. This is as I have got so far: [x.diff.txt](https://github.com/tmux/tmux/files/2025345/x.diff.txt)

nicm:
I have improved the situation with the cursor position but it doesn't really help the problem, because the cursor can still end up on a different line after reflow (tmux was calculating the line correctly but not the column). Perhaps a delay before reflow is the best approach, but I am going to leave this for the moment.

PR #1339, opened 2018-05-06
Beautified the README

sdushantha:
Made the README into markdown so I could make it look better.


ThomasAdam:
This comes up all the time. We're not doing this. See the closed entries here on Github for additional information if you need to.

PR #1358, opened 2018-05-26
Add tests for format strings

sadie-parayno:
Add some tests to check replacing and escaping inside of format strings.


nicm:
Brilliant, applied, thanks!

PR #1359, opened 2018-05-26
tmux.1: Clarify escaping inside and nesting of #{?,,}

sadie-parayno:
Clarify that `,` and `}` must be escaped inside of conditionals and give an example.

Also mention that conditionals can be nested.


nicm:

Issue #1335, opened 2018-05-03
incorrect pane arrangement in newer tmux versions


sassmann:
Running the following script gives an incorrect layout.
#/bin/sh
tmux -f/dev/null  new-session -d -s test
tmux split-window -v -p 55 -t 0
tmux split-window -h -p 67 -t 0
tmux split-window -h -p 50 -t 1
tmux split-window -h -p 67 -t 3
tmux split-window -v -p 25 -t 3
tmux split-window -h -p 33 -t 5
tmux split-window -v -p 33 -t 6
tmux split-window -v -p 50 -t 7
tmux attach-session -t test

Using tmux 2.5 the layout is still as expected. However with tmux 2.6 and 2.7 the layout looks incorrect. Pane sizes are different from what you would get from running the splits from inside the session.
Tested this with xterm and gnome-terminal (terminal size of approx. 390x110).

uname -sp && tmux -V && echo $TERM
Linux x86_64
tmux 2.7
xterm-256color

nicm:
You now need to specify the starting size for detached sessions to new-session or you will get 80x24.

sassmann:
Thanks! Specifying -x and -y and the layout looks right again. May I ask why this is necessary now? Version 2.5 has been able to figure out those values by itself.

nicm:
It is not always possible to know what the size is so we use 80x24.

sassmann:
Without having looked closely yet, would you accept a patch using $LINES and $COLUMNS from the environment (if available) instead of the 80x24?

nicm:
No, it has the actual size from the client but we do not use it because we can't do so consistently. You could make it so that if you do -x- -y- it uses them if possible.

sassmann:
Roger, I'll take a look.

PR #1372, opened 2018-06-07
allow specifying -x- -y- when creating detached session

sassmann:
Creating a new detached session with
tmux new-session -d -x- -y-
will now use the running terminals columns and rows size, instead of
the default 80x24.

This can be useful when setting up detached sessions for later use.
Inspired by https://github.com/tmux/tmux/issues/1335


nicm:

nicm:
Yes it is there.

sassmann:
Ah, you rewrote the commit, that's why I missed it. OK.

Issue #1002, opened 2017-07-09
"bind-key -T root MouseDrag1Border resize-pane -M" doesn't work in a some layout


koie:
* tmux version: master(6995497).
* platform: FreeBSD 11.0
* brief description:
  In some pane layout, "resize-pane -M" doesn't work.
````
  +-----------------+
  |%0   |%3         |
  |-----|-----------|<-- this border is moved downward 1char
  |%4   |%5         |
  |---------------*-|<-- drag down at this border
  |%1         |%2 | |
  |           |   V |
  |           |     |
  +-----------------+
````
* terminal: putty, TERM=xterm(outside) =screen(inside)
* log is attached.
* uname -sp && tmux -V && echo $TERM
  FreeBSD amd64
  tmux master
  xterm
* detail of reprodecing procedure.
````
    tmux split-window -v

    +-----------------+
    |%0               |
    |                 |
    |                 |
    |-----------------|
    |%1               |
    |                 |
    |                 |
    +-----------------+

    tmux split-window -h -p 33

    +-----------------+
    |%0               |
    |                 |
    |                 |
    |-----------------|
    |%1         |%2   |
    |           |     |
    |           |     |
    +-----------------+

    tmux select-pane -t 0
    tmux split-window -h -p 66

    +-----------------+
    |%0   |%3         |
    |     |           |
    |     |           |
    |-----------------|
    |%1         |%2   |
    |           |     |
    |           |     |
    +-----------------+

    tmux select-pane -t 0
    tmux split-window -v

    +-----------------+
    |%0   |%3         |
    |-----|           |
    |%4   |           |
    |-----------------|
    |%1         |%2   |
    |           |     |
    |           |     |
    +-----------------+

    tmux select-pane -t 2
    tmux split-window -v

    +-----------------+
    |%0   |%3         |
    |-----|-----------|
    |%4   |%5         |
    |-----------------|
    |%1         |%2   |
    |           |     |
    |           |     |
    +-----------------+

    set-option -g mouse on
    bind-key -T root MouseDrag1Border resize-pane -M

    +-----------------+
    |%0   |%3         |
    |-----|-----------|<-- this border is moved downward 1char
    |%4   |%5         |
    |---------------*-|<-- drag downward on this border
    |%1         |%2 | |
    |           |   V |
    |           |     |
    +-----------------+
````
* I suppose:
  cmd_resize_pane_mouse_update() selects a target pane for resize, and
  layout_resize_pane() does resize panes.
[...]

PR #1374, opened 2018-06-08
Fix resizing of panes by mouse

da-x:
The iteration of the list pane was not reliable in finding the right pane to resize. Moreover, we actually want to find the inter-pane border that is relevant to the resize action.

Therefore, instead of iterating the list of panes, go over the list of layouts in order to find the window that is next to the border that we are trying to resize. The behavior of a pane border drag is to always move the border, regardless of pane sizes (as long as they don't go over the minimum)

This addresses issue #1002.

### DIFF
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -129,8 +129,8 @@ static void
 cmd_resize_pane_mouse_update(struct client *c, struct mouse_event *m)
 {
 	struct winlink		*wl;
-	struct window_pane	*loop, *wp_x, *wp_y;
-	u_int			 y, ly, x, lx, sx, sy, ex, ey;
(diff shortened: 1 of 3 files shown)


da-x:
It's also the oldest tmux open issue :)

nicm:
Brilliant, works for me. Applied now, thanks!

da-x:
Thanks!

BTW,  in a subsequent change I've made it also possible to resize by dragging corners — it behaves as expected both in '+' and 'T' shaped corners.

https://github.com/da-x/tmux/commit/7874efe521b160e1f69d246a157fb8ac174c6042

It's small but a bit less trivial. Should issue a PR?

nicm:
Yes you can open a new PR or email me a diff and I'll take a look.

PR #1376, opened 2018-06-09
Make it possible to resize multiple panes by dragging corners

da-x:


nicm:
Works great, applied now, thanks!

PR #1377, opened 2018-06-17
last-pane: select the other pane as the last pane in two-pane window

pclouds:
This is basically for convenience because `last-pane` is usually for quickly switching between two panes. Imagine the user creates three panes, then delete one, `last-pane` would refuse to switch to the remaining pane, and the user has to select it explicitly before `last-pane` works again.


pclouds:
> This seems simpler, please try this

Indeed (I was not very familiar with tmux queue handling and went a simpler/safer way). Do I need to update documents or anything else?

nicm:
Applied now, thanks. No need to change any docs.

PR #1385, opened 2018-06-26
fixes gitignore/*.log and pane-base-index

mmjo:
ignore log files
fix pane-base-index upon start

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.diff
 *.patch
 *.core
+*.log
 core
 tags
 .deps/
--- a/window.c
+++ b/window.c
@@ -806,6 +806,10 @@ window_pane_create(struct window *w, u_int sx, u_int sy, u_int hlimit)
 	wp = xcalloc(1, sizeof *wp);
 	wp->window = w;
 
+	if (next_window_pane_id == 0) {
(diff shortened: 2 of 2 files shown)


nicm:
This is not correct, pane-base-index is for window indexes not window IDs.

mmjo:
Are you perhaps mixing up base-index and pane-base-index?
Or am i getting too old for this? ;)

From the manual:

Available session options:
base-index index
	Set the base index from which an unused index should be searched when a new window is created.  The default is zero.

Supported window options (with setw):
pane-base-index index
	Like base-index, but set the starting index for pane numbers.

nicm:
Oh sorry I meant pane not window, but you are still mixing up index and ID :-). The big numbers you see with C-b q are the pane index, it will change if the pane moves; the ID is a unique number prefixed with %, it will never change.

x

mmjo:
I see; but why not start the ID at pane-base-index too?
With the patch, the first pane gets ID 1, same as it's index.
(with setw -g pane-base-index 1 in ~/.tmux.conf)
Minor detail, though.

nicm:
It isn't the same as its index though because the index can change and the ID will not change. I think it is better to keep them separate.

PR #1386, opened 2018-06-27
Use separators for s///: operator other than slash

delphinus:
This enables to use any separator other than `/` in writing statusline
setting with the `s///:` operator.  This is useful to show paths with
replacing `$HOME` to `~`.

```bash
set -g status-left "#{s|/home/delphinus|~|:pane_current_path}"
```

### DIFF
--- a/format.c
+++ b/format.c
@@ -909,6 +909,7 @@ format_replace(struct format_tree *ft, const char *key, size_t keylen,
 	struct window_pane	*wp = ft->wp;
 	char			*copy, *copy0, *endptr, *ptr, *found, *new;
 	char			*value, *from = NULL, *to = NULL, *left, *right;
+	char			 sep;
 	size_t			 valuelen, newlen, fromlen, tolen, used;
 	long			 limit = 0;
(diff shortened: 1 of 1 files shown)


nicm:
This breaks anything starting in s, for example #{session_name}.

Something like this (on top of your change) would work:
~~~~
--- ../tmux/format.c    Wed Jun 27 10:49:32 2018
+++ format.c    Wed Jun 27 10:54:34 2018
@@ -19,6 +19,7 @@
 #include <sys/types.h>
 #include <sys/wait.h>

+#include <ctype.h>
 #include <errno.h>
 #include <fnmatch.h>
 #include <libgen.h>
@@ -976,7 +977,7 @@ format_replace(struct format_tree *ft, const char *key
                break;
        case 's':
                sep = copy[1];
-               if (sep == ':')
+               if (sep == ':' || !ispunct((u_char)sep))
                        break;
                from = copy + 2;
                for (copy = from; *copy != '\0' && *copy != sep; copy++)

delphinus:
Ah, yes. Good point. I fixed and pushed your diff.

nicm:

PR #1389, opened 2018-06-30
Solaris 11.4 switched from AT&T to GNU nroff, so prefers mdoc man pag…

alanc:
…e now

Signed-off-by: Alan Coopersmith <alan.coopersmith@oracle.com>


nicm:
Applied, thanks!

PR #1395, opened 2018-07-05
Add new format that shows rectangle-toggle status

h-michael:
Add format that shows rectangle-toggle status.
When rectangle-toggle is on, return 1.


nicm:
This needs to go into window_copy_add_formats like the others, not a new function.

h-michael:
@nicm 
Thanks for your review.
I fixed it.

PR #1398, opened 2018-07-07
Dynamic split reconfiguration

da-x:
It is a common occurance that users of tmux create a two-level split such as the
following:

```
+---------+---------+
|         |         |
|    0    |    1    |
|         |         |
|         |         |
+---------+---------+
|         |         |
|    2    |    3    |
|         |         |
|         |         |
+---------+---------+
```

An hour later the user has some long output in `3` and wants to resize it
upward, also decreasing the height of `1`. However, what it would also resize
`2`, depending on whether the first split was horizontal or vertical.

This commit adds on the implementation of a mouse-based resize, so that the
splits reconfigure automatically to accomodate the situation, without any
apparent change, as if the splits were 'right' from the start.

Effectively, all attempts to resize a border directly starting from the example
above, will change only the two panes affected, giving the user identical
behavior regardless of how the splits were originally made.

There are cases in which this change in a no-op due to some particular pane
configurations, and the way the algorithm for resplitting works. However,
it should work on most cases, especially in a checkers-style setup (i.e.
a straight NxM).

Open questions:

- Should this behavior made configurable?
- Should it also apply to resizing not via the mouse?
- Is there a better implemenation for it?

### DIFF
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -156,6 +156,8 @@ cmd_resize_pane_mouse_update(struct client *c, struct mouse_event *m)
 	else if (m->statusat > 0 && ly >= (u_int)m->statusat)
 		ly = m->statusat - 1;
 
+	layout_resplit(w, lx, ly);
+
 	for (i = 0; i < nitems(cells); i++) {
 		lc = layout_search_by_border(w->layout_root, lx + offsets[i][0],
 		    ly + offsets[i][1]);
--- a/layout.c
+++ b/layout.c
@@ -614,6 +614,647 @@ layout_resize_layout(struct window *w, struct layout_cell *lc,
 	notify_window("window-layout-changed", w);
 }
 
+struct layout_map {
+	u_int sx;
+	u_int sy;
+	struct layout_cell **screen_map; /* [sx][sy]; */
+	struct layout_cell **cells_array;
+	u_int cells_array_len;
+	u_int cells_len;
+};
+
+#define LAYOUT_MAP_CELL(_x_, _y_) map->screen_map[(_y_) * map->sx + (_x_)]
+
+/*
+ * Build a screen map of the current layout from the windowpane leafs. The
+ * borders will be NULL pointers.
+ */
+static void
+layout_map_build(struct layout_map *map, struct layout_cell *lc)
+{
+	struct layout_cell	*lcchild, *lc_copy;
+	struct layout_cell	**cell;
+	u_int			x, y;
+
+	switch (lc->type) {
+	case LAYOUT_LEFTRIGHT:
+	case LAYOUT_TOPBOTTOM:
+		TAILQ_FOREACH(lcchild, &lc->cells, entry)
+			layout_map_build(map, lcchild);
+		break;
+
+	case LAYOUT_WINDOWPANE:
+		lc_copy = layout_create_cell(NULL);
+		lc_copy->sx = lc->sx;
+		lc_copy->sy = lc->sy;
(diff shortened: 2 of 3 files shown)


nicm:
I will not have time to look at the code until next week but I think mouse
and command resizing should work the same way.

nicm:
This is a nice idea and thanks for doing it, but honestly, I think this is too much extra code for this.

I don't like this idea of flattening the layout and then rebuilding it. I don't think the current layout implementation is really suited to doing this - if this is how layouts should work, perhaps layouts should just be flat from the start (so not a tree).

That would be a bigger change, but then this would be easier to do, although splitting would be harder. 

Although if we were going to rework layouts like that it may be best to make some bigger changes. For example, linking panes to windows in a similar way to how windows are linked to sessions. So layout_cell would become the pane equivalent of struct winlink.

Also there are other things I would like with layouts at the same time, for example adding hints and automatic layouts.

But that is all making it a much bigger change.

da-x:
At first, I also figured that a better structure may allow more flexibility, but I wasn't sure if you'd give up the current one, as it would require rewriting the split logic. So I went for the approach of implementing this feature by only adding new code.

But of course, it is limited. For example, with only horizontal and vertical splits, certain configurations are impossible:
```
+----------------------------------------------------
|                  │                                |
|                  │                                |
|                  │                                |
|                  │                                |
|                  ├─────────────────┬───────────────
|                  │                 │              |
|                  │                 │              |
|                  │                 │              |
|                  │                 │              |
|────────────────────────────────────┤              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
+----------------------------------------------------
```
If the data structure is not a tree, it is solved.

Given the amount of resources I've already invested, I think I'd be satisfied with re-implementing it with a new data structure, maybe something similar to the map that I've already invented with the current one. Not sure I'd be able to go into the other areas that you mentioned. 

Under this new implementation, I can imagine splits as they are now, but what would happen under a 'checkers' board layout when a cell is gone? Is the cell from the left, top, right or bottom going to expand into the new place? Currently it's defined quite well for the splits tree. Also, how is layout-custom.c going to be affected? Layout saving and restore seems to rely on the tree structure.

nicm:
There will often be an obvious cell to take over the space - that is, if there is only one adjacent cell of the same width or height, it gets the space.

I would give each cell an age (which can just be an integer that increases with each new cell rather than an actual timestamp) and if there are multiple candidates then either the oldest or the newest wins.

The custom layout string will need to change, I don't mind breaking backwards compatibility here, although it may be possible to continue to accept the old form.

nicm:
Oh sorry, you meant what if a cell is removed that would mean leaving a gap.

Perhaps either creating such layouts should be forbidden somehow, or we will need to support "unused" space and just draw it filled with dots like when a window is reduced size. Perhaps a flag for splitw could create a pane in the first unused space rather than doing a split.

nicm:
What do other tiling window managers do for these cases?

nicm:
For your example, perhaps tmux could forcibly reduce the size of the top-left pane, so you ended up with:
~~~~
+----------------------------------------------------
|                  │                                |
|                  │                                |
|                  │                                |
|                  │                                |
|-------------------─────────────────┬───────────────
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
|                                    │              |
+----------------------------------------------------

nicm:
Likewise, please comment if you are working on this again and I shall reopen it. Thanks

PR #1404, opened 2018-07-14
Add a command to configure key table transition

midchildan:
I've been using custom key tables to use vim-like modal keybinding in tmux. It works by switching the key table back to prefix mode for every command executed, like below:

```tmux
bind k select-pane -U \; switch-client -T prefix
bind j select-pane -D \; switch-client -T prefix
bind h select-pane -L \; switch-client -T prefix
bind l select-pane -R \; switch-client -T prefix
```

But this approach doesn't work for undefined key sequences, meaning that tmux goes back to root mode if you mistype a command. This PR adds a way to let users specify which key table to transition to after a command is executed. With these changes, the above configuration can be re-written like this:

```tmux
set-next-table -T prefix prefix
bind k select-pane -U
bind j select-pane -D
bind h select-pane -L
bind l select-pane -R
```

This will allow you to keep tmux in prefix mode even if you enter an undefined key sequence.

Here's [a link](https://github.com/midchildan/dotfiles/blob/ca3f0261dddd5846370da50451c5091f70c99600/home/.tmux.conf) to my current tmux configuration, if you're interested.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -112,6 +112,7 @@ dist_tmux_SOURCES = \
 	cmd-set-buffer.c \
 	cmd-set-environment.c \
 	cmd-set-hook.c \
+	cmd-set-next-table.c \
 	cmd-set-option.c \
 	cmd-show-environment.c \
 	cmd-show-messages.c \
--- a/cmd-attach-session.c
+++ b/cmd-attach-session.c
@@ -58,6 +58,7 @@ cmd_attach_session(struct cmdq_item *item, const char *tflag, int dflag,
 	struct winlink		*wl;
 	struct window_pane	*wp;
 	char			*cause;
+	const char		*default_table;
 
 	if (RB_EMPTY(&sessions)) {
 		cmdq_error(item, "no sessions");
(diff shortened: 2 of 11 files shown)


nicm:
I think the idea of having an "All" key which is fired if the key isn't
found in the current key table was better than a new command to set the
default. Someone else requested it and I wrote it but they never tested it,
there will be an issue somewhere, I'll try and find it later.

nicm:
Oh it was Any not All key, try this please: [x.diff.txt](https://github.com/tmux/tmux/files/2195188/x.diff.txt)

midchildan:
It seemed to get stuck in an infinite loop when a matching keybinding wasn't found. I tried fixing the patch, and it now seems to be working: [x.diff.txt](https://github.com/tmux/tmux/files/2195387/x.diff.txt)

nicm:
Ah yes, the code has changed, try this instead please:  [y.diff.txt](https://github.com/tmux/tmux/files/2195535/y.diff.txt)

nicm:

PR #1414, opened 2018-07-25
Enable build on Cygwin

wojt:
Hi,

It seem forkpty signature is different on Cygwin which breaks the build.

Regards,
Wojtek

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -298,7 +298,11 @@ pid_t		 fdforkpty(int, int *, char *, struct termios *,
 
 #ifndef HAVE_FORKPTY
 /* forkpty.c */
+#ifndef __CYGWIN__
 pid_t		 forkpty(int *, char *, struct termios *, struct winsize *);
+#else
+int forkpty(int *, char *, const struct termios *, const struct winsize *);
+#endif
 #endif
 
 #ifndef HAVE_ASPRINTF


nicm:
How has it worked up to now?

wojt:
It had worked in the past but couple of weeks ago I noticed I'm no longer able to build.

nicm:
Right, so are you using a different Cygwin version? Are you using the latest? Did you ask them about it? forkpty should return pid_t.

wojt:
Hi, I did not ask them. I use the latest cygwin x64 and in pty.h there is definition with an int. I guess it should return pid_t, but googling for forkpty gives me different signatures for that function, sometimes it is defined with pid_t, sometimes with int.

nicm:
It should be pid_t, int is an error. So if it used to work they must have changed it.

nicm:
Actually, it doesn't matter, because our declaration should not be used if configure is finding forkpty. That declaration is only used if we are using the forkpty in compat.

Show me config.log please.

nicm:
Also show me your pty.h please.

wojt:
[config.log](https://github.com/tmux/tmux/files/2228005/config.log)

What's interesting is that Cygwin uses newlib and newlib forkpty uses an int instead of pid_t.

wojt:
Unable to attach pty.h here, pasting:
$ cat /usr/include/pty.h
#ifndef __PTY_H__
#define __PTY_H__

#include <_ansi.h>
#include <sys/termios.h>

#ifdef __cplusplus
extern "C" {
#endif

int openpty (int *, int *, char *, const struct termios *,
                     const struct winsize *);
int forkpty (int *, char *, const struct termios *,
                     const struct winsize *);

#ifdef __cplusplus
}
#endif

#endif /* __PTY_H__ */

nicm:
Show me full build output please from ./configure && make clean && make.

wojt:
I just cloned tmux again in order to show you the problem, but this time it built without errors. Not sure what was happening before but I can't reproduce it anymore... Must have been an artifact in my environment. Please reject my pull request.

PR #1434, opened 2018-08-11
Reattach to per-session namespace on OS X 10.10+

fabianishere:
This change will make tmux reattach to the per-session namespace using
the undocumented `_vprocmgr_detach_from_console` function.

Apple uses this approach for their distribution of GNU screen since OS X 10.10
and above, and has the benefit of supporting things like Touch ID support for
sudo in tmux.

See https://github.com/ChrisJohnsen/tmux-MacOSX-pasteboard/pull/70 for more
information.

### DIFF
--- a/compat/daemon-darwin.c
+++ b/compat/daemon-darwin.c
@@ -45,34 +45,19 @@
  * POSSIBILITY OF SUCH DAMAGE.
 */
 
-#include <sys/types.h>
-
-#include <mach/mach.h>
-
-#include <Availability.h>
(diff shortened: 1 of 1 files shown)


fabianishere:
> Hmm, after some testing it seems like using this patch means the commands (pbpaste, pbcopy, etc) do not work after restarting the `loginwindow` process, as a consequence of attaching to the Aqua bootstrap namespace. It might indeed be better to have this option as a separate flag.

This unfortunately cannot be solved within tmux itself.

Issue #1159, opened 2017-11-14
In copy mode, Page down triggers unexpected behavior of line selection


w2w2k2k2:
- your platform (Linux, OS X, or whatever)
Reproduced in tmux 2.5 (on centOS 7 uname Linux x86_64) and 2.6 (macOS Sierra)

- a brief description of the problem with steps to reproduce;
* make sure you have a long history so you can scroll up for multiple pages
* get into vi copy-mode
* select line (usually bind to "V")
* hit PageUp or ctrl-u for a few times, you'll see many lines got selected
* hit PageDown or ctrl-d once
Expected to see: scroll one (or half) page down, selecting one (or half) page less as well
What I see:
* scrolled to bottom immediately
* selection start and end were reversed (selection starting was at the bottom, now it was in the history I scrolled to)

- a minimal tmux config, if you can't reproduce without a config;
Can reproduce without config

- your terminal, and $TERM inside and outside of tmux;
My terminal is iTerm2 Build 3.1.5
$TERM inside tmux: screen-256color
$TERM outside tmux: xterm-256color

nicm:
Can you still reproduce this with master now?

w2w2k2k2:
I'm not sure about master, but upgraded my mac tmux to 2.7. The problem still exists

nicm:
Hmm OK I'll try it again.

amosbird:
I believe the issue should be related to this line

https://github.com/tmux/tmux/blob/a3685486454051d26c820db4ea4cdfc094172c1c/window-copy.c#L400

However those coordinates are a bit confusing that I fail to reason it out.

PR #1440, opened 2018-08-18
fix window-copy page scroll issue.

amosbird:
fix https://github.com/tmux/tmux/issues/1159


nicm:
Seems to work for me, applied to OpenBSD now, will be in GitHub later. Thanks!

Issue #1441, opened 2018-08-19
Supporting unicode escape sequences in config string literals


chrahunt:
It would be  helpful if unicode code point escape sequences could be embedded in tmux configuration string literals. Assuming we use something like the C syntax for universal character names, the following would be considered equivalent:

```
set -g status-right '⌚ %H:%M:%S '
```

```
set -g status-right "\u23F1 %H:%M:%S "
```

or

```
set -g status-right '🕐 %H:%M:%S '
```

```
set -g status-right "\U0001F550 %H:%M:%S "
```

Currently (on master) it appears that the string processing in the config file does not permit any escape sequences other than `\e`, `\t`, `\n`, and `\r` per [here](https://github.com/tmux/tmux/blob/91dec25fc96d17f99ba63115acf1cbbfb058d781/cmd-string.c#L211).

nicm:
Sure, yes, it would...

PR #1443, opened 2018-08-19
Support \u and \U escapes in commands.

chrahunt:
GitHub issue #1441.

This only supports escape sequences in string literals in configuration and as part of interactive commands.


nicm:
Great, applied to OpenBSD now with some minor tweaks (I don't know why you did the utf8_split bit outside the switch, I moved it inside). Will be in GitHub later on. Thanks!

PR #1444, opened 2018-08-19
Add tests for escape characters

chrahunt:
To cover #1443.

Not sure about the behavior of `\t`, it appears to be stripped if alone or at the beginning/end of the input string.


chrahunt:
It appears to impact only the output of `display-message`.

nicm:
display-message goes through utf8_sanitize and will not print unprintable characters.

Issue #1228, opened 2018-01-10
redraw fails with vim


jruffet:
I'm having an issue with tmux 2.6 from stretch-backports (2.6-3~bpo9+1)
Spawning a new tmux with `tmux -f/dev/null -Ltest new` (to be sure, but the issue arises with my conf), vim with stock configuration, when I scroll in vim (pgup/pgdwn) parts of what was displayed and should be redrawn is not.
If I force a tmux redraw (C-b r), it is displayed correctly.
Forcing a vim redraw (C-l) has no effect.
Note that if I split the window in 2 panes, then there is no issue.

nicm:
This issue does not include enough information. Please read the [CONTRIBUTING](https://raw.githubusercontent.com/tmux/tmux/master/CONTRIBUTING) file.

jruffet:
Issue still present in master. All the following info are based on tmux/master.
Within tmux spawned with `tmux -f/dev/null -Ltest new`
```
$ uname -sp && tmux -V && echo $TERM
Linux unknown
tmux 2.6
screen
```
Terminal : konsole
TERM outside tmux : xterm-256color (same term within, if I use my tmux conf)
Log with `tmux -vv -Ltest -f/dev/null new` : [tmux-server-13389.log](https://github.com/tmux/tmux/files/1618514/tmux-server-13389.log)

I hope you have all the info that you need.

nicm:
This is probably https://bugs.kde.org/show_bug.cgi?id=379318, try upgrade Konsole or:
~~~~
set -as terminal-overrides ',*:indn@'
~~~~
Do not use TERM=xterm* inside tmux.

jruffet:
Your workaround works like a charm.
Indeed, konsole shipped with the latest ubuntu is pre-patch.

Sorry about the xterm-256color, that was a mixup. That is actually screen-256color in a tmux session with my .tmux.conf.

Thank you for the very fast and accurate replies.

Issue #1085, opened 2017-09-22
Clearing full lines by scrolling them away broken


22dh22:
## problem / steps
Opening a large file in vim-nox and jumping to bottom (`G`), then top (`gg`), then bottom (`G`), then top (`gg`) causes corrupted output (old and new line content mixed).
This problem does not occur outside of tmux.

## environment
- current tmux master https://github.com/tmux/tmux/commit/ff526e43de84cff3770d32c0b6ea215a7359f521
- up to date Debian stable and sid
- no config
- Konsole, and directly on the virtual console outside X, currently not reproducible on xterm, but I think I saw some corruption there as well some days ago
- `TERM`
  - outside tmux `TERM=xterm` is broken while `TERM=linux` works, probably because of missing features required for the clearing trick
  - inside always `TERM=screen`
  - using `TERM=konsole` or `TERM=konsole-256color` does not fix this

```
$ uname -sp && ./tmux -V && echo $TERM
Linux unknown
tmux master
screen
```

## example
[log and example file](https://github.com/tmux/tmux/files/1325579/log.zip)
I open `/tmp/longfile`, jump `down`+`up`+`down`+`up`, then lines show `ab`.

## links
found Debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=876362

## bisect results
new way to clear added with first chunk of https://github.com/tmux/tmux/commit/0264ef094a79c828e72e03052b9969cd80b28d16
=> introduced into [Debian git](https://anonscm.debian.org/git/collab-maint/tmux.git) with [`749aa5`](https://anonscm.debian.org/git/collab-maint/tmux.git/commit/?id=749aa52e923657cfd09b395eaf3d48f79de521e0)

nicm:
Probably the same as https://github.com/tmux/tmux/issues/933 and https://github.com/tmux/tmux/issues/931.

nicm:
This is a bug in Konsole. May also be a bug in Linux console but since you are not meant to use TERM=xterm but rather TERM=linux it is up to you if you want to see if they will fix it.

22dh22:
The `set -as terminal-overrides ',*:indn@'` did not work but I can confirm that this is fixed with the Konsole version in Debian experimental which is newer than the version of https://bugs.kde.org/show_bug.cgi?id=379318

So not a tmux bug, closing.
Thanks for the quick response!

hefee:
[...]

Issue #933, opened 2017-05-17
My terminal doesn't redraw the screen 


c02y:
- version: tmux master commit 3df4e78492 and another commit from master since I pulled once.
- platform: CentOS6.5 32-bit Qt: 4.6.2 KDE: 4.3.4 (KDE 4.3.4) kde4-config: 1.0
- $TERM: xterm-256color or empty ~/.tmux.conf
- same result with empty ~/.tmux.conf

This problem happens in both in bash or fish-shell, so it is not the problem of shell.

This problem happens in Konsole_v2.3.3/Yakuake_v2.9.8 with new version built from tmux master, but not in gnome-terminal_v2.31.3 with any version of tmux.

This problem happens when using tmux binary built from tmux master, **but** 2.5-rc branch or earlier versions are OK, so it's the problem of new code of tmux.

![recorded](https://cloud.githubusercontent.com/assets/2135996/26136055/362c216c-3aec-11e7-926e-b0274c940767.gif)

[tmux-server-26331.log.txt](https://github.com/tmux/tmux/files/1006374/tmux-server-26331.log.txt)

octoploid:
Yes, this is a known konsole bug. See https://github.com/tmux/tmux/issues/931 for a workaround or a konsole patch.

c02y:
Thanks, `set -as terminal-overrides ',*:indn@'` fix it.

forfrt:
Get fixed. Thanks

ggsalas:
Any update on this? I have the same problem on macos and neovim. Thanks!

PR #1450, opened 2018-08-23
Fixes screen redraw error with vim

paulozulato:
Fixes screen redraw issue when scrolling vim inside tmux, which ends
mixing up new and old line content (as reported at issues #1085, #933
and #1228).

Steps to reproduce this issue:
 1) start tmux 2.8 from a konsole inside a yakuake session with TERM set
    to "xterm" or "xterm-256color";
 2) set tmux TERM to "xterm" or "xterm-256color";
 3) open a long file (whose lines have different lengths) with vi and
    scroll it using PageUp/PageDown keys.

At step 3), content from old lines and new lines will be mixed together.

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -1096,6 +1096,7 @@ tty_clear_area(struct tty *tty, const struct window_pane *wp, u_int py,
 		    px + nx >= tty->sx &&
 		    ny > 2 &&
 		    tty_term_has(tty->term, TTYC_CSR) &&
+		    tty_use_margin(tty) &&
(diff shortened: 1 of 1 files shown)


nicm:
This is wrong, using indn has nothing to do with margins. Did you read the issues you linked to? This is probably a bug in your terminal.

paulozulato:
Sorry, I'd just removed my comment when you replied. I'd just seem this point you have mentioned now and, then, I removed my comment because it was pointless.

PR #1460, opened 2018-08-31
Allow -1 or a large number to easily go to top

mckellyln:
@nicm what do you think ?


nicm:

PR #1464, opened 2018-09-04
Add a space between -f and program filename when invoking ${AWK}

iamleot:
At least on SunOS, nawk-20050424 insists on a space
between -f and the program file name, by penalty of
"/opt/local/bin/nawk: no program filename".

Patch from Hauke Fath via pkgsrc.


ThomasAdam:
Fixed.  Thanks.

iamleot:
Thomas Adam writes:
> Fixed.  Thanks.

Just a nit-pick regarding the changed commit message. :)
Please note that all NetBSD supported in base has nawk version
20121220.  This happened with an older version of lang/nawk on
SunOS, but I would not be surprised if some operating systems ships
with older nawk-s.


Thanks!

PR #1466, opened 2018-09-04
Resize window fixes

mathstuf:


nicm:
Nice one, applied, thanks!

PR #1475, opened 2018-09-10
mouse events: ignore pane borders while zoomed

### LINKED ISSUE #1473: Mouse events ignored in zoom mode where unzoomed is a horizontal pane boundary
Steps to reproduce:

1. Build latest tmux (5aa435a9)
2. Set ./tmp.conf to have only one line - enable the mouse: `set-option -g mouse on`
3. run `tmux -f ./tmp.conf -2 new-session -A -s main`
4. Split top-bottom (C-b ").
5. The bottom pane should be the active one (click it to activate if that's not the case).
6. Zoom it (C-b z)
7. Select text with the mouse at the vertical-middle of the terminal - where the splitter exists when unzommed.

Expected result:
- It's possible to select text.

Actual result:
- No text selected.

This is also visible with an app running in the zoomed pane which uses mouse events - like vim. Selection which starts at the vertical-middle line just doesn't select anything. Selections which start above or below it and then try to cross the vertical-middle threshold change their behavior (like losing mouse tracking).

This issue also exists in tmux 2.8 .

avih:
Previously, when creating a mouse event and settings its window pane,
the pane-search procedure tested pane borders unconditionally, and if
the coordinates were on a border, then this border's pane was set for
the event.

The issue is that in zoom mode borders are irrelevant, and it was
possible to end up with a pane which isn't the active (zoomed) one.

This resulted in mouse events getting ignored/other-issues in zoom mode
if they happen where a pane border exists when unzoomed.

This commit simply skips the border search procedure if the window is
zoomed.

Fixes #1473

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -535,14 +535,20 @@ have_event:
 		m->x = x + m->ox;
 		m->y = y + m->oy;
 
-		TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
-			if ((wp->xoff + wp->sx == px &&
-			    wp->yoff <= 1 + py &&
-			    wp->yoff + wp->sy >= py) ||
-			    (wp->yoff + wp->sy == py &&
-			    wp->xoff <= 1 + px &&
-			    wp->xoff + wp->sx >= px))
-				break;
+		if (s->curw->window->flags & WINDOW_ZOOMED) {
+			/* borders are irrelevant */
+			wp = NULL;
+		} else {
+			/* maybe find a border's pane */
+			TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
+				if ((wp->xoff + wp->sx == px &&
+				    wp->yoff <= 1 + py &&
+				    wp->yoff + wp->sy >= py) ||
+				    (wp->yoff + wp->sy == py &&
+				    wp->xoff <= 1 + px &&
+				    wp->xoff + wp->sx >= px))
+					break;
+			}
 		}
 		if (wp != NULL)
 			where = BORDER;


avih:
FWIW, it's possible to use the patch below instead. It also improves performance because mouse events are arguably more likely to happen inside a pane rather than on a border, so this patch tries to find a pane's "content" area first, and then tries the borders:

```diff
diff --git a/server-client.c b/server-client.c
index ece16a42..5d68789e 100644
--- a/server-client.c
+++ b/server-client.c
@@ -535,22 +535,25 @@ have_event:
 		m->x = x + m->ox;
 		m->y = y + m->oy;
 
-		TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
-			if ((wp->xoff + wp->sx == px &&
-			    wp->yoff <= 1 + py &&
-			    wp->yoff + wp->sy >= py) ||
-			    (wp->yoff + wp->sy == py &&
-			    wp->xoff <= 1 + px &&
-			    wp->xoff + wp->sx >= px))
-				break;
-		}
-		if (wp != NULL)
-			where = BORDER;
-		else {
-			wp = window_get_active_at(s->curw->window, px, py);
+		wp = window_get_active_at(s->curw->window, px, py);
+		if (wp != NULL) {
+			where = PANE;
+		} else if (s->curw->window->flags & WINDOW_ZOOMED == 0) {
+			/* maybe find a border's pane */
+			TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
+				if ((wp->xoff + wp->sx == px &&
+				    wp->yoff <= 1 + py &&
+				    wp->yoff + wp->sy >= py) ||
+				    (wp->yoff + wp->sy == py &&
+				    wp->xoff <= 1 + px &&
+				    wp->xoff + wp->sx >= px))
+					break;
+			}
+
 			if (wp != NULL)
-				where = PANE;
+				where = BORDER;
 		}
+
 		if (where == NOWHERE)
 			return (KEYC_UNKNOWN);
 		if (where == PANE)
```

nicm:
It looks like this breaks using the mouse on unzoomed pane borders.

Please try this instead:

diff --git a/server-client.c b/server-client.c
index ece16a42..8f4be1e4 100644
--- a/server-client.c
+++ b/server-client.c
@@ -535,18 +535,20 @@ have_event:
 		m->x = x + m->ox;
 		m->y = y + m->oy;
 
-		TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
-			if ((wp->xoff + wp->sx == px &&
-			    wp->yoff <= 1 + py &&
-			    wp->yoff + wp->sy >= py) ||
-			    (wp->yoff + wp->sy == py &&
-			    wp->xoff <= 1 + px &&
-			    wp->xoff + wp->sx >= px))
-				break;
+		if (~s->curw->window->flags & WINDOW_ZOOMED) {
+			TAILQ_FOREACH(wp, &s->curw->window->panes, entry) {
+				if ((wp->xoff + wp->sx == px &&
+				    wp->yoff <= 1 + py &&
+				    wp->yoff + wp->sy >= py) ||
+				    (wp->yoff + wp->sy == py &&
+				    wp->xoff <= 1 + px &&
+				    wp->xoff + wp->sx >= px))
+					break;
+			}
+			if (wp != NULL)
+				where = BORDER;
 		}
-		if (wp != NULL)
-			where = BORDER;
-		else {
+		if (where == NOWHERE) {
 			wp = window_get_active_at(s->curw->window, px, py);
 			if (wp != NULL)
 				where = PANE;

nicm:
Looks like this is substantively the same as your original change in the PR anyway.

nicm:
Applied a version of this to OpenBSD now, it will be in GitHub later - please let me know if it doesn't fix it. Thanks!

avih:
> It looks like this breaks using the mouse on unzoomed pane borders.

Hmm.. my bad, didn't try the alternative patch long enough, and I assumed `window_get_active_at` only checks the "content" area. Thanks.

> Applied a version of this to OpenBSD now

Thanks, but I don't understand (I see it's applied in github already). How's OpenBSD related?

Issue #1480, opened 2018-09-18
tmux resize-pane doesn't resize to PANE_MINIMUM


nmeum:
The minimum pane size is defined in `tmux.h` as follows:

```C
/*
 * Minimum layout cell size, NOT including separator line. The scroll region
 * cannot be one line in height so this must be at least two.
 */
#define PANE_MINIMUM 2
```

Creating a new pane with this minimum pane size using `tmux split-window -l 2` is possible and works as expected. Unfortunately, resizing an existing pane created using `tmux split-window` to the minimum pane size with `tmux resize-pane -y 2` does not work as expected. The `tmux resize-pane` command only seems to be able to resize an existing pane to a size of 3 lines.

I can reproduce this with `tmux 2.7` and Git HEAD `2.8-66-g71d2ab18` (using the default configuration and urxvt (`TERM=rxvt-unicode-256color`) on Alpine Linux. In `tmux 1.9` on Debian Jessie both `resize-pane` and `split-window` work as expected with the default configuration.

Is this intended or do you consider this a bug?

nicm:
Not intentional. Funny that -x works though.

nmeum:
I am also wondering if the comment above the `PANE_MINIMUM` macro is actually correct. I don't know much about the tmux internals but is it really necessary for the scrollbar to be at least two lines in height? At least with my configuration it only takes up one line in height.

nicm:
Terminal scroll region cannot be one line. We do support terminals without
scroll region now however so it would be possible to have one line panes
and use the same method (redraw instead of scrolling).

nmeum:
> [ …] so it would be possible to have one line panes and use the same method […]

That would be nice, I personally have a use-case for using tmux with panes which are one line in height.

---

I also investigated the original issue further, the bad behavior was introduced in 2.3.  To be more specific commit f8cc48a43f1fd5fe082fde86b429c2cda5bcf1b0 seems to have introduced this issue.

To be even more specific the following statement added to the `layout_resize_check` function seems to be the issue:

```C
minimum += layout_need_status(lc,
  options_get_number(w->options,
  "pane-border-status") == 1);
```

Commenting out this statement makes `tmux resize-pane -y 2` work as expected again (even on the master branch).

As stated above I don't know enough about the tmux internals to figure out why and if it is necessary to increase the minimum pane height when a border-status is enabled but since this is wrapped in a layout-check-if-statement this might also explain why `-x` works as expected.

nicm:
[...]

PR #1487, opened 2018-09-25
Reduce PANE_MINIMUM to 1

nmeum:
As document in #1480 the cursor placement on scrolling up doesn't seem 100% correct. Otherwise I am not aware of any issues with the changes purposed here.


nicm:
I have applied this to OpenBSD now, it will be in GitHub when it syncs up in an hour or so. It seems OK to me so we can see if it works OK. If so then it will be in tmux 2.9. Thanks!

PR #1495, opened 2018-10-02
compat: ensure imsg hdr is fully initialized

carenas:


nicm:
hdr->len is set in imsg_close.

PR #1545, opened 2018-11-18
New option 'full-search' to control copy-mode history search coverage

derekschrock:
New option 'full-search' to control if copy-mode will search
over the full history.  Defaults to on, when off number of
results and highlighting is not shown.

When dealing with a history of ~20000 lines there's a small,
but noticeable, lag in the amount of time to jump to the first
result and search-again.  >50k+ histories results in multi-second wait.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,9 @@
 CHANGES FROM 2.8 to X.X
 
+* New option 'full-search' to control if copy-mode will search
+  over the full history.  Defaults to on, when off number of
+  results and highlighting is not shown.
+
 * New "terminal" colour allowing options to use the terminal default
   colour rather than inheriting the default from a parent option.
 
--- a/options-table.c
+++ b/options-table.c
@@ -117,6 +117,12 @@ const struct options_table_entry options_table[] = {
 	  .default_num = 0
 	},
 
+	{ .name = "full-search",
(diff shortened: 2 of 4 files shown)


derekschrock:
Do you have any details on how this would work or a patch for testing?

Would tmux init. search over a smaller set (block) of history, jump to that match then continue searching in the background without blocking tmux?  Result totals would be deferred until or removed?

nicm:
I do not think this should be a new option.

If it is to be a simple off/on, it should be two new copy mode commands. Like search-forward vs search-forward-incremental. Maybe search-forward-now or search-forward-immediately? I don't know.

Ideally, copy mode would be more sophisticated. So it should only search and highlight say a couple of screens or a few thousand lines of content then repeat the search to highlight subsequent lines when then become visible.

derekschrock:
Ok.  I'd agree with copy mode being more sophisticated.  Would this also mean the results count would go away or only display the number of results found for the N pages/lines searched?  I think number of results should just go away.

nicm:
I don't think that'll be the most complicated bit. I'd make it display the
number it found so far, but removing it might be OK too.

nicm:
In master we have changed so that marking (highlighting) only searches visible text (it is repeated after scrolling) which should solve this problem.

PR #1548, opened 2018-11-19
GNU/kFreeBSD Fixes

jrtc27:


jrtc27:
That final commit is rather gross, so if you want us to keep that as a local patch in Debian until the mess is fixed (which is likely a while off due to a lack of time and the beast that is glibc) I'm happy for that to be dropped.

ThomasAdam:
Seems to me this mess should be kept out of tmux, and kFreeBSD should be fixed instead.

nicm:
We should use our own strlcpy not libbsd. Using libutil-freebsd is probably
OK. The workaround we don't need.

jrtc27:
In practice, libbsd was already being linked in so this wasn't an error, the main objective was to find a `string.h` with the prototype, as the glibc one doesn't (given it doesn't provide a `strlcpy`). Now, technically this isn't guaranteed, so in theory this could be imposing an extra dependency, but is it not easier to just use libbsd?

jrtc27:
> Seems to me this mess should be kept out of tmux, and kFreeBSD should be fixed instead. It's a niche project at best, punishing tmux for their sloppiness isn't right.

You're entitled to your opinion, but please don't tell me I'm *punishing* you for *providing a patch*. Yes, one of these commits is a hack that doesn't really belong in tmux, but the other two are legitimate issues.

jrtc27:
One other point: I added the configure code to the `*freebsd*` case, but this isn't needed on pure FreeBSD. Are you happy with that, or would you like it split out into its own `*kfreebsd*` case?

nicm:
What is the autoconf OS triple on this platform?

nicm:
Why is the lack of strlcpy not being picked up by AC_REPLACE_FUNCS?

nicm:
Can you show me the config.log please?

jrtc27:
`x86_64-unknown-kfreebsd11.0-gnu`, so `$host_os` is `kfreebsd11.0`, in my case.

Hm, you're right, `HAVE_STRLCPY` isn't defined. The problem simply seems to be that `osdep-freebsd.c` doesn't include `compat.h`, either directly or via `tmux.h`. As the fallback already exists, might as well just use that then and make sure we get `compat.h` included.

nicm:
Yes, try this please:
~~~~
diff --git a/configure.ac b/configure.ac
index 8cd33b6c..1a3ba43c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -478,6 +478,9 @@ if test "x$found_forkpty" = xyes; then
 fi
 AM_CONDITIONAL(NEED_FORKPTY, test "x$found_forkpty" = xno)

+# Look for kinfo_getfile in libutil.
+AC_SEARCH_LIBS(kinfo_getfile, [util util-freebsd])
+
 # Look for a suitable queue.h.
 AC_CHECK_DECL(
        TAILQ_CONCAT,
diff --git a/osdep-freebsd.c b/osdep-freebsd.c
index 61f3f507..a7d02930 100644
--- a/osdep-freebsd.c
+++ b/osdep-freebsd.c
@@ -31,6 +31,8 @@
 #include <unistd.h>
 #include <libutil.h>

+#include "compat.h"
+
 struct kinfo_proc      *cmp_procs(struct kinfo_proc *, struct kinfo_proc *);
 char                   *osdep_get_name(int, char *);
 char                   *osdep_get_cwd(int);

jrtc27:
Isn't `AC_SEARCH_LIBS`'s default behaviour to give an error on failure? I.e. it shouldn't be outside the case statement?

jrtc27:
And is `compat.h` the right thing to include, or should we go with `tmux.h`? That's what the `aix/cygwin/hpux/linux/sunos/unknown` files include, and nobody includes `compat.h` outside of `compat/` and `tmux.h`.

nicm:
Yes, try tmux.h but it may fail because of struct session.

nicm:
AC_SEARCH_LIBS doesn't fail by default as far as I can tell.

jrtc27:
Oh, I see, `sys/proc.h`'s session, ew, ok, `compat.h` it is.

jrtc27:
Updated, confirmed to build fine (and dropped the final patch, though obviously still used locally so tmux actually works).

jrtc27:
> AC_SEARCH_LIBS doesn't fail by default as far as I can tell.

Hm, indeed it doesn't. I'd argue there's no point checking for `kinfo_getfile` on non-FreeBSD systems; would you still prefer it be outside the case?

nicm:
Better to keep platform-specific stuff to a minimum I think.

nicm:
Applied now, thanks!

PR #1558, opened 2018-12-02
configure.ac:use curses m4 file provided by GNU aclocal pkg

andy5995:
The curses test gets added to aclocal.m4 when ./autogen.sh is run. But
your travis log should show you all the details you need to know.


andy5995:
Hmm. tests failed. I've seen this work but not while using ./autogen.sh. (though it worked for me locally). Would you like to see if we can get this working? For my projects, I usually run `./autoreconf -iv` and commit the changed files. Then anybody who tries to build the code from git just runs `./configure` (no ./autogen.sh script), and they don't need the autotools packages installed.

nicm:
What are you trying to do and for what purpose?

andy5995:
Trying to replace the existing test for ncurses in configure.ac with the autoconf macro [AX_WITH_CURSES](https://www.gnu.org/software/autoconf-archive/ax_with_curses.html)

nicm:
What version of autoconf was it added in?

andy5995:
Here's the [revision history](http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=history;f=m4/ax_with_curses.m4)

nicm:
Yes, I don't mind doing this but it looks like the file isn't shipped with at least the autoconf versions I have here?

So tmux would need to ship it, and Thomas is right it is GPL. Perhaps you could ask the authors if they would relicense it under a more free license so we could include it? autoconf is a GNU project though so I suspect that is unlikely.

andy5995:
> Perhaps you could ask the authors if they would relicense it under a more free license so we could include it? autoconf is a GNU project though so I suspect that is unlikely.

Yeah, so unlikely I wouldn't bother. ;)

ThomasAdam:
Late to the party here, sorry about that.

Yes, this is GNU. That won't change, so this change is superfluous,
unfortunately.

As for Travis-CI output, unlike other projects, the failures are all I
check and actively fix. Everything else is just noise.

ThomasAdam:
Closed

PR #1564, opened 2018-12-17
[fix] Fix a crash due buggy log_debug usage

nartes:
1. To reproduce a crash please type in the following inside tmux:
        python -c "import sys; sys.stdout.buffer.write(bytes.fromhex('1b5b3a33346d28'))"

### DIFF
--- a/input.c
+++ b/input.c
@@ -1863,7 +1863,9 @@ input_csi_dispatch_sgr_colon(struct input_ctx *ictx, u_int i)
 				return;
 			}
 		}
-		log_debug("%s: %u = %d", __func__, n - 1, p[n - 1]);
+		if (n > 0) {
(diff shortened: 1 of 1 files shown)


nartes:
please merge the commit without squashing

nicm:
This is the fix, will be in GitHub a bit later. Thanks!
~~~~
--- input.c     19 Nov 2018 13:35:41 -0000      1.139
+++ input.c     17 Dec 2018 21:52:44 -0000
@@ -1862,7 +1862,8 @@ input_csi_dispatch_sgr_colon(struct inpu
                                free(copy);
                                return;
                        }
-               }
+               } else
+                       n++;
                log_debug("%s: %u = %d", __func__, n - 1, p[n - 1]);
        }
        free(copy);

PR #1589, opened 2019-01-28
accept normalized codeset on locale check

mitch-kyle:
Some documentation on [normalized codeset](https://www.gnu.org/software/libc/manual/html_node/Locale-Names.html)

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -218,7 +218,9 @@ main(int argc, char **argv)
 	const struct options_table_entry	*oe;
 
 	if (setlocale(LC_CTYPE, "en_US.UTF-8") == NULL &&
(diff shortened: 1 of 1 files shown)


nicm:
There is no documentation about this at that link, in fact all the examples use "UTF-8"?

This code is only a fallback for when the environment isn't set up properly. Can't you just set LC_ALL, LANG or LC_CTYPE correctly for you system?

mitch-kyle:
Yeah, I'm no longer convinced the issue I was experiencing is actually from this code or even this project. I'll do some tests and I'll be back if I need to be.

Sorry to bother, jumped the gun on this one

PR #1594, opened 2019-02-03
Import OpenSSH's setproctitle() compat shim.

multiplexd:
The current code sets the process title on Linux using `prctl(2)`, which according to the manual page is a Linux-specific system call. This is limited to sixteen bytes, and exposed via `/proc/<PID>/comm` -- some process viewers don't read the `comm` file (instead reading `/proc/<PID>/cmdline`), so any changes to the process title are not visible. From reading the code, I'm also assuming that tmux doesn't support changing the process title on platforms which lack a native `setproctitle()` and aren't Linux.

The way that OpenSSH and Postgresql do this on Linux is to make a copy of `argv` and move `environ` at startup, and then to overwrite the kernel-allocated `argv` and `environ` space with the desired process title, which is then exposed in the `cmdline` file under `/proc`. This also works on some other platforms which don't have a native `setproctitle()`, such as Cygwin and AIX (according to OpenSSH's `configure.ac`), and OpenSSH's code also supports `pstat()` on HP-UX, so I've attempted to import OpenSSH's `setproctitle()` shim as a replacement for the current code.

I've only been able to test this on (x86_64 glibc) Linux, as I don't have access to either a Cygwin or AIX system.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -279,6 +279,7 @@ const char	*getprogname(void);
 #ifndef HAVE_SETPROCTITLE
 /* setproctitle.c */
 void		 setproctitle(const char *, ...);
+void		 compat_init_setproctitle(int argc, char *argv[]);
 #endif
 
 #ifndef HAVE_B64_NTOP
--- a/compat/setproctitle.c
+++ b/compat/setproctitle.c
@@ -1,52 +1,170 @@
+/* Based on conf.c from UCB sendmail 8.8.8 */
+
 /*
- * Copyright (c) 2016 Nicholas Marriott <nicholas.marriott@gmail.com>
+ * Copyright 2003 Damien Miller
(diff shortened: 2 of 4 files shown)


nicm:
I don't want all this stuff in tmux.c.

nicm:
I'm not keen, because the argv[0] modification is a dreadful way to do this. This is not a particularly essential feature, so until Linux gets its act together and adds a sensible, joined-up way to do this, I'm inclined just to stick with prctl as a best effort. tmux only needs 14 or so characters anyway so the length limit is not a problem. If we did add it, it would need to be minimal impact. So adding maybe one or two lines to tmux.c is OK but not this much. TBH I think it best to just leave it as it is.

multiplexd:
Okay, I'll keep this patch in my local tree then.

PR #1595, opened 2019-02-06
Suppress misleading gseq not found warning

whereswaldon:
The `which gseq` command's stderr wasn't being redirected, so it would print an error even when it was working. I was trying to debug 24-bit color support (in a context where it was broken), and I thought I was missing a script dependency for a bit before realizing what was happening.


nicm:
Fixed, thanks.

PR #1601, opened 2019-02-12
fix buffer overrun caused by an ESC key with no following keys

nis8192:
How to replay:
1. in an xterm, run vim (or nvi) in a tmux pane.
2. type following key-sequence in vim.

 i a a ESC I x ESC

after the operation, the first line should be "xaa" and cursor should be located on "x" ,
but tmux generates extra keys `ESC O C', and the cursor will move onto "a".


nicm:
Show logs from reproducing this problem please.

PR #1611, opened 2019-02-18
add a plugin list link

sejimhp:
I started using tmux, I changed various settings, but I did not know that there are plugins.
I thought that it would be nice to have an introduction of plugins so I wanted to add it.

How about this?


nicm:
I don't know anything about these TBH and I don't really want to link to them, sorry.

PR #1620, opened 2019-03-09
Add origin relative mode

FractalBoy:
This pull request enables the use of DECOM/origin mode.

https://vt100.net/docs/vt510-rm/DECOM.html


nicm:
Thanks. Some comments:

- This is incomplete, you are only checking CUP so it is easy to move the cursor outside the region with, for example, CUU or CUD or VPA.
- From looking at xterm, it looks like sending DECOM itself should move the cursor to the new origin.
- There is no point to the change to screen_write_reset because the previous line clears the bit anyway. Also it should be &= not = if it was needed.
- MODE_ORIGIN_RELATIVE should be 0x2000 not 0x100000.
- The checks for behaviour should be in screen_write_* not in input.c.

FractalBoy:
@nicm I made the changes you requested - moving the checks to screen_write_cursormove should fix your first comment as well.

I can't account for why origin mode doesn't work if xterm has code that handles it. In my experience it is broken in tmux.

FractalBoy:
Actually, I must have not installed my changes before testing - that last commit breaks origin mode.

nicm:
I didn't mean it doesn't work in xterm, of course it works in xterm. You should check what xterm does to see how it should work in tmux.

This is still not quite right:
- I can still move the cursor outside the region, for example try this:
~~~~
printf '\033[H\033[2J\033[10;20r\033[?6h\033[30;1Hxyz\033[H'
~~~~
The xyz is printed outside of the region. It should be printed at 20,1 not 50,1. If origin mode is enabled, the cursor should always be inside the region. So you will need to clamp the cursor position in screen_write_cursor_move.

- The DECOM sequence itself should move the cursor to the new origin (so it has an implicit home).

- The cursor is being placed incorrectly when tmux sleeps. You will need to change the code in server_client_reset_state as well to take origin mode into account.

FractalBoy:
Thanks - I'll take a look xterm's CursorSet and other relevant a little closer.

nicm:
I think you are most of the way there just need to add a couple of bits as I mentioned.

FractalBoy:
@nicm I *think* I'm doing what you're asking (except the handling of tmux sleep). The cursor seems to be placed appropriately, but text is off (lower down than it should be). Any ideas/tips?

Without tmux:

image

With tmux:

image

nicm:
Actually, I am wrong - you should not need to change server_client_reset_state, it looks at the screen's cursor position which should be correct.

Not sure what the problem is but it is something to do with the tty drawing code because C-b r fixes it.

nicm:
The problem is that screen_write_collect_flush calls screen_write_cursormove.

nicm:
Try this diff (includes your changes), which makes functions in screen-write.c which don't want to use the origin call screen_write_set_cursor instead of screen_write_cursormove: [x.diff.txt](https://github.com/tmux/tmux/files/2954504/x.diff.txt)

I also made a couple of other minor changes.

You will still need to fix the other cursor movement functions (screen_write_cursorup and so on).

nicm:
All of the callers of screen_write_cursormove outside screen-write.c should be OK because they will be operating on a screen that they constructed (eg for the status line) and which won't have the ORIGIN mode set.

FractalBoy:
That appears to work for me. I'll take a look at the other cursor movements, but I'm not sure if they need to be changed. There doesn't appear to be any special treatment of CUU, CUD, or VPA for origin mode in VT100 documentation.

https://vt100.net/docs/vt510-rm/CUU.html
https://vt100.net/docs/vt510-rm/CUD.html
https://vt100.net/docs/vt510-rm/VPA.html

nicm:
No, but the documentation for DECOM says "The cursor cannot move outside of the margins." To me this means VPA, CUU and CUD will also need to check. But best to try them in xterm and see if it allows them to move the cursor outside the region in origin mode.

FractalBoy:
I used the following sequence, and saw the same results on xterm and tmux (in xterm).

```
printf '\033[10;20r\033[?6h'
printf '\033[40B'
```

image

`printf '\033[40A'`

image

The only difference was when clearing origin mode. I think we need to set the cursor to the new origin when clearing it.

`printf '\033[?6l'`

image

FractalBoy:
That last commit fixed it. Now they work identically.

image

FractalBoy:
VPA also appears to work.

image

nicm:
OK great.

FractalBoy:
@nicm I noticed that this causes an issue with CHA. I'll see if I can figure out how to fix it.

image

nicm:
Are you up to date, do you have commit 303d20a?

FractalBoy:
Oh, I guess not. I'll pull the most recent commits and see if it fixes my issue.

FractalBoy:
Nope, the issue remains. I think we may need to somehow ignore origin mode for HPA.

FractalBoy:
This diff works for me.

[hpafix.txt](https://github.com/tmux/tmux/files/2957914/hpafix.txt)

nicm:
OK good I'll take a look when I get home.

nicm:
I think it might be tidier to make screen_write_cursormove accept -1 to
mean don't-change like screen_write_cursor_set does, then just skip the
origin check if py == -1. You'll need to make its arguments signed.

FractalBoy:
I'll make a new merge request with my changes - I am seeing another slight difference, but I don't know which control sequence is causing it. Here is an example:

```
printf '\033[2;24r\033[?6h'
printf '\0338'
```
In tmux the cursor ends up on the next line (in the bounds), in xterm it goes out of bounds.

image

nicm:
I will fix HPA, here is the change 
[y.diff.txt](https://github.com/tmux/tmux/files/2958180/y.diff.txt)

FractalBoy:
We need a way to ignore origin mode for DECRC.

nicm:
Yes I think just add an argument to screen_write_cursormove, give me a sec I'm going to look at it now.

FractalBoy:
Actually, I think the issue is that we're not saving origin mode in DECSC.

nicm:
Yes, it looks like xterm saves origin mode with DECSC.

RCP is also wrong, and so is DECCOLM although I don't know what to do there, it probably doesn't matter.

nicm:
But even if we save the origin bit with DECSC, the stored cursor position is absolute, so will need a flag to screen_write_cursormove to bypass origin mode in order to restore the cursor.

FractalBoy:
No, because if the origin mode wasn't saved then we clear it in DECRC.

image

nicm:
Yes that's right, we need to save the origin mode flag, but s->cy is the absolute cursor position (not from the origin), so when it is restored with screen_write_cursormove it doesn't need the origin added.

FractalBoy:
Ah, I see. So we need to do both then.

nicm:
Right. I will be out for a couple of hours - if you want to do it it would be great, otherwise I'll look when I get back.

nicm:
Note I have tidied up some local variables and fixed HPA, it will be in Git in a  bit when it next syncs up from OpenBSD so watch out for conflicts when that happens.

FractalBoy:
I'll keep working on it - I see how xterm does it, so I'll see if I can fix it.

FractalBoy:
This works.

[saveom.txt](https://github.com/tmux/tmux/files/2958429/saveom.txt)

FractalBoy:
Okay, this works. I made the default origin mode and turned it off for the controls that we identified as ignoring origin mode.

[saveom.txt](https://github.com/tmux/tmux/files/2958469/saveom.txt)

FractalBoy:
The change is here as well.

https://github.com/FractalBoy/tmux/commit/422f2c8efeaa729fbb5e31210e54511066f9d2ab

nicm:
Does SCP not save origin mode as well?

nicm:
It looks identical to DECSC.

nicm:
OK I have applied this but with some changes, most notably I just added an argument to screen_write_cursor_move instead of playing games with #define. Also I added helper functions for SC/RC. Please try this when it makes it to GitHub and let me know if you still see any problems. Thanks!

FractalBoy:
I still see issues. I don't have time to look at this anymore today, but my master branch works correctly.

nicm:
I probably made a typo, I'll take a look.

FractalBoy:
@nicm CHA/VPA is broken again. Here's a diff with the fix. Just needed to turn off origin mode.

[x.diff.txt](https://github.com/tmux/tmux/files/2959772/x.diff.txt)

nicm:
Fixed, thanks.

FractalBoy:
@nicm I think you made a change based off a mistake I made - the change you made breaks VPA. HPA still works because origin mode was being ignored anyway.

FractalBoy:
Nevermind - neither VPA nor HPA works. You can fix VPA by respecting origin mode, but I don't know how to fix HPA.

FractalBoy:
Ah, I see the reason. You're only respecting CHA, not HPA. CHA works, HPA doesn't.

FractalBoy:
Alright, please use this diff which fixes VPA and adds support for HPA.

[x.diff.txt](https://github.com/tmux/tmux/files/2964290/x.diff.txt)

FractalBoy:
One more issue: the `clear` command seems to work differently between tmux and xterm in origin mode.

In this example, I executed `printf '\033[10;20r\033[?6h'` then `clear`.

image

However, it looks like `printf '\033[2J'` is working fine, which is confusing to me. Do you know if `clear` does something other than `printf '\033[2J'`?

`printf '\033[1J'` also works differently, but not like `clear`. It is fixed when I click on the screen.

image

FractalBoy:
`clear` just outputs `'\033[H\033[2J\033[3J'`, which works, but `clear` itself doesn't. Strange.

nicm:
I have applied your HPA fix, thanks. I must have read the previous change incorrectly.

With TERM=screen or tmux, clear sends \033[H\033[J not 2J, 0J is clear end of screen and 1J is clear start of screen. So \033[H\033[J will not clear the screen in origin mode. You must have checked the xterm terminfo entry where it is 2J.

FractalBoy:
Thanks, that makes sense. However it doesn't address the inconsistencies with 1J (second screenshot).

nicm:
There is a bug with ED 1 but it is not related to origin mode:
~~~~
--- tty.c       14 Mar 2019 09:53:52 -0000      1.318
+++ tty.c       14 Mar 2019 15:52:02 -0000
@@ -1672,7 +1672,7 @@ tty_cmd_clearstartofscreen(struct tty *t
        px = 0;
        nx = screen_size_x(wp->screen);
        py = 0;
-       ny = ctx->ocy - 1;
+       ny = ctx->ocy;

        tty_clear_pane_area(tty, ctx, py, ny, px, nx, ctx->bg);

FractalBoy:
That works!

nicm:
Great, fixed, thanks!

PR #1633, opened 2019-03-16
Add config location ~/.config

teldra:


ThomasAdam:
Thanks, but no.  This has been discussed a few times.  If you want something other than ~/.tmux.conf, symlink it.

PR #1644, opened 2019-03-20
docs: use markdown instead of plain text

DahlitzFlorian:
Use Markdown instead of plain text to make it easier for new contributors and users to read the README files.


nicm:
I'm quite happy with them as they are thanks. If you want to work on some documentation improvements, the man page could perhaps do with revisiting.

Issue #810, opened 2017-03-06
tmux -V returning "master" instead of version number


agrajag9:
When compiling from source in github, `PACKAGE_VERSION='master'` and `PACKAGE_STRING='tmux master'`. This results in a compiled binary that returns `tmux master` to the command `tmux -V` instead of a useful version number. This in turn causes problems for other codebases that rely on determining tmux versions for integration (e.g. powerline).

ThomasAdam:
Seems this changed in c0346701dfa2a4aa4c2ea6f05157406bcfb7a50b

nicm:
This is intentional, programs should not rely on the format of tmux -V.



On 6 Mar 2017 10:48 pm, "agrajag9" <notifications@github.com> wrote:

When compiling from source in github, PACKAGE_VERSION='master' and
PACKAGE_STRING='tmux
master'. This results in a compiled binary that returns tmux master to the
command tmux -V instead of a useful version number. This in turn causes
problems for other codebases that rely on determining tmux versions for
integration (e.g. powerline).

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/tmux/tmux/issues/810>, or mute the thread

agrajag9:
Feature request: version numbering.

nicm:
Releases are versioned, master is not. There is no reason that "2.4" meaning one of 2.4-release or master is any better than "2.4" meaning 2.4-release and master meaning master, and there are advantages to the latter.

ZyX-I:
> This is intentional, programs should not rely on the format of tmux -V.

If programs should not rely on this format how they are supposed to get tmux version? It is much more strong statement then refusing to output version number in git builds.

Also what are advantages of not showing version number on master? I think that best variant is something like `v0.2.0-827-gff6fa39`: this is Neovim version. Basically the output of `git describe`: contains last release version, number of commits since then and actually used hash. This, of course, does not work for non-VCS builds, but most of time it is good.

nicm:
They should not rely on the format always being the same, if they break
because tomorrow we decide to call the next version tmux 'Randy Rabbit'
2017 instead of 2.4 then it is a bug.

I don't want Git to be a build time dependency, so 'x.y' or 'master' is the
most we can do without that.


On 6 Mar 2017 11:46 pm, "Nikolai Aleksandrovich Pavlov" <
notifications@github.com> wrote:

This is intentional, programs should not rely on the format of tmux -V.

[...]

PR #1650, opened 2019-03-24
Add git commit hash to -V option display result

h-michael:
If the user use git command and build at directory has `.git`, `tmux -V` show git commit hash.

ex.
```
$ tmux -V
$ tmux next-2.9 1d6fe43c
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # configure.ac
 
-AC_INIT([tmux], next-2.9)
+AC_INIT([tmux], next-2.9 [m4_esyscmd([./configure.commit])])
 AC_PREREQ([2.60])
 
 AC_CONFIG_AUX_DIR(etc)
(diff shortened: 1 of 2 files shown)


nicm:
No I don't want this thanks.

h-michael:
@nicm 
I'm sorry for not knowing that you don't want that.
I build tmux everyday when new commit is pushed.
But I sometimes face to regressions.
If I can know tmux's commit hash I use, I can more easily to report the issue.

Would  you tell me why you don't want?

nicm:
I don't think the build should have  a dependency on the VCS and I don't think the version string should depend on where you got the code. If you are reporting problems it is best to reproduce with the latest Git anyway.

h-michael:
> I don't think the build should have a dependency on the VCS

I understand your thought.
How do you think adding patch number to each code changes ?

nicm:
That means  a lot of extra work for something that I don't see how can
really be a big issue. If you are running a release, they are versioned
already. If you are running master, then you should be doing git pull
before reporting any issue anyway. If you really must keep track of the
exact commit you built from, then don't do git pull without building, or
make your own note of what commit you built from. Or look at the timestamp
on /usr/bin/tmux.

h-michael:
I see.
related issue #810

I decided to resolve my problem with this comment.
https://github.com/tmux/tmux/issues/810#issuecomment-403418325

Issue #1649, opened 2019-03-23
Cannot bind mouse to status center (empty) area


avih:
I can bind things to `StatusRight` or `StatusLeft`, but when I bind to `Status` e.g. `bind-key -n MouseDown1Status display-message MouseDown1Status` then it's accepted as valid but I can't find a place on the status line to click which triggers the binding.

I'd expect `Status` to be either a global fallback where nothing more specific is defined (which includes the "empty" center part of the status line), or simply the status area between left and right.

Right now, with current tmux master, it seems that clicking the status center area is interpreted as if clicked on the nearest pane char-cell, which I think is unexpected.

nicm:
Works for me.

nicm:
Show full config and logs please.

avih:
I'll bisect the config and let you know the result.

avih:
This is a config which reproduces the issue for me with commit 0292243b :
```
set-option -g mouse on

bind-key -n MouseDown1StatusLeft  display-message StatusLeft
bind-key -n MouseDown1StatusRight display-message StatusRight
bind-key -n MouseDown1Status      display-message Status
```

If I press and hold the left mouse button over the left text of the status bar (`[0] 0:bash*`), it shows "StatusLeft". If I hold it over the right (host and date) it shows "StatusRight", and if I do that at the middle empty area of the bar then nothing happens.

Further more, if I then split horizontally (prefix-%), then clicking at the empty area of the status bar left to the middle selects the left pane, and right to the middle selects the right pane.

Some commits ago (not sure how many, probably some weeks) the StatusLeft and StatusRight bindings behaved the same as now, but the middle empty area simply did nothing (it's not a regression that I know of, I never succeeded binding anything to the middle area of the status bar).

I considered "nothing" a bug at the time but didn't report it, but now that it behaves as if the pane area was clicked (which I also consider a bug), I decided to report it.

avih:
Sorry, I reported the behavior incorrectly.

At the left side (`[0] 0:bash*`) if I click the `[0]` then it shows "StatusLeft", and if I click the `0:bash*` then it shows "Status".

So it seems that that all 3 bindings respond.

The middle empty-area description and change was reported correctly.

I think it's now more unexpected than before, and as far as I can tell there's no way to bind the middle empty area to anything - which is a shame.

nicm:
You are not explaining what you are seeing very clearly.

0:bash is the window list (so it is Status) the remaining text is either
[...]

PR #1653, opened 2019-03-25
Support `StatusDefault` mouse target binding

avih:
Following #1649


avih:
Hmm.. for some reason the "add missing" commit shows at the UI after the "support mouse..." commit, even though it's the earliest of the two. Probably because I authored it later and then moved it to earlier in a `rebase -i`, but it kept the later date...

nicm:

avih:
Thanks :)

avih:
I noticed that you didn't remove the TODO item. Is that intentional?

PR #1670, opened 2019-04-09
Update session activity when a pane is focused.

tafryn:
Focus-in events are a good indication that the client that received the event is now in use. Updating the session activity after such an event allows commands issued outside of a session to affect the newly focused client instead of the last client that saw a key-press.


nicm:

PR #1674, opened 2019-04-09
compat: make sure _PATH_DEFPATH is defined

carenas:
if system doesn't provide paths.h, all _PATH_* variables used should
come instead from compat.h (included from tmux.h)


nicm:
Applied thanks.

PR #1685, opened 2019-04-23
copy-{selection,pipe} now both clear, add -no-clear variants

### LINKED ISSUE #1683: `send -X copy-selection` unexpedtedly also clears it on-screen
Using a recent tmux master (a week or so) and this config:
```
set-option -g mode-keys vi
set-option -g mouse on

bind -Tcopy-mode-vi MouseDragEnd1Pane send -t= -X stop-selection \; send -t= -X copy-selection
```

Seeing that the original binding (in `key-bindings.c`) is:
```
bind -Tcopy-mode-vi MouseDragEnd1Pane send -X copy-selection-and-cancel
```

I wanted to copy it but stay in copy-mode such that the selection is finished but kept visible on screen, and so I changed the command to `stop-selection` + `copy-selection`.

However, while copying and staying in copy-mode, it seems to clear the selection on screen.

If I use this instead:
```
bind -Tcopy-mode-vi MouseDragEnd1Pane send -t= -X stop-selection
```
Then it stays in copy-mode and visible, but the selection isn't copied.

As a work around, it appears that any `copy-pipe` also copies the selection, so this seems to achieve the goal of copying, and staying in copy-mode with the selection visible:
```
# Use dummy copy-pipe to copy and keep the selection visible
bind -Tcopy-mode-vi MouseDragEnd1Pane send -t= -X stop-selection \; send -t= -X copy-pipe ":"
```

However, I think this is undocumented (copy-pipe should be sending the selection to a pipe, not necessarily also copying it internally), and there's also there's the `send -X clear-selection`  command which seems to be clearing the selection on-screen while keeping the paste-buffer with its content unmodified.

Button line, it seems like `copy-selection` also does `clear-selection`, but I think it shouldn't.

avih:
The first commit makes copy-pipe clear the selection (like copy-selection did and still does), and adds two new -no-clear variants.

The second commit is a minor refactor. Not sure if you'll want it, and if not then at least I tried :)

Fixes #1683


nicm:
Looks good, you need to add them to the man page also.

avih:
Ouch. I kept it in mind and for last, and then didn't :) Updating.

avih:
Updated tmux.1

nicm:

avih:
Thanks.

By the way, what's the repository URL of this "Applied to OpenBSD"? does it have some online web-view?

nicm:
http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/tmux/

Issue #1498, opened 2018-10-10
Add the ability to pass environment variables to `new-session`


ismell:
When creating a session I haven't found a clean way of setting up the environment for a session before the initial command is executed. Instead I need to set the environment variable in the global environment, create the session, then use set-environment after the session has been created.

```shell
tmux set-environment -g MY_VAR foo \; \
     new-session -d -s foo watch 'echo $MY_VAR; tmux show-environment MY_VAR' \; \
     set-environment -t =foo MY_VAR foo \; \
     set-environment -g MY_VAR bar \; \
     new-session -d -s bar watch 'echo $MY_VAR; tmux show-environment MY_VAR' \; \
     set-environment -t =bar MY_VAR bar \; \
     set-environment -gu MY_VAR \;

```

I propose adding a new flag to new-session to allow setting the environment. i.e.,

```shell
tmux new-session -d -s foo -e MY_VAR=foo watch 'echo $MY_VAR; tmux show-environment MY_VAR' \; \
     new-session -d -s bar -e MY_VAR=bar watch 'echo $MY_VAR; tmux show-environment MY_VAR' \;
```

I can't use `update-environment` because I only want the session variable to be set on creation and not updated when reattached.

nicm:
Added to TODO file.

nicm:
Try this please: [tmux-environ-arg.diff.txt](https://github.com/tmux/tmux/files/3121768/tmux-environ-arg.diff.txt)

Note that -e is space or comma separated so you can't have environment variable values with spaces or commas in them.

ismell:
I'll try it out this week. Thanks!

nicm:
A variant of the PR was applied instead of my patch so you can now build from master and use -e.

ismell:
Even better. Thanks!

PR #1697, opened 2019-04-26
Add support for providing environment variables to split-window command

christgau:
Here is an alternative implementation for providing environment variables to a new pane created by split-window (see #1498). The PR generally add support for commands to have flags that can occur multiple times. Those must be provided explicitly in the command specification struct. There is a little hacky solution to allow multiple occurrences of the same flag to be present in the args RB tree.  However, the approach allows to have multiple environment variables which can contain spaces and commas to a new pane. This was a must for my use case, where the passed variables are provided are not under user control and do  contain those special characters.

The patch is limited to split-window but other commands may benefit from this extension as well. However, it is PoC and surely needs some review.

### DIFF
--- a/arguments.c
+++ b/arguments.c
@@ -146,13 +146,14 @@ args_print(struct args *args)
 	struct args_entry	*entry;
 	static const char	 quoted[] = " #\"';$";
 	char			*value;
+	struct arg_value	*value_entry;
 
 	len = 1;
 	buf = xcalloc(1, len);
 
 	/* Process the flags first. */
 	RB_FOREACH(entry, args_tree, &args->tree) {
-		if (TAILQ_EMPTY(&entry->values))
+		if (!TAILQ_EMPTY(&entry->values))
 			continue;
 
 		if (*buf == '\0')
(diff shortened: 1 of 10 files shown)


nicm:
I would rather this is supported, but I'm not fond of this design.

I would prefer that multiple arguments be stored in a list (TAILQ) attached to the struct args_entry. It would probably be easiest to add a new struct for the argument values.

I don't think there is any need to list options that allow multiple instances. Just accept multiple for any argument, but have args_get always return the last one that appeared.

I think a first/next API would be nicer than allocating a new list to return the arguments.

nicm:
This also needs to work for new-window, respawn-window and respawn-pane, perhaps new-session also.

Thanks

christgau:
Thanks for your feedback!

> I would prefer that multiple arguments be stored in a list (TAILQ) attached to the struct args_entry. It would probably be easiest to add a new struct for the argument values.

Isn't an STAILQ sufficient for argument values? Those are only added to the last and the list may be iterated only in one direction.

nicm:
There isn't much value in using the different queue types, especially for areas like this where there aren't that many items in the list - at most you'll save a few 10s of bytes. Best to just use a TAILQ.

christgau:
I removed that PoC RB hacky thing by the proposed TAILQ approach - Hopefully, in the way you intended. 😉 

I also added the find first/next mechanism for iterating the values. `args_set` now _adds_ values to the flag, if a value was provided. It does not replace them as before. Don't know if the function it should be renamed therefore. However, if a single value of a flag is needed (and multiple are present), only the last one is considered, like in `args_get`.

As for `args_print` the function does not yet respect the list of arguments but only considers the last one. Should it print all values instead, if more than one is present?

nicm:
Thanks will take a look later. Yes I think it should print them all.

nicm:
Thanks, this mostly looks good. Here [tmux-e.diff.txt](https://github.com/tmux/tmux/files/3124063/tmux-e.diff.txt) is an updated diff with the following changes:

- arg_value -> args_value to match the prefix of the rest of the file.
- Use TAILQ_FOREACH_SAFE to free values.
- Break up args_print a bit more (I might do this separately).
- Rename args_find_first_value just to args_first_value.
- I prefer if the iterator is an argument rather than the return value, change it to be that way round.
- Some trivial renaming of variables and other style bits.
- Don't modify s->environ in spawn.c, modify the copy from environ_for_session instead.

christgau:
> Thanks, this mostly looks good. 

Thank you.

>     * Don't modify s->environ in spawn.c, modify the copy from environ_for_session instead.

My intention here was to prevent the user from overriding some variables which appear to be critical, like `TERM` and `TMUX` which are set in `environ_for_session`, when a the window is split.

nicm:
Well, TBH I think if people are doing that it is their own fault, but if
you really want to avoid it the best thing to do might be to pass the -e
environment into environ_for_session and let it do the merge. I don't know
if I'd bother.

christgau:
> Well, TBH I think if people are doing that it is their own fault

Yeah, that's also a valid argumentation. I'm fine with whatever decision, as long as we can set environment variables for new panes. 😉

nicm:
OK this just needs to also be added to new-window, respawn-window,
respawn-pane and also to tmux.1.

nicm:
I have applied this to OpenBSD now, it will appear in GitHub in an hour or so when it syncs up. Thanks for your work!

Issue #1701, opened 2019-04-30
tmux 2.9: select-pane null dereference


gnachman:
This affects all platforms and does not depend on tmux.conf (I have none). I have confirmed the crash still occurs on master.

tmux version 2.9 crashes in control mode with the command "select-pane" when there are multiple window panes in the current window.

To reproduce the crash:

1. Create two window panes
2. detach
3. `tmux -C`
4. `select-pane -t %0`

## Analysis

The reason is `tty->client` is `NULL` in `tty_window_bigger`. Its caller, `cmd_select_pane_redraw`, iterates over clients and calls passes it the client's tty. Control mode clients ttys don't have a pointer back to the client because the tty is not initialized in `server_client_dispatch_identify` since there isn't a tty attached.

nicm:
What is the impact of this?

rprimus:
Tue Apr 30 18:48:34 BST 2019

Tmux dies (started with `tmux -CC att -dt <session>`).   Please see:
https://gitlab.com/gnachman/iterm2/issues/7758

nicm:
I wasn't asking you, and I wasn't asking what happens, I was asking what the impact is. Is every user going to see this or only a few?

Jesin:
When you said "every user", did you mean "every user of [iTerm2](/gnachman/iTerm2)" or "every user of tmux"? I use tmux and I haven't personally encountered this problem so far (this is the first time I've heard of "control mode" or iTerm2), though a server segfault sounds like it could be pretty bad.

gnachman:
iTerm2 users who use control mode (`tmux -CC`) experience crashes when switching between window panes. I don't have good numbers on how many people this affects as I don't have telemetry.

nicm:
It happens to everyone who uses control mode though?

nicm:
I have made a 2.9a release which fixes this and another bug with layouts.

@gnachman would it be OK if I CC iterm2-discuss@googlegroups.com on RC and release announcements? Clearly nobody who uses iTerm2 + tmux tried this before now.

gnachman:
Provided they use split panes, they'll encounter the bug.

I would be very glad to get RC announcements. I always forget to keep up to date on them and sometimes get surprises. Thank you!

nicm:
Alright, I'll try to remember! Thanks!

PR #1702, opened 2019-04-30
Don't redraw control mode clients.

gnachman:
Fixes issue #1701


nicm:

PR #1731, opened 2019-05-10
Do not reduce window height by status bar height for control mode clients

gnachman:
Prior to commit 646995384d695eed9de1b2363fd2b315ca01785e, the status bar's height was not deducted from window heights in `recalculate_sizes()` for control mode client. That commit moved one of the two checks for control mode into `ignore_client_size`, and the other was lost.

As a result, control mode windows have a usable area that is always smaller than the available space.

### DIFF
--- a/resize.c
+++ b/resize.c
@@ -174,7 +174,7 @@ recalculate_sizes(void)
 	TAILQ_FOREACH(c, &clients, entry) {
 		if (ignore_client_size(c))
 			continue;
-		if (c->tty.sy <= status_line_size(c))
+		if (c->tty.sy <= status_line_size(c) || (c->flags & CLIENT_CONTROL))
 			c->flags |= CLIENT_STATUSOFF;
 		else
 			c->flags &= ~CLIENT_STATUSOFF;


nicm:
I think this is wrong additionally because status_line_size checks STATUSOFF itself. It should just use s->statuslines directly. Also checking for CONTROL in status_line_* couldn't do any harm also.

Does this cause any problems: 
[x.diff.txt](https://github.com/tmux/tmux/files/3165740/x.diff.txt)

~~~~
Index: status.c
RCS file: /cvs/src/usr.bin/tmux/status.c,v
retrieving revision 1.197
diff -u -p -r1.197 status.c
--- status.c	3 May 2019 10:00:48 -0000	1.197
+++ status.c	10 May 2019 09:10:17 -0000
@@ -203,7 +203,7 @@ status_at_line(struct client *c)
 {
 	struct session	*s = c->session;
 
-	if (c->flags & CLIENT_STATUSOFF)
+	if (c->flags & (CLIENT_STATUSOFF|CLIENT_CONTROL))
 		return (-1);
 	if (s->statusat != 1)
 		return (s->statusat);
@@ -216,7 +216,7 @@ status_line_size(struct client *c)
 {
 	struct session	*s = c->session;
 
-	if (c->flags & CLIENT_STATUSOFF)
+	if (c->flags & (CLIENT_STATUSOFF|CLIENT_CONTROL))
 		return (0);
 	return (s->statuslines);
 }
Index: resize.c
RCS file: /cvs/src/usr.bin/tmux/resize.c,v
retrieving revision 1.31
diff -u -p -r1.31 resize.c
--- resize.c	17 Apr 2019 14:43:49 -0000	1.31
+++ resize.c	10 May 2019 09:10:17 -0000
@@ -174,11 +174,12 @@ recalculate_sizes(void)
 		if (ignore_client_size(c))
 			continue;
-		if (c->tty.sy <= status_line_size(c))
+		s = c->session;
+		if (c->tty.sy <= s->statuslines || (c->flags & CLIENT_CONTROL))
 			c->flags |= CLIENT_STATUSOFF;
 		else
 			c->flags &= ~CLIENT_STATUSOFF;
-		c->session->attached++;
+		s->attached++;
 	}
 
 	/* Walk each window and adjust the size. */

gnachman:
I tested it and it works.

nicm:

PR #1733, opened 2019-05-10
Add support for overscore

rbanffy:
Adding support for overline (from ECMA-48) to the status bar styling. Tested on Gnome Terminal and Konsole (also tested on xterm, pandoterm, rxvt, Cool Retro Term, macOS's Terminal, Cathode and iTerm, but none of these support the attribute)

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -790,6 +790,7 @@ grid_string_cells_code(const struct grid_cell *lastgc,
 		{ GRID_ATTR_UNDERSCORE_3, 43 },
 		{ GRID_ATTR_UNDERSCORE_4, 44 },
 		{ GRID_ATTR_UNDERSCORE_5, 45 },
+		{ GRID_ATTR_OVERSCORE, 82 },
 	};
 	n = 0;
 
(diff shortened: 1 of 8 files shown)


nicm:
Looks good, you just need to invent a terminfo capability to turn this on
rather than sending the escape sequence directly. Also don't change the
example file please.

tmux.h
```diff
@@ -445,6 +445,7 @@ enum tty_code_code {
 	TTYC_VPA,
 	TTYC_XENL,
 	TTYC_XT,
+	TTYC_OVERLINE,
```
rbanffy: This is useless right now.

nicm:
Yes you need to define TTYC_something I suggest SMOS (Smos) to match eg smul then add it to tty-term.c as well then you'll be able to use it with tty_term_has and tty_putcode.

grid.c
```diff
@@ -790,6 +790,7 @@ grid_string_cells_code(const struct grid_cell *lastgc,
 		{ GRID_ATTR_UNDERSCORE_3, 43 },
 		{ GRID_ATTR_UNDERSCORE_4, 44 },
 		{ GRID_ATTR_UNDERSCORE_5, 45 },
+		{ GRID_ATTR_OVERLINE, 53 },
```
rbanffy: The 53 here also looks mostly useless because I'm using a literal in tty.c:2203. I'm trying to figure out where this gets connected to the TTYC_ codes that are used by tty_putcode

nicm:
no, that is correct it is for capture-pane

nicm:
You literally need to add about two lines and it's done. Look at something like HIDDEN except where it says TTYC_INVIS or "invis" put your new capability name.

nicm:
I have applied this, but used Smol rather than smol for the capability name. Thanks!

Issue #1732, opened 2019-05-10
[Meta] Backwards compatibility and deprecation warnings


HaleTom:
### Context

This is a **meta**-issue about this project's philosophy of introducing non-backward compatible changes in minor (eg the `Y` in `vX.Y` ) releases.

Individual instances of the pain associated with this philosophy in relation to specific releases have been raised:
* [2.1: Missing backwards compatibility #155](https://github.com/tmux/tmux/issues/155)
* [Tmux 2.8.X to 2.9.X migration #1689](https://github.com/tmux/tmux/issues/1689)

But this issue is not about any particular release's non-backward-compatibility. It is about the strategy of:

1. Generally not allowing for backward-compatibility
2. Having breaking changes released as minor versions

A reasonable expectation in regards to breaking changes is outlined by [semver](https://semver.org/):

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards-compatible manner, and
> 3. PATCH version when you make backwards-compatible bug fixes.

I recognise that this project has chosen it's own versioning, but I can't the definition of what is implied by different types of version bumps.

Please note that I'm **not** suggesting that breaking minor changes should have been a major version bump instead.

I **am** proposing that backward compatibility is considered **usual** and **best practice**.

### Issue
As an example of what occurs on every breaking change, I timed how long it took me to read the 2.8 -> 2.9 issue (Google took me there first), read the doco and examples, and update based on the recent `-style` changes: 23 minutes.

I consider myself a medium-to-advanced level of proficiency. If every starred user takes 20 minutes to do the update, then that's:

Total: 4601 hours or **192 days** of human time.  And that's *only* the starred users. It's probably way more than 10x that, or a **conservative 5.25 human years**.

I use the rolling-release Arch Linux which provides the latest `tmux`, which in turn **requires** me to have a `.tmux.conf` which aligns to the latest version of tmux.

What am I supposed to do when I want to use a fixed-release distribution such as Ubuntu with an older version of `tmux`?

### Work-around
To support older LTS distributions as well as my more more current workstation I am forced to either:

1. Maintain multiple `.tmux.conf` files
1. [Use an ugly hack to allow my config to work across multiple versions](https://stackoverflow.com/a/40902312/5353461):

```
# Version-specific commands [grumble, grumble]
# See: https://github.com/tmux/tmux/blob/master/CHANGES
[...]

PR #1734, opened 2019-05-10
Improve README file

NicolasCARPi:
Hello,

I was reading your messages on #1732 and I thought it was time to try and contribute to tmux myself. Yes I had issues with the latest release and yes I also thought the changelog wasn't verbose enough about it, but you make valid points, it's one thing to complain, and another to actually contribute with testing, documentation and the rest, so here I am! :)

*Side note: my first simili-contribution to this project was creating the [tmux archwiki page](https://wiki.archlinux.org/index.php/Tmux) more than 9 years ago! :D*

Because I don't know how to write C code, my contribution is on the README.

See the before/after below:

Montage-1

Isn't it more sexy? 99% of users reading this file will do it from the GitHub page, so there is no need for keeping it barebones. And it's still very readable in text mode (the advantage of markdown!).

Here is the commit message:

* Use Markdown syntax: readable in text mode and prettier when seen on
GitHub
* Create chapters/headers for clearly separating paragraphs
* Use GNU/Linux instead of just Linux
* Update links for ncurses and libevent
* Use syntax highlighting for code blocks (and drop the '$' which is
confusing for some users)
* Group the contributing bits together and mention the CONTRIBUTING file
* Add a link for the ISC license

Cheers :)
~Nico

PS: yes users will copy paste with the `$` sign and not understand why it doesn't work. Believe it. Also, it makes it easier for copy pasting ;)


nicm:
Thanks for doing this I appreciate the effort but I don't think it's a good idea to ship complex rich text files in the tmux tarball. Most people will be reading this in a terminal text editor not in a web browser so I don't think any form of markup is helpful, especially not for such a short file.

If you could separate out your text changes (but not GNU/Linux please I don't like that), I will apply those.

Possibly the marked up version could be kept and go in a subdirectory if it would be useful.

nicm:
Can GitHub be told to look in a subdirectory for the README file it displays? Then we could put a marked up version in without cluttering up the source tree. It would be nice to put CONTRIBUTING away as well.

ThomasAdam:
On Fri, May 10, 2019 at 03:05:28PM -0700, Nicholas Marriott wrote:
> Can GitHub be told to look in a subdirectory for the README file it
> displays? Then we could put a marked up version in without cluttering up the
> source tree. It would be nice to put CONTRIBUTING away as well.

I don't think so.  But a symlink might do it from the top-level directory.

-- Thomas

nicm:
Yes it says: "If you put your README file in your repository's root, docs, or hidden .github directory..."

So just need to test to see if a file in .github will be used instead of the root README.

nicm:
Well a symlink doesn't help because if is still cluttering the top level source directory with GitHub files. I'll test tomorrow and see if .github/README{,.md} works and if so then @NicolasCARPi's version can go there.

ThomasAdam:
This worked:

https://github.com/ThomasAdam/tmux/tree/ta/test

So it looks like you can.  I'm happy to pick this up if you like, @nicm

nicm:
Great stuff, it would be nice to have the fancy README in there and the
same text in the plain 80-column no-markup /README.

It would also be good to move CONTRIBUTING (and ISSUE_TEMPLATE if possible)
in there and add mark up to it.

NicolasCARPi:
I'm making another PR with README in .github and also moving  the contributing file and add an issue_template. Stay put :)

NicolasCARPi:
See #1735

PR #1735, opened 2019-05-10
Improve the README file

NicolasCARPi:
Now in the .github folder. See discussion on #1734.


ThomasAdam:
Thanks.

The idea was to leave `/README` alone.  This is already part of `EXTRA_DISTS` and the formatting changes here aren't wanted.

NicolasCARPi:
> The idea was to leave /README alone.

I changed it slightly because @nicm said:

> README in there and the same text in the plain 80-column no-markup /README

There is no markup in /README.

nicm:
Thanks. I don't think ISSUE_TEMPLATE should have so much in there - it is the initial text that appears in the text box for users to edit. I think it is inconvenient to put so many instructions in there they will need to delete. The idea was to put the instructions in CONTRIBUTING and a link there in ISSUE_TEMPLATE just as a reminder.

If ISSUE_TEMPLATE is to have more it should be very short, maybe just a list of headings or fields for the user to fill in.

The other changes look good on a quick read although I will try to take a closer look later.

Thanks!

HaleTom:
If I understand it, this will duplicate `README` into a formatted `README.md`.

In the interests of DRY, and given that the markdown version is so readable in a text editor, what is the reason to introduce a maintenance / inconsistency checking overhead later on by keeping the old format?

nicm:
OK. I have made a few changes to README and applied it:

- I like the headings but I don't like using two lines  and ==== I find a bit ugly, I've just changed to "* Heading".
- I tweaked some of the text for conciseness, there is a lot of passive voice (my fault).
- I don't want to include CONTRIBUTING.md or CONTRIBUTING in the tarball so I don't want to link to it from README (.github/README.md is fine).
- Used tabs consistently for indentation.
- I don't really see a need to mention tmux-git ML, few people will want it.

I'll take a look at the MD files in a bit.

nicm:
I have edited CONTRIBUTING.md:

- Use a question and answer form and the second person.

- I don't think it is necessary to mention the ML in here. This is a GitHub file, it is for GitHub issues, I think it is obvious that is where issues will be.

- Expand the text a bit.

nicm:
I've applied README.md but minimal changes:

- I think constantly doing **tmux** is a bit distracting, treat it like any other word.
- Otherwise just change my text inline with ../README.

nicm:
ISSUE_TEMPLATE is fine once I read the file rather than looking at the patch :-). Applied with some minor tweaks. Thanks!

nicm:
I hate that GitHub surrounds all these files with its frames and crap instead of just showing the text, that's why I liked linking to the raw versions but that doesn't work for MD files. Oh well, I suppose we just need to leave it.

nicm:
OK these look good. If you have any additional tweaks please let me know! Thanks!

nicm:
@HaleTom markdown isn't easy to read, it is clutter and I think the main README file is better without it. It hardly ever changes so two versions isn't a problem.

NicolasCARPi:
Hello,

So if I understand, you took my changes and applied them yourself instead of requesting changes from this PR before merging it. It's a bit weird, even if you put my name in the commit message.

Should this PR be closed then?

nicm:
It doesn't seem that weird... it is a pretty normal, you send patches and the committer reviews and applies them. I could have bounced a patch back over to you but it seemed quicker to just commit it, then at least it is in and any further changes can be made on it in tree.

But this PR can be closed if you don't have any further changes. Thanks!

NicolasCARPi:
Ok maybe «weird» isn't the right term. Let's say «uncommon», as with most git project I contributed to, my branch was merged so that I was the author of the commit(s). But if that's your workflow, why not.

My point of view is that the «reward» for a contribution to a project is to appear in the author/contributor list. And with this workflow this doesn't exist.

Anyway, I'm glad the tmux project now has a better README page :) 

~Nico

nicm:
Well, most code changes go into OpenBSD first so they are committed with CVS and then synced up to the portable version automatically later on. Even for stuff that doesn't, it is just much easier for me just to download and apply the patch, review it in emacs and commit it than messing around with git and branches. There is very rarely a diff that doesn't need some small change or other, often code style or whitespace, and it is just a bit of a waste having to go back and forth to people with that sort of minor stuff when the change is otherwise correct. I always try to tell people what I'm done differently if I do anything more than minor tweaks, and if they are operating under a real name then I mention it in the commit message. I'm not sure how much of a reward it really is anyway, but it doesn't seem much different having your name in one place in the commit message than in another...

Anyway, thanks for helping with this!

NicolasCARPi:
@nicm Thanks for the clarification, now I understand better why you did it! :)

PR #1741, opened 2019-05-16
Fix handling "optional" color space identifier parameter for the direct-color SGR sequence

townba:
Fix handling "optional" color space identifier parameter for the direct-color SGR sequence.

ITU T.416-199303, for example, specifies a color space identifier parameter for the direct-color SGR sequence. `tmux` support was added in https://github.com/tmux/tmux/commit/051a29ca03718c8fedd5918ba86556f29672348e but broken as a result of https://github.com/tmux/tmux/commit/4e3d6612845e190a490f40cce79c858dadaee74b.

To test:
* Run the command `printf '\x1B[48:2:255:0:0m(noparam:red)\x1B[48:2::0:0:255m(withparam:blue)\x1B[m\n'`. The output should be in red and blue. (Before this fix, it was only in red.)


ThomasAdam:
This is a mess, and difficult to read.

You might need to rebase this on top of master appropriately.

townba:
> This is a mess, and difficult to read.
> 
> You might need to rebase this on top of master appropriately.

I completely agree. Doing that now.

townba:
*The list of commits was ridiculous, but the "Files changed" showed just the single file.)

PR #1742, opened 2019-05-16
Fix handling "optional" color space identifier parameter for the direct-color SGR sequence

townba:
ITU T.416-199303, for example, specifies a color space identifier parameter for the direct-color SGR sequence. `tmux` support was added in https://github.com/tmux/tmux/commit/051a29ca03718c8fedd5918ba86556f29672348e but broken as a result of https://github.com/tmux/tmux/commit/4e3d6612845e190a490f40cce79c858dadaee74b.

To test:
* Run the command `printf '\x1B[48:2:255:0:0m(noparam:red)\x1B[48:2::0:0:255m(withparam:blue)\x1B[m\n'`. The output should be in red and blue. (Before this fix, it was only in red.)

### DIFF
--- a/input.c
+++ b/input.c
@@ -1940,12 +1940,11 @@ input_csi_dispatch_sgr_colon(struct input_ctx *ictx, u_int i)
 	}
 	if (p[0] != 38 && p[0] != 48)
 		return;
-	if (p[1] == -1)
-		i = 2;
-	else
-		i = 1;
+	i = 1;
 	switch (p[i]) {
 	case 2:
+		if (p[i + 1] == -1)
+			i++;
 		if (n < i + 4)
 			break;
 		input_csi_dispatch_sgr_rgb_do(ictx, p[0], p[i + 1], p[i + 2],
--- a/input.c
+++ b/input.c
@@ -1938,22 +1938,25 @@ input_csi_dispatch_sgr_colon(struct input_ctx *ictx, u_int i)
 		}
 		return;
 	}
-	if (p[0] != 38 && p[0] != 48)
+	if (n < 2 || (p[0] != 38 && p[0] != 48))
 		return;
-	i = 1;
-	switch (p[i]) {
+	switch (p[1]) {
 	case 2:
-		if (p[i + 1] == -1)
-			i++;
-		if (n < i + 4)
+		if (n < 3)
 			break;
-		input_csi_dispatch_sgr_rgb_do(ictx, p[0], p[i + 1], p[i + 2],
-		    p[i + 3]);
+		if (n == 5)
+			i = 2;
+		else
+			i = 3;
+		if (n < i + 3)
+			break;
+		input_csi_dispatch_sgr_rgb_do(ictx, p[0], p[i], p[i + 1],
+		    p[i + 2]);
 		break;
 	case 5:
-		if (n < i + 2)
+		if (n < 3)
 			break;
-		input_csi_dispatch_sgr_256_do(ictx, p[0], p[i + 1]);
+		input_csi_dispatch_sgr_256_do(ictx, p[0], p[2]);
 		break;
 	}
 }


townba:
I considered checking `n` and ignoring the color space parameter (which is essentially what `xterm` does) instead of demanding that it be left unspecified, but I thought I'd start with this approach. Let me know which you'd prefer.

townba:
(This is the sort of change I'm talking about. If there are enough parameters for the color-space form, then we'd assume it's specified [or empty] and skip it. If not, then we'd assume it's the no-color-space form.)

```diff
--- a/input.c
+++ b/input.c
@@ -1943,7 +1943,7 @@ input_csi_dispatch_sgr_colon(struct input_ctx *ictx, u_int i)
        i = 1;
        switch (p[i]) {
        case 2:
-               if (p[i + 1] == -1)
+               if (n >= 6)
                        i++;
                if (n < i + 4)
                        break
```

nicm:
Try this instead please:
~~~~
? 1742.patch
? tmux
? tmux-client-22370.log
? tmux-out-82227.log
? tmux-server-82227.log
Index: input.c
retrieving revision 1.153
diff -u -p -r1.153 input.c
--- input.c	13 May 2019 20:10:23 -0000	1.153
+++ input.c	17 May 2019 05:57:13 -0000
@@ -1940,21 +1940,21 @@ input_csi_dispatch_sgr_colon(struct inpu
 	}
 	if (p[0] != 38 && p[0] != 48)
 		return;
-	if (p[1] == -1)
-		i = 2;
-	else
-		i = 1;
-	switch (p[i]) {
+	switch (p[1]) {
 	case 2:
+		if (p[i + 1] == -1)
+			i = 2;
+		else
+			i = 1;
 		if (n < i + 4)
 			break;
 		input_csi_dispatch_sgr_rgb_do(ictx, p[0], p[i + 1], p[i + 2],
 		    p[i + 3]);
 		break;
 	case 5:
-		if (n < i + 2)
+		if (n < 3)
 			break;
-		input_csi_dispatch_sgr_256_do(ictx, p[0], p[i + 1]);
+		input_csi_dispatch_sgr_256_do(ictx, p[0], p[2]);
 		break;
 	}
 }

nicm:
Please try this instead (last one I sent was wrong):
~~~~
? 1742.patch
? tmux
? tmux-client-54918.log
? tmux-out-15078.log
? tmux-server-15078.log
Index: input.c
retrieving revision 1.153
diff -u -p -r1.153 input.c
--- input.c	13 May 2019 20:10:23 -0000	1.153
+++ input.c	17 May 2019 06:02:39 -0000
@@ -1938,23 +1938,25 @@ input_csi_dispatch_sgr_colon(struct inpu
 		}
 		return;
 	}
-	if (p[0] != 38 && p[0] != 48)
+	if (n < 2 || (p[0] != 38 && p[0] != 48))
 		return;
-	if (p[1] == -1)
-		i = 2;
-	else
-		i = 1;
-	switch (p[i]) {
+	switch (p[1]) {
 	case 2:
-		if (n < i + 4)
+		if (n < 3)
 			break;
-		input_csi_dispatch_sgr_rgb_do(ictx, p[0], p[i + 1], p[i + 2],
-		    p[i + 3]);
+		if (p[2] == -1)
+			i = 3;
+		else
+			i = 2;
+		if (n < i + 3)
+			break;
+		input_csi_dispatch_sgr_rgb_do(ictx, p[0], p[i], p[i + 1],
+		    p[i + 2]);
 		break;
 	case 5:
-		if (n < i + 2)
+		if (n < 3)
 			break;
-		input_csi_dispatch_sgr_256_do(ictx, p[0], p[i + 1]);
+		input_csi_dispatch_sgr_256_do(ictx, p[0], p[2]);
 		break;
 	}
 }

townba:
Incorporated your proposed change with a minor change to bias toward expecting (and ignoring) the parameter. PTAL.

nicm:

PR #1743, opened 2019-05-17
list-keys: make valid for re-input - escape keys and semicolons

avih:
Semicolons as command-separator were not escaped, and neither was the
key itself for special keys: semicolon, single/double quote, backslash,
hash-key '#', dollar '$', tilde '~'.

Now they are escaped and the output of `tmux list-keys` can be used as
input for `tmux source-file ...` to reproduce the bindings.

### DIFF
--- a/cmd-list-keys.c
+++ b/cmd-list-keys.c
@@ -32,6 +32,8 @@ static enum cmd_retval	cmd_list_keys_exec(struct cmd *, struct cmdq_item *);
 static enum cmd_retval	cmd_list_keys_commands(struct cmd *,
 			    struct cmdq_item *);
 
+static const char * escape_key(const char *);
+
 const struct cmd_entry cmd_list_keys_entry = {
 	.name = "list-keys",
 	.alias = "lsk",
@@ -53,6 +55,42 @@ const struct cmd_entry cmd_list_commands_entry = {
 	.flags = CMD_STARTSERVER|CMD_AFTERHOOK,
 	.exec = cmd_list_keys_exec
 };
+
+/* escape <key> such that it's suitable for re-input from list-keys */
+static const char *
+escape_key(const char* key)
+{
+	static char out[34];  /* key_string_lookup_key output (32) + 2 quotes */
+	char last;
+	int len;
+
+	len = strlen(key);
+	if (len == 0 || len > 31)
+		return key;
+	strcpy(out, key);
+	last = out[len - 1];
+
+	/* The following work both for <last> alone and for M-<last> */
+	/* Use single-quotes, except " for single-quote, \ for semicolon */
+	switch(last) {
+	case ';':
+		strcpy(out + len - 1, "\\;");
+		break;
+	case '\'':
+		strcpy(out + len - 1, "\"'\"");
+		break;
+	case '"':  /* FALLTHROUGH */
+	case '\\': /* FALLTHROUGH */
+	case '#':  /* FALLTHROUGH */
+	case '~':  /* FALLTHROUGH */
+	case '$':
+		strcpy(out + len - 1, "' '");
+		out[len] = last;
+		break;
+	}
+
+	return out;
+}
 
 static enum cmd_retval
 cmd_list_keys_exec(struct cmd *self, struct cmdq_item *item)
@@ -83,7 +121,7 @@ cmd_list_keys_exec(struct cmd *self, struct cmdq_item *item)
 		}
 		bd = key_bindings_first(table);
 		while (bd != NULL) {
-			key = key_string_lookup_key(bd->key);
+			key = escape_key(key_string_lookup_key(bd->key));
 
 			if (bd->flags & KEY_BINDING_REPEAT)
 				repeat = 1;
@@ -108,7 +146,7 @@ cmd_list_keys_exec(struct cmd *self, struct cmdq_item *item)
 		}
 		bd = key_bindings_first(table);
 		while (bd != NULL) {
-			key = key_string_lookup_key(bd->key);
+			key = escape_key(key_string_lookup_key(bd->key));
 
 			if (!repeat)
 				r = "";
--- a/cmd-list.c
+++ b/cmd-list.c
@@ -112,12 +112,12 @@ cmd_list_print(struct cmd_list *cmdlist)
 	TAILQ_FOREACH(cmd, &cmdlist->list, qentry) {
 		this = cmd_print(cmd);
 
-		len += strlen(this) + 3;
+		len += strlen(this) + 4;
 		buf = xrealloc(buf, len);
 
 		strlcat(buf, this, len);
 		if (TAILQ_NEXT(cmd, qentry) != NULL)
-			strlcat(buf, " ; ", len);
+			strlcat(buf, " \\; ", len);
 
 		free(this);
 	}
--- a/tmux.1
+++ b/tmux.1
@@ -2354,13 +2354,30 @@ to
 and
 .Em Tab .
 Note that to bind the
-.Ql \&"
+.Ql \&;
 or
 .Ql '
-keys, quotation marks are necessary, for example:
+or
+.Ql \&"
+or
+.Ql \&\e
+or
+.Ql #
+or
+.Ql $
+or
+.Ql ~
+keys, quoting is necessary: semicolon is quoted with backslash, single-quote
+is quoted with double-quotes, and the rest are quoted with single-quotes, for
+example:
 .Bd -literal -offset indent
-bind-key '"' split-window
-bind-key "'" new-window
+bind-key \\;    display-message semicolon-pressed
+bind-key M-\\;  display-message ALT-semicolon-pressed
+bind-key "'"   display-message single-quote-pressed
+bind-key '"'   display-message double-quote-pressed
+bind-key '#'   display-message hash-key-pressed
+bind-key '\\'   display-message backslash-pressed
+bind-key M-'\\' display-message ALT-backslash-pressed
 .Ed
 .Pp
 Commands related to key bindings are as follows:


avih:
I just noticed that space also needs single-quotes (tested working `M-' '` and `' '`), but this patch doesn't quote it.

You can either add it yourself since you never push the original commit anyway, or let me know if you want me to add it (together with any other comments you have).

For now I'm not updating the PR, and it's currently missing single-quotes for the space key.

avih:
Also, this can replace the `case` statement if you want:
```c
	if (last == ';') {
		strcpy(out + len - 1, "\\;");
	} else if (last == '\\') {
		strcpy(out + len - 1, "\"'\"");
	} else if (strchr("\"\\#$~ ", last)) {
		strcpy(out + len - 1, "' '");
		out[len] = last;
	}
```

avih:
Actually, space doesn't need quoting at list-keys as it's printed as `Space`. It can be a quoted literal at tmux.conf, but that's irrelevant to this patch.

avih:
Maybe it would be best to instead just add key names `hash`, `dollar`, `tilde`, `semicolon`, `squote`, `dquote`, `backslash` like there is already with `Space` and others. This would save the need to quote, and I assume the infrastructure for key names already exists anyway.

avih:
> Maybe it would be best to instead just add key name...

Yeah, I think this is a better solution instead of the patch at this PR (still needs tmux.1 update):
```diff
diff --git a/cmd-list.c b/cmd-list.c
index e999c370..fb6c5ac0 100644
--- a/cmd-list.c
+++ b/cmd-list.c
@@ -112,12 +112,12 @@ cmd_list_print(struct cmd_list *cmdlist)
 	TAILQ_FOREACH(cmd, &cmdlist->list, qentry) {
 		this = cmd_print(cmd);
 
-		len += strlen(this) + 3;
+		len += strlen(this) + 4;
 		buf = xrealloc(buf, len);
 
 		strlcat(buf, this, len);
 		if (TAILQ_NEXT(cmd, qentry) != NULL)
-			strlcat(buf, " ; ", len);
+			strlcat(buf, " \\; ", len);
 
 		free(this);
 	}
diff --git a/key-string.c b/key-string.c
index 8442727d..61821cda 100644
--- a/key-string.c
+++ b/key-string.c
@@ -85,6 +85,15 @@ static const struct {
 	{ "KP0",	KEYC_KP_ZERO },
 	{ "KP.",	KEYC_KP_PERIOD },
 
+	/* keys which otherwise require quoting */
+	{ "Tilde",	'~' },
+	{ "Hash",	'#' },
+	{ "Dollar",	'$' },
+	{ "SQuote",	'\'' },
+	{ "DQuote",	'"' },
+	{ "BSlash",	'\\' },
+	{ "SColon",	';' },
+
 	/* Mouse keys. */
 	KEYC_MOUSE_STRING(MOUSEDOWN1, MouseDown1),
 	KEYC_MOUSE_STRING(MOUSEDOWN2, MouseDown2),
```

nicm:
Thanks for this.

I don't really like the idea of adding key names when they already have visible symbols, I think escaping them is better.

Your original patch looks OK, but:

- Please use the same prefix as other functions in the file, so cmd_list_keys_escape_key. It might even be better in key-string.c.
- Don't use strcpy under any circumstances. Use strlcpy or xsnprintf.

You can see the code style here https://man.openbsd.org/style.9 but I will fix it up if its wrong anyway so don't worry about it if you don't want to.

avih:
Hmm... I actually dislike the escaping solution because it's artificial and a guess. It also adds more code and dataflow instead of just more data in an existing template (I think adding more data to be used by existing code is almost always better than adding more code).

And re adding names you could say the same about `Space`, because it can be single-quoted like the other chars I added, so technically it's not required either.

I think the names are also better because quoting is not great at `tmux.conf`, for instance the following don't work, even though common sense (for a shell user, but also for tmux.conf user) suggests they should: `\\` `\"` `\'` `';'` `'M-;'` etc. So If there are solutions to avoid quoting, like with the names which my second patch adds, then I think they would be better.

So I would actually prefer to withdraw this PR if you insist on artificial escaping.

nicm:
Space is different because it is invisible. "M- " is unhelpful and M-Space
is clearly better. M-; is visible and clearer than M-Colon. emacs calls it
M-;.

I don't see here guesswork comes into it. If we want lsk to show output
that can be inserted into .tmux.conf then that is the parser it needs to
target. We already have to escape any command arguments anyway.

Adding additional names seems like a workaround not a fix. I don't mind
accepting the names from the user but I think the output should show the
normal M-; form.

If there are problems with escaping in .tmux.conf then the config and
command parsers should be improved.

avih:
> I don't see here guesswork comes into it

The guesswork with the original patch of this PR is that it takes a technically-arbitrary key string, looks only at the last char, and escapes different chars differently according to what the parser happens to accept currently. There's no inherent correctness in it.

A non-artificial solution would be to escape where the key string is generated, and where it's known that the `M-` or `C-` parts don't need escaping, but the actual key might.

Better yet, if there was some generic procedure to quote arbitrary strings for input.conf which would work equally well for the key argument, the command argument, and any other place, etc, then clearly it should be used, but I don't think I'm familiar with such solution.

If you plan to improve the parsing or introduce such generic escaper/quoter procedure, then this PR should wait till after it happens.

Personally I haven't looked at how the string are stored and what a generic quoter for tmux.conf should do.

nicm:
Yes it would be better to have a common "escape this for the config file please" function. There is already escaping code for the q: format modifier, possibly it would do the job. Or the existing code to escape the arguments could be extended. If the config parser won't accept it's output, it should probably be improved. It is probably unrealistic to expect that anything you use in the shell works in the config or command prompt but it wouldn't be a bad goal to have.

nicm:
If you have a list of example commands that don't parse correctly in the config file but should then if would be helpful, or put them in a regression test (even if it fails now).

nicm:
It is format_quote but it probably quotes too much, unless the config file parser was changed a bit.

avih:
My examples would be the ones I provided (not 100% sure of the `\\` one, need to check again).

In general I'm able to use quoting with complex commands, but the key argument to `bind` triggered some cases I wasn't expecting, especially that backslash quotes a semicolon but doesn't quote other chars, and that single or double quote don't quote semicolon.

avih:
I think we can close this PR for now. The code doesn't go anywhere, and it seems a generic quoter would be a better solution to use, but it first needs to be written.

I'll try to look at quoting related code and see if I can reuse or come up with something which works generally well.

nicm:
Yeah ; is a bit weird IIRC cos it splits up commands and that is done
earlier.

So \\; is processed by the cmd-list parser which will not unescape anything
else. Then the actual command string parser only unescapes inside double
quotes.

The two stage parser is a leftover from before command sequences were
implemented. Ideally it should be merged into one but that is probably a
big change.

An easier fix would be to make the cmd-list.c parser ignore quotes strings
when searching for ;s and possibly make it unescape \\x outside strings
(the cmd.c parser should take care of those inside).

avih:
> The two stage parser is a leftover from before command sequences were
implemented. Ideally it should be merged into one but that is probably a
big change.

Yes, parsing and quoting is complex, ask any shell noob (or veteran).

I'll try to read the manual again, but I don't recall seeing info which explains in details how tmux.conf is being parsed (or alternatively, how should quoting be used in it). Your info about about the two stages and their different quote semantics was new to me, and it does explain things.

nicm:
Yes it is probably not in the manual. Few people have really asked about it
that much IIRC so it probably isn't something that comes up a lot in
practice, but it is irritating figuring out how to escape when you do hit
it.

I don't think it'll be that hard to make cmd_list_parse ignore quoted
strings and to unescape anything else it sees outside and that would be
more consistent at least. I'll take a look at some point. Although will
also need to check that if doesn't break anyone who is already using \;
inside strings.

avih:
I'll need to think about it as well. Maybe introduce some easier form of quoting/grouping (for the same reasons that `$(..)` replaced backticks in shell), or maybe or alter the specification in some way to make it easier to use.

Not off hand though...

avih:
This is a WIP braindump about parsing tmux.conf and splitting it into commands. The general idea I'm currently exploring is replacing some cases of quotes with easier-to-use grouping.

I think the quotes are currently used to achieve two objectives:
- Combine a list of WORDs into a single argument, like the argument of `display-message`.
- When such single argument is a [list of] command - like the command[s] arguments of `if`, then it's a grouping of commands.

I'm currently thinking that the second case could support an additional grouping syntax which is easier to use. Specifically, replace double/single quotes with a matching pair of unquoted `{` and `}`, where newlines are allowed in between without breaking the grouping effect.

This would allow replacing this (from my actual tmux.conf):
```
bind-key -n WheelUpPane \
    if-shell -Ft= "#{mouse_button_flag}" \
        "send-keys -N3 -Mt=" \
        "if-shell -Ft= '#{alternate_on}' \
            'send-keys -N3 -t= Up' \
            'copy-mode -et=; send-keys -t= WheelUpPane'"
```

(note to self: following the above info that `;` is parsed at a different stage, it's unclear to me how this works at all, considering that that the `;` of `copy-mode -et=;` is not not preceded with `\`...).

With this:
```
bind-key -n WheelUpPane {
    if-shell -Ft= "#{mouse_button_flag}" {
        send-keys -N3 -Mt=
        if-shell -Ft= "#{alternate_on}" {
            send-keys -N3 -t= Up
        } {
            copy-mode -et=
            send-keys -t= WheelUpPane
        }
    }
}
```

Or this:
```
bind-key -n WheelUpPane \
    if-shell -Ft= "#{mouse_button_flag}" {
        send-keys -N3 -Mt=
        if-shell -Ft= "#{alternate_on}" {
            send-keys -N3 -t= Up
        } {
            copy-mode -et=
            send-keys -t= WheelUpPane
        }
    }
```

Maybe a way to look at it is that unquoted `{` introduces a form of quoting which can detect matching `}` through possible nesting, and where it's mostly useful when the grouping is of commands.

As I said though, these are yet-incoherent WIP thoughts.

nicm:
I love this idea.

The reason that 'copy-mode -et=; send-keys -t= WheelUpPane' works is because the cmd_string_parse step comes before cmd_list_parse, so cmd_list_parse gets an argv like this:
~~~~
if-shell: argv[0]=if-shell
if-shell: argv[1]=-Ft=
if-shell: argv[2]=#{mouse_button_flag}
if-shell: argv[3]=send-keys -N3 -Mt=
if-shell: argv[4]=if-shell -Ft= '#{alternate_on}' 'send-keys -N3 -t= Up' 'copy-mode -et=; send-keys -t= WheelUpPane'
~~~~
It only cares about ;s which are either on their own (so argv[n]=;) or which terminate a command (argv[n]=command;) - it splits the argv into two at that point. So the ; in argv[4] is doesn't need to be escaped.

if-shell will call cmd_string_parse on its argv[4] (in cmd_if_shell_callback).

So I can explain a bit how it works at the moment:

I was bit inaccurate about cmd_list_parse in my previous comment - it isn't just a leftover. The reason is that we can be given an argv that has been already split by the shell and have to separate it into commands.

The final result of parsing is an argv which is passed to args_parse which calls getopt on it. So everything needs to end up as an argv. But there are several ways to get there:

- If you run "tmux xyz" from the shell, the argv from that client is packed up and sent across the socket and used directly. Splitting of the "xyz" into an argv is left entirely to the shell, but if you do something like "tmux lsk \; killp" then tmux will be given:
~~~~
argv[0]=lsk
argv[1]=;
argv[2]=killp
~~~~
And that is where cmd_list_parse comes in - it splits this into a command sequence of two separate commands.

- If the input is a string then it goes into cmd_string_parse. This is basically everything else. It is called for each line given to tmux in the config file (load_cfg); typing a command at the command prompt; or as string parts of other commands (if-shell "xyz" "abc" will call it on one of xyz or abc).

  cmd_string_parse splits the string into an argv, and then passes it to cmd_list_parse to split into commands. This is the two step process I mentioned earlier.

I don't mind if the shell's argv/argc and a string from tmux follow entirely different codepaths, and that might be best.

So what I would suggest is to merge load_cfg, cmd_string_parse and cmd_list_parse together to produce a single parser that can deal with the config file and possibly also strings (probably they can be treated as a config file of only one line). This could do the {} bit as well. Possibly it could be done with yacc.

Then the argv from the shell can either use a different that just splits it into commands (like cmd_list_parse now), or it could mush the shell's argv together into a string and use the same parser but with a flag that says "expansion and escaping is already done". I would be fine with this being done separately rather than written at the same time.

avih:
Thanks. I need to re-read your last post few times, accompanied by looking at the code. I'll do that later. Sounds promising!

nicm:
I have improved the list-keys output, please open a new PR if you want for your other changes,

avih:
Yeah, seems that now `list-keys` output is suitable for re-input to `source-file`, at least with my bindings. Thanks.

PR #1748, opened 2019-05-20
Add missing quotes in manual

alfunx:
Fixes two small typos, one of which made the second `Ta` literal and broke the formatting.


nicm:

PR #1758, opened 2019-05-26
parsing: support {...} brace strings with nesting

avih:
An unquoted brace pair now quotes verbatim anything which looks like
a tmux "program", while supporting nesting without escaping or special
quoting, as well as allowing multiple lines.

Consequently, { and } now need quoting. Arguments quoting procedure
and relevant built-in key bindings were updated accordingly.


avih:
This implements the second form of the suggestion at https://github.com/tmux/tmux/pull/1743#issuecomment-493657164 (strings only, not full commands) .

avih:
The CI failures should have been on master too, as this PR doesn't use any new functions which tmux master doesn't use at the same file.

Regardless, I also still didn't look at config file syntax tests, if there are any. Some might require modifications if they use unquoted braces.

avih:
Rebased, updated to use the new `yylex_getc1`, and addressed style comments from IRC.

All changes are at the loop of `yylex_token_brace`.

avih:
Also, possible different names for this feature could be brace quotes or block quote. I think maybe block quote is better than other names, because it both states that it's quoting, and also that it contains a (code) block, which is a familiar term from other languages.

nicm:
Applied https://github.com/tmux/tmux/commit/6b332127cae97914d34c39575881fbc87205f4e0. Thanks!

avih:
Thanks!

avih:
For future reference, here's my original docs for brace strings:
```
BRACE STRINGS
     Braces can enclose any group of tmux commands as a string, without the
     usual concerns of escaping and nesting. This is useful with arguments
     which are parsed and executed as commands - like the arguments of
     if-shell.

     For instance, the following commands are identical:

           if-shell true {
               display -p 'brace-dollar-foo: }$foo'
           }

           if-shell true "\n    display -p 'brace-dollar-foo: }\$foo'\n"

     Like single quoted strings, brace strings don't expand during parsing.
     The only differences from single quotes are the fact that brace strings
     may span multiple lines, and the way in which the parser determines where
     the string ends.

     After the opening brace, the parser examines the rest as if it was a tmux
     config file, and looks for braces. Unquoted opening braces increase the
     nesting depth, and unquoted closing braces decrease it. The string ends
     where the nesting depth becomes 0, that is, at the closing brace which
     matches the initial opening brace.

     The string's value is verbatim copy of everything between the initial and
     final braces - including quotes, braces, newlines, comments, etc.

     This means that quotes and brace strings can be used normally inside a
     brace string at any nesting depth. In general, the entire content of any
     valid tmux config file could be placed unmodified between braces to form
     a valid brace string, which would behave as expected if parsed and exe‐
     cuted.
```

PR #1761, opened 2019-05-27
tmux.c: add support for XDG_RUNTIME_DIR

flokli:
When `TMUX_TMPDIR` is not set, try to use `XDG_RUNTIME_DIR` before falling
back to `_PATH_TMP` (`/tmp`).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,8 @@
 CHANGES FROM 2.9 to X.X
 
+* Add support for XDG_RUNTIME_DIR. When TMUX_TMPDIR is not set,
+  try to use XDG_RUNTIME_DIR before falling back to _PATH_TMP (/tmp).
+
 * Add an addition form of string syntax in the configuration file - {}. This
   means commands that take other commands as string arguments can be expressed
   more clearly and without additional escaping. This means that a literal { and
--- a/tmux.c
+++ b/tmux.c
@@ -119,6 +119,8 @@ make_label(const char *label, char **cause)
 
 	if ((s = getenv("TMUX_TMPDIR")) != NULL && *s != '\0')
 		xasprintf(&base, "%s/tmux-%ld", s, (long)uid);
+	else if((s = getenv("XDG_RUNTIME_DIR")) != NULL && *s != '\0')
+		xasprintf(&base, "%s/tmux-%ld", s, (long)uid);
 	else
 		xasprintf(&base, "%s/tmux-%ld", _PATH_TMP, (long)uid);
 	if (realpath(base, resolved) == NULL &&


nicm:
Thanks but I do not want any additional environment variables meaning the same thing.

flokli:
Well, it's not just an additional environment variable, but became a somewhat [standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) location for user-specific runtime files, with much nicer guarantees to confidentiality and cleanup as `/tmp` too.

Yes, there is `TMUX_TMPDIR` as a tmux-specific environment variable , but asking distros to set this might be a bit lame if there's an XDG standard.

nicm:
Yes I know what it is but it isn't a standard which OpenBSD programs follow and I don't want to go down the rabbit hole of trying to support everyone's favourite file locations. We provide one environment variable and that is enough. 

If you are packaging tmux I don't think you should be changing file locations with environment variables. I don't think I would object to a flag to configure for /tmp like SYSCONFDIR (it could default to _PATH_TMP). How do Debian do it, IIRC they change to /var/run? (Or maybe they stopped that? Wasn't there a security issue?)

flokli:
On debian, `tmux` seems to (at least currently) write to `/tmp`, instead of `/run` or `/var/run`.

I fear being able to set `_PATH_TMP` during configure phase wouldn't be enough alone, as it's `/run/user/$(id -u)`, and the uid part would need to be calculated during runtime.

Do you have any suggestions in how to do this in a clean fashion?

nicm:
Sorry forgot to reply.

If you are packaging I'd recommend just patching tmux to use your system default /tmp instead of _PATH_TMP. If users don't like it they can always use -S or TMUX_TMPDIR to move it back to /tmp.

flokli:
@nicm thanks for getting back to me.

The system-wide default for `/tmp` still is `/tmp`, but per-user processes are encouraged to honor `XDG_RUNTIME_DIR` on Linux (which gets set to `/run/user/$(id -u)` through PAM).

While simply hardcoding the default would be just implementing half of the spec, it's also not possible, as that 'default' (`/run/user/$(id -u)`) still contains runtime information (the user id)

I'm not sure if shipping above patch downstream without much chances of this being ever accepted upstream is a good idea either - will have to think about it.

nicm:
I would just do something like this myself but it's up to you:
~~~~
diff --git a/tmux.c b/tmux.c
index 685827d4..0f48c31e 100644
--- a/tmux.c
+++ b/tmux.c
@@ -110,6 +110,7 @@ make_label(const char *label, char **cause)
        char            *base, resolved[PATH_MAX], *path, *s;
        struct stat      sb;
        uid_t            uid;
+       struct passwd   *pw;

        *cause = NULL;

@@ -119,6 +120,8 @@ make_label(const char *label, char **cause)

        if ((s = getenv("TMUX_TMPDIR")) != NULL && *s != '\0')
                xasprintf(&base, "%s/tmux-%ld", s, (long)uid);
+       else if ((pw = getpwuid(getuid())) != NULL)
+               xasprintf(&base, "/run/user/%s/tmux", pw->pw_name);
        else
                xasprintf(&base, "%s/tmux-%ld", _PATH_TMP, (long)uid);
        if (realpath(base, resolved) == NULL &&

PR #1768, opened 2019-05-31
Fix warnings

mathstuf:
---
Feel free to cherry-pick the ones you want.

For the `dsx` and `dsy` commit (c4cc767), it may be better to check `args_has(args, 'x') && args_has(args, 'y')` instead.

### DIFF
--- a/cmd-queue.c
+++ b/cmd-queue.c
@@ -206,6 +206,7 @@ cmdq_get_command(struct cmd_list *cmdlist, struct cmd_find_state *current,
 	u_int			 group = 0;
 
 	TAILQ_FOREACH(cmd, &cmdlist->list, qentry) {
+		shared = NULL;
 		if (cmd->group != group) {
 			shared = xcalloc(1, sizeof *shared);
 			if (current != NULL)
@@ -230,7 +231,8 @@ cmdq_get_command(struct cmd_list *cmdlist, struct cmd_find_state *current,
 
 		log_debug("%s: %s group %u", __func__, item->name, item->group);
 
-		shared->references++;
+		if (shared)
+			shared->references++;
 		cmdlist->references++;
 
 		if (first == NULL)
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -201,7 +201,8 @@ cmd_new_session_exec(struct cmd *self, struct cmdq_item *item)
 				goto fail;
 			}
 		}
-	}
+	} else
+		dsx = 80;
 	if (args_has(args, 'y')) {
 		tmp = args_get(args, 'y');
 		if (strcmp(tmp, "-") == 0) {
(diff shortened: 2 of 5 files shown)


nicm:
Thanks.

What function do we need stdlib.h for in cmd-parse.y?

cmd-queue.c change is wrong because it will set shared to NULL every
loop and the point is to keep it unless the command is in a new group,
but setting it to NULL in the declaration should fix it.

options_array_set and cmd-new-session.c are not problems and I don't see
a warning on any of the compiler versions I use so I'll leave them for
the moment.

mathstuf:
On May 31, 2019 5:39:18 PM EDT, Nicholas Marriott <notifications@github.com> wrote:
>What function do we need stdlib.h for in cmd-parse.y?

free. It's in the commit message :) .

>cmd-queue.c change is wrong because it will set shared to NULL every
>loop and the point is to keep it unless the command is in a new group,
>but setting it to NULL in the declaration should fix it.

Ok, I wasn't sure which was correct. Can groups be reused once closed (i.e., 1, 2, 1)?

>options_array_set and cmd-new-session.c are not problems and I don't
>see
>a warning on any of the compiler versions I use so I'll leave them for
>the moment.

GCC 8.3.something on Fedora 29. They're the compiler being smarter, but not sufficiently smart enough. See each commit message for more details.

nicm:
Oh right I just downloaded the diff I didn't look at the commit message.

I'll add stdlib.h thanks, should have thought of free.

nicm:
Groups are never reused unless it wraps which it won't in one command sequence, it's just an integer that is always increased.

I'll leave the warnings for the GCC bugs for the moment, I don't really use GCC much anymore and they might fix it before it annoys me :-).

nicm:
Thanks!

mathstuf:
On June 1, 2019 2:16:51 AM EDT, Nicholas Marriott <notifications@github.com> wrote:
>I'll leave the warnings for the GCC bugs for the moment, I don't really
>use GCC much anymore and they might fix it before it annoys me :-).

The first one still makes sense (though the alternate fix mentioned in the description may be preferable). Passing just one of -x or -y to new-session uses the other uninitialized.

nicm:
Oh you are right, I think if you don't give -x but do give -y it should
use sx for -x, I'll change that.

cmd-queue.c
```diff
@@ -230,7 +231,8 @@ cmdq_get_command(struct cmd_list *cmdlist, struct cmd_find_state *current,
 
 		log_debug("%s: %s group %u", __func__, item->name, item->group);
 
-		shared->references++;
+		if (shared)
```
mathstuf: Is this conditional still necessary? Or should the `if` at the start of this loop have `!shared ||` added to its condition?

nicm:
No, a cmd group can't be 0, they start at 1.

mathstuf:
Not sure what `cmd->group` has to do with that? Or are you saying that `shared` will always be initialized on the first iteration because the command group will never match 0?

nicm:
Yes that's right, cmd->group will always be nonzero so shared will
always be created. The shared pointer can never be NULL (and if it
is that is a bug we don't want to hide).

mathstuf:
I suppose adding in an `assert(shared && "zero group assignment for command")` would help the compiler figure this out if the warnings pop up in the future.

Issue #1492, opened 2018-09-30
Undercurl (special underline)


blueyed:
What is required for undercurl support in terminals that support it, e.g. kitty?

`printf '\e[4:3mUnderlined\n'` works outside of tmux, but not inside.
I've tried wrapping it in the Ptmux wrapping sequence also without success (`printf '\ePtmux;\e\e[4:3m\e\\Underlined\n' `).

@bfredl left some insights in https://github.com/neovim/neovim/pull/9052#issuecomment-425050320.

nicm:
You will need to add it as an attribute, it can be an extended one like strikethrough (look for GRID_ATTR_STRIKETHROUGH) because it will be rarely used. But you will need to get terminfo capabilities added for it. Underline is smul/rmul, I would suggest initially using something like smuc/rmuc, and see how that flies when it comes to add it to the terminfo entry.

blueyed:
Thanks for the info.
I've looked briefly into it, but it is not trivial (for me).

For reference: https://sw.kovidgoyal.net/kitty/protocol-extensions.html#colored-and-styled-underlines

nicm:
I don't see much need for more types of underline but the extension seems sensible enough, at least by the standards of kitty, so it would not be hard to add at some point.

nicm:
You can try this: [tmux-uc.diff.txt](https://github.com/tmux/tmux/files/2438280/tmux-uc.diff.txt)

You will need to set smuc to the right sequence to turn on the new underlines, so something like:
~~~~
set -as terminal-overrides ',*:smuc=\E[4:3m'
~~~~
This is incomplete however - it is missing the bits needed in grid_string_cells_code (marked XXX). Also, it only supports one type of underline, but the extension adds four.

If we were to add the others then I don't see any reason not to use separate bits in GRID_ATTR_*, but I don't know whether it would be better to add additional terminfo(5) capabilities for each, or to have one capability that takes an argument.

It may be best to wait for a while and see what applications actually use, there is no point in adding support for four different types of underline if they are hardly ever used.

blueyed:
Thanks for looking into this!

1. Is `printf '\e[4:3mUnderlined\n'` supposed to work inside of tmux when using kitty then?
`printf '\e[4:1mUnderlined\n'` works (underlined), but `printf '\e[4:3mUnderlined\n'` prints "nderline" (missing U, not undercurled).
Does kitty's terminfo need a "smuc" entry for this?

[...]

PR #1771, opened 2019-06-01
Underscore color

kmoschcau:
This is a draft for adding colored underscores to tmux.
Right now this does not compile, because I am missing some of the knowledge required to finish this properly.
Mostly I have problems in tty.c.

The discussion that started this can be found in #1492.

Help and advice is welcome.


nicm:
Not sure what your actual questions are but some comments below.

Can you get rid of all the extra // comments that aren't related to your
changes, they make the diff hard to read? Any comments you want to keep
please make them /* */ and complete sentences.

+                               else if (tc->us != 0) // TODO: kmo I have no idea what to use here.
+                                       tty_putcode1(tty, TTYC_SETULC, 0);

AFAIK there is no way to specify the default colour for UL so you will
have to use SGR0 to reset it.

+       // Is this an aixterm colour?
+       if (gc->us >= 90 && gc->us <= 97 && colours < 16)

I think there is no way to specify aixterm colours either - you only
need to worry about 256 and RGB colours.

tty.c
```diff
@@ -2385,6 +2404,54 @@ tty_check_bg(struct tty *tty, struct window_pane *wp, struct grid_cell *gc)
...
+	/* Perform substitution if this pane has a palette. */
+	if (~gc->flags & GRID_FLAG_NOPALETTE) {
+		if ((c = window_pane_get_palette(wp, gc->us)) != -1)
+			gc->us = c;
+	}
+
+	/* Check if this is a 24-bit colour. */
+	if (gc->us & COLOUR_FLAG_RGB) {
+		/* Not a 24-bit terminal? Translate to 256-colour palette. */
+		if (!tty_term_has(tty->term, TTYC_SETRGBU)) { // TODO: kmo
```
kmoschcau: If I understood this correctly, `TTYC_SETRGBU` (which I simply made up) would be another termcap, that specifies whether the target terminal can set the underline color in 24-bit mode. What termcap should I use here?

tty.c
```diff
@@ -2385,6 +2404,54 @@ tty_check_bg(struct tty *tty, struct window_pane *wp, struct grid_cell *gc)
...
+			if (gc->us & 8) {
+				gc->us &= 7;
+				if (colours >= 16)
+					gc->us += 90;
+			}
+		}
+		return;
+	}
+
+	/* Is this an aixterm colour? */
```
kmoschcau: Since we can not specify underline color in aixterm, should this block simply be removed?

tty.c
```diff
@@ -2449,6 +2516,38 @@ tty_colours_bg(struct tty *tty, const struct grid_cell *gc)
...
+
+	/* Check if this is a 24-bit or 256-colour colour. */
+	if (gc->us & COLOUR_FLAG_RGB || gc->us & COLOUR_FLAG_256) {
+		if (tty_try_colour(tty, gc->us, "58") == 0)
+			goto save_us;
+		/* Should not get here, already converted in tty_check_us */
+		return;
+	}
+
+	/* Is this an aixterm bright colour? */
```
kmoschcau: Since we can not specify underline color in aixterm, should this block simply be removed?

tty.c
```diff
@@ -2258,6 +2266,13 @@ tty_colours(struct tty *tty, const struct grid_cell *gc)
 					tty_putcode1(tty, TTYC_SETAB, 0);
 				tc->bg = gc->bg;
 			}
+			if (COLOUR_DEFAULT(gc->us) && !COLOUR_DEFAULT(tc->us)) {
+				if (have_ax)
+					tty_puts(tty, "\033[59m");
+				else if (tc->us != 0) // TODO: kmo I have no idea what to use here.
```
kmoschcau: I am a slight bit lost here. My rough understanding is, that because we do not have a way to specify a default underline color, we have to reset everything and then re-apply everything apart from the underline color.

But I do not think I understand the code here well enough to know how to achieve that.

tty.c
nicm: I would only have Smulc which sets RGB colour and map the 256 colour entry to the nearest RGB colour. I don't think we need two terminfo capabilities for this.

tty.c
nicm: Yes.

tty.c
nicm: Yes you will need to call tty_reset because there is no way to clear the underline colour. That already happens in tty_colours:
~~~~
if (!have_ax && tty_term_has(tty->term, TTYC_OP))
                        tty_reset(tty);

nicm:
Please comment or open a new pull request when you continue working on this. Thanks!

nicm:
I have made some changes to this but I can't get the escape sequence to work with VTE so I can't test, does VTE actually support underscore colour? What terminals apart from Kitty support this?

Here is an updated diff: [x.diff.txt](https://github.com/tmux/tmux/files/3277314/x.diff.txt)

I made these changes:
- The underline colour can only be RGB or 256, it can't be an ANSI colour and it can't be the default colour. So we can use 0 as the value for "not set/default" since one of COLOUR_FLAG_RGB or _256 must be set. This applies to the 59 case in input.c and various checks (eg in grid.c).
- The grid_default_cell definition needs the new member in grid.c.
- Use SGR0 to reset the underline colour to default (add to the existing check in tty_attributes). I think this is better than trying to use the 59 escape sequence.
- tty_check_us only needs to do the palette stuff and convert 256->RGB, since we don't need ANSI or aixterm or default colours and we are going to make the capability RGB only.
- Use the same format for the capability as setaf in xterm-direct rather than separate arguments in tty_colours_us. I'm not a big fan of the format but it is better to use it for this I think.

The following still needs to be addressed:
- We need a way to convert a 256 colour palette entry to RGB for tty_check_us to use. This can be a lookup table added to colours.c like colours_256to16.
- Did you test if the colour actually goes through the palette? If not then that code can be removed.

The capability will look something like this (not tested):
~~~~
set -as terminal-overrides ',*:Setulc=\E[38\:2\:\:%p1%{65536}%/%d\:%p1%{256}%/%{255}%&%d\:%p1%{255}%&%d%;m'
~~~~

nicm:
Hmm VTE should support this since 0.51.2 but I have 0.54.2 and it doesn't work.

nicm:
Ahah, it does work.

nicm:
OK here is a diff that actually works: [y.diff.txt](https://github.com/tmux/tmux/files/3277408/y.diff.txt)

And here is the right Setulc:
~~~~
set -as terminal-overrides ',*:Setulc=\E[58::2::%p1%{65536}%/%d::%p1%{256}%/%{255}%&%d::%p1%{255}%&%d%;m'
~~~~
And this is how to test:
~~~~
printf '\033[4:3m\033[58;5;100mabc\033[0m\n'
~~~~

nicm:
VTE doesn't support palette setting so I can't tell if the 256 colour variant of this should use the palette or not. I assume it does, it would seem sensible.

nicm:
Here is a script to test underline with the palette:
~~~~
printf '\033c'
printf '\033[4:2m\033[58;5;100m\033[38;5;100mabcdef\033[0m' # yellow
sleep 1
printf '\033]4;100;rgb:ff/0/0\007' # now red
echo

nicm:
Here is a version with the 256-to-RGB table added and some other fixes: [z.diff.txt](https://github.com/tmux/tmux/files/3277526/z.diff.txt)

This seems to work OK for me. There are a couple of things with how this works that I'm not delighted about, I'll need to think about them a bit more.

kmoschcau:
I know that gnome terminator supports colored underlines.
https://gnometerminator.blogspot.com/p/introduction.html

nicm:
Is anyone testing this?

kmoschcau:
I did not have the time to do so, no. That's also why my work on it stalled.

nicm:
OK no worries, I will leave it for the moment. Thanks!

kmoschcau:
I just tested your diff and it works for me in gnome-terminator.
The described yellow in your test script looks more green to me, but it successfully changes to red double underscore after a moment.

kmoschcau:
How should I proceed from this? Should I fix my merge request with your diff or do you put it in yourself?

nicm:
No it's OK I'll apply the diff when I get time. Thanks

nicm:
I have applied this now, let me know if you see any problems.

nicm:
Thanks!

PR #1773, opened 2019-06-03
Add "attach-session -x" and "new-session -AX"

cjwatson:
These are roughly the equivalent of "screen -D -r" and "screen -D -R"
respectively (except that new-session also needs to be given a session
name).  The primitive for this was already available via detach-client,
but it wasn't in the most convenient form if you wanted to attach,
detach all other clients, and send SIGHUP to the parents of all other
clients in a single operation.

The best choice of option letter was a little unclear.  detach-client
uses -P, but that felt odd in this context where it's clearly a minor
variation of -d/-D (and it doesn't make sense to require users to type
both).  Both -D and -P were already taken in new-session.  In the end I
felt that making attach-session and new-session be similar was more
important than making either be exactly like detach-client, so I went
for -x/-X by analogy with the keys available in client mode.

### DIFF
--- a/cmd-attach-session.c
+++ b/cmd-attach-session.c
@@ -37,8 +37,8 @@ const struct cmd_entry cmd_attach_session_entry = {
 	.name = "attach-session",
 	.alias = "attach",
 
-	.args = { "c:dErt:", 0, 0 },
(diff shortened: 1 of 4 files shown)


nicm:
Great stuff, applied, thanks!

Issue #1779, opened 2019-06-05
Prevent resize triggering activity


xim:
Hi,

I've noticed that each time I connect to my detached tmux session from a different computer, all the windows are marked with activity. Presumably this happens due to a resize of tmux, which again resizes the windows. The programs/shells redraw to fit the size, and all are marked as having activity.

I'd like a way to get around that, so I can see which windows *actually* have activity. Possible ways of doing this might include:

* Hooks before/after session resize that allow me to disable monitoring before resize, and reactivate it afterwards
* Lazy resizing of windows, so the window only gets resized once I move to it.

## To reproduce

* `tmux -f/dev/null`
* `tmux set -g monitor-activity on`
* `tmux set -g default-command 'exec /bin/bash --norc'` (prevent my exotic .bashrc from doing crazy stuff)
* Open a few extra windows inside screen
* Resize your terminal window

![image](https://user-images.githubusercontent.com/169057/58940850-e2a90b80-877a-11e9-9c5d-24662b91072b.png)

nicm:
No there is no way to avoid this, tmux can't know if the program is updating because of a resize or not. It's best not to use monitor-activity with programs that redraw on resize.

xim:
Although technically correct (the best kind of correct) there are at least two possible workarounds, as mentioned in the initial report:

>     * Hooks before/after session resize that allow me to disable monitoring before resize, and reactivate it afterwards
>     * Lazy resizing of windows, so the window only gets resized once I move to it.

nicm:
Either would be fine with me.

nicm:
I initially thought hooks would be easier but actually I think lazy resize might be.

You may not need to defer the entire resize, just the TIOCSWINSZ. So it is could be enough just to set a flag on windows that aren't the current window in any attached session and do the TIOCSWINSZ when it becomes current. That might not play nicely if the application does any output in the meantime, but it seems like that is a problem we already have since we already defer resizing if it happens quickly.

For hooks you will need to make sure the hooks fire before the resize happens. It would probably mean moving resize to happen from a callback on the global command queue.

xim:
I might look into this at some point if you don't have the time / would prioritize it. Don't have time the coming weeks to dig into tmux internals, but I'll give a heads up here if I start working on it =)

xim:
[...]

PR #1782, opened 2019-06-05
Make window resize evaluation lazy.

xim:
Ref tmux/tmux #1779

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -1255,8 +1255,13 @@ server_client_loop(void)
 	 */
 	focus = options_get_number(global_options, "focus-events");
 	RB_FOREACH(w, windows, &windows) {
+		struct winlink	*wl;
+		TAILQ_FOREACH(wl, &w->winlinks, wentry) {
+			if (wl->session->attached != 0 && wl->session->curw == wl)
+				break;
+		}
 		TAILQ_FOREACH(wp, &w->panes, entry) {
-			if (wp->fd != -1) {
+			if (wl != NULL && wp->fd != -1) {
 				if (focus)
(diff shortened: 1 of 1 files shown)


nicm:
I think this is OK, there is no harm with it sitting with PANE_RESIZE set. This is slightly simpler and also checks if the session is attached:
~~~~
--- server-client.c     25 May 2019 07:18:20 -0000      1.285
+++ server-client.c     6 Jun 2019 08:41:51 -0000
@@ -1339,9 +1339,18 @@ static void
 server_client_check_resize(struct window_pane *wp)
 {
        struct timeval   tv = { .tv_usec = 250000 };
+       struct winlink  *wl;

        if (!(wp->flags & PANE_RESIZE))
                return;
+
+       TAILQ_FOREACH(wl, &wp->window->winlinks, wentry) {
+               if (wl->session->attached != 0 && wl->session->curw == wl)
+                       break;
+       }
+       if (wl == NULL)
+               return;
+
        log_debug("%s: %%%u resize to %u,%u", __func__, wp->id, wp->sx, wp->sy);

        if (!event_initialized(&wp->resize_timer))

xim:
I placed the logic outside the `server_client_check_resize` function so it isn't evaluated per pane. 
c9871a9  is implemented exactly as you proposed, 
4cad13a should be better (unless I misunderstood what happens with that `wp->flags &= ~PANE_REDRAW;` in `server_client_loop`)

nicm:
I like this, will test it for a while.

nicm:
I have applied this now. Thanks!

Issue #1775, opened 2019-06-04
Is it possible to specify custom characters for pane border?


aste4:
The width of default border (basically just a vertical line) is sometimes hard to see. Something similar has been asked in the past https://unix.stackexchange.com/questions/384422/adjust-border-thickness-in-tmux

If I make `pane-active-border-fg`'s color equal to `pane-active-border-bg`'s color, then the border is too thick and uneven - the horizontal border is too thick because character height is greater than the width.

Is it possible to set a custom character for the border? Id imagine there's need to be quite a few - 4 where the T is rotated 90 degrees, +, -, and | 

I would consider changing the | to something ‖ and the corresponding ones accordingly.

nicm:
No there is no practical way to do this without changing the code.

aste4:
Can you elaborate why? 

Where would I make this change in code?

I've been looking at screen-redraw.c (`#define CELL_BORDERS " xqlkmjwvtun~"`) but I don't know how those translate to the borders. I've also found this in my research: https://stackoverflow.com/questions/23368236/tmux-in-putty-displays-border-as-qqqqq-or-xxxx

nicm:
CELL_BORDERS is the ACS characters which are mapped to UTF-8 in tty-acs.c. You can change the UTF-8 characters in the table there, but that will change them for all applications inside tmux as well.

It would also be possible to change them with terminal-overrides but I doubt it will do you much good because you can only specify a single byte character for each ACS character so you wouldn't be able to use UTF-8. And it would also affect every application inside tmux as well.

If you want to change the border characters just for tmux and to use UTF-8 you will need to change screen-redraw.c so that where it writes CELL_BORDERS it instead writes a UTF-8 character with tty_puts, something like:
~~~~
switch (type) {
case CELL_INSIDE:
    break;
case CELL_LEFTRIGHT:
    tty_puts("\342\230\203");
    break;
case CELL_TOPBOTTOM:
    ...

krader1961:
The `CELL_BORDERS` value corresponds to the VT100 alternate character set for specifying line drawing characters. See this [Wikipedia article](https://en.wikipedia.org/wiki/Box-drawing_character#Unix,_CP/M,_BBS) which is just one of a huge number of web pages that talks about this topic.

IMHO, If you find that line drawing characters like vertical-bar are hard to read in your terminal you need to either a) use a different font, or b) use a better terminal.

aste4:
Thank you all for the incredibly fast responses. 

[...]

PR #1785, opened 2019-06-07
added an option to draw an alternate border.

aste4:
The border is slightly thicker than the default one.
`═════╬══════`

see https://github.com/tmux/tmux/issues/1775

This is a draft. I am likely missing other details such as documentation. Please let me know what extra steps I need to take.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -242,6 +242,12 @@ const struct options_table_entry options_table[] = {
 	  .default_num = ALERT_OTHER
 	},
 
+	{ .name = "alternate-border",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_SESSION,
+	  .default_num = 0
+	},
+
 	{ .name = "assume-paste-time",
 	  .type = OPTIONS_TABLE_NUMBER,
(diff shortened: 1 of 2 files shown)


aste4:
As far as I can tell, adding support for drawing the 4 corners (i.e ╗) is not necessary

nicm:
Thanks, but this seems like either too much or too little.

If we are to do this, it should be fully configurable so you can specify whatever characters your want for the borders, not just a choice of two.

But I don't think that this necessary at all. Almost all applications out there use standard ACS single line drawing and I don't think tmux needs a way to be different.

aste4:
If you'll consider putting this feature in, I'll work on making it fully configurable. Not a problem. 

option names could be:

`pane-border-symbol-north-south-east '\xe2\x95\x9a"` (╠)
`pane-border-symbol-south-west-east "byte_string_2"` (╦)
etc.

Let me know!

nicm:
I don't really want this, sorry. I suggest you maintain it privately for the moment and if other people start clamouring for it we can revisit. Thanks.

nicm:
I have added support for this and it will be in tmux 3.2 - `pane-border-lines` option/

PR #1802, opened 2019-06-19
Use KERN_PROC_CWD for osdep_get_cwd on NetBSD when available

iamleot:
KERN_PROC_CWD was recently added on NetBSD-current (in 8.99.42), use it if
available.


krytarowski:
Is this code valid? It tries to prompt for process CWD passing pid_t for process group.

nicm:
Process group ID is always the PID of the group leader.

nicm:
Applied now, thanks!

PR #1805, opened 2019-06-20
Fix minor typos in the manpage and document #{l:}

lacygoill:
I may be wrong, but I think I found a few minor typos in the manpage.
I'm not familiar with the troff syntax, so maybe the formatting is not correct.
I've just tested how the updated manpage was rendered with `$ groff -man tmux.1 | zathura -`, and it looked ok.

I also included a short description for the `#{l:}` format modifier, and used one of the few examples I could find in the codebase:

https://github.com/tmux/tmux/blob/c4a92e57993f3c45944428a6b7cf0a1be421d023/regress/format-strings.sh#L173

There were simpler examples, but for those simply doubling one `#` was enough. So I thought that maybe a complex example would show how useful `#{l:}` really is (i.e. you don't have to worry about missing a special character anywhere, no matter how complex a string is).

Let me know if you want something to be changed.
Feel free not to merge some of the fixes if I made mistakes.

### DIFF
--- a/cmd-find-window.c
+++ b/cmd-find-window.c
@@ -32,8 +32,8 @@ const struct cmd_entry cmd_find_window_entry = {
 	.name = "find-window",
 	.alias = "findw",
 
-	.args = { "CNt:TZ", 1, 1 },
-	.usage = "[-CNTZ] " CMD_TARGET_PANE_USAGE " match-string",
+	.args = { "CNrt:TZ", 1, 1 },
+	.usage = "[-CNrTZ] " CMD_TARGET_PANE_USAGE " match-string",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
@@ -57,30 +57,59 @@ cmd_find_window_exec(struct cmd *self, struct cmdq_item *item)
 	if (!C && !N && !T)
 		C = N = T = 1;
 
-	if (C && N && T) {
-		xasprintf(&filter,
-		    "#{||:"
-		    "#{C:%s},#{||:#{m:*%s*,#{window_name}},"
-		    "#{m:*%s*,#{pane_title}}}}",
-		    s, s, s);
-	} else if (C && N) {
-		xasprintf(&filter,
-		    "#{||:#{C:%s},#{m:*%s*,#{window_name}}}",
-		    s, s);
-	} else if (C && T) {
-		xasprintf(&filter,
-		    "#{||:#{C:%s},#{m:*%s*,#{pane_title}}}",
-		    s, s);
-	} else if (N && T) {
-		xasprintf(&filter,
-		    "#{||:#{m:*%s*,#{window_name}},#{m:*%s*,#{pane_title}}}",
-		    s, s);
-	} else if (C)
-		xasprintf(&filter, "#{C:%s}", s);
-	else if (N)
-		xasprintf(&filter, "#{m:*%s*,#{window_name}}", s);
-	else
-		xasprintf(&filter, "#{m:*%s*,#{pane_title}}", s);
+	if (!args_has(args, 'r')) {
+		if (C && N && T) {
+			xasprintf(&filter,
+			    "#{||:"
+			    "#{C:%s},#{||:#{m:*%s*,#{window_name}},"
+			    "#{m:*%s*,#{pane_title}}}}",
+			    s, s, s);
+		} else if (C && N) {
+			xasprintf(&filter,
+			    "#{||:#{C:%s},#{m:*%s*,#{window_name}}}",
+			    s, s);
+		} else if (C && T) {
+			xasprintf(&filter,
+			    "#{||:#{C:%s},#{m:*%s*,#{pane_title}}}",
+			    s, s);
+		} else if (N && T) {
+			xasprintf(&filter,
+			    "#{||:#{m:*%s*,#{window_name}},"
+			    "#{m:*%s*,#{pane_title}}}",
+			    s, s);
+		} else if (C)
+			xasprintf(&filter, "#{C:%s}", s);
+		else if (N)
+			xasprintf(&filter, "#{m:*%s*,#{window_name}}", s);
+		else
+			xasprintf(&filter, "#{m:*%s*,#{pane_title}}", s);
+	} else {
+		if (C && N && T) {
+			xasprintf(&filter,
+			    "#{||:"
+			    "#{C/r:%s},#{||:#{m/r:%s,#{window_name}},"
+			    "#{m/r:%s,#{pane_title}}}}",
+			    s, s, s);
+		} else if (C && N) {
+			xasprintf(&filter,
+			    "#{||:#{C/r:%s},#{m/r:%s,#{window_name}}}",
+			    s, s);
+		} else if (C && T) {
+			xasprintf(&filter,
+			    "#{||:#{C/r:%s},#{m/r:%s,#{pane_title}}}",
+			    s, s);
+		} else if (N && T) {
+			xasprintf(&filter,
+			    "#{||:#{m/r:%s,#{window_name}},"
+			    "#{m/r:%s,#{pane_title}}}",
+			    s, s);
+		} else if (C)
+			xasprintf(&filter, "#{C/r:%s}", s);
+		else if (N)
+			xasprintf(&filter, "#{m/r:%s,#{window_name}}", s);
+		else
+			xasprintf(&filter, "#{m/r:%s,#{pane_title}}", s);
+	}
 
 	new_args = args_parse("", 1, &argv);
 	if (args_has(args, 'Z'))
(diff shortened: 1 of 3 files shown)


lacygoill:
Sorry, I don't know much about git and I messed up the PR. I was trying to update my fork (because 3 commits have been pushed since I submitted it), and I wanted to include another fix with a second commit, then rebase, but it didn't work as I expected. Now my PR seems to include the last 3 commits; so, I close the PR. I'm going to try to simply generate a diff with `$ git diff`.

lacygoill:
Ok here is the output of `$ git diff`:

    diff --git a/tmux.1 b/tmux.1
    index 6aa6d29f..eb7fd40c 100644
    --- a/tmux.1
    +++ b/tmux.1
    @@ -461,7 +461,7 @@ Will execute
     .Ic if-shell ,
     the shell command
     .Xr true 1 ,
    -.Ic new-window
    +.Ic split-window
     and
     .Ic kill-session
     in that order.
    @@ -475,7 +475,7 @@ commands and their arguments.
     This section describes the syntax of commands parsed by
     .Nm ,
     for example in a configuration file or at the command prompt.
    -Note the when commands are entered into the shell, they are parsed by the shell
    +Note that when commands are entered into the shell, they are parsed by the shell
     - see for example
     .Xr ksh 1
     or
    @@ -2337,9 +2337,7 @@ spreads the current pane and any panes next to it out evenly.
     Make pane
     .Ar target-pane
     the active pane in window
    -.Ar target-window ,
    -or set its style (with
    -.Fl P ) .
    +.Ar target-window.
     If one of
     .Fl D ,
     .Fl L ,
    @@ -4104,7 +4102,7 @@ will expand the format twice, for example
     .Ql #{E:status-left}
     is the result of expanding the content of the
     .Ic status-left
    -option rather than the content itself.
    +option rather than the option itself.
     .Ql T:
     is like
     .Ql E:
    @@ -4140,7 +4138,7 @@ would change
     into
     .Ql bxBxbx .
     .Pp
    -In addition, the first line of a shell command's output may be inserted using
    +In addition, the last line of a shell command's output may be inserted using
     .Ql #() .
     For example,
     .Ql #(uptime)
    @@ -4159,6 +4157,14 @@ global environment set (see the
     .Sx GLOBAL AND SESSION ENVIRONMENT
     section).
     .Pp
    +An
    +.Ql l
    +specifies that any character in a string should be interpreted literally.
    +For example:
    +.Ql #{l:#{?pane_in_mode,#{?#{==:#{session_name},Summer},ABC,XYZ},xyz}}
    +will be replaced by
    +.Ql #{?pane_in_mode,#{?#{==:#{session_name},Summer},ABC,XYZ},xyz} .
    +.Pp
     The following variables are available, where appropriate:
     .Bl -column "XXXXXXXXXXXXXXXXXXX" "XXXXX"
     .It Sy "Variable name" Ta Sy "Alias" Ta Sy "Replaced with"

In addition to the few fixes, and to the documentation of `#{l:}`, the diff removes the documentation of the `-P` flag for the `select-pane` command, because according to the commit [`5f92f92`](https://github.com/tmux/tmux/commit/5f92f92908b81b4ec66682adb84b9ffc8d83c2f7), it should not be documented anymore:

> This makes select-pane -P and -g unnecessary so no longer document them

nicm:
Applied, thanks!

lacygoill:
I've noticed what could be a few errors in the manpage during the last few weeks.
I thought it was a good time to send a patch.

[man.diff.txt](https://github.com/tmux/tmux/files/3409759/man.diff.txt)

If it needs to be reworked, let me know, and I'll try.
But I still don't know much about the nroff syntax.

Feel free to ignore any change you disagree with.

---

'user-keys' is documented in the session options.
But it's not a session option; it's a server one.
I've moved it in the category of the server options right after 'terminal-overrides'.

---

In the next sentence, a comma is missing after `target-session`:

    This section describes the commands supported by tmux.  Most commands
    accept the optional -t (and sometimes -s) argument with one of
    target-client, target-session target-window, or target-pane.
                                 ^

---

There's a dot at the end of the description of the `pane_mode` format variable:

    pane_mode                       Name of pane mode, if any.
                                                             ^

But there are no dots for the other variables. I've removed it to be consistent.

---

The descriptions of the format variables `alternate_on`, `pane_in_mode`, `pane_input_off` , `pane_synchronized` starts with `If`.

    alternate_on                    If pane is in alternate screen
    pane_in_mode                    If pane is in a mode
    pane_input_off                  If input to pane is disabled
    pane_synchronized               If pane is synchronized

But for all the other boolean format variables, the description starts with `1 if`.
I've replaced `If` with `1 if` to be consistent.

---

The description of the 'set-titles-string' option mentions “window title”.
I found the term ambiguous, because “window” can be understood as a tmux window.
Besides, in the description of 'set-titles', the expression “client terminal title” is used.
I replaced “window title” with “client terminal title” to be more accurate.

---

`next-matching-bracket` and `previous-matching-bracket` are not documented, so I added them:

    next-matching-bracket                        %               M-C-f
    ...
    previous-matching-bracket                                    M-C-b

---

In the description of the `display-message` command, the expression “format variables” is used.

    -v prints verbose logging as the format is parsed and -a lists
    the format variables and their values.

But in the FORMATS section, another expression is used which I think refer to the same thing: “Replacement variables”.

    Replacement variables are enclosed in ‘#{' and ‘}', for example ‘#{session_name}'.

I found the usage of two expressions to refer to the same thing inconsistent.
Both of them are used only once in the manpage.
I thought that the first one was more telling, because:

- it's shorter
- you can infer from it that the variable only exists in the context of a format
- from “variable”, you can infer that it can be replaced or evaluated

---

In the description of the `display-message` command, it's said that:

    information is taken from target-pane if -t is given, otherwise the
    active pane for the session attached to target-client.

But I've checked and that doesn't seem to be the case.
It seems that – if `-t` is not given – the information is taken from the current pane, not from the active pane for the session attached to target-client. So I've just removed “for the session attached to target-client”.

---

In the description of the `select-pane` command, target-window is mentioned.

    Make pane target-pane the active pane in window target-window.

But it's not mentioned in the synopsis of the command.

    select-pane [-DLRUdel] [-T title] [-t target-pane]

I've replaced “in window target-window” with “in the current window”.

---

In the synopsis of the `last-pane` command, `[-t target-window]` is mentioned.

    last-pane [-de] [-t target-window]

But it's not mentioned in the following description, and I don't understand what would be the meaning of target-window in this context.
Besides, I tried to provide a window ID (`@123`) to `-t`, but it didn't have any effect.
I've removed it from the synopsis:

    last-pane [-de]

---

The `-N` flag which one can pass to `command-prompt` is not documented.

    command-prompt [-1i] [-I inputs] [-p prompts] [-t target-client]
            [template]

I think `-N` makes the prompt only accept numeric key presses.
I've added the option in the synopsis, as well as a brief description:

    command-prompt [-1Ni] [-I inputs] [-p prompts] [-t target-client]
            [template]

    ...
    -N makes the prompt only accept numeric key presses.

---

In the MOUSE SUPPORT section, it's said that the `{mouse}` and `=` tokens can be used to specify the most recent mouse event:

    In addition, target-session, target-window or target-pane may consist entirely
    of the token ‘{mouse}' (alternative form ‘=') to specify the most recent mouse
    event (see the MOUSE SUPPORT section)  or ‘{marked}' (alternative form ‘~') to
    specify the marked pane (see select-pane -m).

It seems that in the context of this sentence, “the session, window or pane where the most recent mouse event occurred” would be more accurate, because `{mouse}`/`=` seems to take the place of a session, window, or pane identifier.

---

In the FORMATS section, it's said that there are 6 operators to perform comparisons:

    Comparisons may be expressed by prefixing two comma-separated alterna‐
    tives by ‘==’, ‘!=’, ‘<’, ‘>’, ‘<=’ or ‘>=’ and a colon.

But they only seem to work for strings, not for numbers.
I've added “String” in front of “comparisons”:

    String comparisons may be expressed by prefixing two comma-separated alterna‐
    tives by ‘==’, ‘!=’, ‘<’, ‘>’, ‘<=’ or ‘>=’ and a colon.

---

In the description of the `command-prompt` command, it's said that `C-w` deletes the current word:

    Delete current word                            C-w

But in practice, it deletes from the cursor up to the start of the current/previous word.
I replaced the description with this one:

    Delete from cursor to start of word            C-w

---

When you enter copy mode, an indicator is displayed in the upper-right corner of the pane.
But its meaning is not documented; at least I didn't find it.
When I started using tmux, I didn't know what it meant, and it took me some time to understand.
I added this description:

    In copy mode, an indicator is displayed in the upper-right corner of the pane.
    It contains two numbers inside square brackets, separated by a slash.
    The first one is  the number of lines in the scrollback  buffer which are not
    visible and are below the last line of the pane.
    It's a variable number: it can change when you move in the scrollback buffer.
    It gives you a sense of how far from the bottom of the buffer you are.
    The second one is the number of lines in the scrollback buffer which are above
    the first visible line of the pane when you quit copy mode.
    It's a fixed number: it doesn't change when you move in the scrollback buffer.
    It gives you a sense of how big the buffer is.

nicm:
Applied with some minor tweaks, apart from last-pane - the target-window specifies which window to change pane in.

PR #1819, opened 2019-07-03
Inline note in the README

MartyLake:
Ensure that the first time contributor read the prerequisites before trying to build.

Long story: I tried to build the project from WSL and failed multiple times, each time installing a new package. When it finally built, I continued reading the `README` to realize the prerequisites are *after* the build command. I therefore think
that first time contributors would benefit from having the prerequisites listed before the build
command.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -33,15 +33,16 @@ configure with `--enable-utempter` to enable this.
 To get and build the latest from version control:
 
 ~~~bash
+# Note that this requires at least a working C compiler, `make`, `autoconf`,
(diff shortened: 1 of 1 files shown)


nicm:
I think this is fine as it is, putting it as a comment is ugly, and who just starts typing stuff without reading text that is like two lines later on screen?

MartyLake:
I always do that, I just scan to find the section that concerns me, and
then execute from top to bottom. But it's your project, feel free to reject
/ rework / suggest modifications as you want :)

nicm:
I moved the text before the commands, thanks!

PR #1848, opened 2019-07-19
Implement more vi keys for choose tree

### LINKED ISSUE #1846: Feature request: bind-key -T choose-tree
I find some of the functionality in `choose-tree` to use bizarre mappings (I'm so used to vim that I usually hit `/` to search, which works in all tmux output panes *except* `choose-tree`).

Obviously, we could fix this one little issue.

But it would be more interesting to expose a `choose-tree` table for key bindings, à la `copy-mode`, to allow custom `choose-tree` mappings. This would likely require exposure of the native choose-tree functionality as commands (such as `choose-tree-search`, which AFAIK is not available as a tmux command. It could be designed to fail if not in `choose-tree`?).

### Required information

* tmux version (`tmux -V`) 2.9a
* Platform (`uname -sp`) Darwin i386
* $TERM inside and outside of tmux (`echo $TERM`)
  - screen256-color
  - xterm256-color
* Logs from tmux (`tmux kill-server; tmux -vv new`) N/A, surely.

benknoble:
Tested locally.

Closes #1846

(Someone should let me know if master is *not* the appropriate target for this PR.)

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -1019,6 +1019,8 @@ mode_tree_key(struct mode_tree_data *mtd, struct client *c, key_code *key,
 			mode_tree_build(mtd);
 		}
 		break;
+	case '?':
+	case '/':
 	case '\023': /* C-s */
 		mtd->references++;
 		status_prompt_set(c, "(search) ", "",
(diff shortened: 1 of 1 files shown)


nicm:
Applied now, thanks!

Issue #1840, opened 2019-07-14
[lost server] at latest master, Raspbian 9


benlk:
### Issue description

When running `tmux new`, tmux appears to start, but then quits unexpectedly, leaving the terminal in a state with a green bar across the bottom, and no output while typing.

Here's the contents of the shell:
```
[lost server]
             pi@redbird:~/software/tmux$ 
```

![Screen Shot 2019-07-14 at 19 11 16 ](https://user-images.githubusercontent.com/1754187/61190504-3147a080-a66b-11e9-883a-7891c5c2f65e.png)

The green bar goes away after hitting control-C, but the terminal remains borked. Typing new commands doesn't show the commands as typed, and running `ls` results in the results being scattered across the screen:

![Screen Shot 2019-07-14 at 19 12 34 ](https://user-images.githubusercontent.com/1754187/61190514-59370400-a66b-11e9-9124-3d3f49751fae.png)

The weirdness persists until a new shell is opened.

This occurs on a raspberry pi running Raspbian 9, fully updated. I ran into this problem with the raspbian-packaged version of tmux, and decided to try building newer versions.

I've checked out latest master at c9216493cfa3b62b6edc9dae11892a071e56fa5c, but I also saw this under 2.9a and 2.something ancient.

<details>
<summary>A list of strange things seen during `make`: </summary>

The shell output when running make is full of things like this:

> depbase=`echo alerts.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
[...]

PR #1857, opened 2019-07-26
Help clarify "lost server" situation

nealmcb:
As discussed in https://github.com/tmux/tmux/issues/1840, the "lost server" message can be confusing to users.
Return a more helpful string.

### DIFF
--- a/client.c
+++ b/client.c
@@ -202,7 +202,7 @@ client_exit_message(void)
 	case CLIENT_EXIT_TERMINATED:
 		return ("terminated");
 	case CLIENT_EXIT_LOST_SERVER:
-		return ("lost server");
+		return ("Client lost connection to tmux server. Try tmux -vv");
 	case CLIENT_EXIT_EXITED:
 		return ("exited");
 	case CLIENT_EXIT_SERVER_EXITED:
--- a/client.c
+++ b/client.c
@@ -202,7 +202,7 @@ client_exit_message(void)
 	case CLIENT_EXIT_TERMINATED:
 		return ("terminated");
 	case CLIENT_EXIT_LOST_SERVER:
-		return ("Client lost connection to tmux server. Try tmux -vv");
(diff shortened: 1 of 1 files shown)


nicm:
I'm OK with changing it to "server exited unexpectedly" but leave out the advice.

nealmcb:
Thanks for the input. I made that change.

I'd favor "tmux server exited unexpectedly", since many users might not even think of tmux as having a server. Would that be OK?

nicm:
I don't think that's necessary, the user just ran tmux so they know what
the error comes from. The fact that tmux has a server is in the manual.

nealmcb:
OK. Thanks.

krader1961:
Keep in mind that this failure mode can have many causes. For example, if the system goes into "sleep" mode you'll lose the connection (or be unable to establish one) to the tmux server even though it is technically still alive. Similarly a misconfigured firewall or other network problem can induce this failure even though the tmux server is still running. A phrase like "unable to communicate with the tmux server" would be better, IMHO.

nealmcb:
Thanks. Shall I change to `unable to communicate with the tmux server`?

nicm:
No it's fine now

nicm:
There is no firewall for Unix domain sockets and suspend should not
disconnect them, if it does then unexpectedly is fine

nicm:
What would be nice would be document the various messages and their meaning in the manual, not sure maybe a new section near the end.

nicm:
Changed now, thanks!

nealmcb:
Hmm, ahh - actually changed for some reason via essentially identical commit at https://github.com/tmux/tmux/commit/8f40f791d9d45fa70ca5d388eba0f4c2a4cf397b

Spoorthi-kr:
Can anyone please help me out on how to resolve the lost server issue?
I accidently killed the server using tmux kill-server command.

PR #1869, opened 2019-08-13
 Add window-size latest option value

tommie:
Two small warning suppressions (after which building is clean for me).

### DIFF
--- a/cmd-break-pane.c
+++ b/cmd-break-pane.c
@@ -81,6 +81,7 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 	wp->flags |= PANE_STYLECHANGED;
 	TAILQ_INSERT_HEAD(&w->panes, wp, entry);
 	w->active = wp;
+	w->latest = c;
 
 	if (!args_has(args, 'n')) {
 		name = default_window_name(w);
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -259,6 +259,7 @@ cmd_new_session_exec(struct cmd *self, struct cmdq_item *item)
 	memset(&sc, 0, sizeof sc);
 	sc.item = item;
 	sc.s = s;
+	sc.c = c;
 
 	sc.name = args_get(args, 'n');
 	sc.argc = args->argc;
--- a/cmd-new-window.c
+++ b/cmd-new-window.c
@@ -72,6 +72,7 @@ cmd_new_window_exec(struct cmd *self, struct cmdq_item *item)
 	memset(&sc, 0, sizeof sc);
 	sc.item = item;
 	sc.s = s;
+	sc.c = c;
 
 	sc.name = args_get(args, 'n');
 	sc.argc = args->argc;
--- a/cmd-respawn-window.c
+++ b/cmd-respawn-window.c
@@ -59,6 +59,7 @@ cmd_respawn_window_exec(struct cmd *self, struct cmdq_item *item)
 	sc.item = item;
 	sc.s = s;
 	sc.wl = wl;
+	sc.c = cmd_find_client(item, NULL, 1);
 
 	sc.name = NULL;
 	sc.argc = args->argc;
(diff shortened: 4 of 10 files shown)


tommie:
Tested with two clients.

Tested inputs from each client, and resizing happens. Tested setting and changing the window-size option, which also recalculates the size.

nicm:
OK thanks will take a look tomorrow

tommie:
> will take a look tomorrow

Thanks.

There was a bug where setting latest on startup gives the default size 80x24 until the first input is given. Probably because `w->latest` isn't initialized on creation. Updated the PR.

Repro:

```
./tmux -f <(echo set-window-option -g window-size latest) -L tmdev
```

The `break-pane` command doesn't seem to need this initialization, probably because it inherits the size from the current window. Added anyway.

nicm:
I'm not going to fix the warnings like this, they are bugs in GCC and for strftime the #ifdef is worse than the warning.

tommie:
Dropped the warning fix commit.

nicm:
Otherwise I think this looks good, thanks.

Couple of style nits and I deliberately used void * for the latest pointer in struct window because it doesn't take a reference to it, so it must not dereference it.

nicm:
Still not wild about the loop, the problem is that it does it even if nothing needs to change...

nicm:
Well, most sessions will only have <100 or so windows I would guess so its not like it is a big loop.

nicm:
I don't think your idea of checking if the client has changed more recently than the session would work for windows that are linked to multiple sessions.

tommie:
> Couple of style nits

Sorry, I can't see any review comments inline. I'm not very used to GitHub PRs, so sorry if I missed something.

>  I deliberately used void * for the latest pointer in struct window

I didn't change that code: x.diff.txt was using `struct client*`. Changed `spawn_context` and `window` to use `void*` now.

tommie:
> I don't think your idea of checking if the client has changed more recently than the session would work for windows that are linked to multiple sessions.

Ah, I didn't know they could be cross-linked.

> Still not wild about the loop, the problem is that it does it even if nothing needs to change...

I'll need to dig into the code a bit more to see what could be done here. I agree it looks a bit silly since it will mostly do nothing. I'm, of course, open to ideas.

nicm:
Oh maybe I changed it to void * later.

nicm:
Only struct window needs to be void *, not spawn_context because it is on the stack so the client can't be destroyed while it exists.

[x.diff.txt](https://github.com/tmux/tmux/files/3502639/x.diff.txt)

tommie:
Applied.

Since there is a switch now, it would probably be useful to turn WINDOW_SIZE_* into an enum to get a warning if a new value is added.

tommie:
> I don't know what to do to avoid looping over all windows. Perhaps cache the window-size property in the session and only do it if it is latest.

Thinking some more, I think changing all windows in the session is wrong. Say you have an inactive window that is active in another session/client. You don't want that window to change.

So I think it's better to only update the current window (and we only need that if it's `W_S_LATEST`), but also do it when switching to a window that has `W_S_LATEST`.

tommie:
> also do it when switching to a window that has `W_S_LATEST`.

Just moving the `recalculate_sizes` to the bottom of the key handler is enough. I think it's a bit annoying that pressing the prefix key causes a resize, but I suspect that wouldn't be a problem in a more realistic setting.

frosencrantz:
I haven't tried this, but it looks useful.  There hasn't been any activity for almost 2 weeks, I was wondering why this is stalled?

nicm:
I like this, here are a couple more changes but otherwise I think it is seems good: [tmux-latest-xyz.diff.txt](https://github.com/tmux/tmux/files/3597354/tmux-latest-xyz.diff.txt)

nicm:
I have applied this now, thanks!

tommie:
Thanks for providing tmux. My everyday would be much more disorganized without it (and mosh)!

frosencrantz:
@tommie Thank you for this fix.  I now have this turned on all the time.  It makes window-size usable for me.  Before this change I had tried window-size, but it got in the way more than it helped.  I had turned it off, even though through a normal day I will use a tmux session with different size screens.

With 'latest'  the window-size feature is very usable.

Also, thanks nicm for adding the base feature!

Issue #1872, opened 2019-08-15
sort order of paste buffers changed


gobenji:
Commit 7bcc0d16f24 resulted in a change to the order of the buffers when doing `choose-buffer -O name` because the buffer index is no longer 0-padded.

For example, with "bind = choose-buffer -O name -N -F "#{buffer_sample}", the order used to be:
![screenshot-old](https://user-images.githubusercontent.com/254681/63079626-d232bf80-bf79-11e9-9acc-41fa9817f7b5.png)

After that commit, the order is:
![screenshot-new](https://user-images.githubusercontent.com/254681/63079630-d65edd00-bf79-11e9-8b75-9a3fb7fe4b4f.png)

I don't believe it's possible to get the old behavior from config.

nicm:
I like the new names better, use -O time if you want it in the order the buffers were created.

gobenji:
What I don't like with -O time is that it's in fact the reverse order from -O name:

![screenshot-time](https://user-images.githubusercontent.com/254681/63083253-39a13d00-bf83-11e9-87bc-c442be52b532.png)

I like the buffers to appear in chronological order (newest buffer on bottommost line). Can we have a "--reverse" option on that sort :)

nicm:
Yes I am fine with having a way to reverse sort and it would not be hard to add. Either add new sort types reverse-time etc or add a separate -r flag, the former would probably be better because you wouldn't need a new key binding when using the modes interactively.

PR #1875, opened 2019-08-16
For 1872

gobenji:
As suggested in https://github.com/tmux/tmux/issues/1872#issuecomment-521564818

### DIFF
--- a/cmd-choose-tree.c
+++ b/cmd-choose-tree.c
@@ -31,7 +31,7 @@ const struct cmd_entry cmd_choose_tree_entry = {
 	.alias = NULL,
 
 	.args = { "F:Gf:NO:st:wZ", 0, 1 },
-	.usage = "[-GNsw] [-F format] [-f filter] [-O sort-order] "
+	.usage = "[-GNswZ] [-F format] [-f filter] [-O sort-order] "
 	         CMD_TARGET_PANE_USAGE " [template]",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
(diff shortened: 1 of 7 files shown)


nicm:
This looks great, but we can't use qsort_r because it isn't supported on OpenBSD. I'll see if I can get it added, give me a few days. If not we will need to use a temporary global instead.

Here is an updated diff with some minor style fixes and one bug fix (missing breaks inwindow_tree_cmp_session -  the WINDOW_TREE_BY_TIME case would always fall through): [tmux-1875.diff.txt](https://github.com/tmux/tmux/files/3508903/tmux-1875.diff.txt)

nicm:
Actually that's an old version of the diff, try this [tmux-1875-2.diff.txt](https://github.com/tmux/tmux/files/3508925/tmux-1875-2.diff.txt)

nicm:
Well, NetBSD don't have qsort_r either and FreeBSD have substantially changed their qsort code so it is not easy to copy (as well as using a different argument order). Since OpenBSD's qsort calls heapsort it would be a fairly big change to add qsort_r from scratch and I don't have time to do it now. It looks like Solaris doesn't have it as well.

I think it would be best just to use a global for the moment.

nicm:
Here is a version which does this: [tmux-1875-3.diff.txt](https://github.com/tmux/tmux/files/3508978/tmux-1875-3.diff.txt)

gobenji:
> Here is an updated diff with some minor style fixes and one bug fix (missing breaks inwindow_tree_cmp_session - the WINDOW_TREE_BY_TIME case would always fall through): [tmux-1875.diff.txt](https://github.com/tmux/tmux/files/3508903/tmux-1875.diff.txt)

Thanks for the review and style fixes. Good catch on the missing breaks.
Thanks also for the qsort_r() investigation.

My only further comment is regarding:

--- a/window-buffer.c
+++ b/window-buffer.c
@@ -74,6 +74,7 @@ static const char *window_buffer_sort_list[] = {
 	"name",
 	"size"
 };
+static struct mode_tree_sort_criteria window_buffer_sort;
 
Why not have it be a pointer and avoid the memcpy? Same for the other similar
instances.

nicm:
Yes you're right they could just be pointers.

gobenji:
Do you want me to update the pull request or do you want to work from what you've got?

nicm:
It's OK thanks, I'll change it before I commit it.

nicm:
This is in now, thanks!

PR #1883, opened 2019-08-24
Add cursor-down-and-cancel cmd

mckellyln:
@nicm added a and-cancel to cursor-down.


nicm:

PR #1891, opened 2019-08-31
Trivial alphabetization patch for command list

rountree:
The list of commands in the man page was not quite alphabetical, which caused a few moments of confusion.


nicm:

PR #1899, opened 2019-09-04
include header for xmalloc

cyyever:
xmalloc defaults to return int when no prototype is provided,so on some platforms with sizeof(int)=4 high allocated addresses could be truncated,causing invalid memory access.

### DIFF
--- a/compat/asprintf.c
+++ b/compat/asprintf.c
@@ -19,8 +19,10 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 
 #include "compat.h"
+#include "xmalloc.h"
 
 int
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Why?  What's the reason for this change?

cyyever:
> 
> 
> Why? What's the reason for this change?



> 
> 
> Why? What's the reason for this change?

Avoids 

> 
> 
> Why? What's the reason for this change?

Avoids implicit type conversion.

cyyever:
> 
> 
> Why? What's the reason for this change?

The direct reason is that I built tmux on Windows 10 WSL and it core dumped,after I added this include statement,it was OK.

nicm:
Applied now, thanks!

PR #1900, opened 2019-09-05
Fixed referencing in manpage.

emanuele6:
The `pty(4)` manual page, which does not exist, was referenced twice in the tmux manual page.

It should have refered these pages instead:
* in the first instance, the `pty(7)` manual page: where the technical details of pseudo terminals are documented;
* in the second instance, the `pts(4)` manual page: where the pseudo terminal master and slave devices are documented.

I also added `pts(4)` in the `SEE ALSO` section

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -61,7 +61,7 @@ A window occupies the entire screen
 and may be split into rectangular panes,
 each of which is a separate pseudo terminal
 (the
-.Xr pty 4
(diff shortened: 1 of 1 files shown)


nicm:
No this is wrong, pty(4) is correct, look here: https://man.openbsd.org/pty.4

emanuele6:
I see, @nicm

PR #1907, opened 2019-09-11
Look up the ACS->UTF-8 table directly.

koie:
Using `bsearch` to search `tty_acs_table` is time consuming and I think it's better to go directly to the index.
sizeof tty_acs_table is 2KiB.

### DIFF
--- a/tty-acs.c
+++ b/tty-acs.c
@@ -22,61 +22,45 @@
 
 #include "tmux.h"
 
-static int	tty_acs_cmp(const void *, const void *);
-
 /* Table mapping ACS entries to UTF-8. */
-struct tty_acs_entry {
-	u_char	 	 key;
-	const char	*string;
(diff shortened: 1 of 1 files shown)


nicm:
Well does it make much difference? It's not a big table so bsearch will only be doing what 4 or 5 comparisons and making it sparse means it hit like 32 cache lines instead of 4?

koie:
There are many branches when using bsearch, but in any case it is certainly not related to the overall performance.
Microbenchmarking tty_acs_get was 10 microseconds per call (on my slow machine).
Rather than index, linear search may be better depending on the fact that the search character tends to be biased to q and x (and ~).

nicm:
I think with such a small table it will all come out in the wash, unless
you can show this is a hotspot. There aren't usually that many ACS
characters on screen at one time.

PR #1912, opened 2019-09-15
Add push_default and pop_default as style options

### LINKED ISSUE #1909: window-status styles not working properly
```sh
# uname -sp && tmux -V && echo $TERM
Linux x86_64
tmux next-3.1
xterm-256color
# ./tmux
# echo $TERM
tmux-256color
```

Using tmux 2.9a (as well as current tmux 3.1-next), the format styles for the window status seem to be lost when a window status format is specified. See the configuration here:

```
set-option -g status-style fg=colour18,bg=colour20
set-window-option -g window-status-format "#[fg=colour18,bg=colour20] #I:#[default]#W#[fg=colour18,bg=colour20]#F"
set-window-option -g window-status-current-style bg=green,bold
set-window-option -g window-status-last-style bg=yellow
set-window-option -g window-status-style bg=red
```

This produces the following in current tmux, after creating a new session, then immediately creating 3 additional windows:

No colors other than current

Without the format line, this is what appears, and what *should* appear if the format option is set, other than the window number and flags being the same fg/bg as the rest of the status bar:

Screen Shot 2019-09-13 at 10 18 28 AM

Prior to tmux 2.9a (maybe it was in 2.8 as well, but definitely as of 2.7), this worked as intended. The `#[default]` in the format should cause tmux to revert back to the style it would be normally, but that appears to not be working for the non-active window.

[tmux-client-22130.log](https://github.com/tmux/tmux/files/3610604/tmux-client-22130.log)
[tmux-out-22132.log](https://github.com/tmux/tmux/files/3610605/tmux-out-22132.log)
[tmux-server-22132.log](https://github.com/tmux/tmux/files/3610606/tmux-server-22132.log)

johnmiked15:
This adds push_default and pop_default as style options in format strings.

* `#[push-default]` will save the current style as the new default when `#[default]` is specified later
* `#[pop-default]` will revert this and cause #[default] to go back to the base style
* Add `#[push-default]` and `#[pop-default]` around `#{window-status-format}` and `#{window-status-current-format}` in the default `status-format[]`

Fixes #1909

### DIFF
--- a/format-draw.c
+++ b/format-draw.c
@@ -514,7 +514,7 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 	int			 list_state = -1, fill = -1;
 	enum style_align	 list_align = STYLE_ALIGN_DEFAULT;
 	struct grid_cell	 gc;
-	struct style		 sy;
+	struct style		 sy, saved_sy;
 	struct utf8_data	*ud = &sy.gc.data;
 	const char		*cp, *end;
 	enum utf8_state		 more;
@@ -522,8 +522,10 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 	struct format_range	*fr = NULL, *fr1;
 	struct format_ranges	 frs;
 	struct style_range	*sr;
+	struct grid_cell	 current_default;
 
-	style_set(&sy, base);
+	memcpy(&current_default, base, sizeof current_default);
+	style_set(&sy, &current_default);
 	TAILQ_INIT(&frs);
 	log_debug("%s: %s", __func__, expanded);
 
@@ -535,7 +537,7 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 	for (i = 0; i < TOTAL; i++) {
 		screen_init(&s[i], size, 1, 0);
 		screen_write_start(&ctx[i], NULL, &s[i]);
-		screen_write_clearendofline(&ctx[i], base->bg);
+		screen_write_clearendofline(&ctx[i], current_default.bg);
 		width[i] = 0;
 	}
 
@@ -581,7 +583,8 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 			goto out;
 		}
 		tmp = xstrndup(cp + 2, end - (cp + 2));
-		if (style_parse(&sy, base, tmp) != 0) {
+		style_copy(&saved_sy, &sy);
+		if (style_parse(&sy, &current_default, tmp) != 0) {
 			log_debug("%s: invalid style '%s'", __func__, tmp);
 			free(tmp);
 			cp = end + 1;
@@ -594,6 +597,16 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 		/* If this style has a fill colour, store it for later. */
 		if (sy.fill != 8)
 			fill = sy.fill;
+
+		/* If this style pushed or popped the default, update it. */
+		if (sy.default_type == STYLE_DEFAULT_PUSH) {
+			memcpy(&current_default, &saved_sy.gc, sizeof current_default);
+			sy.default_type = STYLE_DEFAULT_BASE;
+		}
+		else if (sy.default_type == STYLE_DEFAULT_POP) {
+			memcpy(&current_default, base, sizeof current_default);
+			sy.default_type = STYLE_DEFAULT_BASE;
+		}
 
 		/* Check the list state. */
 		switch (sy.list) {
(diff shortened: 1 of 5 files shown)


nicm:
Hi

Thanks, this is a good start, however it is not OK to modify a style after it has been parsed. It happens that format_draw uses the same struct style for every call to style_parse but I don't want this to be a requirement. push-default and pop-default need to set an enum in the style and the push/pop then taken care of in format_draw. Also what does it do when pop-default is seen without a preceding push-default?

johnmiked15:
> push-default and pop-default need to set an enum in the style and the push/pop then taken care of in format_draw.
I'll change it to do this instead, stay tuned.

> Also what does it do when pop-default is seen without a preceding push-default?
In this case, it's a no-op.

johnmiked15:
Ok, I updated it to use an enum in `style_parse()` and do the actual pushing and popping in `format-draw.c`. I still have to modify the style after the call to `style_parse()` so that every subsequent call doesn't attempt to save the current style as the default (it only wants to save off the default once). I don't know of another way to make format-draw perform this operation once while using the same struct in each call to `style_parse()`.

nicm:
Looks good, but it doesn't appear to work for me, I am doing this:
~~~~
tmux set window-status-current-format 'default#[push-default bg=red]red#[pop-default]default'

johnmiked15:
> Looks good, but it doesn't appear to work for me, I am doing this:
> 
> ```
> tmux set window-status-current-format 'default#[push-default bg=red]red#[pop-default]default'
> ```

This would have worked with my first pattern, where I save the default in the style struct itself, but now, the default is not updated until the return from `style_parse()`, so it has to be a style option by itself. In the man page update, I tried to make this clear. So you could use:

```
tmux set window-status-current-format 'default#[push-default]#[bg=red]red#[pop-default]default'
```

nicm:
Why not push the style that has been updated?

nicm:
You need a couple of lines in style_tostring as well.

johnmiked15:
> You need a couple of lines in style_tostring as well.

Sure. Let me take care of that, and fix your example.

johnmiked15:
Since `#[push-default]` and `#[pop-default]` do not do anything *by themselves*, your quick example appears to not work because after the `#[pop-default]`, you need to change the style back to using the default. The following demonstrates the functionality of this PR:

```
tmux set window-status-current-format 'default#[push-default bg=red]#[bg=green]green#[default]red#[pop-default]#[default]default
```

push/pop only change what `#[default]` will do later, they don't actually change the style. `#[default]` must be used alone.

nicm:
I think push-default should push the style including any new changes to that style. Why is that not happening, doesn't it copy it to the default after it is updated?

nicm:
Er sorry I mean it should push it BEFORE any changes. That means you'll need to copy it before calling style_parse I think.

johnmiked15:
> I think push-default should push the style including any new changes to that style. Why is that not happening, doesn't it copy it to the default after it is updated?

Ah, I think I see what you're saying. I can make push-default work that way. I think push-default and pop-default should still not actually change the style itself, though.

johnmiked15:
I've just fixed the push behavior to save what it was before.

See the following for a good demonstration:

```
tmux set window-status-current-format 'default#[bg=blue]#[push-default bg=red]red#[bg=green]green#[default]blue#[pop-default]#[default]default'
```

Let me know if there's anything else.

nicm:
Yes I think that push-default in a style should change the default to right before style_parse was called and the current should change as normal to whatever is in the string (so if you start with bg=blue and you do #[push-default bg=red] you end up with default of bg=blue and current of bg=red. Then pop-default takes you back to bg=blue.

It is OK if they change the struct style in format_draw because it owns it, but not if it changes it in style_parse.

I wish I hadn't made it so style_parse is incremental and builds on the passed in style, it is confusing, it would be better if it took in the current and default grid_cell separately and built the style from them. But you don't need to change that.

johnmiked15:
Cool. This is what the PR currently does, other than:

> Then pop-default takes you back to bg=blue.

#[push-default] and #[pop-default] do not actually change the style. You have to use #[default] to actually use what the default is. #[pop-default] will just reset the base grid_cell that's passed into style_parse().

To make it possible to use nested push-defaults with corresponding pop-defaults, format_draw() would need to remember the full stack of push-defaults. This would require storing the defaults as an actual stack when pushing, and pop would actually remove them from that list. If we don't do that, then the only thing #[pop-default] can do is reset the default back to the base grid_cell's style, so it can't go back to bg=blue.

nicm:
Yeah you are right, I think this is OK. I was wondering if push-default could imply default but I think it is better if they are neutral.

nicm:
OK I have applied this to OpenBSD now with some minor changes mostly to tmux.1 - it will be in GitHub when it syncs up in an hour or so, let me know if you see any problems.

johnmiked15:
Looks good. You don't do direct pull request merges here, though?

nicm:
No no merges

Issue #1911, opened 2019-09-15
Bold attribute is losen for color degradation from 256 to 8 colors


shunlir:
### Issue description

for color degradation from 256 color to 8 color, some colors lose the `bold` attribute while others note, the result is inconsistency and confusing.
e.g., `colour34` has issue:
```tmux
setw -g window-status-current-style fg=colour34,bold # fg=green3,bold fallback to 2(green), losing bold
```
while `colour40` is okay:
```tmux
setw -g window-status-current-style fg=colour40,bold # fg=green3,bold fallback to 10 then to 2(green)+bold
```
To reproduce, also ensure the terminal emulator enables bold font/text (or both bold and bright) for ANSI colors
I had a simple look at the source code (https://github.com/tmux/tmux/blob/9b3fefc43529666b1f6adc5efc89f6a5f127f274/tty.c#L2371), seems it interprets `bold` attr as `bright`, meaning more colors, however, the standard indicates it can be both `bold` and `bright`.

### Required information

* tmux version: `2.6`
* Platform: `Linux x86_64`
* $TERM inside tmux is `tmux`, outside is `xterm`
* Logs from tmux (`tmux kill-server; tmux -vv new`).

nicm:
Using 256 colours inside tmux on 8 colour terminals is best effort only.

shunlir:
@nicm, if so, I think it's better to not clear the `bold`/`bright` setting(https://github.com/tmux/tmux/blob/9b3fefc43529666b1f6adc5efc89f6a5f127f274/tty.c#L2380), because it corresponds to SGR code 1, the it means depends on the terminal emulator's capabilities or settings, TMUX shouldn't simply use it for 8 extra colors in 8 colors terminal

nicm:
OK send a diff?

PR #1914, opened 2019-09-18
respect bold/bright attr when fallback from 256 colorname to 8 colors…

shunlir:
respect bold/bright attr when fallback from 256 colorname to 8 colors, since SGR code 1 means bold or bright according to the standard
(https://github.com/tmux/tmux/issues/1911)

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -2374,10 +2374,7 @@ tty_check_fg(struct tty *tty, struct window_pane *wp, struct grid_cell *gc)
 				gc->fg &= 7;
 				if (colours >= 16)
 					gc->fg += 90;
-				else
-					gc->attr |= GRID_ATTR_BRIGHT;
-			} else
-				gc->attr &= ~GRID_ATTR_BRIGHT;
+			}
 		}
 		return;
 	}


nicm:
I think we should change the previous branch of the if not to set the BRIGHT flag either (line 2378).

nicm:
So this so it is the same as bg:
~~~~
Index: tty.c
RCS file: /cvs/src/usr.bin/tmux/tty.c,v
retrieving revision 1.331
diff -u -p -r1.331 tty.c
--- tty.c       5 Aug 2019 06:42:02 -0000       1.331
+++ tty.c       18 Sep 2019 15:04:31 -0000
@@ -2374,10 +2374,7 @@ tty_check_fg(struct tty *tty, struct win
                                gc->fg &= 7;
                                if (colours >= 16)
                                        gc->fg += 90;
-                               else
-                                       gc->attr |= GRID_ATTR_BRIGHT;
-                       } else
-                               gc->attr &= ~GRID_ATTR_BRIGHT;
+                       }
                }
                return;

shunlir:
@nicm Thanks for your careful review, I updated the PR accordingly

nicm:
I have applied this now, thanks!

PR #1915, opened 2019-09-19
WINDOW_TREE_DEFAULT_FORMAT: always display pane_title

blueyed:
With multiple panes it would display currently:

    (0)  - 0: 1 windows (attached)
    (1)  └─> + 1: *Z (2 panes)

But it is certainly better to see the current's pane title.
With this patch:

    (0)  - 0: 1 windows (attached)
    (1)  └─> + 1: *Z (2 panes) ⩫ ~_ (/home/user/Vcs/tmux/tmux -L test)

### DIFF
--- a/window-tree.c
+++ b/window-tree.c
@@ -41,8 +41,7 @@ static void		 window_tree_key(struct window_mode_entry *,
 	"," \
 		"#{?window_format," \
 			"#{window_name}#{window_flags} " \
-			"(#{window_panes} panes)" \
-			"#{?#{==:#{window_panes},1}, \"#{pane_title}\",}" \
+			"(#{window_panes} panes) #{pane_title}" \
 		"," \
 			"#{session_windows} windows" \
 			"#{?session_grouped, " \


blueyed:
btw: it would be nice to document the defaul in the man page - either verbatim, or mentioning that it is complex, with a refererence to the code.

nicm:
I don't actually like having the pane title, it looks cluttered. You can see (or use) the default with tree_mode_format.

blueyed:
Fair enough, but I think it makes it easier to find a window in general (especially with a lot of them), and it's better to have the current pane's title there than nothing. 
I'm having the current directory and current/last command in there.

Feel free to close this though if you do not want to take it.

Do you think the default should be documented?

nicm:
I don't want to include the content but it could mention that you can use
#{tree_mode_format} to see it.

blueyed:
Ok.

Do you think it makes sense to add a config variable/setting to set it?

(Currently I would have to adjust / overwrite the default bindings for chosing a session/window.)

nicm:
I don't really see why, it seem easy enough just changing the key bindings?

PR #1920, opened 2019-09-25
Break line before `screen*`, add `Smol` ANSI sequence to `xterm*` and `screen*`

rbanffy:
A little bit unsure if there aren't more non-ncurses sequences that are nevertheless commonly implemented in popular terminal programs. Smol is my favorite, of course, because it allows beautiful overlined status lines, but there are others that  many terminal programs implement that are dear to other people like me.

In any case, it's nice to break lines for each terminal override rule anyway.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -226,7 +226,8 @@ const struct options_table_entry options_table[] = {
 	  .flags = OPTIONS_TABLE_IS_ARRAY,
 	  .default_str = "xterm*:XT:Ms=\\E]52;%p1%s;%p2%s\\007"
 			 ":Cs=\\E]12;%p1%s\\007:Cr=\\E]112\\007"
-			 ":Ss=\\E[%p1%d q:Se=\\E[2 q,screen*:XT",
+			 ":Ss=\\E[%p1%d q:Se=\\E[2 q:Smol=\\E[53m,"
+	                 "screen*:XT:Smol=\\E[53m",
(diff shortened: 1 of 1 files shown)


nicm:
Can you get rid of all the other commits?

rbanffy:
Sorry. Just did. Local history was a mess.

nicm:
But xterm doesn't have overline??

rbanffy:
@nicm: sadly, no. However, all apps derived from VTE and Konsole do (even though they, for some reason, identify as "xterm*"). It seems "xterm-256color" ended up being a catch-all value for TERM and terminal writers. I wonder if adding the most commonly implemented ANSI sequences that don't have official terminfo names under "*" would be a better approach.

I tried to look into xterm's code to figure out what would be needed for it to support overline, but couldn't find any obvious places. The same goes for iTerm (I foolishly volunteered to do so) - the code is way above my level of cleverness to help.

nicm:
I don't want to add this by default for terminals that don't support it and neither screen nor xterm do. Thomas Dickey already added it to terminfo for tmux and vte.

rbanffy:
You are right. I see you added it in May. I checked my Fedora 31 box, which has an ncurses dated from last month, but it doesn't show anything Smol/Rmol in the vte entries. I guess distros are picking up older versions. :-(

I think we should close this.

PR #1924, opened 2019-09-30
Fixed truncation of long commands in list-keys output.

gpakosz:
With the advent of `display-menu`, we're facing bindings with very long commands.

I stumbled upon this particular one
```
bind-key -T root MouseDown1Pane if-shell -F -t = "#{||:#{mouse_any_flag},#{pane_in_mode}}" "select-pane -t=; send-keys -M" "display-menu -t= -xM -yM -T \"#[align=centre]#{pane_index} (#{pane_id})\" '#{?mouse_word,Search For #[underscore]#{=/9/...:mouse_word},}' 'C-r' {copy-mode -t=; send -Xt= search-backward \"#{q:mouse_word}\"} '#{?mouse_word,Type #[underscore]#{=/9/...:mouse_word},}' 'C-y' {send-keys -l -- \"#{q:mouse_word}\"} '#{?mouse_word,Copy #[underscore]#{=/9/...:mouse_word},}' 'c' {set-buffer -- \"#{q:mouse_word}\"} '#{?mouse_line,Copy Line,}' 'l' {set-buffer -- \"#{q:mouse_line}\"} '' 'Horizontal Split' 'h' {split-window -c '#{pane_current_path}' -h} 'Vertical Split' 'v' {split-window -c '#{pane_current_path}' -v} '' 'Swap Up' 'u' {swap-pane -U} 'Swap Down' 'd' {swap-pane -D} '#{?pane_marked_set,,-}Swap Marked' 's' {swap-pane} '' 'Kill' 'X' {kill-pane} 'Respawn' 'R' {respawn-pane -k} '#{?pane_marked,Unmark,Mark}' 'm' {select-pane -m} '#{?window_zoomed_flag,Unzoom,Zoom}' 'z' {resize-pane -Z}"
```
Which is a stock binding modified to inject `-c '#{pane_current_path}'` to `split-window` calls.

I can source it (tested with 8cf21feefd994dd023b50668e9202b3d888a8c1e) but then `list-keys`' output is truncated and the last `}"` characters are missing.

This is caused by `cmd_list_keys_exec()` using a formatting buffer size `BUFSIZ` allocated on the stack.

On macOS `10.14.6` with Xcode `11.0`, `BUFSIZ` is `1024`
```
$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 11.0.0 (clang-1100.0.33.8)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

For the binding mentioned above, it happens that the number of characters needed to format the escaped command list before calling `cmdq_print(item, "bind-key %s", tmp);` is exactly `1025` characters.

I could have chosen a multiple of `BUFSIZ` but it's only going to postpone the problem.
Instead I used `xrealloc()`. I also did my best to adhere to the coding style but I'm not very familiar writing code in tmux.

### DIFF
--- a/cmd-list-keys.c
+++ b/cmd-list-keys.c
@@ -61,8 +61,9 @@ cmd_list_keys_exec(struct cmd *self, struct cmdq_item *item)
 	struct key_table	*table;
 	struct key_binding	*bd;
 	const char		*tablename, *r;
-	char			*key, *cp, tmp[BUFSIZ];
-	int			 repeat, width, tablewidth, keywidth;
+	char			*key, *cp, *tmp;
+	size_t			cmdlen, tmplen, tmpsize;
+	int			repeat, width, tablewidth, keywidth;
 
 	if (self->entry == &cmd_list_commands_entry)
 		return (cmd_list_keys_commands(self, item));
@@ -101,6 +102,9 @@ cmd_list_keys_exec(struct cmd *self, struct cmdq_item *item)
 		table = key_bindings_next_table(table);
 	}
 
+	tmp = xmalloc(BUFSIZ);
+	tmpsize = BUFSIZ;
+
 	table = key_bindings_first_table ();
 	while (table != NULL) {
 		if (tablename != NULL && strcmp(table->name, tablename) != 0) {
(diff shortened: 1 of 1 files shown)


nicm:
I have applied this to OpenBSD with some changes:
- The table name and key also need to expand the buffer or they could be truncated.
- The size needs to be + 1 because it needs to include the \0.
- Double the size instead of adding to it each time.
- Changed the names of a couple of variables.
It will be in GitHub later on. Thanks!

gpakosz:
Thanks @nicm.

Can this be cherry-picked in 3.0 by chance?

PR #1933, opened 2019-10-10
Fix warnings

mathstuf:
---
The `window-tree` patch might be a bit too cautious for your tastes, but the former is a possibility (without auditing that `value` is never passed in as `NULL`…but then why check it in the code above?).

### DIFF
--- a/options.c
+++ b/options.c
@@ -359,7 +359,7 @@ options_array_set(struct options_entry *o, u_int idx, const char *value,
 {
 	struct options_array_item	*a;
 	char				*new;
-	struct cmd_parse_result		*pr;
+	struct cmd_parse_result		*pr = NULL;
 
 	if (!OPTIONS_IS_ARRAY(o)) {
 		if (cause != NULL)
@@ -408,7 +408,7 @@ options_array_set(struct options_entry *o, u_int idx, const char *value,
 
 	if (OPTIONS_IS_STRING(o))
 		a->value.string = new;
-	else if (OPTIONS_IS_COMMAND(o))
+	else if (OPTIONS_IS_COMMAND(o) && pr != NULL)
 		a->value.cmdlist = pr->cmdlist;
 	return (0);
 }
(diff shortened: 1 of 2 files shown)


nicm:
I don't normally like this kind of change because they are unnecessary and can hide genuine warnings. We are pretty much making the code worse to work around bugs in a particular compiler version.

But options_array_set is a bit confusing so I think we could probably make it better and get rid of the warning at the same time, and I don't think the window-tree.c change can actually cause any confusion.

Please try this and see if it hides the warning in options.c: [x.diff.txt](https://github.com/tmux/tmux/files/3723708/x.diff.txt)

mathstuf:
That does indeed fix the warning in `options.c` and looks cleaner to boot.

> and I don't think the window-tree.c change can actually cause any confusion.

While I tend to agree, the thing is that compilers are pretty decent at pointing out that code isn't that easy to follow (like `options_array_set` here). In that code, there's nothing guarding ensuring that a new enumeration value would actually get all of the necessary `switch` statements updated (well, `-Werror=switch-enum` would, but that's not set here; maybe that would be the better solution?).

There are other warnings still around, but some are unfixable (e.g., passing a non-literal to `strftime` in the formatting code) or is just inherently complex logic (threading through values based on command flag set/unset state). Keeping the warning count low means that the "genuine" warnings don't get lost among the noise.

nicm:
The compiler will warn if an enum value is not used in the switch, that is
easy for it to check.

nicm:
I have applied this to OpenBSD now, it will be in GitHub later. Thanks!

nicm:
BTW -Wswitch is part of -Wall.

format-nonliteral is too useful to turn off even though it means a couple of spurious warnings. It would be possible to hide them with pragma GCC diagnostic push/ignored/pop but that's ugly, and isn't supported on some older GCC versions, and IIRC some compilers will warn about unknown pragmas...

PR #1934, opened 2019-10-11
Fix multiple memory leak issues

ivq:


options.c
```diff
@@ -296,6 +296,7 @@ options_remove(struct options_entry *o)
 	else
 		options_value_free(o, &o->value);
 	RB_REMOVE(options_tree, &oo->tree, o);
+	free((void *)o->name);
```
krader1961: You do not need, and should not include, the cast. The few places in the code which do this cast should remove the cast, IMHO. This cast hasn't been needed for more than two decades.

tmux.c
```diff
@@ -127,6 +127,7 @@ make_label(const char *label, char **cause)
 		free(base);
 		goto fail;
 	}
+    free(base);
```
krader1961: This is a partial fix. Notice the other places that `goto fail;`. This is a situation where it is better to initialize the var to `NULL` where it is defined and then do `if (base) free(base);` in the error path.

options.c
nicm: These are const casts and are necessary.

tmux.c
nicm: Yes that would work, although free(NULL) is always OK so no need for an if guard.

ivq: Note the spaces here. Tab should be used.

nicm:
Applied now, thanks!

PR #1951, opened 2019-10-19
Add formats to get copy-mode cursor position

### LINKED ISSUE #1949: Can't get copy-mode cursor position in shell script
### Issue description

With the following binding:

```tmux
bind -T copy-mode-vi 9 run-shell /tmp/foo.sh
```

And this n `/tmp/foo.sh`:

```sh
#!/usr/bin/env bash

tmux display-message '#{cursor_x} #{cursor_y}'
```

While in copy mode (vi, specifically), the cursor position does not update - it seems to continue to report the cursor position from the time when copy mode was entered.

I've checked the alternate screen positions, and looked through all of the items in FORMATS, and can't find anything that might help me here.

### Required information

Version info:
```
Linux unknown
tmux next-3.1
tmux-256color
```

### Rationale:

In order to avoid the XY problem, I have two use cases I would like this for:

#### Isolating a Whitespace-delimited Word

For my [tmux-plumb plugin](https://github.com/eraserhd/tmux-plumb), I would like to isolate the whitespace-delimited word under the cursor, extract it, then restore the cursor position.  I am able to get the word currently (mostly correctly) using `send-keys -X` and `show-buffer`, but the cursor jumps to after the word.

This use case could alternately be solved by implementing marks.

#### Implementing a Kakoune-like Copy Mode

I would like to write another plugin that makes a copy mode which behaves like [Kakoune](http://kakoune.org/).   Many things are easily doable with `send-keys -X`, but some things will require using `capture-pane`, getting the cursor position, parsing the capture, and then reacting to it.  An example is `<a-i>p`, (select inside paragraph), which selects all lines at the cursor between the preceding and following blank lines. 

I'm willing to do the work, but I'm not sure what the right approach is.  Should `cursor_x` and `cursor_y` be updated in copy mode?  Should there be different variables, e.g. `copy_mode_cursor_x`?  Should a different feature entirely be implemented?

Thank you!

eraserhd:
Closes #1949


eraserhd:
Hmm, looks like I could use the selection coordinates as well.  Let me poke at that a bit later.

(This was pretty easy, @nicm, your pointer was right on, thanks!)

nicm:
Applied now, thanks!

PR #1958, opened 2019-10-31
Implement DECDHL for terminals that support it

FractalBoy:
This merge request attempts to implement DECDHL.

I'm having some trouble clearing it on a new line. DECDHL (top and bottom) should only be in effect for a single line.

Can you offer any pointers?

Note: You must test this in a terminal that supports it, such as PuTTY. Additionally, terminfo doesn't generally seem to include dhtop and dhbot, so you'll need to set terminal-overrides:

```
set -as terminal-overrides ",*:dhtop=\\E#3,*:dhbot=\\E#4"
```


nicm:
Can you be more specific about the problem?

But a larger issue - what is the use of this? What applications use it?

FractalBoy:
This is how DECDHL should behave:

image

This is how it behaves as I've implemented it, so far:

image

Anytime there's a line feed, the DECDHL attribute should be cleared, which is not happening.

The use is purely a nice-to-have. An application I work with uses it. I don't really want to go into further details than that due to potential IP issues.

FractalBoy:
It's a little better now, but still not 100%. Some issues:

1. It doesn't ever respect the bottom part of the text

image

2. Resizing and clicking makes it act strangely. This is what happened after I clicked in the first screenshot:

image

After resizing:

image

nicm:
You clear a bit with x &= ~bit not x &= bit.

FractalBoy:
Oops, typo.

FractalBoy:
Behavior is the same.

nicm:
Your terminal-overrides as the same for both top and bottom.

nicm:
Also in attributes.c you have bottom twice.

FractalBoy:
> Your terminal-overrides has the same for both top and bottom.

That would probably help to be correct. Now it works.

> Also in attributes.c you have bottom twice.

That was fixed in 691bde3f92becfe80fbddae59f7bf7a2dd30bd05

FractalBoy:
I spoke too soon - it was working until I clicked in the window. 

Now it looks like this:

image

nicm:
You still have a lot of work to do here, notably:

1) SGR0 does not clear these attributes so:
- it is not correct for tmux to clear them on SGR0, so you will need to change input_csi_dispatch_sgr;
- tmux can't assume that SGR0 turns them off, so anywhere tmux does SGR0 expecting to reset all attributes will need to change, this is at least tty_stop_tty, tty_reset, tty_invalidate.

2) LF resets, but CR does not - you have both resetting.

3) You will need to be change anywhere where tmux sends LF. This means at tty_cursor, tty_cmd_scrollup. tty_cmd_linefeed should be OK because it is only used after a LF input which should reset.

4) It appears that most (maybe all?) other sequences that move the cursor onto another line also clear these attributes, for example try these:
~~~~
printf '\033[10;1H\033[2J\033#3abc\033[Hdef\n'
printf '\033[10;1H\033[2J\033#3abc\033[Bdef\n'
printf '\033[10;1H\033[2J\033#3abc\033[3;10rdef\n'
printf '\033[10;1H\033[2J\033#3abc\033[2Sdef\n'
~~~~
So similarly to LF you will need to clear these for any of these cursor movement sequences in both input.c and anywhere that tmux itself moves the cursor (mostly it does it via tty_cursor).

5) tmux expects that printing a single width character such as ASCII 'a' will move the cursor to the right one cell. You will need to teach it that for these cells it does not. it may be enough just to put 2 into the data->width member of the cell (as if it was a width 2 UTF-8 character) when it is constructed in input_print but you will need to check that doesn't cause trouble in places that use the cell data but don't apply attributes (copy mode, command prompt, capture-pane).

nicm:
Actually, it appears these sequences are not conventional attributes at all, but rather an attribute of the entire line.

Try this:
~~~~
printf '\033[1;1H\033[2Jabc\033#3\n'
~~~~
The 'abc' is double width even though it was printed before the DECDHL sequence.

So I think these will need to be implemented entirely differently.

FractalBoy:
Is there anything similar already? Should the attribute be set on the grid_line structure instead of the grid_cell?

FractalBoy:
This is particularly interesting to watch:

```
printf 'abc'; sleep 5; printf '\033#3\n'
```

nicm:
Yes you will need a grid line flag. Then you could probably check it and go
into the redraw-entire-line path (tty_draw_pane) for basically everything.

But the difficulty will be if the pane has a different pane to its left or
right, there is no way to only draw part of the terminal width as double.

FractalBoy:
> But the difficulty will be if the pane has a different pane to its left or right, there is no way to only draw part of the terminal width as double.

Yikes, that doesn't seem good. I guess this kind of thing isn't really compatible with panes at all, perhaps it needs to just be turned off if there is a pane to the left or right.

nicm:
Yes I would suggest just displaying double characters like AABBCC if there
is a pane to left or right or if a border needs to be drawn.

TBH I'm not convinced this is worth the complexity.

FractalBoy:
> TBH I'm not convinced this is worth the complexity.

I'm almost convinced as well.

PR #1982, opened 2019-11-23
Remove use of REG_STARTEND

evangreen:
The REG_STARTEND flag is not a part of POSIX standard regular
expressions (see [1]). Avoid using it to maintain tmux's portability to
many operating systems.

Removing it is fairly straightforward. We never used it to limit the
search before the end of the string, we only used it to move search start
offset forward. Instead of passing REG_STARTEND, simply pass string +
start into regexec(). Fix up uses of rm_so and rm_eo to be relative to
start, and Bob's your uncle.

[1] https://pubs.opengroup.org/onlinepubs/009695399/functions/regexec.html

---
Note: I haven't actually tested this patch. Let me know if anything is broken.

### DIFF
--- a/regsub.c
+++ b/regsub.c
@@ -77,10 +77,7 @@ regsub(const char *pattern, const char *with, const char *text, int flags)
 	end = strlen(text);
 
 	while (start <= end) {
-		m[0].rm_so = start;
-		m[0].rm_eo = end;
-
-		if (regexec(&r, text, nitems(m), m, REG_STARTEND) != 0) {
+		if (regexec(&r, text + start, nitems(m), m, 0) != 0) {
 			regsub_copy(&buf, &len, text, start, end);
 			break;
 		}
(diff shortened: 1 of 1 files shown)


nicm:
What platforms don't support this?

You should be able to test it with display-message, something like: "tmux set @foo xyz \; display -p '#{s/z$/q/r:@foo}"

Someone is working on regex searching and if they choose to use REG_STARTEND this might as well stay also so I will leave it for a bit.

ndowens:
https://github.com/tmux/tmux/pull/1986 I proposed an idea for this. Musl does not support REG_STARTEND on their regex.h

nicm:
I will take another look at this soon, I assume this change works for you?

ndowens:
This works for me

> I will take another look at this soon, I assume this change works for you?

Using this patch allows musl to build fine

nicm:

ndowens:
> Applied to OpenBSD now, will be in GitHub later. Thanks!

Thank you much :)

evangreen:
Thank you! And thanks @ndowens for the assist in justification and testing.

PR #1985, opened 2019-11-25
Preserve search highlighting after motions in vi-copy-mode

ericpruitt:
This adds support for preserving highlighting after motions when using vi-copy-mode.


nicm:
Applied now, thanks!

ericpruitt:
Thanks for accepting the change, @nicm. I added a second commit to this pull request because I think "ismotion" might be unset in certain situations in the original code. Or did you see this and the second change isn't required after all?

nicm:
I think you are right, it could be unset if there are marks and the command matches search-* but isn't in the table. It is interesting that the compiler doesn't see this. I think just setting it to 0 when it is declared is enough though.

dewijones92:
I can't see the merge in to master. Is there a reason it hasn't? Thanks :)

ericpruitt:
It's already there: https://github.com/tmux/tmux/commit/c2fde58701d3d0a55ba15ba6004df7ca009175b9

dewijones92:
thanks!! may I ask how I would go about finding that in future?
It wasn't the least apparent to me from this pull request that master included your commits :(

ericpruitt:
You could've looked through commit history for the changes applied around the day this issue was closed.

dewijones92:
Thanks mate. Will do that next time :D

PR #1986, opened 2019-11-27
Fix musl build

ndowens:
Musl's regex.h does not have the needed define of: REG_STARTEND
While, pcre's pcreposix.h gives the definition that is needed.

Though #1982 would be best bet IMO
Signed-off-by: Nathan Owens <ndowens04@gmail.com>

### DIFF
--- a/regsub.c
+++ b/regsub.c
@@ -17,8 +17,11 @@
  */
 
 #include <sys/types.h>
-
+#if defined(__GLIBC__)
 #include <regex.h>
+#else
+#include <pcreposix.h>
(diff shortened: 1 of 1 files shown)


nicm:
No this is not a fix, this will just break it - you might as well just define it to 0. Using PCRE would work but that also means linking against the libraries by changing configure.ac/Makefile.am.

nicm:
I think the other patch is a better solution if we don't need REG_STARTEND, I don't think we need this issue as well so I will close it. Thanks!

PR #1988, opened 2019-11-27
Fix bounds check in layout_parse overflowing to max u_int.

jshap70:
See https://github.com/tmux-plugins/tmux-resurrect/issues/316 for full debug logs of the bug. Basically `layout_parse()` would try to correctly alter the layout to match what is currently available, however it would incorrectly attempt to subtract past 0 to max u_int by not checking bounds. 

Thanks to @DarkKnight288 for finding the issue, and if they wish to submit an MR directly I will gladly let them have it lol.

### DIFF
--- a/layout-custom.c
+++ b/layout-custom.c
@@ -211,9 +211,10 @@ layout_parse(struct window *w, const char *layout)
 		break;
 	}
 	if (lc->type != LAYOUT_WINDOWPANE && (lc->sx != sx || lc->sy != sy)) {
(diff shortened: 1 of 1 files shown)


nicm:
Show full logs please.

nicm:
Oh never mind, this was already fixed after 3.0 in 2cb268d51b71d74cf32e9cd1f67892681a9563e1, that's why I couldn't reproduce.

jshap70:
ah, good catch. guess I'll change positions to proposing a 3.0.1 release... :)

nicm:
Will see what else comes up in the next few days.

Issue #1999, opened 2019-12-02
Solaris does not define _PATH_DEFPATH in paths.h


ericvw:
When attempting to upgrade  `tmux >= 3.0` on Solaris 10, [`spawn.c`](https://github.com/tmux/tmux/blob/master/spawn.c) fails to compile due to `_PATH_DEFPATH` being undefined.

Specifically, the code in question is: https://github.com/tmux/tmux/blob/edf96b06a5aede19e796ace18b4915d973e96e55/spawn.c#L311-L312

From Solaris's [`man paths.h(3HEAD)`](https://docs.oracle.com/cd/E86824_01/html/E54772/paths.h-3head.html), `_PATH_DEFPATH` is not defined.  When `grep`ping through `/usr/include`, I have been unable to find an alternative define that provides something similar to `/usr/bin:/bin`.

Environment:
```
$ uname -sp
SunOS sparc
```

Relevant compilation output:
```
[...]
source='spawn.c' object='spawn.o' libtool=no \
        DEPDIR=.deps depmode=none /bin/sh ./etc/depcomp \
        /scrubbedprefix/bin/cc -DPACKAGE_NAME=\"tmux\" -DPACKAGE_TARNAME=\"tmux\" -DPACKAGE_VERSION=\"3.0a\" -DPACKAGE_STRING=\"tmux\ 3.0a\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"tmux\" -DVERSION=\"3.0a\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DH
AVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -D__EXTENSIONS__=1 -D_ALL_SOURCE=1 -D_GNU_SOURCE=1 -D_POSIX_PTHREAD_SEMANTICS=1 -D_TANDEM_SOURCE=1 -DHAVE_DIRENT_H=1 -DHAVE_FCNTL_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_PATHS_H=1 -DHAVE_STDINT_H
[...]

PR #2001, opened 2019-12-03
Check and define '_PATH_*' macros individually

ericvw:
Some systems (e.g., Solaris) don't define `_PATH_DEFPATH`, even though
the `paths.h` header exists.  Checking each `_PATH_*` individually is
more resilient by ensuring that necessary macros are defined during
compilation.

Closes: #1999


nicm:
Great, applied, thanks!

PR #2020, opened 2019-12-16
Define FNM_CASEFOLD to 0 for compatibility

ericvw:
Some systems (e.g., AIX) don't define `FNM_CASEFOLD` nor support
ignoring case.  To avoid flag value collisions, fallback to setting the
flag to 0 so it doesn't do anything in case the system may define it in
the future.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -21,6 +21,7 @@
 #include <sys/ioctl.h>
 #include <sys/uio.h>
 
+#include <fnmatch.h>
 #include <limits.h>
 #include <stdio.h>
 #include <termios.h>
@@ -167,6 +168,10 @@ void	warnx(const char *, ...);
 
 #ifndef O_DIRECTORY
 #define O_DIRECTORY 0
+#endif
+
+#ifndef FNM_CASEFOLD
+#define FNM_CASEFOLD 0
 #endif
 
 #ifndef INFTIM


nicm:
You can't define if to 0x10 if it isn't supported, that could do anything. It has to be 0.

nicm:
What platforms use FNM_IGNORECASE?

compat.h
```diff
@@ -169,6 +170,14 @@ void	warnx(const char *, ...);
 #define O_DIRECTORY 0
 #endif
 
+#ifndef FNM_CASEFOLD
+#ifdef FNM_IGNORECASE
+#define FNM_CASEFOLD FNM_IGNORECASE
+#else
+#define FNM_CASEFOLD 0x10
```
ericvw: I'm not too keen on this but noticed that `libtar` does something similar (see [libtar:compat/compat.h](https://github.com/tklauser/libtar/blob/6379b5d2ae777dad576aeae70566740670057821/compat/compat.h#L46)).

ericvw:
> You can't define it to 0x10 if it isn't supported, that could do anything. It has to be 0.

I can change it to 0.

> What platforms use FNM_IGNORECASE?

Solaris does as an alias (https://docs.oracle.com/cd/E36784_01/html/E36873/fnmatch-3head.html).

ericvw:
@nicm let me know if you would like me to take an entirely different approach.

My main goal is to get >= 3.0a  working on AIX, which does not define `FNM_CASEFOLD` nor has an option to ignore case.

ericvw:
Given that Solaris does also define `FNMATCH_CASEFOLD`, I'll just default the value to 0 if it is not detected.

ericvw:
I've updated the PR to just fallback defining `FNM_CASEFOLD` to 0 for compatibility.

nicm:
If none of the platforms tmux supports have IGNORECASE but not CASEFOLD, there is no need to add it.

nicm:
Applied now, thanks!

jperkin:
Newer versions of closed source Solaris do have `FNM_CASEFOLD` for compatibility, but older versions of Solaris and illumos platforms do not, failing here recently on SmartOS:

https://us-east.manta.joyent.com/pkgsrc/public/reports/trunk/x86_64/20200119.1207/tmux-3.0anb1/build.log

While newer versions now build ok, the case insensitive functionality is now missing on illumos platforms, so I would recommend using this instead:

```c
#ifndef FNM_CASEFOLD
#ifdef FNM_IGNORECASE
#define FNM_CASEFOLD FNM_IGNORECASE
#else
#define FNM_CASEFOLD 0
#endif
#endif
```

I'll be patching pkgsrc with this in any case.

PR #2030, opened 2019-12-24
Remove reference to non-existent option in tmux.1

sluidfoe:
The first paragraph of the STYLES section in the tmux manpage refers to
"status-left-format," which appears nowhere else in the manpage nor in
the tmux source.


nicm:
Good spot, fixed in OpenBSD now, will be in GitHub later. Thanks!

Issue #2026, opened 2019-12-20
[Feature Request] More format variables for viewing attached clients


tculp:
### Premise

Lately, I have found myself collaborating with multiple clients and sessions attached to the same session group. Since there are many people viewing the same group of windows, it would be nice to know which how many clients are viewing the current window and session group. This information could be part of the window list in the status bar, but format options do not exist for this.

There are several format options that perform similar features, but do not quite fulfill my desire. For example, there is a session_attached variable, to show the number of clients attached to the session, but no session_group_attached variable. Similarly, there is a window_active variable, but no window_attached variable.

Currently, in order to achieve what I would like, I would most likely have to make a script to call many tmux commands and perform several math operations on them.

### Proposal

I think it would make sense for some of the session variables to also be useable for session groups and windows, such as:

     session_group_attached   -   Number of clients session group is attached to
     session_group_many_attached   -   1 if multiple clients attached
     session_attached_list   -   List of clients attached to session
     session_group_attached_list   -   List of clients attached to session group
     window_active_count   -   Number of clients viewing the window
     window_linked_count   -   Number of sessions (or session groups) window is linked to
     window_linked_list   -   List of sessions (or session groups) the window is linked to
     window_active_list   -   List of clients viewing the window
     
Thoughts?

nicm:
OK sure. Anything with a loop needs to be a callback format though (look at the others in format.c).

PR #2034, opened 2019-12-25
Added several format variables for sessions, session groups, and windows

tculp:
session_attached_list - List of clients session is attached to
session_group_attached - Number of clients sessions in group are attached to
session_group_attached_list - List of clients sessions in group are attached to
session_group_many_attached - 1 if multiple clients attached to sessions in group
window_active_clients - The list of clients viewing this window
window_active_clients_count - The number of clients viewing this window
window_active_sessions - The list of sessions on which this window is active
window_active_sessions_count - The number of sessions on which this window is active
window_linked_sessions - The list of sessions this window is linked to (incl. those in session groups)
window_linked_sessions_count - The number of sessions this window is linked to (incl. those in session groups)

Satisfies #2026


nicm:
Excellent -  I have applied this to OpenBSD now, however I have changed the names slightly so that those formats which return a list are called xyz_list (such as window_active_sessions_list) and those that return a count have no suffix (such as window_active_sessions). I think this is more consistent with the existing formats. Let me know if you see any problems. It will be in GitHub when it syncs up in a few hours. Thanks!

PR #2046, opened 2020-01-02
Make move-pane behave like join-pane with marked panes

davidegirardi:
The man says that `move-pane` is "Like join-pane, but src-pane and dst-pane may belong to the same window.".

I actually did not when trying to move a marked pane. This PR fixes it.


nicm:

PR #2051, opened 2020-01-07
Add a highlight-search option

DylanZA:
Highlight-search can make searching unusable on really large buffers. This is due to it trying to look for all occurrences of the result in the whole buffer, which can be very slow. We cannot simply highlight the occurrences in the current view as that requires updating the search as the view changes and while this is possible seemed very complex.

To repro:
set -g history-limit 300000

then run:
$ seq 300000

Searching through this buffer is very sluggish due to the search algorithm.

With highlight-search set to 0 this works much better.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -614,6 +614,12 @@ const struct options_table_entry options_table[] = {
 	  .default_num = 1
 	},
 
+	{ .name = "highlight-search",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_WINDOW,
+	  .default_num = 1
+	},
+
 	{ .name = "main-pane-height",
 	  .type = OPTIONS_TABLE_NUMBER,
 	  .scope = OPTIONS_TABLE_WINDOW,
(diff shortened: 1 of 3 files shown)


nicm:
This came up before and I don't want to do it like this. It either needs to:

- search on demand rather than all at once (I don't mind losing the match
counter).

- search faster.

- have an adaptive limit on the number of matches or (better) the time a
search can run for.

DylanZA:
fair enough.

> - search on demand rather than all at once (I don't mind losing the match
counter).

This is probably the best if you are happy to lose the match counter. Would you be ok with that globally (as in no flags)? Then it could simply search the current window for highlighting and not any further which is at least bounded.


> - search faster.
I don't think this is doable/easily doable.


> - have an adaptive limit on the number of matches or (better) the time a
search can run for.

A bit tricky as then you would have different behaviors randomly (like the search count). I think the first approach is best but would be a UX change

nicm:
Yes, on demand would be great, it needs to work automatically, no new options. Searching just the visible lines should be pretty quick so it could just search the whole lot every time it scrolled (or when you searched again on the top match). I don't mind removing the match counter.

DylanZA:
ok - will try and do this when I get some more time

PR #2067, opened 2020-01-24
Alias FNM_CASEFOLD to FNM_IGNORECASE for Illumos and Solaris

ericvw:
Improve compatibility for case insensitive functionality for older
versions of Solaris and Illumos systems.

CC: @jperkin, thanks for https://github.com/tmux/tmux/pull/2020#issuecomment-577610179.  I didn't know enough about older versions of Solaris nor Illumos.  TIL!


nicm:
Applied now, thanks!

PR #2068, opened 2020-01-24
Document `-d` flag for swap-window

fluffywaffles:
Links to the similar option for break-pane.


nicm:
Applied to OpenBSD now in a slightly different form, will be in GitHub later. Thanks!

PR #2075, opened 2020-02-04
Select_word change for single char words in copy-mode-vi

mckellyln:
@nicm do you mean like this or something completely different ?

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -120,7 +120,7 @@ static void	window_copy_cursor_jump_to_back(struct window_mode_entry *);
 static void	window_copy_cursor_next_word(struct window_mode_entry *,
 		    const char *);
 static void	window_copy_cursor_next_word_end(struct window_mode_entry *,
-		    const char *);
+		    const char *, int);
 static void	window_copy_cursor_previous_word(struct window_mode_entry *,
 		    const char *, int);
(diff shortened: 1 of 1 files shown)


nicm:
No I don't get what you are doing here?

I meant that you can detect a one character word by looking at the cell
before the cursor and seeing if it is a separator character.


How do you reproduce your problem? select-word seems to work OK for one
character words for me.

mckellyln:
If I put the cursor on a one char word in copy-mode-vi and issue a send -X select-word it does not work properly for me, it selects the one char word and the next word after it.

nicm:
Ah yes, I see. I think you will need to check the cell to the left and not move an extra cell forward if it is whitespace.

mckellyln:
I thought instead of duplicating code from next_word or next_word_end, I would just not move right one if called from select_word (if !from_sel_word):
```
	if (keys == MODEKEY_VI && !window_copy_in_set(wme, px, py, separators) && !from_sel_word)
		px++;
```

mckellyln:
I moved around it with send -l lbvhe instead.  I can close or re-factor, whichever you like.
thx
-m

mckellyln:
ok, I'll close this and try a new PR with a different approach.

PR #2080, opened 2020-02-10
Select_word change for single char word in copy-mode-vi

mckellyln:
@nicm here is another way for the single char select_word.  Do you think it makes sense ?


nicm:
I was going to take another look at your other change but I haven't had time yet, I'll take a look at this one as well at some point. Thanks!

mckellyln:
OK, thanks.  The previous PR is at: https://github.com/tmux/tmux/pull/2075
Or suggest some other direction.

nicm:
I have applied this to OpenBSD now, will be in GitHub later. Thanks!

PR #2086, opened 2020-02-20
Add selection_active format - like selection_present but additionally…

mckellyln:
… tells if selection is in sync with cursor or stopped
@nicm what do you think ?
thx,
-m


nicm:

Issue #2028, opened 2019-12-23
Pane border side styles are not configurable using inline styles in pane-border-format


tculp:
### Issue description

Currently, there seems to be no way to configure the style of the outside area of pane borders using inline styles in the pane-border-format variable. This also means there is no way to set the style dynamically based on variables.

My test tmux.conf:

```
set -g pane-border-status bottom
set -g pane-border-style ""
set -g pane-active-border-style ""
set -g pane-border-format "#[#{?#{pane_active},bg=red fill=red,bg=blue fill=blue}]"
```

A screenshot of the result is attached.

<img width="1282" alt="Screen Shot 2019-12-22 at 10 30 24 PM" src="https://user-images.githubusercontent.com/19711877/71334242-4b838580-250b-11ea-9ac9-805503dc24ee.png">

As you can see, the vertical bars are not styled according to the inline style, as well as the right and left side of the bottom-bars.

### Required information

Please provide the following information:

* tmux version (`tmux -V`).
tmux 3.0a
* Platform (`uname -sp`).
Darwin i386
* $TERM inside and outside of tmux (`echo $TERM`).
inside: screen-256color
outside: xterm-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`).
no logs

nicm:
Yes I know, how would it look though? I suppose we could have an "indent" member of the style that controls how many cells left of the pane edge it draws the border. Or always draw it at 0 and have some way to specify border characters in the format.

tculp:
Well since the bottom status bar always draws with 0 implicit margin on both sides, I think it would make sense if the pane border also didn't have any horizontal margin.

I'm not sure of how best to do the sides in code. From a user's pov, maybe the background / foreground of the status would just carry over to the sides, or maybe it would require a push-default. Any ideas?

Note: I just tested changing the border indent and the default style looked better with the indent in place (green text on black background), but my style (with an orange/blue background color) looked fine without the indent. Maybe it would be fine for the indent to be an option, something like pane-border-indent, defaulted to 2.

nicm:
We are moving away from having options for every little thing, so I don't want pane-border-indent, it needs to be expressed in the style, which would also allow it to be used in other contexts.

I don't think it would be very intuitive to make it work out the colour from the format.

I think it would be better to have a style like "fill" (perhaps "indent", or "padding" might be better if it was doing both sides), then the default pane-border-style could be "padding=2".

[...]

PR #2097, opened 2020-03-03
WIP: Support format variables in styles

tculp:
Related to https://github.com/tmux/tmux/issues/2028


nicm:
Some of this looks good and some seems unnecessary.

I think if you are going to store all styles as strings, even if they don't include a format, then it would be best just to get rid of OPTIONS_TABLE_STYLE, options_get_style, options_set_style and so on and just make all of the style options string options instead.

That means there would be no validation when a style option is set, but I think we just have to sacrifice that since there is no practical way to validate styles containing formats without the context where the format will be used later.

Then style_apply can take a format_tree and do the expansion, similar to how you have options_get_parsed_style. As a first step you could just make all of the callers pass in a NULL format, although I see you have done them already.

Parsing styles is expensive so if the style string does not contain a format, I think we should set a flag and cache the result in the struct style.

tculp:
For the sake of cached results, it would probably make sense to keep the OPTIONS_TABLE_STYLE so we can keep track of the cache flag and parsed value. Then options_set_style would do the parsing to determine if there's a format, and setting the flag and cached value appropriately.

nicm:
I don't think that is necessary because style_apply will never be called on
non-style options, there is no need to know the type separate from that.

tculp:
True, but somewhere will need to hold the logic of whether to parse the string or just use a cached value

nicm:
style_apply can do that, it can become something like:
~~~~
void
style_apply(struct grid_cell *gc, struct options *oo, const char *name)
{
    struct options_entry *oe;

    oe = options_get(oo, name);
    if (oe->style_is_cached)
        return (&oe->cached_style.gc);
    /* expand format yadda yadda */
    if (style_parse(&oe->cached_style, ...) != 0)
        return (&grid_default_cell);
    /* maybe this, or maybe something else like strcmp the original form with the expanded one */
    oo->style_is_cached = strstr(s, "#{");
    return (&oe->cached_style.gc);
}
~~~~
Obviously you may need to pass in more arguments or whatever.

nicm:
And reaching into options_entry is a no-no now, so may need some accessors or a helper function to do it in options.c.

tculp:
That would require ALL string options to have the flag and field for the cached value, correct? That seems weird to me.

nicm:
Yes that is fine, there are only a few hundred options it is not a big cost
and it will make the code simpler.

nicm:
It would have to go into the options_entry anyway because there is nothing else you can get out of the options tree, it is the only place you can cache the style.

tculp:
Even if style_apply is doing the logic for pulling the cached string or not, the logic of setting the option to a cached value or not still needs to be somewhere else, and the option still needs to be marked as a style and not a regular string.

nicm:
What does it need to do somewhere else? Why can't style_apply set the
cached flag, like in the example function I gave?

The only extra thing is to clear the cached flag when the option value
changes and that can be done for all options no matter the type, the next
call to style_apply will set the flag again if necessary.

It is much simpler just to make all of them strings, only style_apply needs
to care they are styles.

tculp:
So you mean only check whether it is static or dynamic when it is first applied, not when the option is set? That would prevent us from checking whether it is a valid style at all, as opposed to being able to check when it is set, if it is static

nicm:
Right that is what I said earlier:

> That means there would be no validation when a style option is set, but I think we just have to sacrifice that since there is no practical way to validate styles containing formats without the context where the format will be used later.

tculp:
I was just thinking that we could still do it on static styles, just not dynamic ones. But if you think the validation wouldn't be worthwhile for static styles either, then alright

nicm:
I don't think it is worth the extra complexity, if we can't validate
them all then validating none and having the code be simpler is better.

tculp:
Okay

nicm:
Hi - are you still working on this?

nicm:
Let me know because if you aren't I will take a look at it :-).

nicm:
set `-a` is a bit of a problem, perhaps.

nicm:
Here is a WIP diff with the basics but without all the style_apply calls changed to pass a format and without the special cases (`status-fg` and `status-bg` and some other bits) sorted out: [x.diff.txt](https://github.com/tmux/tmux/files/4539868/x.diff.txt)

nicm:
Updated diff, have committed some bits: [y.diff.txt](https://github.com/tmux/tmux/files/4540015/y.diff.txt)

There is quite a bit of work to make this useful, for example with `pane-border-style` the styles are set up before any borders are drawn, but they need to be changed so they are expanded for each pane.

nicm:
This could be used to let one option like `mode-style` be changed in different places. Nobody has really asked for that much though...

Although looking back at your original issue I'm not sure if it will really solve your problem, perhaps we need other changes as well.

nicm:
Applied some more bits: [z.diff.txt](https://github.com/tmux/tmux/files/4540147/z.diff.txt)

nicm:
Here is a diff  [zz.diff.txt](https://github.com/tmux/tmux/files/4545215/zz.diff.txt) with the various special cases fixed, however it doesn't expand formats for any styles yet because mostly they need further changes to be useful:

- OSC 10 and 11 no longer set the `window-style` option, instead they store the colours in the pane and that is used as the default that can be overridden by the option. This has to happen because appending to the option would potentially mean it could become a huge string.

- `status-fg` and `status-bg` now overrides `status-style` instead of playing games with trying to keep them in sync.

nicm:
Have applied this now, needs further work for each call to `style_apply` to pass a useful format.

nicm:
24316bed495781dd6ae3105a5e34d582f02c31e9 allows formats to be given in `pane-border-style` and `pane-active-border-style` which are useful for some things, for example:
~~~~
set -g pane-border-style '#{?synchronize-panes,bg=yellow,default}'
set -g pane-active-border-style '#{E:pane-border-style} fg=green'
~~~~
But they are not particularly useful for changing individual pane borders, because almost every border is shared by at least two panes - one of them has to win and have its style used but tmux has no way to tell which one that should be (the marked pane reverse attribute works because tmux knows the marked pane should win).

nicm:
I have changed the default `pane-active-border-style`:
~~~~
-         .default_str = "fg=green",
+         .default_str = "#{?pane_in_mode,fg=yellow,#{?synchronize-panes,fg=red,,fg=green}}",

nicm:
I am going to work on adding formats to the other `style_apply` calls later. I am closing this issue for the time being.

Issue #2102, opened 2020-03-06
[Feature Request] Support more operations in format strings


tculp:
### Premise

It would be useful to perform math operations in format strings, such as modulus, addition, multiplication, etc.

### Background

I wish to have my windows in the window-list alternate colors to be able to easily differentiate where they start and stop. This would generally be done with a comparison of 0 and the modulus of the window index / 2 (assuming renumber-windows is set to on). However, there is currently no obvious way within the format system to perform modulus or any other math operation.

nicm:
I'm OK with adding more, it shouldn't be too difficult - take a look at format.c:format_build_modifiers and format.c:format_replace.

tculp:
What do you think would be best for the modifiers? Here are my proposed options:

#{M:equation}  :   The most versatile and understandable option, where everything after an M: (for Math) can be parsed as an entire equation. This is my personal favorite option, due to it's versatility and only using up a single format modifier.

M+ M- M* M/ M%   :   Similar to the above, however this greatly limits the math functionality to a single operation each, and each type of operation would use up a format modifier.

+ - * / %   :   The only issue is / is used by replacement.

++ -- ** // %%   :   A little odd, but doesn't collide with the other format modifiers.

nicm:
I think the second is better, for the first you would need either to make
an infix parser which is going to be a fair bit of extra code, or make it
postfix which is annoying for normal people to use.

The second option can use the existing parsing code.

You could make it one format letter followed  by an operator as the first
argument, so eg:

#{e|+:#{window_index},1}

Or:

#{e|%:#{status-left-length},7}

I would probably make the default integers, if you also want to support FP
then you could add a second argument flags with one to specify an FP result
eg #{e|/|f:...}.

PR #2106, opened 2020-03-09
Added math operations to format strings

tculp:
Satisfies #2102

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ configure
 tmux.1.*
 *.dSYM
 cmd-parse.c
+.vscode
+*.log
--- a/format.c
+++ b/format.c
@@ -73,9 +73,17 @@ struct format_job {
 	RB_ENTRY(format_job)	 entry;
 };
 
-union math_number {
+union format_math_number {
 	long long	l;
-	float		f;
+	double		d;
+};
+
+enum format_math_operator {
+	ADD,
+	SUBTRACT,
+	MULTIPLY,
+	DIVIDE,
+	MODULUS,
 };
 
 /* Format job tree. */
@@ -109,14 +117,6 @@ format_job_cmp(struct format_job *fj1, struct format_job *fj2)
 
 /* Limit on recursion. */
 #define FORMAT_LOOP_LIMIT 10
-
-enum math_operator {
-	ADD,
-	SUBTRACT,
-	MULTIPLY,
-	DIVIDE,
-	MODULUS,
-};
 
 /* Entry in format tree. */
 struct format_entry {
@@ -1819,16 +1819,17 @@ format_replace(struct format_tree *ft, const char *key, size_t keylen,
     char **buf, size_t *len, size_t *off)
 {
 	struct window_pane	 *wp = ft->wp;
-	const char		 *errptr, *copy, *cp, *marker = NULL;
-	char			 *copy0, *condition, *found, *new, *end_character;
-	char			 *value, *left, *right;
-	size_t			  valuelen;
-	int			  modifiers = 0, limit = 0, width = 0, j, buffer_size;
-	struct format_modifier   *list, *fm, *cmp = NULL, *search = NULL, *math_expression = NULL;
+	const char		*errptr, *copy, *cp, *marker = NULL;
+	char			*copy0, *condition, *found, *new, *endch;
+	char			*value, *left, *right;
+	size_t			valuelen;
+	int				modifiers = 0, limit = 0, width = 0, j;
+	struct format_modifier	*list, *fm, *cmp = NULL, *search = NULL, 
+							*mathexp = NULL;
 	struct format_modifier	**sub = NULL;
-	u_int			  i, count, use_floating_point_numbers = 0, nsub = 0, math_precision = 2;
-	enum math_operator operator;
-	union math_number math_left_num, math_right_num, math_result;
+	u_int			  		i, count, use_fp = 0, nsub = 0, mathprec = 2;
+	enum format_math_operator 		operator;
+	union format_math_number 		mathleft, mathright, mathresult;
 
 	/* Make a copy of the key. */
 	copy = copy0 = xstrndup(key, keylen);
@@ -1882,7 +1883,7 @@ format_replace(struct format_tree *ft, const char *key, size_t keylen,
 			case 'e':
 				if (fm->argc < 1)
 					break;
-			    math_expression = fm;
+			    mathexp = fm;
 				break;
 			case 'l':
 				modifiers |= FORMAT_LITERAL;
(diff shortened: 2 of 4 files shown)


nicm:
Thanks - looks good, some comments:

- There are only disadvantages to using float here, please use double.

- Please make the additions to format_replace (inside "if (math_expression != NULL)") a new function ("format_replace_equation"? I don't know.)

- snprintf(NULL, 0, ...) isn't portable. You don't need it here anyway, because you can use xasprintf instead and get rid of the .

- If you did have to use them, use xmalloc and xsnprintf please so you don't do anything random if they fail and don't have to check the return value either.

- It is odd to use switch for a simple 0/1 choice, just use if.

- math_number and math_operator should have a format_ prefix like everything else in the file.

- tmux code style:
  - C style comments only please, no //.
  - Space after `switch` etc so `switch (foo)`.
  - Please keep line lengths to 80 characters or less.
  - No {} for single line if/else/for, only if multiple statements inside.

- I think your variable names are far too long... eg I'd change end_character -> endch, buffer_size -> bufsize, use_floating_point_numbers -> usefp, math_expression -> exp or mexp or mathexp, math_precision -> prec or mprec or mathprec. This isn't Java ;-).

- I'm not tied to "e" as the modifier, it was just a suggestion, if you prefer something else then go for it.

nicm:
> snprintf(NULL, 0, ...) isn't portable.

Actually this might be a lie, but anyway you don't need it.

tculp:
It's interesting that fmod isn't available with the libraries travis is using

tculp:
Thanks for the suggestions, I'll work on it

nicm:
A couple of other things:

- Why do you look over the argv? Why not just check for argc == 1 or 2 or 3 and then look at argv[0] for the op, argv[1] for the flags and argv[2] for the precision? Obviously make sure argc <= 3 when you find 'e'.

- Please use strchr not strcmp for the flags - if we add more flags later we want older versions to ignore them.

I'm not sure if yoy could avoid having both int and double to use different arithmetic, maybe you could just use double throughout and then convert to an int if necessary. Not sure if that would always be the same though, I guess it is easy enough to have them separate.

nicm:
fmod probably needs tmux to link against `libm` on some platforms.

nicm:
libm is always available, so don't worry about using fmod - just some platforms roll it into libc, so you don't need an explicit `-lm`.

tculp:
I personally prefer the longer names. I have had difficulty in other parts of the code due to cryptic variable names (like cx), though none of your suggestions are like that, so I'll change them if you wish.

I loop over argv because I didn't see a need to force the order of the flags, and if we add future flags to the expression they won't have to be in a specific order. Though I see strchr may also serve that purpose as well.

I didn't see any other way of cleanly handling the float/int math without duplication and ensuring the result is always correct.

Should I do anything to fix the Travis tests? I assume that would be somewhere in the makefile.

nicm:
Flags should all be in one argument not in multiple, so it matches the
other modifiers. If we don't enforce order of arguments it is hard to add
new arguments, eg if we wanted to have two number arguments we wouldn't be
able to tell which is which.

Yes please change variable names.

You could just store the number as double and cast it eg (long long)a /
(long long)b in the integer case, but I don't mind either way,  union is
fine.

nicm:
-lm will go in Makefile.am, probably in LIBS but I can add it so don't
worry about it if you want.

tculp:
I'll leave the makefile change to you, since I can't easily test it on other systems.

As for the flags, you made the example (in the feature request) of #{e|/|f:...}, which is using multiple arguments, correct? Do you think another syntax would be better?

nicm:
Multiple arguments yes but one for the operator, one for flags, one for
precision. So if we had flags a, b, c it would be eg #{e|+|abc|3}. Each
flag is a single character.

tculp:
Ah, gotcha.

tculp:
What do you think of the behavior of, if floating point numbers are not being used, just flooring the input numbers and result?

nicm:
I don't mind, I'm OK with that or with separate codepaths for fp and !fp, whichever you like.

nicm:
OK, looks good, thanks. I like this a lot. Please take a look at this (diff against clean tree):

[x.diff.txt](https://github.com/tmux/tmux/files/4314288/x.diff.txt)

This has some mostly cosmetic (rather than with fundamental) changes:

- format_math_operator can be function local.

- Move the entire new function before format_replace so it doesn't need a forward declaration, also make it static so it doesn't need to be declared in tmux.h.

- Get rid of the math prefix on most things and make the log and error messages more consistent. "math equation" seems redundant.

- Instead of separate logging for double/integer, just set precision to 0 for integer by default and use %.*f for logging as well.

- Use strtod for doubles, not strtof. Also check it parsed all of the number.

- Free right/left on failure as well - there are "goto fail" after they are allocated.

- Comment, indentation style.

Two things I'm not sure of:

- It isn't really an equation, maybe we should call it something else. Expression? I don't think it matters.

- I forgot that % will need to be escaped as %%. That's a pain. I think we should change it to something else? We aren't tied to symbols so maybe just `m` for modulus?

Other than that this seems good.

tculp:
I do actually prefer expression, as demonstrated by the fact that I kept accidentally typing it instead of equation.

Since I shouldn't free uninitialized left and right, would it be acceptable to have something like
`	freefail:
	free(right);
	free(left);

	fail:
	format_log(ft, "math expression syntax error: %s", copy);
	return (NULL);` and use `goto freefail;` after allocation, and `goto fail;` before?

Also, what are your thoughts on keeping `%` for modulus, but adding `m` as an alternative option? If you don't like that idea, then I'll just make it `m`. Although, apparently c `%` isn't technically the modulus, but the remainder, so maybe `r` would be better.

nicm:
Please see this which adds some of your later changes as well: 

[y.diff.txt](https://github.com/tmux/tmux/files/4316327/y.diff.txt)

I am happy to have `m` or `r` as modulus as well as `%`.

tculp:
I like that, and mixed in several of your changes

nicm:
OK I have applied this to OpenBSD now with a last couple of changes, most notably:

- Instead of dancing around with floor and ceil, just cast to long long and let the compiler do the truncation.

- The result also needs to be truncated, otherwise something like `#{e|/|115|-4}` (which is -28.75) will be -29 rather than -28 when printed with `%.0f`.

This should be in GitHub in the next few hours, let me know if you see any problems.

If you want to do anything further, it would be nice to: 1) have a regression test for these to go in regress/ 2) have some text to add to https://github.com/tmux/tmux/wiki/Formats.

Thanks for working on this!

tculp:
Sounds good! Thanks for your help!

PR #2112, opened 2020-03-13
Init and reset selflag on stop-selection and clear-selection

mckellyln:
@nicm fyr


topcat001:
I think this works, thanks for reporting the issue! I'll do some more testing on my side.

Btw I found another small issue which fortunately has an easy fix. I'll make a new pull request for it later today.

nicm:
Applied now, thanks!

PR #2123, opened 2020-03-18
Clear selection before search

mckellyln:
@nicm fyr

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -2626,6 +2626,8 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex)
 	u_int				 fx, fy, endline;
 	int				 wrapflag, cis, found;
 
+	window_copy_clear_selection(wme);
+
 	free(wp->searchstr);
 	wp->searchstr = xstrdup(data->searchstr);
 


nicm:
Why would we want to do this?

mckellyln:
If I select a word or line and then just happen to search (/) it highlights the found search string but without this change also leaves my previous selection (totally unrelated to search) highlighted.
It seemed to make more sense to me to clear previous selections so what is highlighted by search is only from search.

nicm:
I don't think this is a good idea, I might want to keep the selection, or be searching while selecting to extend the selection. You can always bind your search keys to clear the selection before searching.

mckellyln:
ok, no problem.  Closing.
Can I bind '/' to prompt for input, then clear-selection, then search ?

mckellyln:
I agree if wanting to highlight until next found item then don't want to clear.

mckellyln:
I think the best answer, at least me, is if the selection is active then don't clear - and search can extend it.  But if the selection is not active (ie was stopped previously) then it should clear.
If I can figure out a binding for the search prompt I'll add this to my config.

mckellyln:
Just a final note, something like this probably works for me:
```
bind-key -T copy-mode-vi '/' command-prompt -p 'search/:' 'if -F "#{selection_active}" "run \"tmux copy-mode ; tmux send -X search-forward %1"\" "run \"tmux copy-mode ; tmux send -X clear-selection ; tmux send -X search-forward %1\""'
```

PR #2131, opened 2020-03-23
Meson port

Salamandar:
As autotools is a bit shady to use, and it's difficult to have *all* required m4 definitions, I found easier to just port Tmux to the Meson build system.
Here it is, feel free to comment/merge.
I also have a branch that is based on `3.1-rc3` instead of master.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -106,7 +106,7 @@ void	warnx(const char *, ...);
 #include "compat/queue.h"
 #endif
 
-#ifdef HAVE_TREE_H
+#if (defined HAVE_TREE_H) || (defined HAVE_SYS_TREE_H)
 #include <sys/tree.h>
 #else
 #include "compat/tree.h"
--- a/compat/meson.build
+++ b/compat/meson.build
@@ -1,0 +1,39 @@
+compat_functions += [
+    'asprintf',
+    'cfmakeraw',
+    'closefrom',
+    'daemon',
+    'explicit_bzero',
+    'fgetln',
+    'freezero',
+    'getdtablecount',
+    'getprogname',
+    'memmem',
+    'reallocarray',
+    'recallocarray',
+    'setenv',
+    'setproctitle',
+    'strcasestr',
+    'strlcat',
+    'strlcpy',
+    'strndup',
+    'strnlen',
+    'strsep',
+    'strtonum',
+]
+
+foreach function : compat_functions
+    macro = 'HAVE_@0@'.format(function.underscorify().to_upper())
+    has_function = cc.has_function(function)
+    config.set(macro, has_function)
+    if not has_function
+        compat_apis += function
+    endif
+endforeach
+
+
+compat_sources = []
+
+foreach api : compat_apis
+    compat_sources += files('@0@.c'.format(api))
+endforeach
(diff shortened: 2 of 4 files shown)


Salamandar:
I'm trying to add the tests from the `regress` subdir, but they seem to need to be run in the correct order…

ThomasAdam:
No thank you -- I want portable tmux and what's in OpenBSD base to remain as close together as possible, and this change doesn't help with that.

Why do you think this is a good idea?

Salamandar:
> Why do you think this is a good idea?

Well, as I said, recent versions of tmux don't build on CentOS. Meson is really modern-built, faster (well, for a small code base like tmux it's not really important).

> I know how to use it;

That's a fair point. But Autotools and Meson easily cohabit (I ported Gimp, and currently am working on D-bus, and you can guess they don't want to drop autotools).

> it allows you to build in the source directory.

That's actually not a good point, because that "bloats" the source dir. A lot of projects forbid this.

> it is already well tested on all the platforms tmux runs on;

Well, it can't build Tmux 3.x on CentOS, so "well tested" is true for old versions (1.8 here), not recent.

> It doesn't work on OpenBSD which a) doesn't have bison, just yacc b) libevent is in the base system so not in pkg-config c) definitely has VIS_DQ in vis.h.

Yes, the bison bit was unnecessary. About libevent and VIS_DQ, I just recopied the logic of configure.ac

The other issues can easily be fixed once I have a BSD system somewhere.

> It doesn't work on CentOS 7.6: "Meson version is 0.47.2 but project requires >=0.50.0."

That also can be easily fixed by not using one or two recent features.

> And it would be a lot more work to do so than you might think.

Please don't assume what I think ;) As I said I already ported Gimp to Meson, so I think I know what difficulties to expect.

nicm:
I didn't assume what you think, I speculated what you *might* think :-).

> As I said I already ported Gimp to Meson

You didn't say this, or if you did I missed it...

> Well, as I said, recent versions of tmux don't build on CentOS. 

You didn't say this either. tmux definitely builds on CentOS, I use CentOS every day. If you can't get it to work please open a separate issue with make output and config.log.

It is a lot of work to change build system and I don't see any pressing reason to do it. tmux is not a large project with multiple subdirectories and complex interdependencies where I know autoconf is a pain in the arse. For a small project like tmux, it works fine, is well tested, is manageable, and the fact that the tarball can be self-contained and doesn't depend on a particular version of an external build tool is a big advantage to overcome.

Salamandar:
> You didn't say this, or if you did I missed it...

I said it higher on my answer, here : 

> That's a fair point. But Autotools and Meson easily cohabit (I ported Gimp, and currently am working on D-bus, and you can guess they don't want to drop autotools).

> It is a lot of work to change build system and I don't see any pressing reason to do it. tmux is not a large project with multiple subdirectories and complex interdependencies where I know autoconf is a pain in the arse. For a small project like tmux, it works fine, is well tested, is manageable, and the fact that the tarball can be self-contained and doesn't depend on a particular version of an external build tool is a big advantage to overcome.

I guess you could tarball the build directory with the ninja files… But anyways, I understand, you don't want to move from Autotools.

nicm:
Yes, I don't think there are good reasons to change at the moment. Thanks for the suggestion anyway.

PR #2142, opened 2020-03-31
Add #{session_cwd} format variable

cruegge:
This patch adds a format variable `#{session_cwd}` that exposes the current session working directory. I could not find any other way to obtain this information. The use case is that I usually have one tmux session per project / repo, with the repo root as working directory so that new windows automatically open in the proper directory, and I would like to have that visible on the status line and in the terminal window title.

### DIFF
--- a/format.c
+++ b/format.c
@@ -2492,6 +2492,7 @@ format_defaults_session(struct format_tree *ft, struct session *s)
 	ft->s = s;
 
 	format_add(ft, "session_name", "%s", s->name);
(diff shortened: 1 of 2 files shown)


nicm:
Looks good, except we do not call it `cwd` we call it `path` eg `pane_current_path`, so it should be `session_path`.

cruegge:
Ok, I changed that.

nicm:
Applied now, thanks!

PR #2151, opened 2020-04-04
map ctrl-j to \r in mode_tree_key

yl790:
The choose-session menu used to have this INLCR-like behavior in some older version of tmux and I got used to it, but now it doesn't have that any more.

Please consider restoring that behavior with this patch.

Signed-off-by: yl790 <yitao@rstudio.com>


nicm:
Hmm, I'm not sure C-j is very typical to use for this. This might be better you do with something like this, rather than changing it for everyone:
~~~~
bind -n C-j if -F '#{m/r:(tree|client|buffer)-mode,#{pane_mode}}' 'send Enter' 'send C-j'

PR #2164, opened 2020-04-12
Document the RGB terminfo capability in the man page.

jody-frankowski:
The terminfo capability RGB has been supported since
fe7a871a23ada0c71fb0886ef99a356c67bf5c0d but there's still no mention of it in
the man page.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -5441,7 +5441,7 @@ The server crashed or otherwise exited without telling the client the reason.
 .Nm
 understands some unofficial extensions to
 .Xr terminfo 5 :
-.Bl -tag -width Ds
+.Bl -tag -width "RGB , Tc"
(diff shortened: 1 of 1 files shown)


nicm:
Good point - I have added some text, however RGB is not a tmux extension but an official terminfo capability, so I have chosen not to add it to the list but instead mentioned that Tc is equivalent. This change should be in GitHub when it syncs up in a few hours.

Thanks!

~~~~
@@ -5534,6 +5534,11 @@
 .Nm
 .Xr terminfo 5
 entry).
+.Pp
+This is equivalent to the
+.Em RGB
+.Xr terminfo 5
+capability.
 .It Em \&Ms
 Store the current buffer in the host terminal's selection (clipboard).
 See the

PR #2198, opened 2020-05-05
Correct pane_path alias in documentation

casr:
`#T` appears to map to `pane_title` in [format.c](https://github.com/tmux/tmux/blob/a43a156/format.c#L176)


nicm:
Fixed now, thanks!

PR #2270, opened 2020-06-11
Add more details about IDs

nicolnt:
I was writing some key bindings and I couldn't figure this out using the manual, so I think that would help others as well 🤙

Tell the reader that IDs are auto increment integers, not just any kind of ID

⚠️ Can someone confirm that IDs doesn't get recycled, otherwise we should specify it.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -830,8 +830,7 @@ windows with a
 .Ql @ ,
 and panes with a
 .Ql % .
-These are unique and are unchanged for the life of the session, window or pane
-in the
+These are unique, auto increment integers and are unchanged for the life of the
 .Nm
 server.
 The pane ID is passed to the child process of the pane in the


nicm:
You should not depend on them being integers or on them being sequential.

nicolnt:
> You should not depend on them being integers or on them being sequential.

As a good practice yes, and it should also be specified 😉 however I think it's always important to reduce room for interpretation. Otherwise confusion may appear.

Say for example a new user see stuff like `0:bash*` in the tab bar, he may probably don't even know what bash is and be confused about which part is the ID. That's an example I can think of.

nicm:
IDs do not appear in the status line by default.

nicm:
Are you confusing the ID and the window index?

nicm:
I don't want to document stuff about the ID that happens to be true now but shouldn't be relied on, you should take the ID as a unique identifier and nothing more, whether it is a number or sequential is not really important.

nicolnt:
I have been confused by these IDs myself, that's why I raised this pull request.

The thing is that these IDs are not just internal stuff, which would not necessarily fit in a manpage, **the user interact with them**. In my opinion it makes total sense to teach her/him about how they behave and what should or shouldn't be done. We should look upon removing any confusion in order to help people not waste their time.

Also the man page can evolve as the the software does, so in case you bring some new kinds of IDs we can update the manual in consequence.

nicm:
We are not documenting that the IDs are sequential numbers because that is not a property that should be relied on, it is not part of how IDs work. Nobody needs to know this in order to use IDs for any purpose they are intended for. We can change the text to clarify that they are entirely unique within the lifetime of the server.

nicolnt:
At least I wanted to tell the user that they were numbers. I just shared my point of view, I don't want to argue so you can do whatever you think is good 😉

Anyways thank you very much for maintaining Tmux, that's one of my top 5 tools! 😸 🤙

nicm:
The problem is, if we are too specific about stuff that doesn't matter, people will write stuff that relies on it and then we can't change it even if we want to.

But tmux.1 already says "numbered with a unique" so I guess we are already saying they are numbers. But I don't really want to say they are sequential because it really doesn't matter.

I don't see what can be made clearer in the current text:
~~~~
     Sessions, window and panes are each numbered with a unique ID; session
     IDs are prefixed with a ‘$’, windows with a ‘@’, and panes with a ‘%’.
     These are unique and are unchanged for the life of the session, window or
     pane in the tmux server.
~~~~
It says they are unique, and it says they don't change while the server is running?

nicm:
There is also a description here: https://github.com/tmux/tmux/wiki/Advanced-Use#unique-identifiers

nicolnt:
I get your point 😉

> But tmux.1 already says "numbered with a unique ID"

Makes sense! I'm closing this issue then. Sorry for wasting your time with that.

PR #2271, opened 2020-06-15
fix dead loop of select-word command

remme123:
when hsize == 0 and oy == 0, select-word over the cy


nicm:
How do you reproduce?

remme123:
1. tmux new
2. bind-key -Tcopy-mode-vi DoubleClick1Pane send-keys -X select-word
3. enter copy-mode
4. double click the empty line below the cursor line

This will cause tmux server to a dead loop in **window_copy_cursor_previous_word_pos**

nicm:
I have applied this now, thnaks!

PR #2288, opened 2020-06-30
-a lists key bindings that do NOT have a note.

Annihilannic:


nicm:

PR #2304, opened 2020-07-12
prevent server crash when event's value is NULL

mmjo:
An events' value may be NULL.
E.g.
`setenv -g -r x`
Then a server crash happens on
`display -p $x`

This simple fix prevents the crash.

Regards, Mike


nicm:

Issue #2313, opened 2020-07-18
Allow spelling "colour" as "color" in config files


verhovsky:
In config files you have to use the American spelling to specify the terminal

```
set -g default-terminal "screen-256color"
```

but then use the correct spelling everywhere else

```
set-option -g status-style fg=colour255,bg=black
```

In some of the checked-in test config files people use "color" in comments but have to use "colour" in the config variables https://github.com/tmux/tmux/search?q=color+colour&unscoped_q=color+colour. This situation causes confusion for some https://superuser.com/a/668027

I prefer the American spelling, and would like to use it in my tmux config files. It would be nice if tmux let you use either spelling and used the Queen-approved one internally

ntnn:
Just to note - `set -g default-terminal "screen-256color"` isn't a decision by tmux devs. That's just the name of the terminfo.

nicm:
OK do you want to open a PR?

PR #2317, opened 2020-07-20
Allow spelling "colour" as "color" in config files

verhovsky:
#2313

Still need to do `ccolour`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 CHANGES FROM 3.2 TO 3.3
 
-* Allow spelling "colour" as "color"
+* Allow spelling "colour" as "color" for the options display-panes-active-colour,
+  display-panes-colour, and clock-mode-colour as well as when specifying colours
+  with colour0 to colour255.
 
 CHANGES FROM 3.1b TO 3.2
 
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+CHANGES FROM 3.2 TO 3.3
+
+* Allow spelling "colour" as "color"
+
 CHANGES FROM 3.1b TO 3.2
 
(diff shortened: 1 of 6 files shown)


nicm:
This is fine but I don't think there is a sensible way to change the option names.

nicm:
I think you will need to make a function that walks a string and converts "colour" to "color" and use that when looking up an option if the first comparison fails. I'm not really sure it is worth it TBH but it would work.

verhovsky:
Is `ccolour` a configurable thing option? What does it do and how do I configure it? I can't find a tmux config that uses it.

nicm:
I actually think it would be better to have an `alternative_name` member in the table entry (`options-table.c`) and check that if the option isn't found, so something like:
```c
struct options_entry *
options_get_only(struct options *oo, const char *name)
{
    ...

    found = RB_FIND(options_tree, &oo->tree, &o);
    if (found == NULL) {
        for (oe = options_table; oe->name != NULL; oe++) {
                if (strcmp(oe->alternative_name, name) == 0)
                        return (options_get_only(oo, oe->name));
        }
        return (NULL);
    }
    return (found);
}
```
ccolour is a terminal feature (cursor colour), I'd probably just leave it TBH.

verhovsky:
@nicm can you do it please? I don't know C or nroff syntax.

nicm:
Try this please: [c.diff.txt](https://github.com/tmux/tmux/files/5094602/c.diff.txt)

Searching the entire table is too expensive so I added a separate table (`options_other_names`). I only added one option (`display-panes-color`) if you want to add the rest.

Thanks

nicm:
I have applied this now, although I don't think it's necessary to mention this is in tmux.1, so I left that out. Thanks!

PR #2322, opened 2020-07-25
add 'delay' option to display-message

theonekeyg:
I think this is necessary addition, because changing `display-time` every time if you need more/less time to display-message something seems inconvenient


nicm:
Looks good but I think you should use -1 to mean "use the option" and then 0 can mean no delay (wait for key press), like the option value.

theonekeyg:
Fixed, i also realized i was using spaces, and this repo uses tabs, so i converted my spaces to tabs.

nicm:
Applied now with some minor tidying. Thanks!

Issue #1919, opened 2019-09-24
Add support for querying background color via OSC 11


blueyed:
`'\e]11;?\007'` can be used to query the background color of a terminal (OSC 11), but it is not supported in tmux: https://github.com/tmux/tmux/blob/06683cdd592b877ecdd3d30dc69d398402b73ef2/input.c#L2379-L2398

It would be nice if tmux would respond to it with the current window's background color.

This is used by Neovim to automatically set its background setting to "dark" or "light".

(I've also tried wrapping it in the "Ptmux" sequence (`printf '\ePtmux;\e\e]11;?\007\e\\'`) without success - likely because tmux interprets the response then itself somehow/parly?)

nicm:
Yes, this would not be hard to add, but would be best to add it for 10 and 12 also.

nicm:
Try this please - it will only respond to ? if you have set the colour in the first place: [x.diff.txt](https://github.com/tmux/tmux/files/3652793/x.diff.txt)

PR #2343, opened 2020-08-10
Reply for colour queries for OSC 10 and 11.

mgoral:
Hi, I revisited closed issue #1919 and implemented replies for OSC codes 10 and 11, which seems to be forgotten. Please have a look, thanks!

### DIFF
--- a/input.c
+++ b/input.c
@@ -2465,6 +2465,10 @@ input_osc_colour_reply(struct input_ctx *ictx, u_int n, int c)
     u_char      r, g, b;
     const char  *e = ictx->input_end == INPUT_END_BEL ? "\007" : "\033\\";
 
+    /* Don't respond if color wasn't set by user. */
+    if (c == 8)
+        return;
+
     if (~c & COLOUR_FLAG_RGB)
         return;
 
--- a/input.c
+++ b/input.c
@@ -2459,6 +2459,19 @@ input_osc_parse_colour(const char *p, u_int *r, u_int *g, u_int *b)
 	return (1);
 }
 
+static void
+input_osc_colour_reply(struct input_ctx *ictx, u_int n, int c)
+{
+    u_char      r, g, b;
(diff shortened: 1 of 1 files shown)


nicm:
I'm not wild about this because it only works if the user has set the colour and there is no practical way for it to work otherwise. So is it really that useful? What applications use this?

krader1961:
This is an example of a change that should have been discussed in an issue before a pull-request (i.e., code change) was submitted for review. As @nicm asks, what is the use case? If I create a tmux session, without explicitly setting a background color for the session, what should interrogating the background color return?

This feels like a situation where a feature is implemented as if tmux itself was not present in the use case. Which suggests the person proposing the change does not understand how text terminals, such as a DEC VT100, work or they don't fully understand what it means to multiplex multiple clients onto a virtualized terminal.

mgoral:
@nicm @krader1961: Neovim uses OSC to automatically set its background option (`:set background`), which can be used e.g. by color themes to adapt to either light or dark background.

The problem is that some xterm-compatible terminals already support these codes, but running tmux intercepts these queries. If tmux shouldn't respond to these queries, then why is it allowed to change pane colors by OSC in first place? Why implement OSC codes only partially? After reading #1919 I felt that implementing querying of background/foreground colors is something wanted.

Maybe implementation can be improved for the use case when background isn't explicitly set in tmux by passing through these queries to the underlying terminal. What do you think?

nicm:
It is a lot more useful being able to set than query. You can't pass the query to the underlying terminal because there may be several with different background colours.

mgoral:
Right, so do you see if this can be reliably implemented somehow? Maybe respond only when user sets background in tmux? Or is it better to leave it as it is now (don't respond at all)?

nicm:
The options are to do what your diff does and only respond when the user has set the colour, or to never respond.

mgoral:
@nicm can you have a look if this is now acceptable?

nicm:
Applied now, thanks!

PR #2378, opened 2020-09-12
unbind -q

mmjo:
`unbind -a copy-mode` in ~/.tmux.conf will bark when you reload
added a -q option to unbind to make it quiet

`unbind -a` did unbind of both root and prefix tables and `unbind -an` was not honored.
fixed that so `unbind -a` deletes prefix; unbind -an deletes root table.

Did a small refactoring of 
```
if (! args_has('a'))
  ...
else
  key = UNKNOWN

if (key== UNKNOWN) {
  ...
  return
}

code that runs when ! args_has('a')
```

### DIFF
--- a/cmd-queue.c
+++ b/cmd-queue.c
@@ -25,10 +25,6 @@
 
 #include "tmux.h"
 
-/* Command queue flags. */
-#define CMDQ_FIRED 0x1
-#define CMDQ_WAITING 0x2
-
 /* Command queue item type. */
 enum cmdq_type {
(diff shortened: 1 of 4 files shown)


nicm:
OK but don't mess with the flags like this, just don't call cmdq_error.

mmjo:
> OK but don't mess with the flags like this, just don't call cmdq_error.

Not calling was the easiest thing to do.  :)
Calling cmdq_error, with the new flag, has the advantage of leaving a trace in the log, when -q is used.

PR #2381, opened 2020-09-16
unbind -q

mmjo:
now without touching item->flags


nicm:

PR #2382, opened 2020-09-16
Fixes all warnings

mmjo:
May I suggest adding -Werror to CFLAGS?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,16 @@
 *.core
 core
 tags
+TAGS
+ID
+GTAGS
+GRTAGS
+GSYMS
+GPATH
+cscope.out
+cscope.in.out
+cscope.po.out
+cscope.files
 .deps/
 compat/.dirstamp
 aclocal.m4
--- a/client.c
+++ b/client.c
@@ -284,7 +284,7 @@ client_main(struct event_base *base, int argc, char **argv, int flags, int feat)
 
 	/* Save the flags. */
 	client_flags = flags;
-	log_debug("flags are %#llx", client_flags);
+	log_debug("flags are %#lx", client_flags);
 
 	/* Initialize the client socket and start the server. */
 	fd = client_connect(base, socket_path, client_flags);
@@ -891,7 +891,7 @@ client_dispatch_wait(struct imsg *imsg)
 			fatalx("bad MSG_FLAGS string");
 
 		memcpy(&client_flags, data, sizeof client_flags);
-		log_debug("new flags are %#llx", client_flags);
+		log_debug("new flags are %#lx", client_flags);
 		break;
 	case MSG_SHELL:
 		if (datalen == 0 || data[datalen - 1] != '\0')
@@ -944,7 +944,7 @@ client_dispatch_attached(struct imsg *imsg)
 			fatalx("bad MSG_FLAGS string");
 
 		memcpy(&client_flags, data, sizeof client_flags);
-		log_debug("new flags are %#llx", client_flags);
+		log_debug("new flags are %#lx", client_flags);
 		break;
 	case MSG_DETACH:
 	case MSG_DETACHKILL:
--- a/client.c
+++ b/client.c
@@ -284,7 +284,7 @@ client_main(struct event_base *base, int argc, char **argv, int flags, int feat)
 
 	/* Save the flags. */
 	client_flags = flags;
-	log_debug("flags are %#lx", client_flags);
+	log_debug("flags are %#llx", client_flags);
 
 	/* Initialize the client socket and start the server. */
 	fd = client_connect(base, socket_path, client_flags);
@@ -891,7 +891,7 @@ client_dispatch_wait(struct imsg *imsg)
 			fatalx("bad MSG_FLAGS string");
 
 		memcpy(&client_flags, data, sizeof client_flags);
-		log_debug("new flags are %#lx", client_flags);
+		log_debug("new flags are %#llx", client_flags);
 		break;
 	case MSG_SHELL:
 		if (datalen == 0 || data[datalen - 1] != '\0')
@@ -944,7 +944,7 @@ client_dispatch_attached(struct imsg *imsg)
 			fatalx("bad MSG_FLAGS string");
 
 		memcpy(&client_flags, data, sizeof client_flags);
-		log_debug("new flags are %#lx", client_flags);
+		log_debug("new flags are %#llx", client_flags);
 		break;
 	case MSG_DETACH:
 	case MSG_DETACHKILL:
--- a/cmd-unbind-key.c
+++ b/cmd-unbind-key.c
@@ -32,8 +32,8 @@ const struct cmd_entry cmd_unbind_key_entry = {
 	.name = "unbind-key",
 	.alias = "unbind",
 
-	.args = { "anT:", 0, 1 },
-	.usage = "[-an] [-T key-table] key",
+	.args = { "anqT:", 0, 1 },
+	.usage = "[-anq] [-T key-table] key",
 
 	.flags = CMD_AFTERHOOK,
 	.exec = cmd_unbind_key_exec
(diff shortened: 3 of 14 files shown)


nicm:
Thanks.

This is only all the warnings for your particular compiler version. Some recent GCC versions have loads of warnings from -Wmaybe-unitialized most of which are complete lies. -Werror would break anyone using those. Or anyone using any other compiler that has new warnings, whether they are right or wrong.

>    M client.c (6)
>    M control.c (2)

This is only correct if uint64_t is unsigned long which it is not always. Seems like a compiler problem, it should probably accept either unsigned long or unsigned long long for `ll` on platforms where they are both 64 bits. Or we could use PRIx64, but I kind of hate that.

>    M format.c (23)

Fixed the unused variable. I'm not going to add #pragma goo.

>    M input-keys.c (3)

I think if we are to have a default here it should be a fatalx.

>    M menu.c (2)

I think this is right, fixed.

>    M resize.c (2)

I think this might be right but the code could probably do with being reorganized. There is a lot of common bits with default_window_size.

>    M server.c (5)

I think this is right but I don't think cmdq_append should be called either. Fixed.

>    M tty-keys.c (2)

I think this is wrong. The only way terminator can leave the loop without being set is if end == len and the function will then return before it is used.

>    M utf8.c (5)

Fixed.

>    M window-copy.c (6)

Unused functions can just be removed entirely. 

The window_copy_search_jump looks wrong. sbuf is always set if regex is true and only used if regex is true. However, the warning could possibly by elimated by moving the free(sbuf) up directly after the regcomp block since sbuf is not needed after regcomp. Can you try that?

>    M window-customize.c (4)

window_customize_build_options I think the compiler is wrong, but I don't think it is possible for it to know that. This code is a bit too complicated I think.

window_customize_set_option I think this might be right, does this fix it instead though?
~~~~
@@ -1031,9 +1031,10 @@ window_customize_set_option(struct clien
                return;

        oe = options_table_entry(o);
-       if (~oe->scope & OPTIONS_TABLE_PANE)
+       if (~oe->scope & OPTIONS_TABLE_PANE) {
+               scope = WINDOW_CUSTOMIZE_NONE;
                pane = 0;
-       if (oe != NULL && (oe->flags & OPTIONS_TABLE_IS_ARRAY)) {
+       } if (oe != NULL && (oe->flags & OPTIONS_TABLE_IS_ARRAY)) {
                scope = item->scope;
                oo = item->oo;
        } else {

mmjo:
On Wed, Sep 16, 2020 at 11:39:34AM -0700, Nicholas Marriott wrote:
> Thanks.
You too.
 
> This is only all the warnings for your particular compiler version. Some recent GCC versions have loads of warnings from -Wmaybe-unitialized most of which are complete lies. -Werror would break anyone using those. Or anyone using any other compiler that has new warnings, whether they are right or wrong.

Yes, -Werror would only be feasible for gcc.
I do think that there should be no warnings at all.
At least from gcc or whatever you choose as 'the' compiler for tmux.
(given the default settings you get when running make)
Over time the warnings accumulate and will distract.
Or make you ignore or overlook them.

Initializing some variables, even when logically not needed, does no harm. 
The -O2 setting will likely not execute any code if unused.

> > M client.c (6)
> > M control.c (2)
> 
> This is only correct if uint64_t is unsigned long which it is not always. Seems like a compiler problem, it should probably accept either unsigned long or unsigned long long for `ll` on platforms where they are both 64 bits. Or we could use PRIx64, but I kind of hate that.

You're right. Unsigned long could be 32 bits.
I didn't know about the PRI macros, C89 did not have these ;)
They are indeed ugly.

Alternative, cast (uint64_t) to unsigned long long.
This does not generate a warning:
```
#include <stdio.h>
#include <stdint.h>
int main () {
	uint64_t u = UINT64_MAX;
	printf ("%llx\n", (unsigned long long)u);
}
```

On a side note. It looks as if uint64_t is too small for some flags.
E.g. CLIENT_ACTIVEPANE is defined as 0x80000000ULL

> > M format.c (23)
> Fixed the unused variable. I'm not going to add #pragma goo.

Haha, yes these pragmas are ugly.
Would you accept a separate nowarn.c file that is made with (additionally):
  -Wno-format-y2k -Wno-format-nonliteral
containing:
```
static size_t
nowarn_strftime(char *s, size_t max, const char *fmt, const struct tm *tm)
{
   return strftime(s, max, fmt, tm);
}
```

> > M input-keys.c (3)
> I think if we are to have a default here it should be a fatalx.

Yes, i considered that too; it would be a programming error.

> > M menu.c (2)
> I think this is right, fixed.

Reasoning: if the local pointer is not initialized, it may contain any value.
The test for NULL then fails and the random pointer will be dereferenced.

> > M resize.c (2)
> I think this might be right but the code could probably do with being reorganized. There is a lot of common bits with default_window_size.
Oh yes, i see.
The code in these cases also do no differ too much.
I'll see if i can understand that enough to refactor it.

> > M server.c (5)
> I think this is right but I don't think cmdq_append should be called either. Fixed.
Not sure about that. Aren't we now missing out on an error message?
The cmdq_append code looks for c == NULL though.

> > M tty-keys.c (2)
> I think this is wrong. The only way terminator can leave the loop without being set is if end == len and the function will then return before it is used.
Indeed, but gcc is not aware of that logic.
What is more wrong, initializing terminator or having a warning?
I would say the latter.

> > M window-copy.c (6)
> The window_copy_search_jump looks wrong. sbuf is always set if regex is true and only used if regex is true. However, the warning could possibly by elimated by moving the free(sbuf) up directly after the regcomp block since sbuf is not needed after regcomp. Can you try that?
I will have a look.

> > M window-customize.c (4)
> window_customize_build_options I think the compiler is wrong, but I don't think it is possible for it to know that. This code is a bit too complicated I think.
Too much oo! ;)
If o is not initialized and oo2 == NULL then o gets compared to NULL.
So i think o should be initialized to NULL.

> window_customize_set_option I think this might be right, does this fix it instead though?
> 
> ```
> @@ -1031,9 +1031,10 @@ window_customize_set_option(struct clien
>                 return;
> 
>         oe = options_table_entry(o);
> -       if (~oe->scope & OPTIONS_TABLE_PANE)
> +       if (~oe->scope & OPTIONS_TABLE_PANE) {
> +               scope = WINDOW_CUSTOMIZE_NONE;
>                 pane = 0;
> -       if (oe != NULL && (oe->flags & OPTIONS_TABLE_IS_ARRAY)) {
> +       } if (oe != NULL && (oe->flags & OPTIONS_TABLE_IS_ARRAY)) {
>                 scope = item->scope;
>                 oo = item->oo;
>         } else {
> ```
It fixes the warning. Which is weird, because setting scope inside the
`if ... _PANE' does not seem to influence the warning.

In that same `if ... _PANE' oe->scope is referenced.
Later oe is checked for != NULL.
I think that the same check is needed before oe->scope.

Regards, Mike

nicm:
-Werror is only feasible for a specific compiler *and* a specific compiler version. We can't control what gcc or clang versions everyone is using. Warnings also sometime depend on the system headers (such as deprecated and unused return value warnings).

The compiler I care about is the clang version supported on OpenBSD -current and there are only three warnings when it is used (printf format strings that are not const). Warnings do not accumulate because they are fixed.

> Initializing some variables, even when logically not needed, does no harm. 

I don't agree. Initializing variables when not needed prevents compilers ever working out that they are used uninitialized. This can hide bugs if compiler detection improves, and introduce bugs if the code is changed. The same applies to unnecessary casts.

I will look at your other comments tomorrow.

nicm:
I have used a (unsigned long long) cast for %llx and %llu with uint64_t in a few places already in tmux so I suggest doing that in any others as well.

I think we should just leave strftime alone TBH, I don't think there is a sensible way to get rid of the warning.

mmjo:
On Thu, Sep 17, 2020 at 02:48:01PM -0700, Nicholas Marriott wrote:
> -Werror is only feasible for a specific compiler *and* a specific compiler version. We can't control what gcc or clang versions everyone is using. Warnings also sometime depend on the system headers (such as deprecated and unused return value warnings).

Agreed.
-Werror is useful when only using gcc, like in personal projects.


> The compiler I care about is the clang version supported on OpenBSD -current and there are only three warnings when it is used (printf format strings that are not const). Warnings do not accumulate because they are fixed.

I have no experience with clang.
Does clang present itself as gcc? 
Asking this, because i only see a reference to gcc in Makefile.am


> > Initializing some variables, even when logically not needed, does no harm. 
> 
> I don't agree. Initializing variables when not needed prevents compilers ever working out that they are used uninitialized. This can hide bugs if compiler detection improves, and introduce bugs if the code is changed. The same applies to unnecessary casts.
Hmm, i slightly disagree as to the relative importance of this argument,
versus the problems that exist when having many warnings.
But i am fine with your choice, as you said that (most) warnings are to be fixed.

> I will look at your other comments tomorrow.
No hurry.

Regards, Mike

nicm:
> On a side note. It looks as if uint64_t is too small for some flags. E.g. CLIENT_ACTIVEPANE is defined as 0x80000000ULL

Client flags are now uint64_t. Anywhere they are still 32 bits needs to be changed. Where did you see this?

I have applied a couple of your other fixes. Can you try again now and see what warnings you have left?

If you show me the output of `make >/dev/null` it will only show the warnings.

nicm:
Try again once the changes have synced to GitHub I mean, it will be an hour or so probably.

mmjo:
On Thu, Sep 17, 2020 at 02:52:51PM -0700, Nicholas Marriott wrote:
> I have used a (unsigned long long) cast for %llx and %llu with uint64_t in a few places already in tmux so I suggest doing that in any others as well.

Ok. Can fix that.

> I think we should just leave strftime alone TBH, I don't think there is a sensible way to get rid of the warning.

I find the solution of a nowarn.c not so bad.
For gcc i can then add the -Wno* CFLAGS for nowarn.c only to Makefile.am.
But it would also need to vary with compiler.
Does clang have flags to ignore the warnings?

I will make a pull request, feel free to ignore it if you don't like it.

Regards, Mike

mmjo:
On Fri, Sep 18, 2020 at 04:24:21AM -0700, Nicholas Marriott wrote:
> > On a side note. It looks as if uint64_t is too small for some flags. E.g. CLIENT_ACTIVEPANE is defined as 0x80000000ULL
> 
> Client flags are now uint64_t. Anywhere they are still 32 bits needs to be changed. Where did you see this?

client.c has functions having an int flags as parameter:
- client_connect
- client_main

I was wondering more about the ULL suffix. Suspecting that an unsigned long long 
was maybe 16 bytes. But it is the same as unsigned long (8 bytes).
At least at my platform. So it might be a problem on other platforms.

Btw. Would you want all flags, with #defined values, converted to enums?

> I have applied a couple of your other fixes. Can you try again now and see what warnings you have left?
> 
> If you show me the output of `make >/dev/null` it will only show the warnings.

client.c: In function ‘client_main’:
client.c:287:27: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘uint64_t {aka long unsigned int}’ [-Wformat=]
  log_debug("flags are %#llx", client_flags);
                       ~~~~^
                       %#lx
client.c: In function ‘client_dispatch_wait’:
client.c:894:32: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘uint64_t {aka long unsigned int}’ [-Wformat=]
   log_debug("new flags are %#llx", client_flags);
                            ~~~~^
                            %#lx
client.c: In function ‘client_dispatch_attached’:
client.c:947:32: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘uint64_t {aka long unsigned int}’ [-Wformat=]
   log_debug("new flags are %#llx", client_flags);
                            ~~~~^
                            %#lx
format.c: In function ‘format_pretty_time’:
format.c:1406:24: warning: ‘%y’ yields only last 2 digits of year [-Wformat-y2k]
  strftime(s, sizeof s, "%h%y", &tm);
                        ^~~~~~
format.c: In function ‘format_find’:
format.c:1485:5: warning: format not a string literal, format string not checked [-Wformat-nonliteral]
     strftime(s, sizeof s, time_format, &tm);
     ^~~~~~~~
format.c: In function ‘format_expand1’:
format.c:2432:3: warning: format not a string literal, format string not checked [-Wformat-nonliteral]
   if (strftime(expanded, sizeof expanded, fmt, tm) == 0) {
   ^~
resize.c: In function ‘recalculate_size’:
resize.c:362:2: warning: ‘sy’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  log_debug("%s: @%u new size %u,%u", __func__, w->id, sx, sy);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
resize.c:362:2: warning: ‘sx’ may be used uninitialized in this function [-Wmaybe-uninitialized]
tty-keys.c: In function ‘tty_keys_next’:
tty-keys.c:1141:14: warning: ‘terminator’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  *size = end + terminator;
          ~~~~^~~~~~~~~~~~
tty-keys.c:1100:15: note: ‘terminator’ was declared here
  size_t  end, terminator, needed;
               ^~~~~~~~~~

Regards, Mike

nicm:
> client.c has functions having an int flags as parameter: - client_connect - client_main

OK I will look at these.

> I was wondering more about the ULL suffix. Suspecting that an unsigned long long
> was maybe 16 bytes. But it is the same as unsigned long (8 bytes).
> At least at my platform. So it might be a problem on other platforms.

We do not support any platforms where long long is not 64 bits.

> Btw. Would you want all flags, with #defined values, converted to enums?

No, flags should be \#define.

nicm:
I'm going to close this now, please open a new PR with any additional fixes. Thanks!

PR #2385, opened 2020-09-18
fix warnings: added unsigned long long casts of client_flags for llx …

mmjo:
as discussed, fixes 3 warnings


nicm:
Applied to OpenBSD now along with some other changes for 64-bit client flags, will be in GitHub later. Thanks!

PR #2386, opened 2020-09-18
Extend CMSG_DATA test needing _XOPEN_SOURCE 600 for c99

ericvw:
On Solaris, the Sun Studio 12.6 C compiler defaults to c99 which
requires `_XOPEN_SOURCE 600` to define `_XPG4_2` in order to define
`CMSG_DATA`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -329,7 +329,9 @@ AC_SEARCH_LIBS(socket, socket)
 AC_CHECK_LIB(xnet, socket)
 
 # Check for CMSG_DATA. Some platforms require _XOPEN_SOURCE_EXTENDED (for
-# example see xopen_networking(7) on HP-UX).
+# example see xopen_networking(7) on HP-UX).  Other platforms require
+# _XOPEN_SOURCE 600 when compiling with c99 (for example see standards(7)
+# on Solaris).
 XOPEN_DEFINES=
 AC_MSG_CHECKING(for CMSG_DATA)
 AC_EGREP_CPP(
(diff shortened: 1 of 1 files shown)


ThomasDickey:
It's not a complete test:
- it doesn't check if `CMSG_DATA` is defined without the particular value of `_XOPEN_SOURCE`
- it doesn't check if `_XOPEN_SOURCE` might be defined to another value already.

ericvw:
> it doesn't check if `CMSG_DATA` is defined without the particular value of `_XOPEN_SOURCE`

If I am following `configure.ac` correctly, the first test for `CMSG_DATA` occurs without any `_XOPEN_SOURCE` defines at: https://github.com/tmux/tmux/blob/ca5afb34bfdcaddbaed12e7375b48c05019e4c18/configure.ac#L333-L346

Expanding the diff context above this change should reveal this test already exists and will not set `XOPEN_DEFINES` if `CMSG_DATA` is available without the macros being set.

> * it doesn't check if `_XOPEN_SOURCE` might be defined to another value already.

The tests for `CMSG_DATA` is the only place in `configure.ac` for determining which `_XOPEN_SOURCE` macros should be defined. https://github.com/tmux/tmux/blob/ca5afb34bfdcaddbaed12e7375b48c05019e4c18/configure.ac#L331-L369

There is already an existing test that doesn't check if any of the macros are already set.  For this PR, I was extending the check by keeping along with the same pattern.

@nicm, let me know if you would like me to send a separate PR if you feel like this type of check is necessary.  I will need some guidance because I haven't worked on the maintainer side of the autotools stack that much.

nicm:
This looks fine, applied now, thanks!

nicm:
HP-UX has supported UNIX 03 since 11i v3 which was apparently released in 2007 so we could perhaps drop the _XOPEN_SOURCE_EXTENDED test now. Although it doesn't do any real harm to leave it.

PR #2388, opened 2020-09-19
Prevent #() from being expanded in #() output

cowens:
Formats are recursively expanded. It is dangerous for a shell commands
output to be able to run further shell commands, so expansion of #()
is allowed only once.

### DIFF
--- a/format.c
+++ b/format.c
@@ -2471,14 +2471,7 @@ format_expand1(struct format_tree *ft, const char *fmt, int time)
 			format_log(ft, "found #(): %s", name);
 
 			if (ft->flags & FORMAT_NOJOBS) {
-				out = format_expand1(ft, fmt, time);
-				/* ugly, but works */
-				if (len - off < 3) {
-					buf = xreallocarray(buf, 2, len);
-					len *= 2;
-				}
-				buf[off++] = '#';
-				buf[off++] = '(';
+				out = xstrdup("");
 				format_log(ft, "#() is disabled");
 			} else {
 				ft->flags |= FORMAT_NOJOBS;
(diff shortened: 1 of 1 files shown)


nicm:
Why are you still expanding #() when it is disabled? It should be removed.

cowens:
Can you give me a test case where it shouldn't expand `#()`?  I don't see how FORMAT_NOJOBS does anything today. It looks like the only thing that sets it is the bison code. I never played much with yacc or bison and I don't really understand the grammar it is building. The code I have written works for all of the cases I am aware of.

Looking at cmd-parse.y again, it looks like when a thing is "expanded" it is supposed to build a new tree with the FORMAT_NOJOBS flag set and then expands it again, but that doesn't seem to actually work (because #(thing) can expand to #(evil) which executes). Maybe I am just fixing it in the wrong location and the real bug is in the bison code (and I am just masking it by fixing it in expand1).

nicm:
I'm not saying it doesn't work, I'm asking why you are changing the
behaviour so that FORMAT_NOJOBS expands #() again instead of just removing
it? It is intended to that if jobs are disabled then #() is entirely
ignored. If you have #(sh foo) and foo does "echo x#(sh bar)y" the result
should be "xy" and "bar" should not be executed. Is it not enough to just
set FORMAT_NOJOBS for the  call to format_expand in format_job_get, why are
you changing the bit that handles the flag in format_expand1?

I think the use in cmd-parse.y is for %if where we do not want it to start
commands. So if you do "%if #(sh foo)" the result should be an empty string
and "foo" should not be executed.

cowens:
I think I became confused and forgot that the goal of this wasn't to get the actual contents of the shell command to the status bar, but to prevent the execution of the nested shell command.  If the user wants to see the actual contents of the shell command, they will have to escape them with the bit of perl I gave initially (which will prevent expansion). I am fixing now.

cowens:
I noticed as I was preparing the other one that my git config was still set for work, so I am going to kill this request, unwind my commits on the branch, recommit them (probably squashing them while I am at it), and then submit a second pull request.

PR #2389, opened 2020-09-21
escaping now works with styles

cowens:
This wound up being significantly more complicated and Rube Goldbergian than I expected. I am still not sure if this is the right way.

### DIFF
--- a/format-draw.c
+++ b/format-draw.c
@@ -509,10 +509,11 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 	size_t			 size = strlen(expanded);
 	struct screen		*os = octx->s, s[TOTAL];
 	struct screen_write_ctx	 ctx[TOTAL];
-	u_int			 ocx = os->cx, ocy = os->cy, i, width[TOTAL];
+	u_int			 ocx = os->cx, ocy = os->cy, i, n, width[TOTAL];
 	u_int			 map[] = { LEFT, LEFT, CENTRE, RIGHT };
 	int			 focus_start = -1, focus_end = -1;
 	int			 list_state = -1, fill = -1;
+	int                      even;
 	enum style_align	 list_align = STYLE_ALIGN_DEFAULT;
 	struct grid_cell	 gc, current_default;
 	struct style		 sy, saved_sy;
@@ -547,6 +548,39 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 	 */
 	cp = expanded;
 	while (*cp != '\0') {
+		if (cp[0] == '#' && cp[1] != '[' && cp[1] != '\0') {
+			for (i = 1; cp[i] == '#'; i++) {}
+			if (cp[i] == '[') {
+				even = !(i % 2);
+				cp += i + (even ? 1 : -1);
+				if (!sy.ignore) {
+					n = i;
+					for (i = i/2; i > 0; i--) {
+						utf8_set(ud, '#');
+						screen_write_cell(&ctx[current],
+							&sy.gc);
+					}
+					width[current] += n/2;
+					if (even) {
+						utf8_set(ud, '[');
+						screen_write_cell(&ctx[current],
+							&sy.gc);
+						width[current]++;
+					}
+				}
+			} else {
+				width[current] += i;
+				cp += i;
+				for (; i > 0; i--) {
+					utf8_set(ud, '#');
+					screen_write_cell(&ctx[current],
+						&sy.gc);
+				}
+				continue;
+			}
+
+		}
+
 		if (cp[0] != '#' || cp[1] != '[' || sy.ignore) {
 			/* See if this is a UTF-8 character. */
 			if ((more = utf8_open(ud, *cp)) == UTF8_MORE) {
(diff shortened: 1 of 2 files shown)


nicm:
I think the format.c change can be done more concisely - if we find #*[ then we can copy what is in the source buffer already, and if we don't we can fall through as it did before and carry on from the next character.

Note that for xreallocarray, the second and third arguments are multiplied together - so it will always be 2*len because we want to double the buffer size each time.

How about this?

I'm not sure about the format-draw.c changes, I think you have the right idea - will take a better look later on.

~~~~
--- format.c    16 Sep 2020 18:37:55 -0000      1.262
+++ format.c    22 Sep 2020 06:31:09 -0000
@@ -2502,8 +2502,30 @@ format_expand1(struct format_tree *ft, c
                                break;
                        fmt += n + 1;
                        continue;
-               case '}':
                case '#':
+                       /*
+                        * If ##[ (with two or more #s), then it is a style and
+                        * can be left for format_draw to handle.
+                        */
+                       ptr = fmt;
+                       n = 2;
+                       while (*ptr == '#') {
+                               ptr++;
+                               n++;
+                       }
+                       if (*ptr == '[') {
+                               format_log(ft, "found #*%zu[", n);
+                               while (len - off < n + 2) {
+                                       buf = xreallocarray(buf, 2, len);
+                                       len *= 2;
+                               }
+                               memcpy(buf + off, fmt - 2, n + 1);
+                               off += n + 1;
+                               fmt = ptr + 1;
+                               continue;
+                       }
+                       /* FALLTHROUGH */
+               case '}':
                case ',':
                        format_log(ft, "found #%c", ch);
                        while (len - off < 2) {

nicm:
I don't really understand why your change to format_draw is doing so much, isn't it enough just to draw `##` as `#`? Then `##[` will become a `#` and a `[`, and `###[` will become `#` and the `#[` processed as a style. Why do you need to count them? Maybe I am misunderstanding what you are trying to do.

Something like this? Can you give me an example of what case this doesn't solve?
~~~~
--- format-draw.c       16 May 2020 16:26:34 -0000      1.20
+++ format-draw.c       22 Sep 2020 06:54:51 -0000
@@ -547,6 +547,13 @@ format_draw(struct screen_write_ctx *oct
         */
        cp = expanded;
        while (*cp != '\0') {
+               if (cp[0] == '#' && cp[1] == '#') {
+                       utf8_set(ud, '#');
+                       screen_write_cell(&ctx[current], &sy.gc);
+                       cp += 2;
+                       continue;
+               }
+
                if (cp[0] != '#' || cp[1] != '[' || sy.ignore) {
                        /* See if this is a UTF-8 character. */
                        if ((more = utf8_open(ud, *cp)) == UTF8_MORE) {

nicm:
Here is a diff with both changes, I have left your format_trim_* changes alone for the moment: [y.diff.txt](https://github.com/tmux/tmux/files/5259577/y.diff.txt)

cowens:
That change seems to break octothorps on their own being escaped:

```
set-option -g status 3
set-option -g status-format[0] "commented ##[fg=red]not red"
set-option -g status-format[1] "one octothorp # two octothorps ####"
set-option -g status-format[2] "red octothrop #[fg=red]###[fg=default] blue octothorp #[fg=blue]#"
```

Which looks like this with my change:

Screen Shot 2020-10-05 at 09 38 11

and this with yours:

Screen Shot 2020-10-05 at 09 39 03

cowens:
Oh, and I am counting them because we have to walk to the end of octothorpes anyway to see if there is bracket at the end and it is easy to say that the escaping is (int)(count/2) and not have to walk over that section repeatedly (for instance if there is a format "######", my code only has to look at those six characters and say, oh it should be "###", but a character-by-character method would have to look at "######", escape the second octothorpe, then "####", escape the second octothorpe, "##", and escape the second octothorpe.

nicm:
I'll have another look later on. I think it would be better to have simpler code even if it has to walk repeatedly.

nicm:
OK I see what you are doing, we need to walk the #s in format_draw because we only want to convert ## to # if they end with [, otherwise we are repeating what the format_expand code will already have done. That's why mine turns #### into # rather than ##.

nicm:
OK here is another diff, I have used your method in format_draw but reordered it and tidied it up a bit, please have a a look: [abc.diff.txt](https://github.com/tmux/tmux/files/5328007/abc.diff.txt)

cowens:
That looks like it works.  Do you want me to update the pull request with this code, or will you just apply the patch directly?

nicm:
It's OK thanks, I will apply it directly. I need to look over the format_trim_* bits as well. Thanks.

nicm:
I finally had some more time to look at this, sorry for the long delay!

I have tidied up the format_width/format_trim_* changes a bit but left the logic the same as yours. Here is an updated diff: [tmux-style-escape.diff.txt](https://github.com/tmux/tmux/files/5457869/tmux-style-escape.diff.txt)

This also adds `w` format modifier which evaluates to the width. Here is a test script that uses this to check the output: [testx.sh.txt](https://github.com/tmux/tmux/files/5457930/testx.sh.txt)

There appears to be a couple of errors in the trim code:

1) If V is `##[bg=yellow]0123456` then `#{=-3:V}` is `3456` and not `456`.

2) If V is `###[bg=yellow]0123456` then `#{=3:V}` is `###[bg=yellow]01` and not `#[bg=yellow]012` or `###[bg=yellow]012` (`#{-3:V}` is `#[bg=yellow]456` for the same input so it would be better if the `#`s were compressed in all cases).

Don't have any more time to look at these right now but will have another look later on.

cowens:
I'll see if I can carve some time out this weekend to take a look too.

nicm:
Actually I think my test script was wrong for one of these, and a minor change to `format_trim_right` (changing `width += (old_n / 2)` to `width += (old_n / 2) - even`) seems to fix the other. Here is an updated diff and test script. If you have some time to check over these and see if they look correct (particularly if the test script is checking the right things), it would be great. I'll make it into a proper regression test script later on.

[tmux-style-escape2.diff.txt](https://github.com/tmux/tmux/files/5458598/tmux-style-escape2.diff.txt)

[testx.sh.txt](https://github.com/tmux/tmux/files/5458599/testx.sh.txt)

nicm:
I have committed a more extensive test that also tests what actually appears when the text is displayed in the status line. It is in the regress directory and the commit is here: https://github.com/tmux/tmux/commit/ce2b6ff40e13fd25f359cfe3faad49b094aad115

The regress tests don't print anything and just exit 1 but there are three printfs commented out you can uncomment if you want. It defaults to using ../tmux but you can run it manually with a different tmux by doing, for example:
~~~~
env TEST_TMUX=/usr/src/usr.bin/tmux1/tmux sh style-trim.sh
~~~~

nicm:
I have applied this now to OpenBSD, it should be in GitHub soon. Please let me know if you see any problems. Thanks!

PR #2390, opened 2020-09-21
Prevent #() from being expanded in #() output

cowens:
Formats are recursively expanded. It is dangerous for a shell commands
output to be able to run further shell commands, so expansion of #() is
allowed only once.

### DIFF
--- a/format.c
+++ b/format.c
@@ -2474,8 +2474,10 @@ format_expand1(struct format_tree *ft, const char *fmt, int time)
 				out = xstrdup("");
 				format_log(ft, "#() is disabled");
 			} else {
+				ft->flags |= FORMAT_NOJOBS;
 				out = format_job_get(ft, name);
 				format_log(ft, "#() result: %s", out);
+				ft->flags &= ~FORMAT_NOJOBS;
 			}
 			free(name);
 


nicm:
I don't really like turning the flag off and on again. Please take a look at this which gives the format a separate set of internal expand flags (and uses it for the time flag also) and makes all internal callers use format_expand1. It is a bit more invasive but I think it is better in the long run: [x.diff.txt](https://github.com/tmux/tmux/files/5259411/x.diff.txt)

nicm:
Does this work for you?

cowens:
Sorry, work got insane for a bit.  It doesn't produce the same results as my version.  Using the config

    set-option -g status-format[0] "[#(cat ~/something.txt)] [#(cat ~/something.txt | perl -pe 's/\\x23/\\x23\\x23/g')] [#(cat ~/nested.txt)]"
    set-option -g status-interval 1

and the files something.txt

    #(echo evil)

and nested.txt

    [#I] [#(echo evil)] [#H]

I get this with an unpatched tmux 3.1b

     [evil] [#(echo evil)] [[0] [evil] [generic]]

With my patch I get

    [] [#(echo evil)] [[0] [] [generic]]

With your patch I get

    [] [] []

nicm:
Ah yeah that idea was wrong. Please try this instead: [tmux-format2.diff.txt](https://github.com/tmux/tmux/files/5327825/tmux-format2.diff.txt)

cowens:
That seems to work. If I understand correctly, you have added another level of indirection on top of the format tree to hold the loop and flags instead of the format tree holding them directly.  Then you can create new things at that layer of indirection and leave format tree untouched. When processing continues, the old version of the state will be used. That lets you also refactor some other code that has to have its own discrete versions of the loop and flags.

nicm:
Yes I have added a state struct to hold working state during format_expand instead of putting it into struct format_tree which means we can change the flags by creating a new state. I have applied this now with some additions to also put the time into the state struct. It should be in GitHub in a few hours - please let me know if you see any problems. Thanks!

PR #2392, opened 2020-09-21
Refactor resize.c

mmjo:
refactoring of default_window_size and recalculate_size

Number of lines in resize.c go from 417 to 322

From issue #2382 
>  > M resize.c (2)
> I think this might be right but the code could probably do with being reorganized. There is a lot of common bits with default_window_size.

Regards, Mike

### DIFF
--- a/resize.c
+++ b/resize.c
@@ -92,130 +92,124 @@ ignore_client_size(struct client *c)
 	return (0);
 }
 
-void
-default_window_size(struct client *c, struct session *s, struct window *w,
-    u_int *sx, u_int *sy, u_int *xpixel, u_int *ypixel, int type)
+static int
+clients_calculate_size(int type, struct window *w,
+	int (*skip_client) (struct client *),
+	u_int *sx, u_int *sy, u_int *xpixel, u_int *ypixel)
 {
 	struct client	*loop;
 	u_int		 cx, cy, n;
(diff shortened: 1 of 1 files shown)


nicm:
I like it. Please take a look at this which has some minor tidying and also adds a load of comments: [z.diff.txt](https://github.com/tmux/tmux/files/5259820/z.diff.txt)

mmjo:
On Tue, Sep 22, 2020 at 07:59:24AM +0000, Nicholas Marriott wrote:
> I like it. Please take a look at this which has some minor tidying and also adds a load of comments: [z.diff.txt](https://github.com/tmux/tmux/files/5259820/z.diff.txt)

Will look at it later.

But at first glance, the diff has nothing to do with this resize.c refactoring.
What is it goal?

Regards, Mike

nicm:
Sorry I added the wrong file, I'll fix it later.

nicm:
Try this instead please: [zz.diff.txt](https://github.com/tmux/tmux/files/5265909/zz.diff.txt)

mmjo:
Looks good.

Regards, Mike

nicm:
Great. OpenBSD is about to lock for release but I will apply this once it unlocks.

nicm:
Applied now, thanks!

PR #2395, opened 2020-09-23
adds macro NOWARN and function nowarn_strftime

mmjo:
Makefile (.am) contains a specific rule addition to compile nowarn.c with -Wno...
So no #pragma GCC goo ;)

What are your thoughts on this?


nicm:
We're not adding two files and 30 lines of code to get rid of two useless warnings, just ignore them.

Issue #2396, opened 2020-09-24
Yacc requirement should be in readme


themightyoarfish:
In #2127 it was reported that tmux compilation fails because yacc is not found. It was said this would be added to the instructions. 
https://github.com/tmux/tmux#dependencies does not mention this requirement. Could it be added there? Could the autotools setup detect the absence and refuse configuring?

Issue #2127, opened 2020-03-20
Tmux compile issue


descendingnerdliness:
Kubuntu 18.04 After downloading the master, compiling results in the following.
```
/bin/bash ./etc/ylwrap cmd-parse.y y.tab.c cmd-parse.c y.tab.h `echo cmd-parse.c | sed -e s/cc$/hh/ -e s/cpp$/hpp/ -e s/cxx$/hxx/ -e s/c++$/h++/ -e s/c$/h/` y.output cmd-parse.output -- yacc  
./etc/ylwrap: line 175: yacc: command not found
Makefile:848: recipe for target 'cmd-parse.c' failed
make: *** [cmd-parse.c] Error 127

```
Similarly, tmux-3.1-rc3 yields this compile result:
```
/bin/bash ./etc/ylwrap cmd-parse.y y.tab.c cmd-parse.c y.tab.h `echo cmd-parse.c | sed -e s/cc$/hh/ -e s/cpp$/hpp/ -e s/cxx$/hxx/ -e s/c++$/h++/ -e s/c$/h/` y.output cmd-parse.output -- yacc  
./etc/ylwrap: line 175: yacc: command not found
Makefile:847: recipe for target 'cmd-parse.c' failed
make: *** [cmd-parse.c] Error 127
```
I have installed the two prerequisites. Configure seemed to complete normally. 
Please tell me what else I am missing?

* Platform (`uname -sp`).
Linux x86_64

nicm:
You now also need to install `bison`. I'll add it to the instructions.

descendingnerdliness:
OK, Thanks! 
bisonc++ 6.01 is in the Ubuntu 18.04 repo. Will that do?

nicm:
No that is for C++. You should definitely have bison, try:
~~~~
apt-get install bison
~~~~
If that doesn't work, try byacc instead of bison.

descendingnerdliness:
It ran with the Ubuntu versioning of bison 2:3.0.4.dfsg-1build1.
Thanks!

PR #2399, opened 2020-09-25
Adding YACC to dependencies in README

Shanduur:
As stated in #2127 and #2396, YACC is required for compilation. This pull request is created to resolve those issues.


nicm:
I have made a change similar to this, thanks!

PR #2400, opened 2020-09-27
Updated new-window default format in help

gpakosz:
The default format for `new-window` is `#{session_name}:#{window_index}.#{pane_index}`


nicm:
Applied now, thanks!

Issue #911, opened 2017-05-04
Feature request: Hyperlinks


egmontkob:
Recently two popular terminal emulators, `GNOME Terminal` and `iTerm2` have implemented a brand new feature: explicit hyperlinks.

Unlike the existing functionality of most terminal emulators of automatically detecting URLs that appear onscreen, this time it's like hyperlinks on web pages: the link target is explicitly specified by the OSC 8 escape sequence and the visible text can be an arbitrary piece of text.

We believe this feature can provide a noticeable productivity boost for plenty of users, and hence we're hoping to get some widespread recognition and support, in which `tmux` joining the game would be a key factor!

Ideally `tmux` would make its presence transparent to the user and make sure hyperlinks work exactly as if `tmux` weren't in the game. This includes autogenerating or mangling the `id` parameter in some trivial way, see the docs linked below for why and how, but that's a piece of cake. The ~~tough~~challenging bits is probably coming up with a memory-efficient way of storing a target URL for some of the character cells.

Please see https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda for details about this feature.

nicm:
I'm not going to implement this but if you come up with a patch I'll look
at it, although it will need to be kept fairly simple. It can just use the
existing extended cells mechanism used for UTF-8. Also there will need to
be some way to detect it even if it just means inventing a terminfo
capability.



On 4 May 2017 10:11 pm, "egmontkob" <notifications@github.com> wrote:

Recently two popular terminal emulators, GNOME Terminal and iTerm2 have
implemented a brand new feature: explicit hyperlinks.

Unlike the existing functionality of most terminal emulators of
automatically detecting URLs that appear onscreen, this time it's like
hyperlinks on web pages: the link target is explicitly specified by the OSC
8 escape sequence and the visible text can be an arbitrary piece of text.

We believe this feature can provide a noticeable productivity boost for
plenty of users, and hence we're hoping to get some widespread recognition
and support, in which tmux joining the game would be a key factor!

Ideally tmux would make its presence transparent to the user and make sure
hyperlinks work exactly as if tmux weren't in the game. This includes
autogenerating or mangling the id parameter in some trivial way, see the
docs linked below for why and how, but that's a piece of cake. The
toughchallenging
bits is probably coming up with a memory-efficient way of storing a target
URL for some of the character cells.

Please see https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3c
b5feda for details about this feature.

[...]

PR #2403, opened 2020-09-30
Apply patch for OSC 8 hyperlinks

sencer:
Is there a reason this patch is not included in HEAD?


nicm:
It is not complete. Hardly any terminals support it. Hardly any applications support it.

dufferzafar:
For people who need more context, please see: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

It has a list of terminals & applications that do support it. @nicm I think saying "hardly any" for that list doesn't seem justified.

glensc:
@dufferzafar: I think a correction is in order: "doesn't seem justified anymore"

for posterity, the original PR seems to be at:
- https://github.com/tmux/tmux/issues/911

@sencer: you should remove your "this patch" change from the readme.md changes, if you want this to be merged. i'm pretty sure upstream will never accept it in this form. in fact, you should never create pull requests from your master branch as the branch is directly linked with PR so you can't modify the branch for your own purposes.

sencer:
@glensc I added that part after @nicm's response.

glensc:
@sencer remove it if you want the change to be merged. it describes this specific pull request, so such text belongs to the commit message and/or pr body, not the final readme (result of the merge), as that would describe the project.

yesnomaybeyes:
> It is not complete. Hardly any terminals support it. Hardly any applications support it.

gcc, coreutils, and systemd do at least

nicm:
Right, but they use it for noddy stuff. Nobody seriously needs URLs in gcc or systemd. What about major programs like irssi/weechat? emacs/vim? mutt? What's the status of `less -R` support? What major terminals support it apart from VTE and iTerm2?

Anyway the patch is not finished, it needs:

- It needs changes to support the `id` parameter (not sure how this will look, can we ignore it?).

- Existing links need to be reused instead of added (fairly easy).

- It will need to be added as a terminal feature in tty-features.c.

- I don't like that it keeps adding links forever. Old ones need to be freed. I am not sure whether it is better to make the list of links per-pane instead of global, or to collect them somehow from the global list.

dustymabe:
> Right, but they use it for noddy stuff. Nobody seriously needs URLs in gcc or systemd. What about major programs like irssi/weechat? emacs/vim? mutt? What's the status of `less -R` support? What major terminals support it apart from VTE and iTerm2?

It's a simple use but I use this all the time with `alias ls='ls --color=auto --hyperlink=auto'` in my `.bashrc` and running `ls` inside GNOME Terminal. Everything I `ls` now (PDFs, Pictures, Documents, etc) is just a right-click+open hyperlink away from being opened. I explicitly have to leave tmux now and open a new gnome-terminal window if I want the feature.

It might not be a compelling enough anecdote, but it's one datapoint.  I do appreciate you considering it even though it is more "fringe" at this point.

sencer:
@glensc, which part you didn't understand exactly? I modified README after nicm said he is not interested in merging this PR. So I am not interested in merging it anymore. So no I am not removing it - and where the patch came from is written in that part, if you are interested prepare a PR yourself.

kevr:
> > Right, but they use it for noddy stuff. Nobody seriously needs URLs in gcc or systemd. What about major programs like irssi/weechat? emacs/vim? mutt? What's the status of `less -R` support? What major terminals support it apart from VTE and iTerm2?
> 
> It's a simple use but I use this all the time with `alias ls='ls --color=auto --hyperlink=auto'` in my `.bashrc` and running `ls` inside GNOME Terminal. Everything I `ls` now (PDFs, Pictures, Documents, etc) is just a right-click+open hyperlink away from being opened. I explicitly have to leave tmux now and open a new gnome-terminal window if I want the feature.

> It might not be a compelling enough anecdote, but it's one datapoint. I do appreciate you considering it even though it is more "fringe" at this point.

It's definitely a useful feature. Seeing as how every other shell program out there just tries to regex links and expects that nothing is wrapped, this makes it really false positively ridden for opening links through shell content. Example: `tmux-urlview` + Mutt.

Having a standard for links in general would be absolutely fantastic and definitely a step that Linux needs to make sooner or later.

stuaxo:
Having programming tools is a perfect use for it.

I want to click the file in the stack trace and go straight to the file in my IDE.

lilydjwg:
I have made [a wrapper for ripgrep](https://github.com/lilydjwg/search-and-view) that displays matches with hyperlinks that will bring to the exact place the match is found. It works well with this patch.

thornycrackers:
Kitty supports this https://sw.kovidgoyal.net/kitty/kittens/remote_file.html  and the project has 10.7k stars so i'd say it would be handy to have this feature.

PR #2404, opened 2020-09-30
Make parse_window_name() aware of quoted commands

gpakosz:
I noticed the following:

```
$ tmux setw -g remain-on-exit on
$ tmux new-window "yes | head -n1"
```
```
→ ~ tmux list-panes -s -t tmux-bug -F '#{pane_dead} #{pane_tty} #{pane_id} #{pane_start_command}'
0 /dev/ttys015 %76 
1 /dev/ttys016 %77 "yes | head -n1"
```
image

As you can see, the command is now quoted, which tmux didn't do before but I'm not sure when the behavior changed.

The proposed change adjusts the `name` string in `parse_window_name()` to trim `"` characters.

### DIFF
--- a/names.c
+++ b/names.c
@@ -142,6 +142,9 @@ parse_window_name(const char *in)
 	char	*copy, *name, *ptr;
 
 	name = copy = xstrdup(in);
+	if (*name == '"')
+			name++;
+	name[strcspn (name, "\"")] = '\0';
 	if (strncmp(name, "exec ", (sizeof "exec ") - 1) == 0)
 		name = name + (sizeof "exec ") - 1;
 


nicm:
This is wrong if `name` is a single `"`.

I think you could just do this:
~~~~
if (*name == '"')
    name++;
name[strcspn (name, "\"")] = '\0';
~~~~

gpakosz:
> This is wrong if `name` is a single `"`.

True. But can it be?

Applied and tested your suggestion. It works.

Can this land in `3.2`? 🙏

nicm:
I don't know if it can be or not, but there is no harm in expecting it. I can probably put it into 3.2.

gpakosz:
> I can probably put it into 3.2.

Yes please. I believe it's harmless. Is there a new RC coming?

nicm:
No next will be 3.2 release, no more RCs.

nicm:
Appled now, thanks.

PR #2405, opened 2020-09-30
Add distribution package installation instructions

carlwgeorge:


nicm:
Who does this help? Surely every RHEL user knows how to use yum, every Debian user apt, and so on. None of them have an unusual package name for tmux or anything.

carlwgeorge:
I've seen users that will jump straight to manual compilation steps if distro packages aren't mentioned.  If you like, I can reduce this down to just mentions of the distribution packages, without the package management commands.

nicm:
OK, I don't want to clutter the readme with this, so instead I have added a table to the start of https://github.com/tmux/tmux/wiki/Installing and changed to mention this page from README.md.

nicm:
Thanks!

carlwgeorge:
Works for me, thanks for adding the info to the wiki.

PR #2406, opened 2020-10-02
Fix build if getdtablesize() is not available

fornwall:
Aligns with what is done in compat/closefrom.c.

Fixes building for Android which lacks getdtablesize().

### DIFF
--- a/compat/imsg.c
+++ b/compat/imsg.c
@@ -73,7 +73,13 @@ imsg_read(struct imsgbuf *ibuf)
 again:
 	if (getdtablecount() + imsg_fd_overhead +
 	    (int)((CMSG_SPACE(sizeof(int))-CMSG_SPACE(0))/sizeof(int))
-	    >= getdtablesize()) {
+	    >=
+#ifdef HAVE_SYSCONF
+	        sysconf(_SC_OPEN_MAX)
+#else
+		getdtablesize()
+#endif
+	) {
 		errno = EAGAIN;
 		free(ifd);
 		return (-1);


nicm:
I think this should be done with a compat/getdtablesize.c and a test and AC_LIBOBJ in configure.ac.

nicm:
Actually we already have some local changes to imsg* so maybe this is OK, I'll take a look tomorrow.

fornwall:
> Actually we already have some local changes to imsg* so maybe this is OK, I'll take a look tomorrow.

Great - let me know if I can help out!

Related to that: Do you know how relevant the `HAVE_SYSCONF` fallback is? That is, are there any relevant systems where sysconf(3) is not available?

I've submitted some unconditional getdtablesize->sysconf changes to some other packages in the past and would like to know if I've hurt their portability enough to warrant going back and adding the relevant fallback.

nicm:
I expect any systems with getdtablesize but not sysconf are probably so old that they don't matter, although I don't know for sure, and there is no difference AFAIK. But upstream imsg*.c uses getdtablesize and it is probably not worth the bother of changing it when we can just use a libobj replacement. Does this work for you? [tmux-getdtablesize.diff.txt](https://github.com/tmux/tmux/files/5326492/tmux-getdtablesize.diff.txt)

fornwall:
> Does this work for you? tmux-getdtablesize.diff.txt

Yes, this fixes the Android build as well.

nicm:
Great stuff, applied now, thanks!

PR #2413, opened 2020-10-09
fix note for "previous-window" default key binding

theStack:


nicm:

PR #2448, opened 2020-10-31
Add support for fuzz testing

snizovtsev:
This patch introduces a new entrypoint for fuzzing escape sequence parser.
It is enabled by `--enable-fuzzing` configure flag and shouldn't be used for release builds.
I consider this as a groundwork before integrating with google/oss-fuzz project.


nicm:
Looks good thanks.

But one wider question - what does adding this get us? It is great that you have used this and have found some bugs, but is anyone else going to use it? What is the oss-fuzz project going to do with it? Do they run automated tests? I don't want to add this if it is just going to sit there and nobody will ever use it again.

Also a few comments on the diff:

- Is there a way to get away without the change to tmux.c? How about replacing main entirely by doing this in compat.h:
~~~
#ifdef FUZZING_BUILD
#define main tmux_main
#endif
~~~~
Then providing a stub main in input-fuzzer.c, for example:
~~~~
#undef main
int
main(void)
{
    exit(0);
}
~~~~

- FUZZING_LIBS is set but never used? It would be better to do without it if possible.

- I'd make all the variables FUZZER or FUZZING instead of a mixture. NEED_FUZZER, --enable-fuzzer, FUZZER_BUILD? or FUZZING, I don't mind.

snizovtsev:
> But one wider question - what does adding this get us? It is great that you have used this and have found some bugs, but is anyone else going to use it? What is the oss-fuzz project going to do with it? Do they run automated tests?

Yes, it's like CI but for security. For popular open source projects, Google offers their compute resources for running fuzzers continuously. If it starts crashing under sanitizer they will send an email to project maintainers including the details. Projects like OpenSSL, FFmpeg, LibreOffice, and ImageMagick are fuzzing via OSS-Fuzz. They also offer [rewards](https://www.google.com/about/appsecurity/patch-rewards/) for projects integrated into OSS-Fuzz. 

The process starts when [the manifest](https://github.com/google/oss-fuzz/compare/master...snizovtsev:master#diff-d57fe362ba2a8c65984f2920b647fcee2c1eb810505ffdca73ec5ce5ede1bce2) is merged into https://github.com/google/oss-fuzz repository, but only with maintainer's consent. I'm not sure they will accept tmux but I hope so.

Note that I'm not affiliated with Google, the initiative is mine.

> How about replacing main entirely by doing this in compat.h:

Then `input-fuzzer` binary will link, but tmux will not. The problem is that linking with FUZZING_LIBS implies main() symbol declared outside. But we may want to build both binaries to reproduce crash under sanitizer.

Maybe 

```
#ifdef NEED_FUZZING
#define main __weak main
#endif
```
in `compat.h`? 

> FUZZING_LIBS is set but never used? It would be better to do without it if possible.

It is used in oss-fuzz manifest to alternate between fuzzer engines (libfuzzer, AFL, honggfuzz):
```
./configure \
    --enable-fuzzing \
    FUZZING_LIBS="${LIB_FUZZING_ENGINE} -lc++" \
```
> I'd make all the variables FUZZER or FUZZING instead of a mixture.

It makes sense. I've also renamed AC_FUNC_REALLOCARRAY since we don't own autoconf namespace.

nicm:
OK makes sense, will take another look in a bit, thanks.

nicm:
This looks good to me now, let me know what happens with the ossfuzz upstream issue. Thanks!

snizovtsev:
tmux is accepted now: google/oss-fuzz#4593

nicm:
OK I have applied this now with one change - recallocarray takes more arguments than reallocarray, so the same test can't be used for both. I just split it into two separate tests. Let me know if I have broken anything. Thanks!

snizovtsev:
Thank you! Oss-Fuzz also refers to initial corpus currently located in my personal profile [1]. This is just some random valid inputs picked from other projects' test suites, that helps fuzzer to converge faster. What about moving this under https://github.com/tmux ?

[1] https://github.com/snizovtsev/tmux-fuzzing-corpus

nicm:
I have imported this as https://github.com/tmux/tmux-fuzzing-corpus. Will you need to modify it in future?

snizovtsev:
OK Now all references are under `github.com/tmux`.

> Will you need to modify it in future?

Well, I think it's unlikely. It is processed by `build.sh` anyway, so I can place hooks there.

Thank you for taking the time.

PR #2453, opened 2020-11-04
Haiku first support.

devnexen:
forkpty compat layer based on hpux mainly minus few unsupported ioctl calls.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -58,6 +58,12 @@ if IS_NETBSD
 AM_CPPFLAGS += -D_OPENBSD_SOURCE
 endif
 
+# Set flags for Haiku.
+if IS_HAIKU
+AM_CPPFLAGS += -D_BSD_SOURCE
+endif
+
+
 # List of sources.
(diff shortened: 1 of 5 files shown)


nicm:
Thanks:

- Does this work? I'm not going to apply a work-in-progress.

- Don't do `#ifdef __HAIKU__` for `IMAXBEL` or other termios bits. Please define the missing bits to 0 in compat.h like we do for `ECHOPRT`.

- Don't use `AC_DEFINE` for compiler flags like `_BSD_SOURCE`, add it to `AM_CPPFLAGS` in Makefile.am like we do for NetBSD.

- Please use C style comments (`/* ... */`) not C++ style (`//`);

devnexen:
Yes it does. addresses the rest of your remarks.

nicm:
Great stuff - I have applied this now with a few minor tweaks, let me know if I have introduced any mistakes. Thanks!

devnexen:
so far so good ;-)

PR #2457, opened 2020-11-05
Change default string for status-right

coughingmouse:
Date format was not internationally recognizable. Especially `%b` which East Asians will have hard time understanding. Changed it to be ISO 8601 compliant.

I don't know what I'm doing.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -662,7 +662,7 @@ const struct options_table_entry options_table[] = {
 	  .scope = OPTIONS_TABLE_SESSION,
 	  .default_str = "#{?window_bigger,"
 			 "[#{window_offset_x}#,#{window_offset_y}] ,}"
-			 "\"#{=21:pane_title}\" %H:%M %d-%b-%y",
+			 "\"#{=21:pane_title}\" %Y-%m-%d %H:%M",
 	  .text = "Contents of the right side of the status line."
 
 	},


nicm:
I think it would be more disruptive to change it now after 10 years than live with it. It is easy to configure anyway. If it was to change, I think the date should be removed entirely.

coughingmouse:
Keeping the date was what would keep it from being very disruptive (lengthwise one character diff). How can it be so disruptive to make it useful for more people, if it was already so useless?

It is easy to tinker around for one single machine, but keep in mind that international tmux users probably have more than one to configure. And they probably have other packages to configure for international support (encoding, font, ), again, for every machine. You can't assume users to even configure the feature that should have been given well in the first place.

nicm:
It's not useless and I don't like the date with only numbers. Either it stays with `%b` or it goes entirely.

nicm:
What we could do, is put the locale in a format variable, then it could use an alternative format if `%b` is unsuitable. Although it seems to me that it is up to libc really to pick something sensible for `%b` for the locale.

coughingmouse:
It's not even a matter of like & dislike for non-European population. `%b` is completely useless for them. You're forcing people to learn 12 words that don't even correspond to their etymology. Besides, it is still useless for [pretty much everyone outside the US, the Philippines, and Malaysia](https://en.wikipedia.org/wiki/Date_format_by_country#Usage_map) because of the order of year, month, and date.

If you like arbitrary Latin alphabet combinations instead of simple numerals for month, you could easily configure every tmux you have yourself, as you suggested, instead of rendering it impossible to read for others.

nicm:
`%b` is meant to be a locale-specific representation, it is hardly tmux's fault if the representation in your locale is wrong.

Besides, even if it is wrong, you're being a bit melodramatic. Anyone who doesn't understand the months in English at all is going to have trouble with much of the default output or, for example, the tmux man page. And there have hardly been queues of people complaining about it - in fact you are the first.

coughingmouse:
It is tmux's fault to use `%b` the wrong way. You're assuming that `%d-%b-%y` is a valid local-specific representation (albeit possibly wrong order) but it is not. In East Asian languages, year, month and date all have local-specific representations (e.g. 2020년 10월 19일). But current implementation's `%d-%b-%y` aesthetically looks like crap (19-10월-20; year and date are not only reversed but indistinguishable).

nicm:
Yes, you are right. But it is unfortunate there isn't more variety of locale-aware options. There is no really good one-stop option here:

- 2020-02-01 is ugly and has the least important information first. Including the year at all is a little silly for a status line which is meant to be readable at a glance (who doesn't know what year it is?).

- 01-02-2020 will confuse Americans.

- Anything without the full year like (01-02-20 or 20-02-01) will also confuse everyone eventually.

I am inclined to either remove the date entirely and have only either `%H:%M` or `%a %H:%M`.

coughingmouse:
There ***are*** a few sets of superb one-stop options here, that the world knows and uses already: [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601). Of these, ones that are more human-friendly are of our concern:

* `--MM-DD` is the "without year" variation of the standard. But it is less commonly used and recognized, and the part that it does not include year can be disruptive to some existing tmux users. At least that is why I didn't push this be used.
* Although `YY-MM-DD` is officially a valid format, it surely will confuse Americans and some other people like you've said, but not everyone. This is the default format used in along with `YYMMDD` here.
* `YYYY-MM-DD` is the most neutral looking and intuitive format which is recognizable to everyone. It's also one of the most widely used and recognized format, as well as the one that most closely resembles tmux's original date format. (WLOG)
It is so commonplace that there's also `%F` in C's strftime() which represents this format alongside `%R` which is for `%H:%M`, both of which I did not use for clarity (which now that I think of it might not have been necessary).
It's a very rude and ignorant---in a racist way---thing to say that a language or its characteristics is inferior (especially for the wrong reason, which is very often the case) and your saying that it "looks ugly" does seem quite ludicrous to me. Not just because your concept of beauty in this case might be misleading, but also because the fact that its order is how it is usually represented in East Asian countries and in contrast to your opinion looks normal to a very large population of East Asian countries should pretty much settle this dispute. FYI it looks fine to me. Not just fine, actually; I don't think there could be a better way to represent date without radically changing calendaring system as a whole. 
I dare you; tell me that we also need to make stopwatches show time in `MM:SS:HH` format instead of `HH:MM:SS` because hours are usually meaningless in stopwatches, and we have to put important ones first. If you wish to think of it your way, thinking of it this way might help you understand: with `YYYY-MM-DD HH:MM` you can just read the date at the end and continue reading hour and minute without having to read the rather obvious month and year at all.

As far as I know, this is not a personal project; it is an open source project meant to be used by everyone whose design should be determined by personal tastes. I think it should depend on what is most universally accepted within functionality. Currently, this is a feature COMPLETELY not functional for East Asian users. This was the point for the whole time and I don't see how you can't agree on the fact that it's something that needs a fix.

nicm:
Whether a date format is ugly or not is subjective and I am being neither rude nor racist, and it is obvious that `YYYY-MM-DD` and indeed all your other suggestions have significant usability problems for a status line. I don't think there is any point in discussing this further.

Issue #2460, opened 2020-11-12
[Feature Request] Official AppImage build


michaellee8:
I can see from https://github.com/tmux/tmux/issues/1172 that @nelsonenzo already provides an AppImage build in https://github.com/nelsonenzo/tmux-appimage, and there is an official version of AppImage build in 3.0. However, there are no offical ones from other versions.

By any chance an "official" AppImage build can be provided for newer versions? It would be great if people who need to work in environment with no package manager granted/available to get the latest tmux working without compiling from source (sometimes not even possible if you don't have apt-get). I know there is a third-party one from @nelsonenzo but a official hence verified build would be more trusted.

I guess I can help to do this if this suggestion can be accepted. Perhaps a build job in Travis CI would be enough.

nicm:
Your best bet is to talk to @nelsonenzo, if you automatically build them and then open an issue or email me then I will add them to the release.

michaellee8:
@nicm Please look at my PR

nicm:
Moved to PR.

PR #2465, opened 2020-11-14
Provide an AppImage generation script

michaellee8:
Partially implement #2460 

This us a script I wrote to generate an AppImage for tmux.  It reuses the script I previously wrote for vifm. To make sure that it will work on 256 color terminal (looks like libncurses5 don't have that support yet), while not missing support for older distros, I prevented using system dependencies as in https://github.com/nelsonenzo/tmux-appimage, instead I compile latest version of both libevent and libncurses6 from source to make sure it works for both old and new systems. It also includes its own latest terminfo database so that it should work with newer terminals.

Ideally this script should be ran in an old system to ensure backward compability. It will need pkg-config, autotools, automake, autoconf, curl and imagemagick. It does most things in a /tmp directory that will delete itself automatically when the script exits and only leave the tmux.appimage built in repo root. The script is also designed to be invoked in repo root, somehing like `./tools/genappimage.sh`.

This script is also designed to run in CI to automatically trigger a AppImage build and upload it to GitHub Release in each CI tag build. I have an example for vifm here https://github.com/michaellee8/vifm/blob/master/.travis.yml.


nicm:
OK thanks. The best way to do this would be for you to create a GitHub repository and attach the images you automatically build to it. If you then let me know when there is a new one - either by opening an issue or automatically by email, I will attach them to the releases. But I would suggest you talk to @nelsonenzo who is already building these, because there is little point in both of you doing it.

michaellee8:
> OK thanks. The best way to do this would be for you to create a GitHub repository and attach the images you automatically build to it. If you then let me know when there is a new one - either by opening an issue or automatically by email, I will attach them to the releases. But I would suggest you talk to @nelsonenzo who is already building these, because there is little point in both of you doing it.

Just as previously mentioned, I actually hope that there would be an AppImage build provided officialy by you so that people can know that it is verified by you and therefore are able to trust that. There are some people (like me) who would tend to prevent using non-officialy provided binaries since there are always security concerns there. There are too many ways to get yourself hacked these days so trusting a random guy like me and download and execute a third party provided binary to get usable console access on your 5 years old production server storing confidential data is not ideal. That's why I designed this script to be self-contained without much dependencies so it can be ran in most environments. I just hope that people could gain most convenience without compromising security.

nicm:
OK I don't see how I am any more trustworthy than you unless they are built just as openly, but I don't mind them being built automatically from tmux. I don't want them automatically attached to releases though, I will do that manually.

If you give me something I can apply to .travis.yml and/or .github that will build the packages then I will take a look.

Please keep helper scripts and whatnot in a new directory under .github, that way it is nicely separated.

The tarballs are the releases, so better if it can build from tarballs rather a tag. I don't know if it can be triggered by a release tarball upload; if not then there is a GitHub API to get the list of releases perhaps it could use.

I think best if the packages are stored in a new repository (I can create one for it like tmux/appimage-packages), best would be committed to the repository but if that can't work then added as releases to it would be OK. Or it could open a tmux issue and add the package as a file.

michaellee8:
@nicm I am happy to help but how do you suppose to retrieve the automatically built packages if you are going to attach them manually to GitHub releases?

nicm:
They will need to go into a repository or somewhere I can download them from.

michaellee8:
@nicm It seems that it is possible to upload the build artifacts into a S3/GCS bucket that you wish to, which you can collect that from there but you will need to add the S3 credentials in the repo setting. 

https://docs.travis-ci.com/user/uploading-artifacts

Actually it should be possible to upload it to any destination you want, with some scripts. Unfortunately it does not provide it's own way to store build artifacts like what GitLab does, it requires you to use a external provider.

nicm:
If it can attach to a release in the tmux repository, can't it create a release in a different repository under the same project?

michaellee8:
You mean organization? I think so with some custom scripts, I can help with that.

nicm:
There must be a way for me to pass private credentials securely to a Travis CI build, if I can do that I could pass an ssh key which would let it commit to a dedicated repository. That way it can clearly be seen that the package I attach to the issue is the same as the one that the script built.

nicm:
Or using GitHub Actions instead of Travis CI? I don't know if that would do it.

nicm:
The problem with using a repo is that presumably in order to add new packages it would need to checkout the old ones which might be big and take a while, so a release on a new repo might be better.

nicm:
Or maybe we should just attach it to the main tmux release, I suppose it can't do any harm, I can just delete it if anything is wrong...

nicm:
Forget uploading it - let's just attach to the tmux releases, that's fine, anything else would be too much of a pain in the arse I think. If you make a PR with the changes needed for .travis.yml and .github then I'll take a look. Thanks!

michaellee8:
Ok, I will provide the required changes in a few days.

ThomasAdam:
I'm happy to move our CI to GitHub Actions if that's easier?

michaellee8:
Actually I read your .travis.yml and I can see something like 60 combinations in job matrix, and actually we only need to do the release process on exactly one of them. I guess for convenience stake I can make the release process specifically for GitHub Actions which will not mess up with your current testing infrastructure and you may not even need to provide your own GITHUB_TOKEN since GitHub Actions should have provided one (unsure, let me work that out).

> The tarballs are the releases, so better if it can build from tarballs rather a tag. I don't know if it can be triggered by a release tarball upload; if not then there is a GitHub API to get the list of releases perhaps it could use.

My understanding is that CI builds are only triggered in push, pull request and tag. It is also hard to get the release tarball at the tag time because that tarball should have been not released yet at the time that the "tag build" runs.

nicm:
You could get the latest release from https://api.github.com/repos/tmux/tmux/releases/latest and if it does not have an AppImage package, download the tarball and build one.

michaellee8:
> You could get the latest release from https://api.github.com/repos/tmux/tmux/releases/latest and if it does not have an AppImage package, download the tarball and build one.

In my understanding the tarball should not have exist yet at the time the tag build is run. I suppose the tarball is something you upload manually?

nicm:
Yes but you could just check on every push and it will be there soon enough.

michaellee8:
@nicm I have a draft implementation here, it should not require further setup to work once you merged this PR. Every time you create a release (from the GitHub website for example) it will trigger a GitHub action job that will automatically build the AppImage and then upload it to the Release. What you need to is just to publish it.

nicm:
OK looks good, thanks.

This doesn't matter but I would probably just have downloaded libevent like ncurses rather than using git. Also ncurses does support pkg-config like libevent, but you need extra flags, look at https://github.com/tmux/tmux/wiki/Installing#building-dependencies.

A few things that do need to be looked at:

- I thought you were going to download the release tarball? Where does the tmux source code come from? I can probably write something to work out what release to build for if you want. The reason I would prefer it to look at the releases and work out what to build is mainly for release candidates where I don't create a new release each time, instead I delete the tarball and upload a new one - I would like to just be able to delete the AppImage package as well and it will automatically build a new one.

- This is not enough:
    ~~~~
    # Contributed by michaellee8 <ckmichael8@gmail.com> in 2020
    ~~~~
    Can you please either add a simple "Public domain" (like example_tmux.conf), or add your copyright and a suitable ISC license statement like the C files. I don't mind which. This is only necessary for the genappimage.sh script, not main.yml.

- I think the generated AppImage file needs at least the tmux version in the name. Previously they were called eg `tmux-3.0a-x86_64.AppImage` - including both version and arch seems sensible to me.

-  There is some commented out stuff in the script, can you tidy it up and remove anything that isn't necessary? Or at least add a comment noting why it is commented/when it should be uncommented?

Thanks!

nicm:
Are you going to work on this any further? Thanks!

michaellee8:
> Are you going to work on this any further? Thanks!

Yes, but I have so many things to do recently, I will get back to this soon.

PR #2476, opened 2020-11-24
feat(commands): add command to get session cwd

orthros:
Adds new command: "get-cwd" which will return the
current working directory for the given session.

There are various workflows where it is beneficial
to be able to `cd` to the "root" of a given `tmux` 
session. 

Consider working in a large monorepo
e.g golang/x/build and "scoping" your 
working directory to a subdirectory of that tree.
It may be beneficial, after working in that directory
to be able to `cd` to the session's "root" workspace
with `cd $(tmux gcwd)` .

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -84,6 +84,7 @@ dist_tmux_SOURCES = \
 	cmd-display-panes.c \
 	cmd-find-window.c \
 	cmd-find.c \
+	cmd-get-cwd.c \
 	cmd-if-shell.c \
 	cmd-join-pane.c \
 	cmd-kill-pane.c \
--- a/cmd-get-cwd.c
+++ b/cmd-get-cwd.c
@@ -1,0 +1,65 @@
+/* $OpenBSD$ */
+
+/*
+ * Copyright (c) 2007 Nicholas Marriott <nicholas.marriott@gmail.com>
+ *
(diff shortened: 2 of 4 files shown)


nicm:
This is already available in the `session_path` format variable.

orthros:
Interesting! I had no idea about that; thanks!

Looking at it though I'm somehow not able to format out the `session_path` format variable. 

This is after running `tmux -vv -Ltest -f/dev/null new` as per the doc.

```bash
$ tmux list-sessions -F '#{session_name} foo #{session_path} bar'
0 foo  bar
```

I suspect I have an old version of `tmux` as I do not have the `-f` flag for `list-sessions` 

```bash
$ tmux list-sessions -f '#{session_attached}'
tmux: unknown option -- f
usage: list-sessions [-F format]
```

If I were to have an  updated version of `tmux` however, would running:

```bash
tmux list-sessions -f '#{session_attached}' -F '#{session_path}'
```

produce the desired result? Thanks!

Thanks!

nicm:
~~~~
tmux display -p '#{session_path}'

nicm:
You need master or 3.2-rc for `session_path`.

orthros:
Fantastic! I'll close this and eagerly await the official release of 3.2 (am currently on 3.1c)

PR #2491, opened 2020-12-03
Added PANE_SYNCRONIZE flag for duplicate input to selected panes

Rashpil93:
Hi. In this request, I made it possible to select panes to sync input. I think it will be useful in some cases. I also added the ability to customize the highlight of the selected pane or panes.

### DIFF
--- a/cmd-set-option.c
+++ b/cmd-set-option.c
@@ -149,9 +149,24 @@ cmd_set_option_exec(struct cmd *self, struct cmdq_item *item)
 
 	/* Change the option. */
 	if (args_has(args, 'u')) {
-		if (o == NULL)
+		if (strcmp(name, "synchronize-panes") == 0 && scope == OPTIONS_TABLE_WINDOW) {
+			struct window *w = target->w;
+			struct window_pane *wp;
+			TAILQ_FOREACH(wp, &w->panes, entry) {
+				if (options_get_number(wp->options, "synchronize-panes")) {
+					o = options_get_only(wp->options, name);
+					if (options_remove_or_default(o, idx, &cause) != 0) {
+						cmdq_error(item, "%s", cause);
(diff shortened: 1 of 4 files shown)


nicm:
How is this any different from using `select-pane -d` and `-e`?

Rashpil93:
> How is this any different from using `select-pane -d` and `-e`?

PANE_INPUTOFF prohibits input to pane.

PANE_SYNCRONIZE allows you to select specific panes for input synchronization and input synchronization will only work in those PANES that have PANE_SYNCRONIZE. 
In other panes it will also be possible to enter something and it will not be synchronized with other panes.

Also, setting pane-synchronized-border-style allows you to select a border style for PANE_SYNCRONIZE and you can immediately see which panes are synchronized.

It also remains possible to select all PANEs for synchronization via select-window -S.

nicm:
OK. But I think this should be done differently:

- Instead of removing `synchronize-panes` and adding new flags to `select-pane` and `select-window`, we should change `synchronize-panes` to be a pane option so it can be set individually with `set-option -p` (look at, for example, `allow-rename`).

- `pane-synchronized-border-style` is then unnecessary because you can leave the check for `synchronized-panes` in the default `pane-active-border-style`.

- Instead of `select-window -S`, I suggest adding a flag to `set-option` to unset a pane option on all panes in the target window.

Also please note than "synchronize" is always spelt with an "h".

nicm:
Is this finished now? Thanks

Rashpil93:
Hi. Yes

nicm:
OK I have applied this, but I changed `set-option` to add an extra flag (`-U`) to unset pane options instead of making `synchronize-panes` work differently. It is in OpenBSD now and will be in GitHub later on. Thanks!

PR #2525, opened 2021-01-06
window-copy: remove unused variable

mathstuf:
---
Last usage was removed in c68baaad98807963da954bc1aa77c22f2df05ff0.

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -3009,7 +3009,7 @@ window_copy_search_marks(struct window_mode_entry *wme, struct screen *ssp,
 	struct screen			*s = data->backing, ss;
 	struct screen_write_ctx		 ctx;
 	struct grid			*gd = s->grid;
-	int				 found, cis, which = -1, stopped = 0;
+	int				 found, cis, stopped = 0;
 	int				 cflags = REG_EXTENDED;
 	u_int				 px, py, i, b, nfound = 0, width;
 	u_int				 ssize = 1, start, end;
@@ -3074,9 +3074,6 @@ again:
 			}
 
 			nfound++;
-			if (px == data->cx &&
-			    py == gd->hsize + data->cy - data->oy)
-				which = nfound;
 
 			if (window_copy_search_mark_at(data, px, py, &b) == 0) {
 				if (b + width > gd->sx * gd->sy)


krader1961:
FWIW, I've found that enabling, and paying attention to, most compiler warnings tends to keep these types of oversights from occurring. Also, using tools like `oclint` is a good idea. Having said that, it is not immediately obvious that the block that assigns to `which` should be eliminated -- thus allowing the var to be removed. Is there any possibility the change that eliminated any uses of that var should have retained (or introduced) at least one use? In other words, is the fact the var is no longer used a mistake? With a commit that recent there is the possibility orphaning the var was a mistake.

@mathstuf, Did you actually review commit c68baaa and confirm the var is no longer relevant? Here I am not talking about "relevant" in the sense that there is at least one use of the var in the code. Rather, relevant in the sense that the condition that results in its assignment is still meaningful and should be acted on.

mathstuf:
That is why I dug into the history of the variable (should have mentioned). Unfortunately, the lack of comments makes it hard to follow without knowing what the original usage of the variable is meant to do (the name is…not exactly evocative).

Reading the code more closely than "fixing a warning" scanning, I *believe* that `which` was the count of the search results and the currently selected one (so "this is the 10th out of the 45 results we found" that showed up in the upper-right). Since this is the explicit goal of the commit in question, it really should just be removed if that is correct.

> FWIW, I've found that enabling, and paying attention to, most compiler warnings tends to keep these types of oversights from occurring.

I agree, however numerous PRs of mine have been rejected due to "that is just the compiler being pedantic". That is where most of my warning analysis for tmux now goes :/ .

krader1961:
> Since this is the explicit goal of the commit in question, it really should just be removed if that is correct.

Yes, that was my impression as well; although, I spent only a few minutes looking at that change.

> I agree, however numerous PRs of mine have been rejected...

I feel your pain. I do like that the project owner is very opinionated and quickly closes idiotic issues and proposed changes. However, sometimes, I think that individual is too hasty to reject beneficial changes. Such as leveraging lint tools to identify problems.

nicm:
I fix warnings and am concerned with code quality all the time and if you think otherwise you have really not been paying attention.

tmux has extensive warnings enabled, the only one people keep trying to fix that we deliberately don't pay attention to is GCC's `-Wmaybe-uninitalized` because it is extremely inaccurate and the only way to fix it is to initialize variables unnecessarily which hides warnings when the code changes or the compiler improves.

mathstuf:
Sorry for being slightly hyperbolic, the discussions around those just seem to stick around more in the mind.

I agree that compilers are not wrong in some cases. However, they should probably at least get a comment that "this variable is used, the compiler is just not smart enough to see it". The problem is that if a new codepath is added where it *is* now uninitialized, the warning isn't new and is just as easy to blow off as "the compiler being silly" again even though it is now a *real* warning.

I do wish that compilers would be smarter, but it is also the case that C is not as simple as most developers expect. Compilers are free to reorder code and may be warning about codepaths that only exist post-optimization and don't (naívely) reflect the literal source code that developers are looking at. For example, there's a warning in the codebase today about a potentially uninitialized variable of the pattern:

```c
if (cond)
  var = 1;
else
  some_other_code();
more_code();
if (cond)
  use(var); // warning about potentially uninitialized use of `var`
```

Sure, the code doesn't *look* like `cond` would change (it's a simple `data->var` lookup), but the compiler is probably also taking into account the lack of a mutex to stop some *other* thread (that it cannot prove does not exist) or even `more_code()` from messing with `cond` by either changing `data` or `data->var` itself. Since it cannot prove to itself that `cond` is actually invariant here, it fires the warning. I think doing `int res = cond;` then using `res` would suffice there since the lookup is then in stack memory which the compiler can convince itself "yeah, there's no way that's changing"[1], but haven't tested it.

[1]Within the abstract C machine at least; random pointers could be computed or stack arrays overflowed, but that falls into UB which is just considered "nope, won't happen" in C land.

nicm:
> I agree that compilers are not wrong in some cases. However, they should probably at least get a comment that "this variable is used, the compiler is just not smart enough to see it". 

I don't agree with you here. Adding this kind of comment is clutter and makes code worse. If we add a comment now, when do we remove it? When the compiler fixes the warning? What if they change it and it warns in a different place, do we add a new comment? What about people using a different compiler? It is a distraction and makes code harder to read and understand.

The best way to get rid of these warnings is (if possible) to change the code around so that the compiler can work out that it is not used or is initialized, which usually makes it easier for people to understand as well. This is what we did to remove warnings in resize.c, and in some other places.

> The problem is that if a new codepath is added where it is now uninitialized, the warning isn't new and is just as easy to blow off as "the compiler being silly" again even though it is now a real warning.

Right, but if you initialize it and then move the code around, the compiler can't tell the variable is unitialized even if it is obvious - so instead of getting a warning you end up with a null pointer deref.

> For example, there's a warning in the codebase today about a potentially uninitialized variable of the pattern:

What codebase? tmux? What warning?

> Sure, the code doesn't look like cond would change (it's a simple data->var lookup), but the compiler is probably also taking into account the lack of a mutex to stop some other thread (that it cannot prove does not exist) or even more_code() from messing with cond by either changing data or data->var itself.

I don't really know what you are saying here, AFAIK no C compiler at the moment takes multithreaded memory access into account in uninitialized access warnings (it would be pretty much impossible) so mutexes don't come into it. I don't think any check allocated or global data except for the most simple cases either.

nicm:
What compiler are you using and seeing these warnings? I don't see anything to worry about with clang 10.0.1, clang 12.0.0, or gcc 9.3.0.

mathstuf:
gcc 10.2.0 (in Fedora 33)

krader1961:
> I fix warnings and am concerned with code quality all the time and if you think otherwise you have really not been paying attention.

@nicm, I know you are concerned with code quality. That is one reason I switched from `screen` to `tmux`. I, also, do not see any compiler warnings about the unused `which` variable (on macOS using its clang compiler). The lack of a compiler warning for the unused `which` local variable surprises me. Which makes me wonder if other compiler flags or a `make lint` build target (that runs a tool like oclint) might not have some value.

nicm:
Compiler warnings are not perfect or even that good. The trouble with lint and variants is that they are often worse and you are likely to have a large number of false positives to dig through. I have run splint on the codebase occasionally, but by all means run other tools if you like. Static analysis tools like scan-build are better but even it has a number of false positives. Coverity is meant to be quite good too but it is not free and setting it up for a free project turned out to be a pain IIRC.

nicm:
Here are their instructions for Travis CI which would probably not be that hard to integrate since we already use it perhaps I will do it some point, https://scan.coverity.com/travis_ci.

nicm:
I haven't forgotten about this but I don't have GCC 10 and I haven't had time to build it.

nicm:
OK I took a look at GCC 10.2 warnings. One of them is correct (in server-client.c).

The others are because GCC doesn't seem to understand for this warning when a switch has a case for every value of an enum. I have fixed these in a couple of ways.

Two of them (in the options code) there are obvious "default" groups of values, so I just grouped them together as a default case. This means we won't see any warnings if we add new values to the enum but that is OK I think. 

In window-copy.c there is a default value that can be assigned to the variable.

That leaves only the two warnings for strftime for which there isn't really a good solution, and a silly warning for `%y` which we can remove by doing `-Wno-format-y2k`, although we'll have to see if that fails on any older GCC versions.

These changes will be in GitHub later on. I also applied your fix a while back so it should be there too. Thanks!

PR #2546, opened 2021-01-22
Add copy-mode commands to explicitly turn rectangle selection on or off.

wm-noble:
Currently, users can only toggle rectangle selection, but cannot set it explicitly. This would allow you add something like the following to .tmux.conf:

    bind -T copy-mode-vi v send -X rectangle-off \; send -X begin-selection
    bind -T copy-mode-vi C-v send -X rectangle-on \; send -X begin-selection

to make `copy-mode-vi` even more similar to vim's visual mode keybindings, rather than requiring you to press space after toggling rectangle selection. Default behavior is unchanged.

I've tested this out locally and it works as expected.


nicm:
Great stuff, applied to OpenBSD now, will be in GitHub later. Thanks!

PR #2553, opened 2021-01-30
Add a new option to detach-on-destroy

sencer:
A new option, "smart", to not detach on destroy if there is any detached tmux
sessions. If all available sessions are attached, destroying detaches.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -71,6 +71,9 @@ static const char *options_table_window_size_list[] = {
 static const char *options_table_remain_on_exit_list[] = {
 	"off", "on", "failed", NULL
 };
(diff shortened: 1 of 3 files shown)


nicm:
Good, but "smart" is not very meaningful, how about "all-attached" or something like that?

sencer:
Yes, I didn't like `smart` that much either. I don't see why you suggest `all-attached` though? Is `off-when-detached` too verbose?

nicm:
Oh yeah I was thinking of something else so all-attached doesn't make much sense.

What about just no-detached?

sencer:
Sounds good!

nicm:

PR #2558, opened 2021-02-06
Report client focus in 'client_flags'

da-x:
For the case when a long running command has ended running in some tmux session's pane, this lets me rig a script that tells whether this pane is currently focused on any attached client. I can see if its session attached in one of the clients, and then see if that client has focus. If no currently focused client sees this pane, I can send myself an popup/email or some other kind of notification.


nicm:

Issue #2280, opened 2020-06-22
3.2 release candidates discussion


nicm:
### Ask or discuss any quick questions here about 3.2 release candidates only please.

Bug reports please open a new issue with full information, see [CONTRIBUTING](https://github.com/tmux/tmux/blob/master/.github/CONTRIBUTING.md).

3.2-rc is a release candidate for 3.2.

**[The CHANGES file is here](https://raw.githubusercontent.com/tmux/tmux/3.2/CHANGES).**

---

There will be a **lot** of changes in this release, these are the most notable, for more see the CHANGES file:

- How data is sent to control mode clients has been completely revamped to both be more fair with multiple panes and to prevent huge amounts of data being backed up; this feature is used by default when tmux is attached with the latest iTerm2 nightly builds.

- Configuration file parsing has changed slightly: the contents of the new {} syntax introduced in 3.1 must now be valid tmux command syntax; and to allow formats to be annotated, strings given with quotes may now contain newlines (leading spaces and comments are stripped).

- A new customize mode available with `C-b C` (`C-b S-c`) which allows options and key bindings to be browsed and changed interactively.

- Support for extended keys offered by some terminals (xterm, mintty, iTerm2), see https://github.com/tmux/tmux/wiki/Modifier-Keys#extended-keys.

- A `pane-border-lines` option to change the characters used to draw the pane border separators.

- How UTF-8 data is stored has been rewritten to reduce memory use for characters in the BMP.

- The message log (`C-b ~`) has been changed to be per server instead of per client and to have some useful content.

- Client flags have been extended (`refresh-client -F` becomes `-f` and also added for `attach-session` and `new-session`), and a `read-only` flag added (same as `-r` to `attach-session`).

- A new `active-pane` client flag that if given allows a client to have its own active pane for each window rather than being tied to the server's active pane.

- Improved `command-prompt` completion including showing a menu of completions.

- All style options can now be formats, for example the default `pane-active-border-style` now changes colour depending on `pane_in_mode` and `synchronize-panes`.

- Search match count has come back in copy mode, as well as performance improvements and additional styles for marking of search terms.

- Try `$XDG_CONFIG_HOME/tmux/tmux.conf` as well as `~/.config/tmux/tmux.conf`.

- Window and pane hooks such as `window-layout-changed` and `pane-exited` are now window or pane options instead of session options.

[...]

PR #2561, opened 2021-02-08
add pipe, pipe-no-clear and pipe-and-cancel commands

laktak:
As suggested in #2280 ;-)

Please let me know if this is what you had in mind.


nicm:
Looks good, but if it is not creating a buffer then it doesn't need the prefix argument?

laktak:
Right, I removed the prefix.

nicm:
Great stuff, applied to OpenBSD now, will be in GitHub later. Thanks!

PR #2571, opened 2021-02-16
chore: define default terminal capabilities overrides

HallexCosta:
- [x] Setting default terminal overrides not to break when showing true colors in tmux.

## TMUX Version: Pre-release (3.2-rc3)

**Broken colors**
GIF 15-02-2021 23-55-43

**True colors**
GIF 15-02-2021 23-22-38

**Tmux with neovim**
GIF 15-02-2021 22-10-18

> P.S.: My tmux does not work the configuration file (~/ .tmux.config), I am not aware of the reason for the problem, it simply ignores this file, I cannot set override settings in the terminal or configure keybinds.

**.tmux.config**
```bash
# (~/.tmux.config)

set-option -ga terminal-overrides ",xterm-360color:Tc"

bind-key r source-file ~/.tmux.config \; display-message "~/.tmux.config reloaded"
```

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -305,7 +305,7 @@ const struct options_table_entry options_table[] = {
 	  .type = OPTIONS_TABLE_STRING,
 	  .scope = OPTIONS_TABLE_SERVER,
 	  .flags = OPTIONS_TABLE_IS_ARRAY,
-	  .default_str = "",
+	  .default_str = ",xterm-256color:Tc",
 	  .separator = ",",
 	  .text = "List of terminal capabilities overrides."
 	},


nicm:
No. There is no guarantee that terminals which use `xterm-256color` have RGB colours. Configure this in `~/.tmux.conf` using `terminal-overrides` or `terminal-features`.

HallexCosta:
I tried to use the .tmux.config configuration file, but I can't get the terminal settings and keybinds to work, I don't know if using WSL gets in the way of reading the file, the only way it worked for me was this.

>System:
WSL version: 1.0
WSL Kernel: Ubuntu 18.04 LTS
Tmux: 3.2-rc3 
Neovim: 0.4.4

P.S.: Tested in the tmux version ( 2.6,  3.1b, 3.1c ), but I was also not successful.

nicm:
The file is called .tmux.conf not .tmux.config.

HallexCosta:
> The file is called .tmux.conf not .tmux.config.

**Solved!**
Lack of attention mine, sorry my english not very god.

PR #2573, opened 2021-02-18
Fix chdir warnings during compile

ericcurtin:
It's about an ignored return value, log errno if chdir fails

### DIFF
--- a/job.c
+++ b/job.c
@@ -113,8 +113,11 @@ job_run(const char *cmd, struct session *s, const char *cwd,
 		sigprocmask(SIG_SETMASK, &oldset, NULL);
 
 		if (cwd == NULL || chdir(cwd) != 0) {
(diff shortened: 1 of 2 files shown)


nicm:
I think it should just _exit(1) like the other failures.

ericcurtin:
fatal does an exit(1) right? This code ok?

nicm:
That's ok but fatal will automatically add errno so just `chdir(\"/\") failed` is enough.

nicm:
I'll do it tomorrow, don't worry about updating if you don't want to.

ericcurtin:
Done, I was in front of the computer with a terminal still open so it only took 10 seconds to repush. Feel free to edit again!

nicm:
Applied a similar change to OpenBSD now, will be in GitHub later. Thanks!

Issue #2575, opened 2021-02-19
Please add windows support for tmux


soroshsabz:
ITNOA

I think it is very good and useful for windows community to add support windows for tmux, and tmux can run directly on PowerShell in windows.

How we can help you for doing this?

nicm:
This is unlikely to happen, the Windows terminal is just too different. Also see https://github.com/tmux/tmux/issues/2368

nicm:
I suppose we could do it without ANSI escape sequences, but it would not be easy.

krader1961:
Note that the Windows 10 console has support for ANSI escape sequences; although, it has to be enabled by calling SetConsoleMode() with the ENABLE_VIRTUAL_TERMINAL_PROCESSING flag. See https://docs.microsoft.com/en-us/windows/console/setconsolemode as well as numerous StackOverflow questions on this topic. The [Elvish shell](https://elv.sh) enables that mode to produce styled output without having to otherwise know anything about MS Windows. There are, of course, other quirks such as a different signal handling model that may be problematic for tmux.

In short, it's probably possible to make a native Windows 10 version of tmux, without littering the code with lots of `#if WIN32` directives, but it will have some limitations. Such as not being able to dynamically enable/disable tracing via SIGUSR2. It's also wholly impractical to support older versions of MS Windows.

soroshsabz:
I think Windows 10 support is just enough. 😊

skyline75489:
As @krader1961 pointed out, Windows console now supports VT sequences just as normal Linux terminals. And there's the brand new [Windows Terminal](https://github.com/microsoft/terminal) as a replacement to conhost.exe to improve the overall console experience on Windows. You can use a large amount of ANSI escape sequences and it would work as expected, if you set the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag.

The major thing stopping tmux from being available on Windows is that tmux depends on ncurses, which does not have Windows support, for now. @nicm I'd like to purpose reopening this issue.

nicm:
Are you going to work on it? If so then feel free to ask me questions, I will reopen the issue when things are clearly moving.

skyline75489:
I might give it a try to see if it's possible to create an MinGW build of tmux. I'm an outside contributor of Windows Terminal so I know it's possible to have tmux on Windows and I'd personally love to see it happen.

nicm:
[...]

PR #2582, opened 2021-02-24
Organize a bunch of headers into compat.h

skyline75489:
Supports #2575

### DIFF
--- a/client.c
+++ b/client.c
@@ -17,9 +17,6 @@
  */
 
 #include <sys/types.h>
-#include <sys/socket.h>
-#include <sys/un.h>
-#include <sys/wait.h>
 #include <sys/file.h>
 
 #include <errno.h>
--- a/client.c
+++ b/client.c
@@ -18,7 +18,6 @@
 
 #include <sys/types.h>
 #include <sys/socket.h>
-#include <sys/uio.h>
 #include <sys/un.h>
 #include <sys/wait.h>
 #include <sys/file.h>
--- a/cmd-find.c
+++ b/cmd-find.c
@@ -18,7 +18,6 @@
 
 #include <sys/types.h>
 
-#include <fnmatch.h>
 #include <limits.h>
 #include <stdlib.h>
 #include <string.h>
--- a/cmd-if-shell.c
+++ b/cmd-if-shell.c
@@ -18,7 +18,6 @@
  */
 
 #include <sys/types.h>
-#include <sys/wait.h>
 
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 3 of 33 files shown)


skyline75489:
@nicm How do like the approach of this PR? Believe me I hate `#ifdef` just as much as you do. We probably still gonna have them but I'll try limit the usage of `#ifdef` to the minimum.

nicm:
I think you should make `compat-win32.h` for the structs and whatnot.

I don't mind moving headers about if necessary but I won't do it until you have something working and we can see what it looks like.

skyline75489:
Well I have to be honest, there's a loooong way before I can visually have "something working". The code compiles with a bunch of `#ifdef`. Linking still fails, because there's a lot of things need to be ported to Windows.

Overall I think the porting process can be divided into the following steps:

* [X] Make it compile with all the header tweaks. 
* [ ] Port tty related code with ConPTY
* [ ] Port socket related code with WinSock
* [ ] Port event related code with libevent if possible
* [ ] Port everything else
* [ ] Compile & Link should work.
* [ ] It would like 100% crash but let's see how to fix them
* [ ] Fix a lot of things that does not work
* [ ] Maybe we can make it run if nothing blocks it

 It's like my day 2 of the porting process, so I think I can call it a day by just passing the compilation.

skyline75489:
Having `compat-win32.h` is good idea. Will do. 

My local repo is a mess (but hey, it compiles), so I just picked the header changes here.

nicm:
Yes I'm sure it will need a lot of tidying to be finished, do feel free to feed back smaller changes if they are improvements for all platforms but I don't want to do all these header changes or start adding `#ifdef WIN32` right now. Thanks

skyline75489:
Would it be OK if I still do the imports organizing, which is in fact "improvement for all platforms"? I found some imports should be decided by `compat.h` for better compatibility. If it's cool with u, I'll drop the win32 things in this PR and only leave the reorganizing .

nicm:
What do you mean imports?

skyline75489:
I mean `#includes`. Sorry, wrong language (`#Objective-C`). Pretty must just to have most Linux headers in `compat.h`, and drop the Win32 things

nicm:
I don't want to move them all about at the moment, they are fine as they are for every Unix platform.

I meant more if you have to move code around or add helper functions or anything that can apply to any platform, rather than stuff that is only needed for Windows.

We can move the includes around pretty easily later and it would be a good first step but I'd like to wait until you have something a bit further along.

skyline75489:
That's very fair. Thanks for the feedback. I'll close this for now.

nicm:
Great thanks, let me know if you run into any problems or have any questions. If you get it building and working I can take a look and we can work out how best to tidy it up to merge it at that point.

PR #2589, opened 2021-02-28
Add conditional check for del_curterm

skyline75489:
The same check exists in line 530:

https://github.com/tmux/tmux/blob/ba9f89c44e3294284493178fb34f864aafb399cd/tty-term.c#L530

### DIFF
--- a/tty-term.c
+++ b/tty-term.c
@@ -695,7 +695,10 @@ tty_term_read_list(const char *name, int fd, char ***caps, u_int *ncaps,
 		(*ncaps)++;
 	}
 
+#if !defined(NCURSES_VERSION_MAJOR) || NCURSES_VERSION_MAJOR > 5 || \
+    (NCURSES_VERSION_MAJOR == 5 && NCURSES_VERSION_MINOR > 6)
 	del_curterm(cur_term);
+#endif
 	return (0);
 }
 


ThomasDickey:
which terminfo implementation (for which tmux actually works) lacks `del_curterm`?

nicm:
I think was added because del_curterm crashed on some old ncurses (maybe only on Solaris?), and it was never worth figuring out why. But why do you need it?

ThomasDickey:
Offhand, it's used for freeing memory that's no longer needed.    ncurses 5.6 predates tmux by a few years, anyway.

nicm:
No I meant why did @skyline75489 need to add the conditionals.

ThomasDickey:
That was essentially my question :-)

skyline75489:
Well I do not have del_curterm on Windows so I need the check to remove it.

skyline75489:
Actually the entire terminfo database is not available on Windows but that's another issue. 

I figured this is a missing check for other platforms, too. So I added it here. But if it's not really needed， feel free to ignore this PR.

nicm:
For Windows without terminfo you will need to add compat versions of setupterm/del_curterm/tigetstr/etc that work from a fixed terminfo (`ansi` I expect).

nicm:
Yes we should either add this ifdef or remove the other one, I'll take a look during the week.

nicm:
If you can work out how to do without SCM_RIGHTS then terminfo is not a big problem.

nicm:
The first del_curterm should have been removed and there is no harm in adding the ifdefs for the others. Thanks!

PR #2590, opened 2021-03-01
Guard possible NULL active pane in default_window_name

skyline75489:


nicm:
Applied to OpenBSD (with another fix also) now, will be in GitHub later. Thanks!

PR #2599, opened 2021-03-05
Use evutil_* for socket operation

skyline75489:
This is the first step towards socket operation compatibility on Win32.

It should have no effect on *nix OSes.

### DIFF
--- a/client.c
+++ b/client.c
@@ -59,7 +59,7 @@ static struct client_files client_files = RB_INITIALIZER(&client_files);
 
 static __dead void	 client_exec(const char *,const char *);
 static int		 client_get_lock(char *);
-static int		 client_connect(struct event_base *, const char *,
+static evutil_socket_t  client_connect(struct event_base *, const char *,
 			     uint64_t);
 static void		 client_send_identify(const char *, const char *,
 			     char **, u_int, const char *, int);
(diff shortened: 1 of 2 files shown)


skyline75489:
I don't really know C style indentions. Seems the project is using tabs? Pardon me for maybe doing something wrong.

skyline75489:
Hi @nicm , just want you to know that after some investigation and experiments I found there's almost no way to add support of Windows without adding tons of hacks inside tmux. I decided to pivot to adding support for tmux control mode in WIndows Terminal, which also will help the overall experience with tmux on Windows.

Thanks you for all the helps and explanation. I'll close this PR now. If someone in the future is interested in this, I bet they can benefit from the explanation you patiently wrote.

nicm:
OK no worries, thanks.

nicm:
Control mode is based around what iTerm2 has needed so far, but there is no reason we can't extend it if other terminals need different things, let me know.

skyline75489:
Just want to point there's still a long way to implement tmux Control Mode in Windows Terminal, because issues in the infrastructure layer. But generally it's possible. I have hacked an early POC together and found a lot of problems on the Terminal side 😅 

Thanks again for being supportive of my work. I'll try to push tmux CC on Windows as far as I can go.

Issue #2603, opened 2021-03-08
Previous search result and current search result are highlighted if they are next to one another


aaronjensen:
### Issue description

From `tmux -vv -Ltest -f/dev/null new`:

`echo eeeeeeeeeee`
`C-b [`
`C-r eee`

<img width="226" alt="CleanShot 2021-03-08 at 08 49 43@2x" src="https://user-images.githubusercontent.com/8588/110337180-3e0ccb00-7feb-11eb-9c68-b89ef2469789.png">

Both the preceding and current search results are highlighted when the cursor is at the beginning of one result and after another.

[tmux-client-6078.log](https://github.com/tmux/tmux/files/6102124/tmux-client-6078.log)
[tmux-out-6080.log](https://github.com/tmux/tmux/files/6102125/tmux-out-6080.log)
[tmux-server-6080.log](https://github.com/tmux/tmux/files/6102126/tmux-server-6080.log)

### Required information

```
Darwin i386
tmux next-3.3
tmux-256color
```

topcat001:
This is an interesting bug! I'll take a look this evening and report back. Thanks you for reporting.

topcat001:
Actually, this seems expected given the current code. If we search for `eee` then the cursor is positioned just after the match, and it highlights `eee`. However in that case the cursor is already on the next match and hence it is also highlighted. In other words, the highlight space is `eee.`. I don't see how we can consistently change this without also changing the conventions for positioning the cursor.

aaronjensen:
@topcat001 yeah, I'm sure it has to do with https://github.com/tmux/tmux/commit/df7fbcd7a5a775586522380a70530721d1f2c151 being introduced, which is what introduced the two different (forward and reverse) conventions. 

It seems like it should only be highlighting the term based on the current search direction. That is, if the direction is forward it should highlight the term before. If the direction is backwards, it should highlight the term beginning with the cursor point.

topcat001:
Yes this now works correctly.

PR #2604, opened 2021-03-08
Search again in VI mode behaves like VI

aaronjensen:


window-copy.c
```diff
@@ -3083,10 +3101,19 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
-	if (direction)
+	/*
+	 * Behave according to mode-keys. If it is emacs, search forward leaves the
+	 * cursor after the match. If it is vi, the cursor remains at the beginning of
+	 * the match, regardless of direction.
+	 */
+	keys = options_get_number(wp->window->options, "mode-keys");
+	if (direction) {
+		if (keys == MODEKEY_VI)
+			window_copy_move_right(s, &fx, &fy, wrapflag);
```
aaronjensen, line 3112: In order to fix what https://github.com/tmux/tmux/commit/df7fbcd7a5a775586522380a70530721d1f2c151 was fixing as well, I need to move right here the length of the current search result. 

@nicm is there a way for me to get that here?

window-copy.c
```diff
@@ -3083,10 +3101,19 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
 		endline = gd->hsize + gd->sy - 1;
+	}
 	else {
-		if (again)
+		if (keys == MODEKEY_VI || again)
```
aaronjensen, line 3116: Aside: the name of `again` is confusing to me and it doesn't appear to be set when I would expect. For example, AFAICT, an initial `C-s` sets `again` to 1, but it's not searching "again".

nicm:
`again` is meant to be true if it is repeating the search, for example `search-again` or an incremental search. It looks like it might be an error that it is set in both cases in `window_copy_cmd_search_forward_incremental`, `window_copy_cmd_search_backward_incremental` doesn't set it for both.

IIRC the length of the match should be returned from `window_copy_search_jump` in `foundlen`. Is that what you need?

nicm:
You can't know the length of the current search result before you have searched because. If you want the length next time you will need to save it like the other bits and pieces (`searchstr` etc).

Don't you just want it to work like it did before? Isn't it enough just to make the changes only happen if != VI?

aaronjensen:
Before we didn't have regex searching. Now, with regex searching, it's possible for `n`, or `search-again` to jump to something contained within the current search. It looks to be, according to the commit message, the primary thing you were trying to fix with https://github.com/tmux/tmux/commit/df7fbcd7a5a775586522380a70530721d1f2c151 ("...so that regex searching doesn't overlap when searching forwards")

As far as I can tell, in order to do that, we need to either have the cursor already to the right of the search result, or we need to have it jump to the end of the search result before `search-again`.

window-copy.c
```diff
@@ -3097,26 +3118,40 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
+	 * cursor after the match. If it is vi, the cursor remains at the beginning of
+	 * the match, regardless of direction.
+	 */
+	keys = options_get_number(wp->window->options, "mode-keys");
+	if (direction) {
+		if (keys == MODEKEY_VI) {
+			if (data->foundlen <= 0)
+				window_copy_move_right(s, &fx, &fy, wrapflag);
+			else
+				for (i = 0; i < data->foundlen; i++)
```
aaronjensen, line 3132: @nicm could you provide some guidance on this please? I'd like to only move all the way right of a match if the cursor is currently positioned at the beginning of the match.

I'm having a hard time wrapping my head around the various coordinate systems and data structures that tmux uses, so I don't know where to begin when it comes to figuring this out.

window-copy.c
```diff
@@ -3097,26 +3118,40 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
-	    wrapflag, direction, regex, &foundlen);
+	    wrapflag, direction, regex);
 	if (found) {
 		window_copy_search_marks(wme, &ss, regex, visible_only);
-		if (foundlen != 0) {
+		if (keys == MODEKEY_EMACS && data->foundlen != 0) {
 			/* Adjust for wrapped lines eating one right. */
-			i = data->cx + foundlen;
+			i = data->cx + data->foundlen;
 			while (i > gd->sx - 1) {
```
aaronjensen, line 3150: Also, @nicm can you explain what's going on here, please?

nicm: I didn't write this bit, @topcat001 did, he can probably tell you.

nicm:
I think you should walk the searchmarks arrays to find the start or end of the current match and position the cursor where you want it to be for the search.

aaronjensen:
> I think you should walk the searchmarks arrays to find the start or end of the current match and position the cursor where you want it to be for the search.

As I mentioned, I don't really know what coordinates mean what, so it's really hard for me to just do this. Many of the variables have truncated names, their comments aren't clear enough for me, there's the concept of grids which I don't fully understand and there's the concept of what's onscreen vs off. 

Is there some documentation somewhere that I could read through to familiarize myself with the concepts in code?

What value points to the current position of the cursor? 

What coordinate system is that in? 

Is it the same coordinate system that `fx` and `fy` are in? Could I simply assign those new values?

How do I tell if I'm at the beginning of a searchmark? Is there no concept of the "current" search mark?

nicm:
> Is there some documentation somewhere that I could read through to familiarize myself with the concepts in code?

No just the code. You need to know how tmux represents the grid. There are two coordinate systems.

An x,y can either be offset from the grid which is the entire pane contents, look at grid.c. It goes from 0 to hsize + sy - 1 where hsize is the size of the history and sy is the size of the visible portion. fx and fy are based on grid coordinates in this function.

The other coordinates are only on the visible screen and they go from 0,0 to sx - 1,sy -1 where sx and sy are the visible width and height.

> What value points to the current position of the cursor?

The cursor is data->cx and data->cy. You can see in window_copy_search it converts the cursor position on the visible screen into a grid position by adding on the amount of history that is above it (ie, the total size of the history - the copy mode scrolled up offset in data->oy):
~~~~
        fx = data->cx;
        fy = screen_hsize(data->backing) - data->oy + data->cy;
~~~~

> Is it the same coordinate system that fx and fy are in? Could I simply assign those new values?

You can assign fx,fy new values and that is where window_copy_search_jump will search from.

> How do I tell if I'm at the beginning of a searchmark? Is there no concept of the "current" search mark?

The searchmark array is a big array with one byte for each character of the total content (history and visible). Adjacent bytes in a match have the same value with zero meaning not a match, so if you have "abctexttextdef" then array might contain "00011112222000".

You can use window_copy_search_mark_at to get an element from the array, you can use that to walk backwards or forwards from fx,fy or anywhere near it to find the start or end of the match. If you do this you don't need to worry about wrapping or line lengths, because a element in an array cannot be part of a match unless it is OK for the cursor to end up there.

nicm:
Try something like this:
~~~~
--- window-copy.c       9 Mar 2021 13:07:50 -0000       1.320
+++ window-copy.c       11 Mar 2021 17:22:10 -0000
@@ -64,6 +64,8 @@ static int    window_copy_search_rl(struct
 static int     window_copy_last_regex(struct grid *, u_int, u_int, u_int,
                    u_int, u_int *, u_int *, const char *, const regex_t *,
                    int);
+static int     window_copy_search_mark_at(struct window_copy_mode_data *,
+                   u_int, u_int, u_int *);
 static char    *window_copy_stringify(struct grid *, u_int, u_int, u_int,
                    char *, u_int *);
 static void    window_copy_cstrtocellpos(struct grid *, u_int, u_int *,
@@ -3105,6 +3107,24 @@ window_copy_search(struct window_mode_en
                endline = 0;
        }

+       u_int at, start;
+       if (data->searchmark != NULL &&
+           window_copy_search_mark_at(data, fx, fy, &start) == 0 &&
+           data->searchmark[start] != 0) {
+               if (direction) {
+                       // move right
+               } else {
+                       log_debug("XXX %u %u %u", data->searchmark[start], fx, fy);
+                       while (window_copy_search_mark_at(data, fx, fy,
+                           &at) == 0) {
+                               log_debug("XXX %u %u %u", data->searchmark[at], fx, fy);
+                               if (data->searchmark[at] != data->searchmark[start])
+                                       break;
+                               window_copy_move_left(s, &fx, &fy, wrapflag);
+                       }
+               }
+       }
+
        found = window_copy_search_jump(wme, gd, ss.grid, fx, fy, endline, cis,
            wrapflag, direction, regex, &foundlen);
        if (found) {

aaronjensen:
Thank you, I appreciate the detailed response. I'll give this a shot and report back as soon as I have a chance.

nicm:
It is one of the most complicated parts of tmux to change, copy mode and searching has been edited a lot over time and so it is quite untidy and it has to deal with a lot of complicated things like regexs and marking matches and not be too slow.

window-copy.c
topcat001: When a line is wrapped, and you do a cursor right, then the cursor will stop at one past the end before moving on to the next line (it displays a '$' in that case). That's what I mean by "eating" one right. This means that if the line wraps n times, you need to insert n additional rights in order for the cursor to be positioned just after the match in this case. Does that help?

aaronjensen:
@nicm okay, this seems to work well for me. It behaves as identically to vim and emacs both as far as I've been able to tell. It does away with `again` when reverse searching by, in all cases, scanning to the left if we are on a search mark until we're not on that one anymore, and then scanning again to left until the beginning of the current search mark if we landed on one.

In vi mode, it even respects moving to the middle of a search mark and then doing an `n` or `N`. It does what vim does, that is, it treats the current mark as current and bypasses it when searching. This improves on the current reverse search again behavior in master right now.

I'll also probably look at #2603 and at least fix it in vi mode. In emacs mode, I believe it would require persisting the last search direction.

window-copy.c
```diff
@@ -3097,19 +3116,60 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
 	wrapflag = options_get_number(wp->window->options, "wrap-search");
 	cis = window_copy_is_lowercase(str);
 
-	if (direction)
+	if (direction) {
+		keys = options_get_number(wp->window->options, "mode-keys");
+		/*
+		 * Behave according to mode-keys. If it is emacs, search forward leaves the
```
aaronjensen, line 3122: What is the standard tab width and line width in this codebase?

nicm:
https://man.openbsd.org/style

...

"Indentation is an 8 character tab. Second level indents are four spaces.
All code should fit in 80 columns.

Do not add whitespace at the end of a line, and only use tabs followed by
spaces to form the indentation. Do not use more spaces than a tab will
produce and do not use spaces in front of tabs."

window-copy.c
topcat001: I think https://man.openbsd.org/style is what we follow.

nicm:
Looks good on a quick read, will take a better look next week

nicm:
Thanks!

window-copy.c
aaronjensen: Thanks, formatting adjusted accordingly.

aaronjensen:
Also updated to fix search highlighting for Emacs. It's pretty strict in that if you're searching forward it will only highlight the term if the cursor is just after it. From my testing I didn't see a way to keep the highlights visible like they are in vi mode, so this seems to behave fine, but let me know if I'm missing something.

nicm:
This reads fine to me, @topcat001 does it look OK to you?

topcat001:
Hi I did a little bit of testing mostly in vi mode, and I see one problem. Let's suppose the text on screen is `aaa|aaa|aa` and the search pattern is `aaa`. The pipes are not part of the text but "virtual", added to separate generations of search marks (so for this search the search marks array will be `111|222|00`. Now you see the issue: a backwards search from the end will put the cursor on the 6th `a` whereas I think we want it on the 4th `a` - that's what vim does.

This can be fixed during the character by character backward search by checking if we are inside a marked region and if so, by moving to the start of the region like we do in the other instances. Hope this makes sense. @aaronjensen  it should be relatively simple to fix it.

topcat001:
I think I got the order wrong. The initial cursor positioning happens before marking, so the adjustment to the start of the marked region should happen after `window_copy_search_marks` has run.

topcat001:
Was just testing emacs mode. The above issue is present in this case as well. I noticed a second issue in this case:  while searching backwards by pressing `N` during a forward `C-s` search, when the cursor is on the first `a` the first triplet `aaa` is not highlighted. I think this due to the check in `window_copy_update_style` checking only the `searchtype`; it should check for the actual search direction instead (which can be reverse even if `searchtype` is `WINDOW_COPY_SEARCHDOWN`).

aaronjensen:
I saw this as well, but since I believe it reproduces on master I considered it a separate issue. I'll see if I can look at it a bit, but I think the only way to really solve it may be to compute search marks first and then jump based on them, but I haven't put a lot of thought into it.

aaronjensen:
Oh, @topcat001 how is N bound in emacs mode? It seems like the problem you're describing may be that that does not end the cursor in the same place the initial search does. I don't fully understand that repro.

topcat001:
You are right about the first issue being present in master. I'm okay with it being unaddressed in this commit. I have a few more updates coming up and I can fix it there.

`n` and `N` are bound to `search-again` and `search-reverse`, respectively in both vi and emacs modes. To reproduce it, you can use my example pattern, and then start a forward search using `Ctrl-S`. Then do `N` and you'll see the highlighting does not work correctly when the cursor is positioned at the start of the pattern.

If on the other hand you use `Ctrl-r` to start the search, and continue with `n` then the highlighting is correct. So `Ctrl-s` and then `N` behaves differently from `Ctrl+r` followed by `n`. They both search backwards and should behave identically.

aaronjensen:
> They both search backwards and should behave identically.

I agree, though should the behavior actually be that N should move the cursor and keep the match the same the way that C-r does?

topcat001:
I think it's a little confusing if it doesn't. Consider the following pattern: `xxxxbcdxxxbcdyyyy`. If I now search forward for `bcd` using `Ctrl-s`, it finds and highlights `bcd` in magenta (using default colours). If I copy the selection using `Ctrl-W` then the magenta selection is copied.

Now if I move the cursor to the last `y` and then press `N`, the cursor jumps to the `b` and `bcd` highlighted in cyan. If I press `N` again it jumps to the first match, again in cyan. So, nothing is highlighted in magenta now. If I press `Ctrl-W` now I know the match under the cursor would be copied (determined by `window_copy_match_at_cursor`) but it's not magenta. In master this scenario works they way it should: the current match is highlighted in magenta and the other one in cyan, and it appears broken currently.

I think one quick way to fix this is store the actual match direction in `data` perhaps and use that in `window_copy_update_style` instead of `searchtype`. Then the behaviour is consistent.

topcat001:
So the `reverse-` command functions could set a `data->reverse` flag (and the forward functions would unset it) and then the check both `searchtype` and `reverse` to determine if it is really searching "down" in `window_update_style` in the if. I believe then we'll get the correct behaviour.

topcat001:
Also, note that highlighting and marks generation always searches forward, so any backward search should be consistent with that regarding cursor positioning, otherwise we will see glitches like this.

aaronjensen:
@topcat001 Thanks, I see the problem now. I originally stored direction in search data, but found `searchtype`, thinking it would have what I needed. Thanks to you, I see that it doesn't change when using search-reverse because that's actually how you know what to do with search-reverse. I updated the commit to store `direction` as `searchdirection` and use it instead.

topcat001:
Thanks for the update. This works fine! I'm ok with `searchdirection`. If @nicm is ok I have no further comments.

aaronjensen:
About the other issue of reverse searching, I believe one approach would be to reverse search as we do now, then apply search marks, then scan backwards until we're at the beginning of the mark, similar to how I scan before jumping. I think that'd work... should I add that in or do you want to cover that later?

aaronjensen:
That edge case exists with forward searching as well. If your cursor is on the a at the end of the second group: `aaa|aaa|aa` and you `C-s  aaa` it selects the end `a|aa`. This is actually consistent with how Emacs works. Emacs always considers the position of the cursor when determining the next jump.... it does not appear to resolve marks from the beginning of the buffer and jump you along them. 

It's not how vim works, however. Vim does jump you along resolved marks.

Also, Vim doesn't consider the point you are at as part of the search, which I actually regressed in this patch on the initial search. I'll work on correcting that.

topcat001:
Yes, I did notice the forward case as well. Getting the emacs marking behaviour would require changes to `window_copy_search_marks` so I'm okay with keeping the vim marking behaviour at the moment, unless you feel strongly about it.

The cursor is first positioned during the initial search by `window_copy_search_jump`, then `window_copy_search_marks` does the marking. This is around line 3175 in `window_copy.c` with your patch applied. Then the line wrap ("eating right" comment) adjustment happens.

One idea is to replace the block in lines 3176-3185 (master+this PR's patch) with code that adjusts the cursor depending on `mode-keys`, `direction`, and where the cursor happens to be situated relative to a marked area. Obviously this is a little tricky and will need some testing, but I think its probably not too much trouble. Please feel free to give it a try if you want, thanks.

aaronjensen:
@topcat001 I gave it a shot. I switched to using `window_copy_move_right` and `window_copy_move_left` to compute the cursor position so I did not have to do with the wrapping markers, but that may be a bad idea. It allowed me to get rid of `foundlen` and be consistent with how I positioned the cursor after the search.

You'll also note that I call `window_copy_search` recursively in the instance that a forward search leaves us in the middle of a mark. That's an indicator that we actually need to search again, so we do.

Please give it a solid test.

aaronjensen:
Oh, and I amended the first commit to correct the regression when the first search in vim mode did not move the cursor right once: https://github.com/tmux/tmux/pull/2604/files#diff-188efdf3fcfd0c57782cab5dadca1cc3cc7c94018f342ad58f5d1f4e1b1c9634R3127-R3143

topcat001:
Thank you for the update. I'm looking at it right now and will report back soon.

topcat001:
Hi, I had a chance to test the changes. The recursive call to search can sometimes crash due to infinite recursion. One way to reproduce it is by doing an incremental search in emacs mode while using a pattern like `aaa|aaa|aa` and typing `aa...` at the prompt. If you do this you'll easily see why it recurses infinitely.

I think in the recursive call we should make two changes:
- If we are not at the start of a highlighted area, move to one past the end (much like how emacs forward search works) before starting the nested search.
- Call `window_copy_search_jump` instead of `window_copy_search` in line 3191. Since the marking is already done we do not want to do it again, and also avoid running the other marking logic.

Apart from this issue and some minor style nits it looks good. However I am unable to test it fully yet because of the above crash issue.

aaronjensen:
@topcat001 thanks for the testing and the suggestion. Corrected. Feel free to comment inline on the style nits so I can learn to be more consistent w/ the expected style and/or use GitHub to propose the changes.

topcat001:
Thanks, looks ok after some initial testing, I have added a couple of style comments inline. In addition to those, there are some long lines (>80 columns). You can find them in vim for example by searching for `\%>80v.\+`. I'll keep testing but for now this looks good.

aaronjensen:
@topcat001 did you submit the review? I do not see the style comments.

As for the long lines, the indentation is pretty deep, so breaking the lines down produces some pretty rough looking code, but I can take a crack at it.

topcat001:
Sorry, review started. I can still see a few lines with a character in column 81. One suggestion to avoid ugly code due to deep indents is to factor them out into small functions. For example, the code which moves right till it reaches the end of the marked area, and similar code to find the beginning could be written as small functions. You can see this pattern elsewhere in the code.

aaronjensen:
@topcat001 I refactored some of the cursor movement into separate functions. It probably warrants some more testing because so much moved around. I also fixed a couple of potential issues having to do with variables being set conditionally when they should not have been (like `keys`).

window-copy.c
```diff
@@ -3097,27 +3152,97 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
+				data->cy = fy - screen_hsize(data->backing) +
+				    data-> oy;
+			}
+		}
+		else {
+			/*
+			 * When searching backward, position the cursor at the
+			 * beginning of the mark.
+			 */
+			if (window_copy_search_mark_at(data, fx, fy,
```
aaronjensen, line 3230: I tried to extract this too but ran into a strange bug I couldn't track down or consistently reproduce on my first reverse search (the cursor would end up after the mark, rather than at the beginning)

window-copy.c
```diff
@@ -72,15 +74,20 @@ static int	window_copy_search_marks(struct window_mode_entry *,
 		    struct screen *, int, int);
 static void	window_copy_clear_marks(struct window_mode_entry *);
 static void	window_copy_move_left(struct screen *, u_int *, u_int *, int);
+static void	window_copy_move_right(struct screen *, u_int *, u_int *, int);
+static void	window_copy_move_before_search_mark(
```
aaronjensen, line 78: I wasn't sure if these were necessary, but it seemed like other ones that weren't strictly necessary were declared, so i followed suit.

window-copy.c
```diff
@@ -3048,31 +3068,66 @@ window_copy_search_jump(struct window_mode_entry *wme, struct grid *gd,
...
+static void
+window_copy_move_before_search_mark(struct window_copy_mode_data *data,
+    u_int *fx, u_int *fy, int wrapflag)
+{
+	struct screen  *s = data->backing;
+	u_int		at, start;
+
+	if (window_copy_search_mark_at(data, *fx, *fy, &start) == 0 &&
+	    data->searchmark[start] != 0)
+		while (window_copy_search_mark_at(data, *fx, *fy, &at) == 0 &&
```
topcat001, line 3085: Enclose the `while` in braces.

window-copy.c
```diff
@@ -3048,31 +3068,66 @@ window_copy_search_jump(struct window_mode_entry *wme, struct grid *gd,
...
+static void
+window_copy_move_after_search_mark(struct window_copy_mode_data *data,
+    u_int *fx, u_int *fy, int wrapflag)
+{
+	struct screen  *s = data->backing;
+	u_int		at, start;
+
+	if (window_copy_search_mark_at(data, *fx, *fy, &start) == 0 &&
+	    data->searchmark[start] != 0)
+		while (window_copy_search_mark_at(data, *fx, *fy, &at) == 0) {
```
topcat001, line 3102: Enclose the `while` in braces.

window-copy.c
```diff
@@ -3048,31 +3068,66 @@ window_copy_search_jump(struct window_mode_entry *wme, struct grid *gd,
...
+window_copy_move_before_search_mark(struct window_copy_mode_data *data,
+    u_int *fx, u_int *fy, int wrapflag)
+{
+	struct screen  *s = data->backing;
+	u_int		at, start;
+
+	if (window_copy_search_mark_at(data, *fx, *fy, &start) == 0 &&
+	    data->searchmark[start] != 0)
+		while (window_copy_search_mark_at(data, *fx, *fy, &at) == 0 &&
+		       at > 0) {
```
topcat001, line 3086: I think we should drop the test for `at > 0`. It is true that there is no marking "before" `a==0` but there can be a position to move to before it if the mark starts at `a==0`. If you position a marked area to start at the top left of the visible screen and search up (easiest to check this in vi mode), you'll see that the search gets "stuck" there because of this check.

window-copy.c
aaronjensen: Ah, so the rule is more than 1 line, rather than more than one statement? I got it, thanks.

window-copy.c
aaronjensen: Got it, updated.

window-copy.c
krader1961: I can't speak for the code style preferred by @topcat001 or @nicm but I always require braces around code blocks that follow `if`, `else`, `while`, `for`, etc. In my four decades of programming I have personally found two bugs due to missing braces. More importantly, I find the inconsistency makes code harder to read. So even this simple block requires braces, IMHO:

```
if (true) {
    do_something();
}
```

The only exception is when the block fits on a single line:

```
if (true) do_something();
```

topcat001: We try to follow [style(9)](https://man.openbsd.org/style): "Use a space after keywords (if, while, for, return, switch). No braces are used for control statements with zero or only a single statement unless that statement is more than a single line, in which case they are permitted". Typically OpenBSD source has braces around all multiline statements, and omits them for simple one-liners.

window-copy.c
topcat001: Interesting. I combined the "move_before_start" and "move_to_start" functions and it seems to work fine. Please feel free to test the attached diff and incorporate the change if it works for you. The changed function is `window_copy_move_before_search_mark`. I will test this further with your latest diff.
[a.diff.txt](https://github.com/tmux/tmux/files/6176775/a.diff.txt)

aaronjensen: This diff doesn't use fx/fy to change the cursor position. I also don't know the code well enough to know what happens if you move before the 0th position in the grid and then move right again with wrap off. Would it end up on the 2nd character of the mark?

And I'm personally not a fan of passing a boolean to a function to fundamentally change its behavior, I see that as two functions. I'm good leaving this as-is, if you are.

topcat001: Sounds good. I'm ok with this being as it is now.

topcat001:
Looks good. One very minor point is that if the cursor is in the middle (not the start) of a highlighted area in vi mode and I search backwards then because of line 3184 it will jump to the previous highlight and not to the start of the current highlighted area. This differs from vi. However I'm ok with the current behaviour.

window-copy.c
```diff
@@ -3097,27 +3153,97 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
+			}
+		}
 		endline = gd->hsize + gd->sy - 1;
+	}
 	else {
-		if (again)
-			window_copy_move_left(s, &fx, &fy, wrapflag);
+		/* Scan to just before current search mark. */
+		if (data->searchmark != NULL)
+			window_copy_move_before_search_mark(data, &fx, &fy,
```
topcat001, line 3184: Just noting a (minor) deviation from vim here, mentioned in my last comment.

aaronjensen: Hmm, good catch. It was my intent to emulate vim's behavior with this. eb360c1 should do that.

window-copy.c
```diff
@@ -3097,27 +3134,94 @@ window_copy_search(struct window_mode_entry *wme, int direction, int regex,
...
+				 * When there are no search marks, start the
+				 * search after the current cursor position.
+				 */
+				window_copy_move_right(s, &fx, &fy, wrapflag);
+			}
+		}
 		endline = gd->hsize + gd->sy - 1;
+	}
 	else {
-		if (again)
```
aaronjensen, line 3103: It's unclear to me what this flag was for in this case. In what situations do we not want to backtrack before searching again?

window-copy.c
topcat001: Perfect! Less code is more!

window-copy.c
topcat001: This made sense in the context of the old behaviour. In `master`, vi mode, if you have the pattern `xxx|b|bbxxxbbbbyyy` and the cursor is on the "piped" `|b|` and you search `?bbb` then the cursor won't move and you will get the `|b|` as the first match, since `again==0`. The `n`ext search will have `again==1` and will jump backward (and wrap). Since this is not what vim does (and now we do what vim does) we don't need this flag anymore.

topcat001:
@nicm @aaronjensen I can't see any obvious issues, so ok from me.

window-copy.c
```diff
@@ -3048,31 +3066,50 @@ window_copy_search_jump(struct window_mode_entry *wme, struct grid *gd,
...
+window_copy_move_after_search_mark(struct window_copy_mode_data *data,
+    u_int *fx, u_int *fy, int wrapflag)
+{
+	struct screen  *s = data->backing;
+	u_int		at, start;
+
+	if (window_copy_search_mark_at(data, *fx, *fy, &start) == 0 &&
+	    data->searchmark[start] != 0) {
+		while (window_copy_search_mark_at(data, *fx, *fy, &at) == 0) {
+			if (data->searchmark[at] != data->searchmark[start])
```
topcat001, line 3084: We should also break if `wrapflag` is 0 and we are at the end of the grid. Otherwise this will cause an infinite loop. This can be reproduced by starting a program which displays a string at the bottom edge of the pane (such as another tmux) and searching for that string.

aaronjensen: Ah, how do I detect the end of the grid?

topcat001: Same as `window_copy_move_right`: `fx == gd->sx - 1; fy == gd->hsize + gd->sy - 1`.

aaronjensen: Amended dd4796d

topcat001: This is now fixed, thanks!

aaronjensen:
@topcat001 @nicm Does this look good now? Anything else needed to get this merged?

Thanks!

nicm:
I'll try to have a look this week, I expect it will be fine.

nicm:
Applied to OpenBSD now, will make to GitHub later. Thanks both!

aaronjensen:
Thank you!

PR #2614, opened 2021-03-15
(feature): Add client-detached message to control-mode

### LINKED ISSUE #2598: Add a command/flag to continually display-message
### Issue description

This might be out of scope for tmux as a project but I'm gonna ask for it anyways.

At the moment tmux has a `display-message` command that can be used to print the result of some format string to stdout. I'd like to ask for an extension to this command either as a new-flag for it or as a new command altogether which can re-calculate and print that message at a set interval.

The use case for this is because in my status-line config I've got a little icon (next to the date) which changes depending on the active panes command.

demo

The way I've implemented this at the moment is to have a [script](https://github.com/mohkale/dotfiles/blob/b7a64e624777545fe28f33466f56c236398db1a9/prog/tmux/cmds/tmux-icons) periodically run `tmux display -t "$session" -p '#{pane_current_command}'` which I then pipe elsewhere to calculate an appropriate icon to display on the status-line.

My current approach works well enough (although I'm not a fan of spawning so many tmux commands continuously) however it isn't very convenient or maintainable. Simply looking at my message-log the vast majority of logs seem to just be me querying this value.

Screenshot_20210305_045638

I'd like to replace my script with a single call to `tmux display-message`, maybe of the form `tmux display-message -r 5 -t "$session" -p '#{pane_current_command}` which every 5 seconds (`-r` for repeat cause there's already a `-I` flag which I don't want to confuse with) prints out the value of the format string passed to it. This would mean the command should never exit, which personally I'm fine with but there may be complications in tmux itself preventing that.

Thoughts?

### Required information

* tmux version (`tmux -V`): tmux next-3.3
* Platform (`uname -sp`): Linux unknown
* $TERM inside and outside of tmux (`echo $TERM`): st-256color (outside) screen-24bit (inside)
* Logs from tmux (`tmux kill-server; tmux -vv new`). Not really necessary here.

mohkale:
Closes #2598

This commit adds an extra message to command-mode to indicate when a client has detached from tmux.

I wasn't sure whether there was any need to distinguish between the current client detaching and an external client detaching (such as the session changes message has both %session-changed for the current command-mode client and %client-session-changed for another client doing so) so this commit doesn't distinguish between the current and another client.

@nicm 

I believe this is what you meant when you said to take the same route as subscription-changed. I've tested on my own machine and this does echo message prior to a client detaching. Please let me know if there're any other improvements to be made.

### DIFF
--- a/control-notify.c
+++ b/control-notify.c
@@ -176,8 +176,8 @@ control_notify_client_detached(struct client *cc) {
 	struct client	*c;
 	TAILQ_FOREACH(c, &clients, entry) {
 		if (CONTROL_SHOULD_NOTIFY_CLIENT(c)) {
-            control_write(c, "%%client-detached %s", cc->name);
-        }
+			control_write(c, "%%client-detached %s", cc->name);
+		}
 	}
 }
 
--- a/control-notify.c
+++ b/control-notify.c
@@ -172,6 +172,16 @@ control_notify_client_session_changed(struct client *cc)
 }
 
 void
+control_notify_client_detached(struct client *cc) {
+	struct client	*c;
+	TAILQ_FOREACH(c, &clients, entry) {
+		if (CONTROL_SHOULD_NOTIFY_CLIENT(c)) {
+            control_write(c, "%%client-detached %s", cc->name);
+        }
+	}
+}
+
+void
 control_notify_session_renamed(struct session *s)
 {
 	struct client	*c;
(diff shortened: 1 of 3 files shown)


nicm:
I don't think you need to distinguish own client because you will get `%exit`.

mohkale:
@nicm 

All right. Lemme know if anything else needs to be done.

nicm:

Issue #2592, opened 2021-03-02
3.2 release candidates discussion (part 2)


nicm:
### Ask or discuss any quick questions here about 3.2 release candidates only please.

Bug reports please open a new issue with full information, see [CONTRIBUTING](https://github.com/tmux/tmux/blob/master/.github/CONTRIBUTING.md).

3.2-rc is a release candidate for 3.2.

**[The CHANGES file is here](https://raw.githubusercontent.com/tmux/tmux/3.2-rc/CHANGES).**

---

There will be a **lot** of changes in this release, these are the most notable, for more see the CHANGES file:

- How data is sent to control mode clients has been completely revamped to both be more fair with multiple panes and to prevent huge amounts of data being backed up; this feature is used by default when tmux is attached with the latest iTerm2 nightly builds.

- Configuration file parsing has changed slightly: the contents of the new {} syntax introduced in 3.1 must now be valid tmux command syntax; and to allow formats to be annotated, strings given with quotes may now contain newlines (leading spaces and comments are stripped).

- A new customize mode available with `C-b C` (`C-b S-c`) which allows options and key bindings to be browsed and changed interactively.

- Support for extended keys offered by some terminals (xterm, mintty, iTerm2), see https://github.com/tmux/tmux/wiki/Modifier-Keys#extended-keys.

- A `pane-border-lines` option to change the characters used to draw the pane border separators.

- How UTF-8 data is stored has been rewritten to reduce memory use for characters in the BMP.

- The message log (`C-b ~`) has been changed to be per server instead of per client and to have some useful content.

- Client flags have been extended (`refresh-client -F` becomes `-f` and also added for `attach-session` and `new-session`), and a `read-only` flag added (same as `-r` to `attach-session`).

- A new `active-pane` client flag that if given allows a client to have its own active pane for each window rather than being tied to the server's active pane.

- Improved `command-prompt` completion including showing a menu of completions.

- All style options can now be formats, for example the default `pane-active-border-style` now changes colour depending on `pane_in_mode` and `synchronize-panes`.

- Search match count has come back in copy mode, as well as performance improvements and additional styles for marking of search terms.

- Try `$XDG_CONFIG_HOME/tmux/tmux.conf` as well as `~/.config/tmux/tmux.conf`.

- Window and pane hooks such as `window-layout-changed` and `pane-exited` are now window or pane options instead of session options.

[...]

Issue #2563, opened 2021-02-08
Feature request: Hyperlinks (continued)


Mellbourn:
This feature request is a continuation of https://github.com/tmux/tmux/issues/911 which was opened in May 2017 and locked in March 2020.

tmux does not implement hyperlinks using the OSC 8 format, see https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda 

The previous feature request discusses many points
* Is it simple or not to implement? Some answers imply that it would be simple.
* Some answers make it sound that it should already work "FAIK tmux has always ignored unknown escape sequences.", but it does not work (at least on iTerm and Terminal on macOS)
* Whether or not terminal hyperlinks will be widely implemented or used. In my opinion, it has become a common feature and well supported by many terminal emulators, see link above.

There exists at least one fork of tmux supporting hyperlinks https://github.com/dandavison/tmux

So, I think that this feature deserves a second look. If there is a better forum to bring this up, please advice. The previous request is the top Google hit if you search for "tmux hyperlinks"

nicm:
https://github.com/tmux/tmux/pull/2403#issuecomment-707674012

PR #2621, opened 2021-03-21
Add support for hyperlinks

wm-noble:
This is mostly just the patch from #911, with an extra commit to add the link ID to `grid_extd_entry` and copy it to and from grid cells so that the hyperlink is not lost when entering copy mode and scrolling the window.

I realize you're on the fence about this change. Consider this another attempt to persuade you to adopt it since I would personally find it quite useful.


bluss:
Further discussion: #2563 and https://github.com/tmux/tmux/issues/2592#issuecomment-791622592, https://github.com/tmux/tmux/pull/2403#issuecomment-707674012 - any chance that some of those concerns can be addressed?

I think for a minimal version it would be ok to ignore the `id` parameter.

wm-noble:
Sure, I can look into these concerns. @nicm, could you clarify what you meant by "Existing links need to be reused instead of added"? Is that referring to links with the same ID as existing links?

nicm:
Links with the same content. You will probably need a tree by content as well as by id.

PR #2628, opened 2021-03-27
Fix musl compile error in proc.c

siho22:
```
proc.c:20:10: fatal error: sys/queue.h: No such file or directory
   20 | #include <sys/queue.h>
```

This was already fixed in 7b085136, but got reintroduced with ce5de765.


nicm:
Applied, thanks!

PR #2631, opened 2021-04-01
Fix missing comma punctuation

finn0:
While mentioning different options for appearance and behavior of tmux in the man page, we've missed comma punctuation to separate list of options. After changes, manpage looks like this:
```console
OPTIONS
     The appearance and behaviour of tmux may be modified by changing the
     value of various options.  There are four types of option: server
     options, session options, window options, and pane options.
```


nicm:
Applied, thanks!

finn0:
Thanks! for creating and maintaining tmux.

PR #2648, opened 2021-04-13
man: fix prefix/suffix confusion

mathstuf:
The example and the code both look for a prefix, not a suffix.


nicm:
But in `N/w`, `w` is a suffix, no?

mathstuf:
Ah, I was seeing it "attached" via `:` to the window name, not via `/` to the "command name". This makes sense.

PR #2651, opened 2021-04-14
Add fallbacks to the run tests to fix crosscompiling

hasso:


nicm:
Applied, thanks!

voltagex:
Left a comment on the commit itself, this doesn't seem to compile: https://github.com/tmux/tmux/commit/bb6242675ad0c7447daef148fffced882e5b4a61#commitcomment-49559990

nicm:
autoconf version?

voltagex:
> 
> 
> autoconf version?

autoconf (GNU Autoconf) 2.69 - from Alpine

nicm:
OK not sure what is up, I will revert this for the moment.

lopsided98:
I'm not sure how to fix this. When I opened the link to the commit comment a few minutes ago, the comment was there, but now it seems to have disappeared, so I don't even know exactly what the error was. The user has deleted their account, so I can't ask them.  I don't see how this patch could cause an error, and it works fine for me with autoconf 2.70.

nicm:
Alright well I shall apply it again and see if anyone else complains.

lopsided98:
As long it doesn't break native builds, it can't really hurt since cross-compiling doesn't work without it either.

nicm:
Done. Thanks!

lopsided98:
Thanks! I noticed though that there is another instance of `AC_RUN_IFELSE` introduced in https://github.com/tmux/tmux/commit/b2588eed03640dc74ffdbebd10c719064b575291 that will need a similar patch.

PR #2655, opened 2021-04-15
Check whether strtonum function actually works

tgyurci:
The `strtonum` function is available from macOS 11.0, and it is declared in the corresponding Xcode SDK.  But when using the newest SDK with macOS 10.x then `configure` detects falsely that `strtonum` is available.  The build will complete (with warnings about the unavailable) but when the resulted `tmux` binary tries to call `strtonum` then it crashes with Dyld error message: Symbol not found: _strtonum.

This patch checks whether `strtonum()` really works by executing the configure test binary.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -150,9 +150,17 @@ AC_REPLACE_FUNCS([ \
 	strlcpy \
 	strndup \
 	strsep \
-	strtonum \
 ])
 AC_FUNC_STRNLEN
+
+AC_MSG_CHECKING([for working strtonum])
+AC_RUN_IFELSE([AC_LANG_PROGRAM(
+         [#include <stdlib.h>],
+         [return (strtonum("0", 0, 1, NULL) == 0 ? 0 : 1);]
+ )],
+ [AC_DEFINE(HAVE_STRTONUM) AC_MSG_RESULT(yes)],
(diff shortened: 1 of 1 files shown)


nicm:
Where *is* `_strtonum`? Would we be better to link against whatever library has it?

tgyurci:
I don't know the internal details, I'm just guessing that `_strtonum` is the real implementation from the base os and `strtonum` is just a stub provided by `stdlib.h` from the SDK.  I believe that the binary should link to `strtonum` and to not ignore the warning about not yet implemented function:

```
arguments.c:369:7: warning: 'strtonum' is only available on macOS 11.0 or newer [-Wunguarded-availability-new]                                        
        ll = strtonum(value->value, minval, maxval, &errstr);                                                                                         
             ^~~~~~~~                                                                                                                                 
./compat.h:309:12: note: 'strtonum' has been marked as being introduced in macOS 11.0 here, but the deployment target is macOS 10.15.0                
long long        strtonum(const char *, long long, long long, const char **);                                                                         
                 ^                                                                                                                                    
arguments.c:369:7: note: enclose 'strtonum' in a __builtin_available check to silence this warning                                                    
        ll = strtonum(value->value, minval, maxval, &errstr);
             ^~~~~~~~                                                      
```

nicm:
Applied now, thanks!

PR #2660, opened 2021-04-17
Avoid unportable `==` test(1) operator in configure

iamleot:
The `==` test(1) operator is not POSIX and not all test(1) implementations support it.
Consistently use `=` test(1) operator as done for the rest of `configure.ac`.

Noticed when updating `misc/tmux` package in [pkgsrc](https://www.pkgsrc.org/) via `pkgsrc/mk/check/check-portability.sh` script.


nicm:
Great, applied now, thanks!

iamleot:
Thank you!

Issue #2672, opened 2021-04-22
3.2 regression: enum values need to fit in 32 bits


ericvw:
Previous issue: https://github.com/tmux/tmux/issues/864 back in 2017.  When this surfaced, https://github.com/tmux/tmux/issues/864#issuecomment-295849226 was mentioned as the fix on `master`.

On SunOS Sparc (`uname -sp`), for tmux >= 3.2, typing in the terminal does not result in any output.

`git bisect` indicates this merge introduces the issue https://github.com/tmux/tmux/compare/a46916b45250ef21d22d7a1fc1bb7a89bd0a3645...588865152a1bb70d77758b68a2869b794aa31dd4.  

The C language standard requires that all values fit in an int.  However, it is implementation-defined behavior for the compiler to choose either signed or unsigned `int`.  [`gcc` defaults to `unsigned int` if there are no negative values ](https://gcc.gnu.org/onlinedocs/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html#Structures-unions-enumerations-and-bit-fields-implementation)specified, which may be why this issue didn't surface earlier.

It appears the path forward is to either change `KEYC_BASE` to a different value or use global constants.  

@nicm, how would you like to proceed?  I'm happy to contribute a patch once I know how you would like to address this.

nicm:
key_code is explicitly unsigned:

    typedef unsigned long long key_code;

If KEYC_BASE is being stored in an int, that code should be changed to use key_code.


From: "Eric N. Vander Weele" ***@***.***>
Reply to: tmux/tmux ***@***.***>
Date: Thursday, 22 April 2021 at 15:04
To: tmux/tmux ***@***.***>
Cc: Nicholas Marriott ***@***.***>, Mention ***@***.***>
Subject: [tmux/tmux] 3.2 regression: enum values need to fit in 32 bits (#2672)


Previous issue: #864<https://github.com/tmux/tmux/issues/864> back in 2017. When this surfaced, #864 (comment)<https://github.com/tmux/tmux/issues/864#issuecomment-295849226> was mentioned as the fix on master.

On SunOS Sparc (uname -sp), for tmux >= 3.2, typing in the terminal does not result in any output.

git bisect indicates this merge introduces the issue a46916b...5888651<https://github.com/tmux/tmux/compare/a46916b45250ef21d22d7a1fc1bb7a89bd0a3645...588865152a1bb70d77758b68a2869b794aa31dd4>.

[...]

PR #2673, opened 2021-04-22
Change special key code constants to macros instead of enums

ericvw:
The C standard specifies that enumeration values must fit within a
signed integer.  However, it is implementation defined what type the
compiler may choose.  For the Sun Studio compiler, that type is `int`
whereas `gcc` defaults to `unsigned int`.

This change avoids value overflow by changing the constants to be
defined as macros.

Closes: #2672


ericvw:
@nicm, this is just a draft/preview of the approach I took.  I had to expand `KEYC_MOUSE_KEY` manually.  Before I continued with getting everything formatted appropriately, I wanted to get your feedback before polishing up the patch.

nicm:
I think it is unnecessary to have an enum at all, why not just:
~~~~
#define KEYC_FOCUS_IN (KEYC_BASE + 0)
#define KEYC_FOCUS_OUT (KEYC_BASE + 1) (or KEYC_FOCUS_IN + 1 if you prefer that)
...
~~~~
Unless you can come up with some cpp trick you will need to expand out `KEYC_MOUSE_KEY` which is unfortunate, but you are having to do that anyway.

Or I wonder if we could squeeze the special keys in somewhere else... there is space in Unicode from 0x80 to 0xa0 and the PUAs at 0xe000, 0xf0000 and 0x100000, but would using them be asking for trouble? Maybe it would be OK, what do you think?

We could swap things around so 0x00 to 0xff are special keys and Unicode is encoded higher up. But it is kind of unintuitive so it seems like it would be really asking for bugs.

nicm:
Maybe we use something like `0x10F000` in the middle of the highest private use area, it kind of seems unlikely anything is going to be sending Unicode keys from the PUA range and will happen to hit the same ones tmux is.

Sorry I only thought of this after you had done this!

nicm:
Something like this, does the compiler still complain?
~~~~
--- a/tmux.h
+++ b/tmux.h
@@ -109,11 +109,16 @@ struct winlink;
 #define VISUAL_ON 1
 #define VISUAL_BOTH 2

-/* Special key codes. */
-#define KEYC_NONE            0x00ff000000000ULL
-#define KEYC_UNKNOWN         0x00fe000000000ULL
-#define KEYC_BASE            0x0001000000000ULL
-#define KEYC_USER            0x0002000000000ULL
+/* No key or unknown key. */
+#define KEYC_NONE            0x000ff000000000ULL
+#define KEYC_UNKNOWN         0x000fe000000000ULL
+
+/*
+ * Base for special (that is, not Unicode) keys. An enum must be at most a
+ * signed int, so these are based in the highest Unicode PUA.
+ */
+#define KEYC_BASE            0x0000000010e000ULL
+#define KEYC_USER            0x0000000010f000ULL

 /* Key modifier bits. */
 #define KEYC_META            0x00100000000000ULL

nicm:
Actually this change is not enough on its own because there are places in the code that check for example `< KEYC_BASE`; these need to change or some Unicode keys will be incorrectly excluded.

nicm:
How about this? [tmux-keys.diff.txt](https://github.com/tmux/tmux/files/6364565/tmux-keys.diff.txt)

ericvw:
> I think it is unnecessary to have an enum at all, […]

I agree the enum is not necessary.  Instead of hardcoding incrementing offsets, I found it easier to use the enum so I could script generating the `#define`'s in my editor.

> How about this?
> [tmux-keys.diff.txt](https://github.com/tmux/tmux/files/6364574/tmux-keys.diff.txt)

I'm trying this right now…

ericvw:
Using the `tmux-keys.diff` patch, I am able to now type in the terminal of the `tmux` session.

However, upon startup I receive the following initial upon invocation before shell initialization:

```
(null):0: unknown key: C-Up                                                                                                   [0/0]
(null):0: unknown key: C-Down
(null):0: unknown key: C-Left
(null):0: unknown key: C-Right
(null):0: unknown key: C-Up
(null):0: unknown key: C-Down
(null):0: unknown key: C-Up
(null):0: unknown key: C-Down
```

nicm:
Hmm that's weird, I'll take a look.

nicm:
Whoops I must have made a change after I tested, can you try this instead please: [tmux-keys2.diff.txt](https://github.com/tmux/tmux/files/6382127/tmux-keys2.diff.txt)

ericvw:
> can you try this instead please: [tmux-keys2.diff.txt](https://github.com/tmux/tmux/files/6382127/tmux-keys2.diff.txt)

This works!  I'm going to locally patch with this for now until a new release is cut.

Thanks for the help because I was getting a bit outside my depth in the special key handling :).

nicm:
Great, I have applied this now, let me know if you see any problems. Thanks!

Issue #2677, opened 2021-04-24
Neovim doesn't render lines correctly after switching from a maximized pane 


gregorias:
#### Actual Behavior

When I switch from a maximized pane with neovim to another pane, the neovim pane starts rendering new lines offset by one. The video explains it:

https://user-images.githubusercontent.com/722385/115953149-f8985480-a4e9-11eb-9f95-6669de33a6b9.mp4

In the video: 

1. I open two vertically split panes with neovim in the left pane.
2. I maximize the neovim pane.
3. I switch to the pane on the right side. You can see that the top bar disappears after a second.
4. I switch back to the neovim pane and change cursorlines. You can see that cursorline renders a line that was previously one line up.
5. After going through the entire file, I refresh neovim with <Ctrl-l> and scroll again.

#### Some Debugging

* The bug was introduced by [this commit from June 2020](https://github.com/tmux/tmux/commit/d919fa1ed0ea3b167ffc811abba26a2dbcd20631). Previous commits do not exhibit this behavior.
* I can "fix" this issue by maximizing and minimizing the terminal window. I suspect that forces some kind of a full refresh.
* I don't have this issue with vim.
* The bug appears on suckless and alacritty terminals.

### Required information

Please provide the following information:

* tmux version: [3.2](https://github.com/tmux/tmux/commit/43c292fa91b22fd8422cc1c4b67e2d2ac3fab767).
* Platform (`uname -sp`): Linux
* $TERM inside and outside of tmux: inside (screen-256color), outside (st-256color)
* Logs from tmux (`tmux kill-server; tmux -vv new`):  (Sorry this file is too large, 30+ MB, I can work on making it smaller in case it's necessary for further debugging)

nicm:
Try without a .tmux.conf and see if that gives you a smaller server log, also you can gzip it.

nicm:
If a pane is resized multiple times but ends up the same size as it started (or there is a risk this happened) tmux will resize it to height - 1 and back to force the application to redraw. It is probable neovim is not handling this correctly but perhaps there is another problem or it is not necessary in this case.

gregorias:
[tmux-server-15849.log](https://github.com/tmux/tmux/files/6369318/tmux-server-15849.log) here's a server log from a session that triggered this bug.

[...]

Issue #2260, opened 2020-06-05
Zoomed panes are not redrawn fully when switching sessions


lukas-reineke:
### Issue description

When switching to a session with a zoomed pane, the content of that pane does not get re-drawn fully.
The part of the pane that is re-drawn is not related to where the split would be. Its always drawn from the top, but stops at the size the pane would be when not zoomed.

It only happens when:
* the pane is displaying a TUI program like `vim` or `htop`. Normal shell output like `ls` is fine
* the pane does not take up the whole space from top to bottom. (there is a horizontal split. other splits don't affect the outcome)

Examples:
pane X has the problem when zoomed:
```
+-------+
|   X   |
+-------+
|       |
+-------+
```
```
+---+---+
|   | X |
+---+---+
|       |
+-------+
```

pane X has no problem:
```
+---+---+
|   |   |
|   | X |
|   |   |
+---+---+
```
```
+---+---+
|   |   |
+---+ X |
|   |   |
+---+---+
```

### Required information

* No configuration
* tmux version: `next-3.2`
* Platform `linux`
* $TERM: `xterm-kitty` (tested with other emulators as well)

### Screenshots
![DeepinScreenshot_select-area_20200605144442](https://user-images.githubusercontent.com/12900252/83841321-96046180-a73b-11ea-9690-f8f68954ac98.png)
Zoomed pane before switching sessions.

![DeepinScreenshot_select-area_20200605144508](https://user-images.githubusercontent.com/12900252/83841344-a3215080-a73b-11ea-9769-17f1bb78d534.png)
Zoomed pane after switching sessions, and then switching back

nicm:
This (and #2252) are both due to the fact that we resize the grid immediately but we don't do `TIOCSWINSZ` to tell the program until the end of the event loop.

In this case, `switch-client` unzooms the window before selecting it and zooms it again afterwards. It has to do this if the active pane changes because the zoomed pane will be different. But the consequence if the active pane hasn't changed is that the grid is resized smaller when unzoomed then back to its previous size when it is zoomed again, but the application is never told it has been smaller, which matters for programs using the alternate screen because content has been lost and they need to redraw.

There are a few possible things we can do:

1) We can avoid unzooming if the active pane hasn't changed. This is a good idea and similar to what the fix for #2252 does, although it is in a different place (window resizing).

[...]

PR #2678, opened 2021-04-24
Revert "Change how panes are resized so that the code is clearer and if the pane"

gregorias:
This reverts commit d919fa1ed0ea3b167ffc811abba26a2dbcd20631.

This commit fixes https://github.com/tmux/tmux/issues/2677 and probably
reopens https://github.com/tmux/tmux/issues/2260


nicm:
Hold your horses.

PR #2692, opened 2021-05-01
Fix compiler warnings: maybe-ununitialized, format

jtroo:
Compiling the latest master in my environment, I get the warnings below. The first two are not fixable, but the rest are fixed in this PR.

```
format.c: In function ‘format_find’:
format.c:3359:5: warning: format not a string literal, format string not checked [-Wformat-nonliteral]
 3359 |     strftime(s, sizeof s, time_format, &tm);
      |     ^~~~~~~~
format.c: In function ‘format_expand1’:
format.c:4455:3: warning: format not a string literal, format string not checked [-Wformat-nonliteral]
 4455 |   if (strftime(expanded, sizeof expanded, fmt, &es->tm) == 0) {
      |   ^~
format.c: In function ‘format_replace’:
format.c:4202:12: warning: ‘fm’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 4202 |    value = format_search(fm, wp, new);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~



server.c: In function ‘server_tidy_event’:
server.c:166:28: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘uint64_t’ {aka ‘long unsigned int’} [-Wformat=]
  166 |     log_debug("%s: took %llu milliseconds", __func__, get_timer() - t);
      |                         ~~~^                          ~~~~~~~~~~~~~~~
      |                            |                                      |
      |                            long long unsigned int                 uint64_t {aka long unsigned int}
      |                         %lu



tty-term.c: In function ‘tty_term_read_list’:
tty-term.c:724:3: warning: ‘s’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  724 |   xasprintf(&(*caps)[*ncaps], "%s=%s", ent->name, s);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~



window-buffer.c: In function ‘window_buffer_get_key’:
window-buffer.c:315:2: warning: ‘wp’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  315 |  format_defaults(ft, NULL, s, wl, wp);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
window-buffer.c:315:2: warning: ‘wl’ may be used uninitialized in this function [-Wmaybe-uninitialized]
window-buffer.c:315:2: warning: ‘s’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

### DIFF
--- a/format.c
+++ b/format.c
@@ -4199,7 +4199,7 @@ format_replace(struct format_expand_state *es, const char *key, size_t keylen,
 			value = xstrdup("0");
 		} else {
 			format_log(es, "search '%s' pane %%%u", new, wp->id);
-			value = format_search(fm, wp, new);
+			value = format_search(search, wp, new);
 		}
 		free(new);
 	} else if (cmp != NULL) {
--- a/server.c
+++ b/server.c
@@ -163,7 +163,8 @@ server_tidy_event(__unused int fd, __unused short events, __unused void *data)
     malloc_trim(0);
 #endif
 
-    log_debug("%s: took %llu milliseconds", __func__, get_timer() - t);
+    log_debug("%s: took %llu milliseconds", __func__,
+        (unsigned long long)(get_timer() - t));
     evtimer_add(&server_ev_tidy, &tv);
 }
 
--- a/tty-term.c
+++ b/tty-term.c
@@ -697,7 +697,7 @@ tty_term_read_list(const char *name, int fd, char ***caps, u_int *ncaps,
 		ent = &tty_term_codes[i];
 		switch (ent->type) {
 		case TTYCODE_NONE:
-			break;
+			fatalx("invalid code in tty_term_codes");
 		case TTYCODE_STRING:
 			s = tigetstr((char *)ent->name);
 			if (s == NULL || s == (char *)-1)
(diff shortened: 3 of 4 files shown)


nicm:
Do not use PRIu64, use a cast please.

format.c I think this should be using `search` not `fm`.

In tty-term.c it would be better to make TTYCODE_NONE a fatalx.

i think the window-buffer.c change is OK.

Thanks

jtroo:
For the change in tty-term.c, the compiler (`gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0`) isn't smart enough to detect that `s` won't be uninitialized with the fatalx call. The fatalx call is added in the latest update, but the `s` pointer is still initialized to NULL for the compiler to stop generating a warning.

nicm:
OK I think we should leave the warning in that case.

nicm:
What about this instead of `fatalx`? Does this fix the warning?
~~~~
--- a/tty-term.c
+++ b/tty-term.c
@@ -697,7 +697,7 @@ tty_term_read_list(const char *name, int fd, char ***caps, u_int *ncaps,
                ent = &tty_term_codes[i];
                switch (ent->type) {
                case TTYCODE_NONE:
-                       break;
+                       continue;
                case TTYCODE_STRING:
                        s = tigetstr((char *)ent->name);
                        if (s == NULL || s == (char *)-1)

nicm:
I applied it with the `continue`, let me know if you still see warnings. Thanks!

Issue #987, opened 2017-06-28
Feature request: next-word-vi and next-word-end-vi commands for copy-mode-vi


Grazfather:
tmux 2.4 on any platform

This has been broken as far as I can remember.

I use vi mode for text selection, but its w (`copy-mode-vi w` -> `send-keys -X next-word`) doesn't use the vim definition of a word.

In vim, if I am on the 's' in 'some_func()', hitting `w` will move to '('. In tmux, it'll move to 'f'. The next press of `w` will go past the end of the word to the end of ().

This is because the vi bindings still leverage the normal tmux commands, which use emacs semantics

Adding vim-style word-moving semantics command (e.g. `next-word-vi`) would be a great addition.

nicm:
You can change word-separators option.

Grazfather:
Indeed you can (I didn't know, now at least I can remove "_").

That still doesn't seem to work correctly. If I set it, for instance, to _only_ "-", it still won't navigate `this-test-case` properly: It moves to the "t", _not_ to the "-" as it should.

nicm:
Yes it works like emacs. I am not opposed to adding further differences when in vi mode but I'm not planning to do it myself, someone familiar with vi will need to do it.

Grazfather:
Fair enough. Not sure if the `next-word` should add a check for if it's in vi mode (It already does this check for something else) or if new commands should be added to provide the vi semantics, and have them bound by default in `copy-mode-vi`. I already have an idea about how it should be done, but haven't otherwise looked at the code before.

I renamed and edited the issue, hoping we can keep it as a feature request, and hoping someone can hop on it, otherwise I will take a look when I have the time.

nicm:
It should have a check for MODEKEY_VI like the existing check in window_copy_cursor_next_word_end().

nicm:
Please open a PR or send an email if you have a change.

janderland:
For people searching for a solution to this "problem", from @Grazfather own [dotfiles](https://github.com/Grazfather/dotfiles/blob/master/tmux.conf)...

```
# Set the word separators to get closer to vim's
set-option -g word-separators " -()@.,[]{}:="
```

PR #2693, opened 2021-05-03
Better word-boundary navigation in Vi-style copy mode

wm-noble:
This addresses #987.

Word navigation in ViM treats whitespace, letters, and non-letter symbols as 3 separate character classes, where letters and symbols are treated equivalently but separately (so `@#$` is considered a "word" just like `hello`, and `hello@#$` with no whitespace would be 2 "words"), and whitespace works similarly to how `word-separators` works currently. To implement this kind of navigation in `copy-mode-vi`, I've added 2 new options:

- `vi-word-whitespace` which defaults to `" \t"`
    - Since this works similarly to how `word-separators` works normally, and the default for `word-separators` is `" "`¹, perhaps we should re-use that option instead and avoid introducing a new one? Doesn't matter to me.
- `vi-word-symbols` which defaults to `"+-*/^=()[]{}@#$%&:;<>.,/\\|?~`'\""`
    - This seems to be all the symbols according to ViM by default.

Adding these 2 new options and changing the behavior in `copy-mode-vi` to use them would be a breaking change for anybody that uses that mode, but shouldn't affect anybody that uses emacs mode, given that `copy-mode-test-emacs.sh` passes with both the old and new code. Anybody using Vi mode would find themselves with navigation that works like the default navigation in ViM, which I believe would be a welcome change for most, and can always be customized to work like the current navigation mechanism by making `vi-word-symbols` empty and making `vi-word-whitespace` the same as `word-separators`.

¹ I am confused as to why the documentation currently says `" @_-"` is the default for `word-separators`, when the code in `options-table.c` seems to differ, and it doesn't seem to work that way as shown by the Emacs copy mode test, so I modified the documentation accordingly. Lmk if I'm misunderstanding.


nicm:
Thanks. Looks good to me aside from style nits, maybe @topcat001 also would like to take a look since he wrote the grid-reader stuff?

I am not really inclined to have two different sets of options for vi and emacs for something as minor as word movement, so I think we should pick one and use it for both. I think the vi mechanism is more sophisticated and can probably be configured to be close enough to emacs, so I think you should just add `word-whitespace` as well as `word-separators` and make it behave like vi for both.

You can just delete the defaults from tmux.1, it is too much effort to keep them in sync and it is trivial to see them from tmux itself.

What about vi keys in the status line? Do you intend to change it also?

topcat001:
No obvious issues I can see, but if you don't mind I'd like to test the changes a little bit. I'll report back tomorrow if I find any issues.

wm-noble:
Yes, with separate options for whitespace and symbols, it would be easy to preserve the current behavior by simply making the symbols list empty. I'd just like to clarify how much of an appetite you have for breaking changes. There are many details to decide on:

1. Keep the current default for `word-separators` as just whitespace, add a new option `word-symbols` with the symbols, but only use it in Vi mode. This would change nothing for people using emacs mode.
2. Similar to (1), but use the symbols regardless of mode. This simplifies the code a bit, but may result in a bit more confusion since people who want to keep the current behavior for emacs mode would have to look up what changed and set `word-symbols` empty.
3. Redefine `word-separators` to be the symbols, and add a new option `word-whitespace` to take its place, as you seemed to suggest. This would probably be a bit more confusing for people who currently customize `word-separators`.
4. Similar to (3), but don't bother with a `word-whitespace` option at all, and just hardcode it as `" "`. For anybody with an existing setting for `word-separators` that includes whitespace, we can just ignore that whitespace. This would probably be the cleanest solution, just wanted to check that you're cool with redefining the meaning of an existing option like this.

nicm:
I'd go with 4, there is no whitespace except spaces anyway.

topcat001:
Functionally seems fine so I think I'm ok with it, modulo @nicm's comments/questions above, thanks!

wm-noble:
Ok I did 4, changed the default of `word-separators`, and made the status line consistent with copy mode, which involved adding a new key flag to unambiguously convey when there is no good conversion from certain vi commands to equivalent emacs commands since the current implementation assumes a 1-to-1 relationship.

Also patch what I think was a bug introduced in e2d01795d that allowed modified characters such as `M-w` that default in the big switch-case in `status_prompt_key` (in emacs mode) to cause a crash, and this seems like a sensible fix.

status.c
```diff
@@ -1360,8 +1360,10 @@ status_prompt_key(struct client *c, key_code key)
 		return (0);
 	if (key <= 0x7f)
 		utf8_set(&tmp, key);
-	else
+	else if (KEYC_IS_UNICODE(key))
```
topcat001, line 1363: Nice! I hit this issue causing a crash but didn't get around to fixing it yet.

nicm:
OK looks great. Here is a diff on top of yours with some small changes: [aaaa.diff.txt](https://github.com/tmux/tmux/files/6439916/aaaa.diff.txt)

- For `KEYC_VI`, 0x2... comes after 0x1... :-).

- I like gotos and giant switches as much as the next person, but let's not go too far in combining the two - add helper functions to avoid the gotos into the switch in `status_prompt_key`.

- Some C99 additions are still dangerously controversial and that includes stdbool.h, so replace them with int.

- The rest are minor style nits (fill comments, break lines after commas if possible, don't start lines with operators, etc) - not all of these are in your new code.

It still seems to work for me on a cursory test but can you please check I haven't broken anything? Particularly in `status.c`.

I am a little concerned about `WHITESPACE` being too generic and going to make something blow up, but the most likely candidate is ncurses.h and it is OK so I think we can go with it and see what happens.

Thanks

wm-noble:
Cool, points well taken. Applied your diff as-is. Found some corner cases that aren't working quite right but I'll finish patching them up as soon as I can.

wm-noble:
Ok, everything works as expected now in Vi mode on the status line and in copy mode. Based on my experiments, pretty sure it works appropriately for Emacs mode as well, where "forward word" is really more similar to "end of word" in Vi.

topcat001:
Thanks! I gave it a spin and looks good so far. After it gets merged
I'll do more thorough testing.

wm-noble:
I spoke too soon. There was still an edge case in Emacs mode I wasn't aware of where doing `M+f` to go forward a word doesn't work properly when the cursor starts on a whitespace and  is followed by at least 1 more whitespace. Tested it manually and all the edge cases now appear to be covered the same way in both modes.

nicm:
Great stuff, applied now, thanks!

wm-noble:
Thank you for maintaining this!

jmeurin:
This broke selecting URLs with select-word
bind -T copy-mode C-u {
  send-keys -X search-backward "(https|http)://[0-9a-z_]+"
  send-keys -X select-word
  send-keys -X copy-selection-and-cancel
}

with http://foo in the panel, 
at f03b27c7 (before this), the clipboard would contain http://foo
at 022d0210  (this commit),it onl contains http

nicm:
You should be able to do without select-word since copy will copy the current match by default.

Or remove / and : from your word-separators.

wm-noble:
Ya this does break from the prior default behavior as I explained above. I have no issues with changing back the default value for `word-separators` to something that causes less friction for people. This new default is just what I prefer since it's consistent with vim, and nicm was cool with it so that's what I did :)

Issue #2694, opened 2021-05-03
`tmux list-keys` does not terminate when pipe fails


josephjang:
### Issue description

When the output of `tmux list-keys` is pipelined into any command which does not read, it is expected for the tmux client process to terminate since it fails to write to stdout.

`tmux` v3.1 works as expected.
```
$ tmux -vv -Ltest -f/dev/null list-keys | true
$ 
```

But the [latest release](https://github.com/tmux/tmux/tree/3.2) or the current snapshot (2e7ec8c), the client and the server process does not terminate.

```
$ tmux -vv -Ltest -f/dev/null list-keys | true

```

### Required information

Please provide the following information:

* tmux version (`tmux -V`): tmux 3.2
* Platform (`uname -sp`): Linux x86_64 (`uname -sm`)
* $TERM inside and outside of tmux (`echo $TERM`): xterm-kitty
* Logs from tmux:
  * stuck at 1620024496 and killed the client with TERM at 1620024626.
  * [tmux-client-483028.log](https://github.com/tmux/tmux/files/6413336/tmux-client-483028.log)
  * [tmux-server-483030.log](https://github.com/tmux/tmux/files/6413339/tmux-server-483030.log)

PR #2695, opened 2021-05-03
Fix a regression where client doesn't exit when stdout write fails

josephjang:
This PR addresses https://github.com/tmux/tmux/issues/2694.

When the output of tmux client is pipelined and the latter process fails
without reading all the output, tmux is supposed to terminate silently.

Due to the changes introduced by the commit e40831a, tmux client fails
to exit when writes to stdout fails. `client_write_error_callback`
correctly cleans up the event relevant to the client file for stdout. But
the problem is that exit-checking code `client_file_check_cb` is called
before the clean-up. As the result, it fails to recognize the write failure
and the tmux client remains in the event loop and the tmux server does
the same.

This patch moves the call for `client_file_check_cb` after the clean-up
in `file_write_error_callback`, which is the same to the code before e40831a.
~~The patch also changes the order of the call for the call-back in
`file_write_callback`, which does not change the logic and is also more
consistent.~~

### DIFF
--- a/file.c
+++ b/file.c
@@ -503,14 +503,14 @@ file_write_error_callback(__unused struct bufferevent *bev, __unused short what,
 
 	log_debug("write error file %d", cf->stream);
 
-	if (cf->cb != NULL)
-		cf->cb(NULL, NULL, 0, -1, NULL, cf->data);
-
 	bufferevent_free(cf->event);
 	cf->event = NULL;
 
 	close(cf->fd);
 	cf->fd = -1;
+
+	if (cf->cb != NULL)
+		cf->cb(NULL, NULL, 0, -1, NULL, cf->data);
 }
 
 /* Client file write callback. */


nicm:
Thanks but `file_free` can free `cf` so you are you sure you are not introducing a use-after-free here in `file_write_callback`?

josephjang:
> Thanks but `file_free` can free `cf` so you are you sure you are not introducing a use-after-free here in `file_write_callback`?

Thanks for checking. I've overlooked the line which frees `cf`. Changed the patch to keep the original order in the `file_write_callback`.

josephjang:
I think it would be nice that `file_free` is called in the `file_write_error_callback` as well to be consistent with `file_read_error_callback`. But, I'm not sure yet it's safe and anyway, it's out of the scope of this issue. :)

nicm:
Great, applied, thanks!

Issue #2398, opened 2020-09-24
[Feature request] Support systemd socket activation


msharov:
There is currently a race condition in trying to attach to the same session of tmux from two processes; there is no way to guarantee that only one tmux is launched. This happens, for example, when launching tmux as a persistent service and then attaching to it from an xterm. Both of these occur at login and while X startup provides some delay between the invocations, it is not a hard guarantee.

The problem could be solved if tmux could be started by socket activation. Systemd would launch tmux when the socket is accessed and pass its file descriptor via sd_listen_fds. Tmux could use the fd name, obtained using sd_listen_fds_with_names, as the name of the session. Subsequent starts would connect to the socket, set in the environment, instead of launching a new tmux, thus eliminating race conditions.

nicm:
You should not be able to end up with two tmux listening on the same socket path. Please show logs (both servers and the clients which start them) from this happening.

msharov:
You are right, I didn't think of that. Just specifying the socket would be sufficient to prevent two tmux servers being launched. (No, I did not observe this happening and would not know how to set it up to catch such a timing error)

However, this does not entirely solve the problem, because it is still undetermined which invocation launches the server. I want to always end up with the tmux service running as the server and with the xterm running the client, but if the xterm-invoked tmux got to locking the socket first, it would fork off the server and the systemd tmux service would end up as a client, connected to an unparented server process.

With socket activation, tmux server would explicitly be launched as such, using the -D option (which you ought to document on the man page), and the xterm activation would only be able to launch the client. This way the tmux server process would be properly supervised by systemd and there would be no ambiguity about when and by what it is launched.

nicm:
You don't need to specify the socket, you should not be able to get two servers running which think they are listening on the same socket whether or not you specify the socket.

I think there are a couple of things you can do:

- Set it up so that X is always started after tmux, there may be some way to say that it is a dependency. Most init systems have some way to express dependencies.

- Wrap your xterm tmux in a script that stops it starting tmux, for example something as simple as `while ! tmux has 2>/dev/null; do sleep 1; done; tmux new` would probably work.

msharov:
systemd certainly does have dependency management. Unfortunately it does not work here because a systemd service is considered started immediately after it is forked off, while the tmux control socket is not created until a little later. One way to solve this would be to launch tmux as type notify and use sd_notify API in tmux to tell systemd when it is really ready.

[...]

PR #2741, opened 2021-06-14
Add abstract socket name support

s1341:
This PR adds support for specifying abstract socket names using `-S`. This allows the use of tmux on unrooted Android.

### DIFF
--- a/client.c
+++ b/client.c
@@ -160,7 +160,7 @@ retry:
 			goto retry;
 		}
 
-		if (lockfd >= 0 && unlink(path) != 0 && errno != ENOENT) {
+		if (lockfd >= 0 && (path[0] != '@' && unlink(path) != 0) && errno != ENOENT) {
 			free(lockfile);
 			close(lockfd);
 			return (-1);
--- a/client.c
+++ b/client.c
@@ -118,6 +118,10 @@ client_connect(struct event_base *base, const char *path, uint64_t flags)
 	}
 	log_debug("socket is %s", path);
 
+	if (path[0] == '@') {
+	    sa.sun_path[0] = '\0';
+	}
+
 retry:
 	if ((fd = socket(AF_UNIX, SOCK_STREAM, 0)) == -1)
 		return (-1);
--- a/client.c
+++ b/client.c
@@ -118,9 +118,11 @@ client_connect(struct event_base *base, const char *path, uint64_t flags)
 	}
 	log_debug("socket is %s", path);
 
+#ifdef __ANDROID__
 	if (path[0] == '@') {
 	    sa.sun_path[0] = '\0';
 	}
+#endif
 
 retry:
 	if ((fd = socket(AF_UNIX, SOCK_STREAM, 0)) == -1)
@@ -160,7 +162,13 @@ retry:
 			goto retry;
 		}
 
-		if (lockfd >= 0 && (path[0] != '@' && unlink(path) != 0) && errno != ENOENT) {
+		if (lockfd >= 0 &&
+			(
+#ifdef __ANDROID__
+			    path[0] != '@' &&
+#endif
+			    unlink(path) != 0
+			) && errno != ENOENT) {
 			free(lockfile);
 			close(lockfd);
 			return (-1);
--- a/server.c
+++ b/server.c
@@ -115,10 +115,11 @@ server_create_socket(int flags, char **cause)
 		errno = ENAMETOOLONG;
 		goto fail;
 	}
-	unlink(sa.sun_path);
 
 	if (socket_path[0] == '@') {
 	    sa.sun_path[0] = '\0';
+	} else {
+	    unlink(sa.sun_path);
 	}
 
 	if ((fd = socket(AF_UNIX, SOCK_STREAM, 0)) == -1)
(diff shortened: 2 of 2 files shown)


nicm:
What does this do and how does it work?

s1341:
It adds support for abstract Unix socket names. See https://man7.org/linux/man-pages/man7/unix.7.html under ‘abstract’. 

It allows one to specify a socket name starting with an ‘@‘. Internally the ‘@‘ gets converted to a null byte. 

Abstract sockets are useful because they do not require a backing file/inode. Additionally, when running on android via adb, the shell does not have permission to create normal Unix sockets, as selinux blocks this. It can, however create abstract sockets. This allows tmux to be used without root via adb.

nicm:
So where does the `@` come into it?

nicm:
So where does the `@` come into it?

nicm:
I mean, did you pick `@`? How are people meant to know to use `@`?

nicm:
I think this needs a check in `server_update_socket` as well.

nicm:
Also the unlink in server_create_socket will need to be conditional I think.

And should we still create /tmp/tmux-XXX? I guess not?

Thanks

nicm:
Oh maybe we don't create it with -S anyway, so that might be OK.

s1341:
‘@‘ is the ‘standard’ way to represent abstract socket names in user interfaces. 

The unlink will silently fail. Also, as far as I can tell, no additional socket is created with ‘-S’. 

I’ll check the ‘server_update_socket’ function.

s1341:
I looked at ‘server_update_socket’. It should just exit early in the case of an abstract socket as stat will fail. I don’t think it requires any special handling.

nicm:
What will happen to `@foo` if I do:
~~~~
touch @foo
tmux -S @foo new
~~~~

s1341:
Ok. Sold. I will make the unlink conditional and add special case code in ‘server_update_socket’.

nicm:
Great thanks.

s1341:
@nicm: Sorry for the delay. Finally got a chance to make the changes we discussed.

nicm:
OK thanks, will take a look. What are the security implications here? What happens if two users open an abstract socket using the same name?

s1341:
Good question. I think it is the same as when using a regular socket path... Not really sure how to go about checking...

s1341:
From here: https://unix.stackexchange.com/questions/579612/unix-domain-sockets-for-non-root-user, it seems that anyone can connect:
```
Unlike regular Unix socket paths, abstract Unix socket names have different semantics, as anybody can bind to them (if the name is not already taken), and anybody can connect to them.

Instead of relying on file/directory permission to restrict who can connect to your socket, and assuming that eg. only root could create sockets inside some directory, you should check the peer's credential with getsockopt(SO_PEERCRED) (to get the uid/pid of who connected or bound the peer), or the SCM_CREDENTIALS ancillary message (the get the uid/pid of who sent a message over the socket).

This (replacing the usual file permission checks) is also the only sane use of SO_PEERCRED/SCM_CREDENTIALS IMHO.
```

nicm:
Well, we can't have that. I think you will need to use `setsockopt(SO_PEERCRED)` when the client connects and make sure the euid is 0 or matches the server. But only do this for abstract sockets because otherwise it will break for people who use shared filesystem sockets.

nicm:
I am concerned about the point in the top comment on this page regarding containers: https://utcc.utoronto.ca/~cks/space/blog/linux/SocketAbstractNamespace?showcomments

But I am not sure this is a genuine issue for tmux since it is not a network program. Does it matter if UID 99 outside one of these containers can access tmux sockets for UID 99 inside and vice versa? Or are they the same and it is not really a security boundary? I don't know anything about this form of Linux container. How do other programs deal with this?

I'm not sure this is going to fly TBH, it seems like abstract sockets are very poorly thought out and very likely to introduce security problems. Can you solve your Android problem another way? Surely there is somewhere to create temporary files?

s1341:
Unfortunately, it's not an issue of having a place to put the file... selinux doesn't allow creation of sockets by the shell user/process at all. In my experience, abstract sockets are the only workable solution.

nicm:
I think what you would need to do is generate a secure (unguessable) name for the abstract socket and store it in a file on disk with appropriate permissions, then later clients could read the file to discover the abstract socket to use, but file system permissions would stop other users from finding it out. But it seems rather complicated to me.

Does Android support multiple users? Maybe we can make this only work on Android.

s1341:
I think making it work only on Android is the way to go. Should I modify the PR to accomplish that?

nicm:
Sure do that and we can see how it looks, thanks.

s1341:
Ok. I'll take a look...

s1341:
I pushed a fix with some conditional compilation for Android.

nicm:
OK, another problem. How does the lock file work? If you give `-S @foo` it is going to create `@foo.lock` in the current directory. What happens if that directory is not writable? What if the client and server are started in different directories? We will need a different way to lock starting the server.

s1341:
hrm. Good point. I think that we'll need to place the lock in `/data/local/tmp`... It is the only temp directory guaranteed to be accessible to the shell user.

msharov:
You do not need the lockfile with abstract sockets. It is one of their best features; the socket is created atomically and deleted when closed, eliminating any race conditions.

Regarding socket access: you really do have to check the connection uid in the server. You can't just generate an "unguessable" socket name because all active abstract sockets can be easily listed with `netstat -l`.

nicm:
Without a lock file, another client could start the server after connect and before bind, causing the original client to fail.

msharov:
You might also consider implementing socket activation (#2398) as an alternative solution. It would allow using abstract sockets and eliminate the race condition you mention. Just a suggestion...

PR #2755, opened 2021-06-22
Fix strtonm run test when crosscompiling

malbarbo:
Same fix as in https://github.com/tmux/tmux/pull/2651 for the `strtonum` function introduced in https://github.com/tmux/tmux/commit/b2588eed03640dc74ffdbebd10c719064b575291.


nicm:
Applied, thanks!

Issue #2737, opened 2021-06-10
3.2a questions and comments


nicm:
### Ask or discuss any quick questions here about 3.2a please.

Bug reports please open a new issue with full information, see [CONTRIBUTING](https://github.com/tmux/tmux/blob/master/.github/CONTRIBUTING.md).

**[The 3.2a CHANGES file is here](https://raw.githubusercontent.com/tmux/tmux/3.2a/CHANGES).**

The previous issue for 3.2 is [here](https://github.com/tmux/tmux/issues/2645).

As always, thanks to the many people who contributed to both this and to previous releases.

---

Changes from 3.2 to 3.2a are:

* Add an "always" value for the "extended-keys" option; if set then tmux will forward extended keys to applications even if they do not request them.

* Add a "mouse" terminal feature so tmux can enable the mouse on terminals where it is known to be supported even if terminfo(5) says otherwise.

* Do not expand the filename given to -f so it can contain colons.

* Fixes for problems with extended keys and modifiers, scroll region, source-file, crosscompiling, format modifiers and other minor issues.

---

There are a **lot** of changes in the 3.2 release, these are the most notable, for more see the CHANGES file:

- How data is sent to control mode clients has been completely revamped to both be more fair with multiple panes and to prevent huge amounts of data being backed up; this feature is used by default when tmux is attached with the latest iTerm2 nightly builds.

- Configuration file parsing has changed slightly: the contents of the new {} syntax introduced in 3.1 must now be valid tmux command syntax; and to allow formats to be annotated, strings given with quotes may now contain newlines (leading spaces and comments are stripped).

- A new customize mode available with `C-b C` (`C-b S-c`) which allows options and key bindings to be browsed and changed interactively.

- Support for extended keys offered by some terminals (xterm, mintty, iTerm2), see https://github.com/tmux/tmux/wiki/Modifier-Keys#extended-keys.

- A `pane-border-lines` option to change the characters used to draw the pane border separators.

- How UTF-8 data is stored has been rewritten to reduce memory use for characters in the BMP.

- The message log (`C-b ~`) has been changed to be per server instead of per client and to have some useful content.

- Client flags have been extended (`refresh-client -F` becomes `-f` and also added for `attach-session` and `new-session`), and a `read-only` flag added (same as `-r` to `attach-session`).

- A new `active-pane` client flag that if given allows a client to have its own active pane for each window rather than being tied to the server's active pane.

- Improved `command-prompt` completion including showing a menu of completions.

[...]

PR #2763, opened 2021-07-02
Format display popup command if single item passed

ClashTheBunny:
I commented on the 3.2 release Q&A here: https://github.com/tmux/tmux/issues/2737#issuecomment-872671383

This would fix that exact comment.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -376,7 +376,7 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 	if (argc == 0)
 		shellcmd = options_get_string(s->options, "default-command");
 	else if (argc == 1)
-		shellcmd = argv[0];
+		shellcmd = format_single_from_target(item,argv[0]);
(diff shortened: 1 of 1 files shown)


nicm:
We do not do this for other commands which take shell commands like this, so I think we shouldn't do it here either. You can wrap the command with `run -C` instead.

ClashTheBunny:
I've tried a bunch of things, including adding more paths and escaping and shells and haven't figured it out.  Is there an example?

Would there be the possibility of exposing `format_single_from_target` to the config format, and then there wouldn't be a 'some things like run do templating and most other things don't' situation?

nicm:
run -C "popup 'echo #{version}'"

ClashTheBunny:
:facepalm: I totally was thinking of it backwards.  That works perfectly:
```
bind '`' run -C "display-popup -d '#{pane_current_path}' -xC -yC -w'90%' -h'90%' -E 'tmux attach -t #S-popup || tmux new -s #S-popup'"
```

PR #2765, opened 2021-07-03
Improve error message if socket dir has too liberal permissions

ukleinek:
tmux expects that the directory containing the socket used to
communicate with the server process to be owned by the user and that
others have no permission to access it.

If this fails however, the emitted error message is misleading as the
directory exists just fine:

	uwe@taurus:~$ chmod o+x /tmp/tmux-1000
	uwe@taurus:~$ tmux a
	error creating /tmp/tmux-1000 (Permission denied)

Rework error handling in make_label() to emit an error message that
talks about unsafe permissions in this case instead.

Debian-Bug: https://bugs.debian.org/990215
Fixes: f32fd2df69d4 ("Improve error message if creating socket parent directory fails [...]")

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -207,16 +207,19 @@ make_label(const char *label, char **cause)
 	free(paths);
 
 	xasprintf(&base, "%s/tmux-%ld", path, (long)uid);
-	if (mkdir(base, S_IRWXU) != 0 && errno != EEXIST)
+	if ((mkdir(base, S_IRWXU) != 0 && errno != EEXIST) ||
+	    lstat(base, &sb) != 0) {
+		xasprintf(cause, "error creating %s (%s)", base, strerror(errno));
 		goto fail;
-	if (lstat(base, &sb) != 0)
-		goto fail;
+	}
 	if (!S_ISDIR(sb.st_mode)) {
 		errno = ENOTDIR;
+		xasprintf(cause, "%s is not a directory", base);
 		goto fail;
 	}
 	if (sb.st_uid != uid || (sb.st_mode & S_IRWXO) != 0) {
 		errno = EACCES;
+		xasprintf(cause, "unsafe permissions for %s", base);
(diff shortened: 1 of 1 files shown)


nicm:
Nice, applied now with some minor tweaks. Thanks!

ukleinek:
I'm disappointed about how this went. I'm not the author of the resulting commit according to git, my commit log was discarded, the reference to the Debian bug is gone, this Pull Request (not issue as is written in the pushed commit log) is "closed with unmerged commits", the closing commit has no reference to this PR (at least not one that github or the git tools would create a link for) and you didn't even make the effort to write my name correctly :-\

This is not how to motivate people to continue contributing to a project.

nicm:
The commit message reads:
~~~~
    Improve error reporting when the tmux /tmp directory cannot be created
    or used, GitHub issue 2765 from Uwe Kleine-Koenig.
~~~~
Which both points to this issue and credits you by name.

ukleinek:
@nicm: I wonder if you have read my message and tried to understand my frustration.

nicm:
I did read it but you are credited and this issue is clearly pointed to in the commit message for anyone who may want to follow up further. Other than that you seem to be asking me to follow your process rather than mine, which doesn't really seem reasonable.

OpenBSD uses CVS so I cannot merge this pull request directly. This also means it does not have author metadata so it is necessary to include the author in the commit message. I also need to apply the diff to review and test it and for a relatively small change it is not worth the time going back and forth with minor changes or worrying particularly about having an extensive commit message, I'm sure you do not want to waste either your time or mine. I do not use GitHub to track changes once a commit is made, so the state of this PR after the commit is made is immaterial.

It is also practice to keep commit messages to ASCII so I cannot include the ö in your name which is UTF-8, it is my understanding that oe is an acceptable alternative but if you prefer something else then apologies, although it seems a pretty minor point.

PR #2802, opened 2021-08-02
Update the session environment when the latest active client changes

ncfavier:
This allows one to switch between multiple clients attached to the same session and have the environment update without needing to reattach.

(If you need to credit me but can't use my name, please use my GitHub username instead)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ CHANGES FROM 3.2a TO 3.3
 * Add different command histories for different types of prompts ("command",
   "search" etc). From Anindya Mukherjee.
 
+* Update the session environment when the latest active client changes.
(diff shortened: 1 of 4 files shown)


nicm:
Thanks. I am not sure this should happen automatically. 

I think this would be better done by adding a hook when the active client changes and a way to trigger an environment update from it. Maybe just by doing `attach`? Or a new flag to it or some other command, maybe `set-environment`.

ncfavier:
Fair. I'll look into that

nicm:
OK cool, it should be easy enough to add a hook but I am not sure firing a command will have the right client and session, but give it a go and if you have trouble let me know.

PR #2803, opened 2021-08-02
Add client-active hook

ncfavier:
Replaces #2802

This seems to work fine, I can run `set-hook -g client-active attach` to achieve the desired behaviour.


nicm:
OK I will take a look tomorrow.

nicm:

PR #2824, opened 2021-08-14
Minor cleanups

a1346054:


nicm:
Applied most of these but not the sh changes which seem unnecessary. Thanks!

a1346054:
True, it is not strictly necessary, however it has undefined behavior in POSIX

https://github.com/koalaman/shellcheck/wiki/SC2166

nicm:
Yes I saw that but it seems like the way they are used here is fine. And they are only for testing anyway.

PR #2851, opened 2021-08-24
Add -Wno-format-nonliteral.

MasahikoSawada:
This is my first contribution to tmux. Please correct me if I'm sending the change to the wrong place.

When building tmux source code on OSX 10.15.7 I got the following warnings:

```
file.c:209:36: warning: format string is not a string literal [-Wformat-nonliteral]
                evbuffer_add_vprintf(cf->buffer, fmt, ap);
                                                 ^~~
file.c:216:36: warning: format string is not a string literal [-Wformat-nonliteral]
                evbuffer_add_vprintf(cf->buffer, fmt, ap);
                                                 ^~~
2 warnings generated.
log.c:112:22: warning: format string is not a string literal [-Wformat-nonliteral]
        if (vasprintf(&fmt, msg, ap) == -1)
                            ^~~
1 warning generated.
xmalloc.c:126:21: warning: format string is not a string literal [-Wformat-nonliteral]
        i = vasprintf(ret, fmt, ap);
                           ^~~
xmalloc.c:155:26: warning: format string is not a string literal [-Wformat-nonliteral]
        i = vsnprintf(str, len, fmt, ap);
                                ^~~
```

This PR adds `-Wno-format-nonliteral` flag to suppress these warnings.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,7 +29,7 @@ AM_CFLAGS += -Wmissing-prototypes -Wstrict-prototypes -Wmissing-declarations
 AM_CFLAGS += -Wwrite-strings -Wshadow -Wpointer-arith -Wsign-compare
 AM_CFLAGS += -Wundef -Wbad-function-cast -Winline -Wcast-align
 AM_CFLAGS += -Wdeclaration-after-statement -Wno-pointer-sign -Wno-attributes
-AM_CFLAGS += -Wno-unused-result -Wno-format-y2k
+AM_CFLAGS += -Wno-unused-result -Wno-format-y2k -Wno-format-nonliteral
 AM_CPPFLAGS += -DDEBUG
 endif
 AM_CPPFLAGS += -iquote.


ThomasDickey:
If you've used the *printf* attribute supported by gcc and clang, then those warnings do not appear.

nicm:
/usr/include/stdio.h:
~~~~
int      vsnprintf(char *, size_t, const char *, __va_list)
                __attribute__((__format__ (printf, 3, 0)))
                __attribute__((__nonnull__ (3)))
                __attribute__((__bounded__(__string__,1,2)));
~~~~

nicm:
The only way anyone suggested to avoid it before was using `#pragma GCC diagnostic` which is a bit ugly and IIRC is hard to get to work across clang, gcc and old versions of both.

ThomasDickey:
in ncurses and other programs, I do this with a macro `GCC_PRINTFLIKE` (see [configure-check](https://github.com/ThomasDickey/my-autoconf-snapshots/blob/master/AcSplit/CF_GCC_ATTRIBUTES) and [example](https://github.com/ThomasDickey/ncurses-snapshots/blob/6156f5598461297a1795f6e538d8e185a75637e3/include/curses.h.in#L709)).  That's worked with [gcc](https://invisible-island.net/personal/lint-tools.html#tool_gcc) since the mid-1990s.  clang works with that (though it has its own clang-specific declarations).

nicm:
Are you sure it suppresses the warning when the format is not a literal? I'm pretty sure it doesn't.

nicm:
cc wouldn't even know evbuffer_add_vprintf had a format argument if it didn't have attribute printf :-).

ThomasDickey:
sorry - "not a literal" (it won't do **_that_** inside), but all of the calls to the function which use the format (presumably not always a variable) would be checked.

ThomasDickey:
What I noticed earlier (and prompted my comment) was this line:

log_vwrite(const char *msg, va_list ap)

which might be improved with a *printflike* declaration :-)

ThomasDickey:
The [gcc documentation](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html)]notes that *In C99 mode, the functions snprintf, vsnprintf, vscanf, vfscanf and vsscanf are also checked.* so I'd assume that the `va_list` would not confuse it.

nicm:
Ah yes I should add it to those, thanks.

MasahikoSawada:
Thank you for the comments.

I agreed that it's better to suppress these warnings not by adding this flag.

nicm:
OK I added these, and it gets rid of some of the warnings with clang 11 - I guess because now the compiler can track the `fmt` argument through between functions. But it still leaves  two warnings where I am genuinely using a nonliteral format, will see what I can do about these as well.

nicm:
This gets rid of the last two with clang 11 but I'm not really wild about it: [tmux-format-nonliteral.diff.txt](https://github.com/tmux/tmux/files/7044637/tmux-format-nonliteral.diff.txt)

ilya-bobyr:
```patch
--- tmux.h	25 Aug 2021 07:09:30 -0000	1.1136
+++ tmux.h	25 Aug 2021 07:26:19 -0000
@@ -93,6 +93,18 @@ struct winlink;
 #define DEFAULT_XPIXEL 16
 #define DEFAULT_YPIXEL 32
 
+/* Don't complain about format arguments. */
+#if __clang__ || __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
+#define no_format_nonliteral(x) \
+	_Pragma ("GCC diagnostic push") \
+	_Pragma ("GCC diagnostic ignored \"-Wformat-nonliteral\"") \
+	do { x; } while (0) \
+	_Pragma ("GCC diagnostic pop")
+#else
+#define no_format_nonliteral(x) \
+	do { x; } while (0)
+#endif
+
 /* Attribute to make GCC check printf-like arguments. */
 #define printflike(a, b) __attribute__ ((format (printf, a, b)))
 ```
 
 _minor_
 Maybe put `Pragma`s inside the `do / while` block as well?

nicm:
Yes they should probably go inside. I will look at this again at some point.

nicm:
I have applied this now, Thanks all!

nicm:
I actually changed it to work a different way entirely but the outcome is the same :-).

PR #2873, opened 2021-09-10
tmux.1: Version specific configuration example

ilya-bobyr:
While there are many ways how one can make a config file compatible with
multiple versions of tmux, it would be nice to provide an example most
people can refer to and copy from.

It would also document an "official" way of how the "multi-version
config files" should be written, stating support for this approach in
the future.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -710,6 +710,26 @@ Conditionals may be given on one line, for example:
 .Bd -literal -offset indent
 %if #{==:#{host},myhost} set -g status-style bg=red %endif
 .Ed
+.Pp
+You may also use
+.Ql %if
+/
+.Ql %elif
+/
+.Ql %endif
+to make your configuration file compatible with different version of tmux.
+Even if the commands have different arguments in different versions.
+For example, say you want a green status line if running a development
+build, blue if version 3.1 or abov, and red otherwise:
+.Bd -literal -offset indent
+%if #{m/r:^next-,#{version}}
+set -g status-style bg=green
+%elif #{&&:#{m/r:^[0-9]+\.[0-9]+$,#{version}},#{e|>=|f:#{version},3.1}}
+set -g status-style bg=blue
+%else
+set -g status-style bg=red
+%endif
+.Ed
 .Sh COMMANDS
 This section describes the commands supported by
 .Nm .


nicm:
I don't really want to document this like this in the man page, I think an FAQ entry would be more appropriate, we can use your example.

I'd like to have an example for before `m` but I don't have time to test one at the moment.

Maybe:
~~~~
--- a/FAQ.md
+++ b/FAQ.md
@@ -449,6 +449,38 @@ The passthrough escape sequence is no longer necessary for changing
the cursor
 colour or style as tmux now has its own support (see the `Cs`, `Cr`, `Ss` and
 `Se` capabilities).

+### How can I make .tmux.conf portable between tmux versions?
+
+For `set-option`, the `-q` flag to suppress warnings about unknown options
+since tmux 0.3, for example:
+
+~~~~
+set -gq mode-bg red
+~~~~
+
+Since tmux 2.3, the running server version is available in the `version` format
+variable. This can be used with `if-shell`, `if-shell -F` (since tmux 2.0) or
+`%if` (since tmux 2.4) to check for specific server versions.
+
+The `m` (since tmux 2.6) and `m/r` (since tmux 3.0) modifiers are most useful
+for this. For example to show a green status line if running a development
+build, blue if version 3.1 or above and red otherwise:
+
+~~~~
+%if #{m/r:^next-,#{version}}
+set -g status-style bg=green
+%elif #{&&:#{m/r:^[0-9]+\.[0-9]+$,#{version}},#{e|>=|f:#{version},3.1}}
+set -g status-style bg=blue
+%else
+set -g status-style bg=red
+%endif
+~~~~
+
+Note that on OpenBSD version numbers the tmux version number tracks the OpenBSD
+version, see [this FAQ
+entry](https://github.com/tmux/tmux/wiki/FAQ#how-often-is-tmux-released-what-is-the-vers
ion-number-scheme)
+for information on tmux version numbers.
+
 ### Why don't XMODEM, YMODEM and ZMODEM work inside tmux?

 tmux is not a file transfer program and these protocols are more effort to

ilya-bobyr:
Your version has way more details.
I like it better :)

I must admit, I did not know there is a FAQ section.
Nor did I ever look at the Wiki section of this repo.
It actually contains a lot of useful information!

I think it would be good to add a link to the wiki to the `Documentation` section of the `README` file.
Let me send a PR :)

ilya-bobyr:
Here: https://github.com/tmux/tmux/pull/2875

nicm:
I have applied this FAQ change now. Thanks!

PR #2890, opened 2021-09-16
Add support for hyperlinks

wm-noble:
A continuation of #2621 with feedback incorporated. Still have a couple issues to nail down and am looking for advice.

1. I made the links per-pane rather than global, which I think makes the most sense. There is a bug in this implementation though (see the `TODO` in `tty.c`) because `server_client_get_pane` returns the active pane, not the pane that the grid cell is in. Is there any way to get that pane without having to plumb that information through at every callsight for `tty_attributes`?
2. Right now each pane only frees the URL's when the pane is destroyed, but we could garbage-collect them more eagerly with some simple reference counting on the "attribute IDs". Is there a good place in the code to hook into whenever a line goes outside the history, and it's grid cells get freed?
3. Added support for the ID parameter.
4. URIs and ID parameters are re-used when possible.
5. One more `TODO` in `input.c` about writing to the buffer while parsing the escape sequence.
6. The variable names for the RB trees are pretty long but I struggled to come up with better ones, mainly cuz of the 2 different notions of 'ID'. Suggestions appreciated.


nicm:
Looks good:

- In `input_osc_8`, I think it is best not to write into `p`. You could set a sensible limit on the length of the ID (256 bytes?) and copy it on to the stack?

- I suggest instead of tying this to a pane and putting the members directly into `struct window_pane`, you create a `struct hyperlinks` with the three members you are adding and then put a `struct hyperlinks *` member in `struct window_pane`. This means 1) all the data structures (and all the tree goop) except a forward declaration for `struct hyperlinks` can go into `hyperlinks.c` out of the way 2) hyperlinks can (maybe later) be made to work in popups which do not have a `struct window_pane` 3) it will help with the next point.
 
- Do not use `server_client_get_pane`, it is to work out the active pane, not for redrawing. You need to make `tty_attributes` take a `struct hyperlinks *` (like it does `colour_palette`).

nicm:
Also:

- Don't forget `Hlr`, `Hla`, `Hls` need to go into the `TERMINFO EXTENSIONS` section of `tmux.1`.

- I think I would make the feature `hyperlinks` not `hyperlink`, but maybe that is just me.

- I think we should pass both URI and ID through utf8_strvis.

- Minor style nits:

1) Standalone comments should look like one of these:
~~~~
/* A sentence. */

/*
 * A sentence. And another one.
 */
~~~~
So full punctuated sentences and if multiline then open and closing `*` on their own line. See the start of https://man.openbsd.org/OpenBSD-current/man9/style.9#DESCRIPTION.

 2) I would avoid long comments inside structs unless it is something really critical, instead put it together in the head-of-struct comment.

 3) You are missing a space before `final_param` in the declarations at the top of `input_osc_8`.

 4) `\0` is not a "NULL terminator", it is a "nul terminator" or (better) just `\0`. `NULL` is a pointer.

nicm:
You parsing in `input_osc_8` with `strpbrk` is clever but I think it could do with a hint about what it is doing, maybe just a comment like this:
~~~~
               final_param = (*end == ';'); /* first ; is the end of keys and start of URI */
~~~~
I'd probably remove or shorten the comment about `sizeof "id="` or just something like:
~~~~
               if (end - start >= (sizeof "id=") && /* id mustn't be empty */
~~~~

wm-noble:
I'm confused why anything would need to go through `utf8_strvis`. Where do you think think that should happen and why? Other than that, everything done as suggested. Also:

1. I moved `struct hyperlinks` from window panes to screens. It seems like that's the logical container for it.
2. I added per-screen namespaces to all the parameter IDs, e.g. `param_id=<screen-namespace>.<raw-param-id>`, where `<namespace>` is a 3-digit hex number. The purpose of the ID parameter is to let the terminal emulator know if two non-adjacent hyperlinks are logically a single hyperlink. It seems like that shouldn't happen across screens, but it's not a huge deal if it does, so I didn't worry much about namespaces being re-used on rare occasions.
3. Still 1 more TODO in `tty.c`. It seems like `tty_hyperlink` ends up getting called pretty much any time something is printed to the screen (right?), so maybe we don't want to constantly preface everything with `\E]8;;\E\\`? Although it's unclear to me if that would open up some edge cases where we somehow fail to "close" a hyperlink sequence.

nicm:
Looks good, some small comments:

- You need a license and copyright on hyperlink.c.

- It seems odd having `hyperlink_put_inverse` return the same thing (`attr_id_dest`) in two different ways just to save a couple of lines of code in `hyperlink_put`.

- `hyperlink_init` would be better using the return value than returning with a pointer?

- `hl->ns % (16 * 16 * 16)` - would `hl->ns & 0xfff` be clearer since what you are doing is limiting it to three hex digits?

- This line is too long:
~~~~
+               param_id_curr = RB_MIN(param_id_to_attr_ids, &uri_curr->attr_ids_by_para\
m_id);
~~~~

- In `hyperlink_reset`, you should probably be using `RB_FOREACH_SAFE`.

- You do not need the cast to int in `attr_cmp` because it will be implicitly cast as the return type is int. (TBH doing comparison functions like this is not really good practice because it can be incorrect with `INT_MIN`/`INT_MAX`/`UINT_MAX` but I already do it all over the place so I wouldn't worry about it.)

- `utf8_stravis` - I don't think tmux should send invalid UTF-8 to the outside terminal even if the application inside requests it, so I think it would be wise to pass the hyperlink and the ID through the sanitizing functions. This shouldn't be much more than replacing the `xstrdup` in `hyperlink_put` with `utf8_stravis` and using `utf8_strvis` in `hyperlink_write_namespaced`.

nicm:

nicm:
I'm going to close this for the moment, please feel free to get in touch if you continue to work on it. Thanks!

PR #2899, opened 2021-09-18
configure.ac: add cross-compile defaults for TERM tests

rfrancoise:
3.3 added a few more `AC_RUN_IFELSE` calls that need a cross-compile default value, otherwise cross-compiling just fails.


nicm:
Applied, thanks!

PR #2930, opened 2021-10-13
Add popup-border-lines

afh:
This PR adds the `popup-border-lines` option and the `-b` flag to `display-popup` as suggested in [@nicm's comment on #2927](https://github.com/tmux/tmux/pull/2927#issuecomment-939905955):
> How about adding a popup-border-lines option also to allow -B to be set globally? Or if you want to be really clever it could have the more of the same values as pane-border-lines (none, single, double, heavy, simple?)

To test:
* Build and run tmux `./tmux -vv -f/dev/null -Ltest`
* Then test out the different popup borders (see `test_popup_border.sh` and screenshots below for example commands)

<details><summary><code>test_popup_border.sh</code>: Script to test various <code>border-popup-lines</code> combinations</summary>

```
#!/bin/ksh
./tmux set -g mouse on

./tmux set -gw popup-style bg=blue,fg=brightwhite
./tmux set -gw popup-border-style bg=cyan,fg=brightmagenta

popup_border_lines="single rounded heavy double simple padded none"

for style in $popup_border_lines; do
  clear;echo "\$ ./tmux popup -b $style \"echo popup -b $style\""
	# ./tmux display -p '#{popup-border-lines}'
  ./tmux popup -b $style "echo popup -b $style"
done

for style in $popup_border_lines; do
  clear; echo "\$ ./tmux popup -B -b $style \"echo popup -B -b $style\""
  ./tmux popup -B -b $style "echo popup -B -b $style"
done

for style in $popup_border_lines; do
  clear; echo "\$ ./tmux set -gw popup-border-lines $style"
  ./tmux set -gw popup-border-lines $style
  echo "\$ ./tmux popup \"echo popup-border-lines $style\""
  ./tmux popup "echo popup-border-lines $style"
  #./tmux set -gwU popup-border-lines
done

for style in $popup_border_lines; do
  clear; echo "\$ ./tmux set -gw popup-border-lines $style"
  ./tmux set -gw popup-border-lines $style
  echo "\$ ./tmux popup -B \"echo -B popup-border-lines $style\""
  ./tmux popup -B "echo -B popup-border-lines $style"
  #./tmux set -gwU popup-border-lines
done

clear; echo "\$ ./tmux set -gw popup-border-lines fancy"
./tmux set -gw popup-border-lines fancy
echo "\$ ./tmux popup -b fancy \"echo popup -b fancy\""
./tmux popup -b fancy "echo popup -b fancy"

#./tmux display -p '#{popup-border-lines}'
./tmux set -gwU popup-border-lines
```

</details>

* [tmux-client-62018.log](https://github.com/tmux/tmux/files/7336273/tmux-client-62018.log)
* [tmux-out-62030.log](https://github.com/tmux/tmux/files/7336275/tmux-out-62030.log)
* [tmux-server-62030.log](https://github.com/tmux/tmux/files/7336277/tmux-server-62030.log)

| Popup Border Lines | 📸 |
| --- | --- |
| `single` | image|
| `rounded` | image|
| `heavy` | image|
| `double` | image|
| `simple` | image|
| `padded` | image|
| `none` | image|
| Invalid value | image|

| Without `-B` option | With `-B` option |
| --- | --- |
| image|image|
| image | image|

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -52,9 +52,10 @@ const struct cmd_entry cmd_display_popup_entry = {
 	.name = "display-popup",
 	.alias = "popup",
 
-	.args = { "BCc:d:e:Eh:t:w:x:y:", 0, -1, NULL },
-	.usage = "[-BCE] [-c target-client] [-d start-directory] "
-		 "[-e environment] [-h height] " CMD_TARGET_PANE_USAGE " "
+	.args = { "Bb:Cc:d:e:Eh:t:w:x:y:", 0, -1, NULL },
+	.usage = "[-BCE] [-b border-lines] [-c target-client] "
+		 "[-d start-directory] [-e environment] [-h height] "
+		 CMD_TARGET_PANE_USAGE " "
 		 "[-w width] [-x position] [-y position] [shell-command]",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
@@ -354,10 +355,12 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 	struct tty		*tty = &tc->tty;
 	const char		*value, *shell, *shellcmd = NULL;
 	char			*cwd, *cause, **argv = NULL;
-	int			 flags = 0, argc = 0;
+	int			 lines = -1, flags = 0, argc = 0;
 	u_int			 px, py, w, h, count = args_count(args);
 	struct args_value	*av;
 	struct environ		*env = NULL;
+	struct options		*o = s->curw->window->options;
+	struct options_entry	*oe;
 
 	if (args_has(args, 'C')) {
 		server_client_clear_overlay(tc);
@@ -393,6 +396,23 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 	if (!cmd_display_menu_get_position(tc, item, args, &px, &py, w, h))
 		return (CMD_RETURN_NORMAL);
 
+	value = args_get(args, 'b');
+	if (args_has(args, 'B'))
+		lines = POPUP_LINES_NONE;
+	else if (value == NULL)
+		lines = options_get_number(o, "popup-border-lines");
+	else {
+		cause = NULL;
+		oe = options_get(o, "popup-border-lines");
+		lines = string_choice_from_options(options_table_entry(oe),
+		    value, &cause);
+		if (cause != NULL) {
+			cmdq_error(item, "border-lines %s", cause);
+			free(cause);
+			return (CMD_RETURN_ERROR);
+		}
+	}
+
 	value = args_get(args, 'd');
 	if (value != NULL)
 		cwd = format_single_from_target(item, value);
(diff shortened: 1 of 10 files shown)


cmd-display-menu.c
```diff
@@ -393,6 +396,23 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 	if (!cmd_display_menu_get_position(tc, item, args, &px, &py, w, h))
 		return (CMD_RETURN_NORMAL);
 
+	value = args_get(args, 'b');
+	if (args_has(args, 'B'))
+		lines = POPUP_LINES_NONE;
+	else if (value == NULL)
+		lines = options_get_number(o, "popup-border-lines");
+	else {
+		cause = NULL;
```
afh, line 405: 💥 Not seeting `cause` to `NULL` here crashes the server, I'm not sure why.
@nicm Do you see what might be the issue here?

options.c
```diff
@@ -989,28 +989,38 @@ options_from_string_flag(struct options *oo, const char *name,
 	return (0);
 }
 
+int
+string_choice_from_options(const struct options_table_entry *oe,
```
afh, line 993: ℹ️ Did some refactoring and extracted part the code from `options_from_string_choice` into this separate method, so it could be used in `cmd-display-menu.c` to convert a `-b` flag string argument to the proper popup-border-lines `int`.

screen-redraw.c
```diff
@@ -31,57 +31,9 @@ static void	screen_redraw_draw_pane(struct screen_redraw_ctx *,
 static void	screen_redraw_set_context(struct client *,
 		    struct screen_redraw_ctx *);
 
-#define CELL_INSIDE 0
```
afh, line 34: ℹ️ The removed code here was put into `tmux.h` so it could also be used by `screen-write.c`

nicm:
> I flipped these two lines around as it resulted in the wrong corner character being displayed (see screenshot below). Could this cause an issue somewhere else in tmux or could this have been an issue in tmux all along

This is an existing bug, but use of that character is quite rare. Try, for example:
~~~~
set pane-border-status top
set pane-border-lines heavy
resize-window -x 20
~~~~

afh:
Thanks for your continued thorough reviews and patience on this PR, @nicm, much appreciated.

> I think we should use BOX_LINES_SINGLE instead of using -1 (or add a BOX_LINES_DEFAULT if you prefer that, but it would need to be above the numbers used for the choices)?

I went with -1 `BOX_LINES_DEFAULT` and -1 means use the option fallback instead of the argument in certain cases.

> In popup.c: int flags and int lines should go together int flags, lines.

I'm not sure where exactly you'd like to group `flags` and `lines` together in `popup.c`. Can you please tell me the line-numbers?

> OK sure an enum would be better probably.

There is now `enum box_lines` and `enum pane_lines`. I hope I found all uses of these 🤞 

---

- [x] Best way is probably to do the option lookup in popup_display […]
- [x] Space after switch in screen_write_box_border_set: switch (popup_lines). Also the indentation looks off.
- [x] Likewise options_find_choice( has an extra space after the bracket in tmux.h.
- [x] Please use /* ... */ comments and not C++-style // comments.
**Didn't even notice as my muscle memory just took over :)**
- [x] Brackets around the value in return for the new tty-acs.c functions.

nicm:
Great stuff, applied now with some minor tweaks. Thanks!

Issue #2933, opened 2021-10-15
Trim menu item's name if it exceeds screen width?


097115:
### Issue description

If a menu item's name is longer than the screen's width, then the menu is not displayed. May be it would be great to trim menu items name in such a case? It seems that the limit should be the screen's width (like,`tput cols`) minus 8 characters.

(Sure, nobody actually uses such long names for their menu items :) But if a menu is auto-generated...)

### Required information

* tmux version: 264fe7f, reports as `tmux next-3.4`

097115:
@afh,

Thanks for your interest!

Yeah, it seems to be working. However, it trims the key, too -- like `(1)` in this example:
![](https://i.imgur.com/brFoiJu.png)

Also, as you mentioned, adding something like `...` at the end would be a nice little touch :)

Anyway, great work, and thanks again!

afh:
Thanks for trying this out. I did not think of the keys, thanks for bringing this up!
Let me re-iterate on the change in the PR.

afh:
@097115 I've updated #2936, so that the menu key will still be visible and truncated menu items show a `>` at the end.

PR #2936, opened 2021-10-16
Truncate long menu items instead of not showing menu

afh:
to `address` #2933.

Adding a truncation indicator like `...` at the end would be a nice touch for improvement.

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -57,7 +57,7 @@ menu_add_item(struct menu *menu, const struct menu_item *item,
 	struct menu_item	*new_item;
 	const char		*key, *cmd;
 	char			*s, *name;
-	u_int			 width;
+	u_int			 width, max_width;
 	int			 line;
 
 	line = (item == NULL || item->name == NULL || *item->name == '\0');
@@ -100,6 +100,8 @@ menu_add_item(struct menu *menu, const struct menu_item *item,
 	new_item->key = item->key;
 
 	width = format_width(new_item->name);
+	max_width = c->tty.sx - 8;
+	width = (width < max_width) ? width : max_width;
 	if (width > menu->width)
 		menu->width = width;
 }


nicm:
Two similar issues here:

1) What happens if `strlen(s)` is less than 3 here?
    ~~~~
        max_width = c->tty.sx - 8;
        if (strlen(s) - 3 > max_width) {
                s[max_width - 1] = '>';
                s[max_width] = '\0';
        }
    ~~~~

2) `c->tty.sx` can be less than 8 (`menu_prepare` only checks if it is at least 5). I think the code is safe even in this case, but relying on overflow for a bounds check does not make the code obvious to anyone else reading it.

afh:
I'm not entirely sure whether I understand what you mean by "overflow to bound_max"...
I've integrated your proposal with minor stylistic changes. Is there anything else that should be addressed from your end?

nicm:
I mean that in this line:
~~~~
max_width = max_width - strlen(key) - 3;
~~~~
`max_width` can overflow if it is less than `strlen(key) - 3`; in that case, this subsequent line:
~~~~
if (slen > max_width) {
~~~~
Will still be false, but it is much clearer to be explicit about what you intend to happen rather than letting `max_width` overflow and relying on the fact that `SIZE_MAX` is very big to make this condition work.

Also `max_width` could end up being zero which would mean you would write before the string.

I actually think my example is wrong and you should do this:
~~~~
    if (slen > max_width) {
         if (max_width != 0)
              s[max_width - 1] = '>';
         s[max_width] = '\0';
    }
~~~~
Since you want to end up with an empty string if there is no room at all.

But what happens now if there is no room for the key?

I would be inclined to remove the key first if there is limited space, so do something like:

- If `keylen >= max_width || slen >= max_width - keylen`, don't add the key.

- If `slen >= max_width` truncate the string and add `>`.

Does this make sense?

nicm:
I don't think you need `menu_key` to be separate, you could just use `key` as a marker for "key should be added".

I am not sure this is quite right, because it is checking for `max_width > klen` won't it trim the string even if there would be room for it without the key? Say if `slen` is 100, `max_width` is 105 and `klen` is 10,  then `max_width > klen` so it will add the key, but that will only leave width 95 for the string, so it will be trimmed, even though if the key wasn't there it would fit into the space.

How about this instead? [x.diff.txt](https://github.com/tmux/tmux/files/7379824/x.diff.txt)

nicm:
Take a look at this one please: [y.diff.txt](https://github.com/tmux/tmux/files/7381399/y.diff.txt)

afh:
Interestingly the patch didn't apply to this branch in my git workspace, so I applied your changes manually...

I like the way things are implemented, especially the use of "dynamic truncation" in the `xasprintf` format (`xasprintf(&name, "%.*s%s", (int)max_width, s, suffix);`); I was wondering how to do that, very nice!

I'd like to disagree with the notion that a key should only be added if there is space for the entire menu item text **and** the key. I think there is a minimum length for a menu item text to provide useful/enough information and in certain cases the key could help convey more context. Some (possibly rather theoritcal and extreme) examples:

* The first two window title menu entries are `Swap Left (l)` and `Swap Right (r)`. Assuming a menu width of 8 only displaying the key when there is space for item name and key tmux would show `Swap Le>` and `Swap Ri>`. Displaying the key with the item name in this particular case would be better IMHO as tmux would show `Swap (l)` and `Swap (r)`.
* For wide menus, 80 characters and above, I think truncating the menu item text short is okay as the meaningfulness (hopefully) doesn't rely on the last 4-14 (or so characters) that a key may take up, so `Lorem ipsum dolor (C-M-Right)` is likely better as it shows the key than just `Lorem ipsum dolor sit amet`.
* On the other hand a very narrow menu width of only 14 containing items `Swap Left (M-C-Left)` and `Swap Right (M-C-Right)` should probably omit the keys and show `Swap Left` and `Swap Right` instead of `Sw> (M-C-Left)` and `S> (M-C-Right)`

At this point I'd like to extract the menu item text computing into its own function and start writing table based unit tests to figure out a/the right algorithm and propose a refinement on how tmux could best render menu item text. If you'd like to include these tests in tmux that [wc]ould be nice, but I'd be perfectly okay with these only helping during development.

What are your thoughts on this, @nicm? :)

PR #2941, opened 2021-10-20
Add -T title argument to display-popup

afh:
Given that popups and menus in tmux are quite similar and share common code it seems useful to add the possibility to set a title on a popup with the `-T` argument.

For the implementation I added the capability to render a title to `screen_write_box`, this would potentially allow for a title position of either top or bottom (or even on the sides if we wanted to go really crazy). What are your thoughts on that, @nicm?

Not sure if `popup_data` is the best place to keep

To test:

* Build and run tmux `make -j$(sysctl -n hw.ncpu); ./tmux -vv -f/dev/null -Ltest`
* Display a popup with a title: `./tmux popup -T 'Popup with Title'
* Verify that the menu title refactoring did not introduce any regression by displaying a menu with a title

<details><summary><code>test_popup_title.sh</code> Test Script</summary>

```
#!/bin/ksh

clear; echo "\$ ./tmux popup -T '#[align=centre]Visor' echo \"Popup with Title\""
./tmux popup -T '#[align=centre]Popup Title' echo "Popup with Title"
clear; echo "\$ ./tmux menu -T '#[align=right]Menu Title' Menu '' '' with '' '' title 'M-C-Right' ''"
./tmux menu -T '#[align=right]Menu Title' Menu '' '' with '' '' title 'M-C-Right' ''
```

</details>

| Popup with Title | Menu with Title |
| --- | --- |
|2021-10-20-100405_905x562_scrot|2021-10-20-100418_905x562_scrot|

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -52,10 +52,10 @@ const struct cmd_entry cmd_display_popup_entry = {
 	.name = "display-popup",
 	.alias = "popup",
 
-	.args = { "Bb:Cc:d:e:Eh:t:w:x:y:", 0, -1, NULL },
+	.args = { "Bb:Cc:d:e:Eh:t:T:w:x:y:", 0, -1, NULL },
 	.usage = "[-BCE] [-b border-lines] [-c target-client] "
 		 "[-d start-directory] [-e environment] [-h height] "
-		 CMD_TARGET_PANE_USAGE " "
+		 CMD_TARGET_PANE_USAGE " [-T title] "
 		 "[-w width] [-x position] [-y position] [shell-command]",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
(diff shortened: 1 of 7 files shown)


nicm:
This warning looks like a false positive to me, the code is:
~~~~
        struct window_customize_itemdata        **item_list;
...
        data->item_list = xreallocarray(data->item_list, data->item_size + 1,
            sizeof *data->item_list);
~~~~
It is allocating an extra item onto an array of `struct window_customize_itemdata *`, so it wants to expand the list by the size of one item, which is `sizeof *data->item_list` (that is, `sizeof (struct window_customize_itemdata *)`).

If you remove the `*` you will be getting `sizeof (struct window_customize_itemdata **)` instead (although as it happens the size is the same because all pointers are the same size on all the platforms tmux supports).

Basically every use of `xreallocarray` in the codebase does this so it is funny it is only complaining about this one.

afh:
Thanks looking into the warning, explaining what is going on, and confirming this is indeed a false positive, @nicm. I may not have come across other occurrences of `xreallocarray` while perusing tmux's code, so this may not be the only one.

Glad you like the changes proposed in this PR and thanks for applying them already.
The minor changes make much sense, I wasn't sure about `const char *` and went with `char *`, still familiarising myself with the code-base.I agree that the lifecycle of title is clearer if copied.

Thanks for putting the time and effort into these reviews, I find them very helpful to get more acquainted with the tmux source and dusting off my rusty C :)

PR #2959, opened 2021-10-31
Add cursor-colour option

afh:
To test:
* Build and run tmux: `./tmux -vv -f/dev/null -Ltest`
* Set the different colours for the cursor: `./tmux set cursor-colour $(awk 'BEGIN{srand();printf ("#%x%x%x", rand()*255, rand()*255, rand()*255)}')` (see below)

ℹ️ If the changes in this PR are applied the **Options for default cursor colour.** item can be removed from the [Contribution — Small things](https://github.com/tmux/tmux/wiki/Contributing#small-things) wiki page

<details><summary><code>test_cursor_colour.sh</code></summary>

```
#!/bin/ksh

echo "Cursor should be default. Enter to continue"; read
./tmux set cursor-colour crimson
echo "Cursor should be crimson. Enter to continue"; read
echo -en "\033]12;pink\a"
echo "Cursor should be pink. Enter to continue"; read
echo -en "\033]12;#ff00ff\a"
echo "Cursor should be #ff00ff. Enter to continue"; read
echo -en "\033]112\a"
echo "Cursor should be crimson again. Enter to continue"; read
./tmux set -u cursor-colour
echo "Cursor should be default again. Enter to continue"; read
```

</details>

### DIFF
--- a/input.c
+++ b/input.c
@@ -2442,7 +2442,7 @@ input_top_bit_set(struct input_ctx *ictx)
 }
 
 /* Parse colour from OSC. */
-static int
+int
 input_osc_parse_colour(const char *p)
 {
 	double	 c, m, y, k = 0;
--- a/input.c
+++ b/input.c
@@ -137,10 +137,12 @@ static void	input_reset_cell(struct input_ctx *);
 static void	input_osc_4(struct input_ctx *, const char *);
 static void	input_osc_10(struct input_ctx *, const char *);
 static void	input_osc_11(struct input_ctx *, const char *);
+static void	input_osc_12(struct input_ctx *, const char *);
 static void	input_osc_52(struct input_ctx *, const char *);
 static void	input_osc_104(struct input_ctx *, const char *);
 static void	input_osc_110(struct input_ctx *, const char *);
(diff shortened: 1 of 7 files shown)


nicm:
Looks good but:

- The stray change in `options.c` is still there:
    ~~~~
    -                       if (options_get_number(w->options, "automatic-rename"))
    +                       int option = options_get_number(w->options, name);
    ~~~~

- Do not call `input_osc_parse_colour` in `screen_set_cursor_colour`, instead add an `input_osc_12` and `input_osc_112` to `input.c` which do the parsing and call `screen_set_cursor_colour` (you can copy `input_osc_10` and `input_osc_110`).

afh:
Oh, nice! Why didn't I think of that :) I've made the changes you requested.

nicm:
I have applied this now with a couple of minor changes:

- You forgot to remove the call to `input_osc_parse_colour` in `screen_set_cursor_colour`.

- Because the colour could now be an ANSI colour, `input_osc_colour_reply` also needs to call `colour_force_rgb`.

Thanks!

afh:
Great! Thanks for making those minor changes, @nicm and apologies for not catching these on my end.

PR #2962, opened 2021-10-31
arguments: handle invalid states

mathstuf:
Since C enums are not strongly typed enough, any actual value can occur
(at least without whole-program analysis), so explicitly handle invalid
states as crashes instead of undefined behavior.

---
I *think* `fatalx` makes sense here, but may that is too harsh (since if they happen, we're in UB anyways by falling off the end of a function not returning `void`).

### DIFF
--- a/arguments.c
+++ b/arguments.c
@@ -108,6 +108,7 @@ args_value_as_string(struct args_value *value)
 	case ARGS_STRING:
 		return (value->string);
 	}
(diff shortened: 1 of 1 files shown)


nicm:
I don't mind this, but the fatalx is better outside the switch, a default
case will hide the warning if members are added to the enum.

nicm:
Applied, thanks!

PR #2977, opened 2021-11-16
Move option default-terminal to session options in man page

trygveaa:
CHANGES says that the default-terminal option was changed to a session
option in version 2.1, so it should be under session options in the man
page.


nicm:
It was changed to a server option.

trygveaa:
Oh, right I see that in the history of the man file now. Then `CHANGES` is wrong though, it says `'default-terminal' is now a session option.`. Also, all examples use `-g` instead of `-s`, so I guess they are not updated.

PR #3003, opened 2021-12-06
treat C1 control characters as being zero width

### LINKED ISSUE #3002: incorrect character width for c1 control characters
### Issue description

tmux appears to believe that characters such as `U+0099` have a width of 1, but as far as i can tell, this is incorrect (they should have a width of 0). this leads to display corruption if these characters are printed - for instance, you can reproduce this by running `echo -e 'foo\xc2\x99bar'` and moving the copy mode selection around. i can reproduce this on latest git (commit 10b3cd17), under both alacritty and xterm.

running this code:

```c
#include <utf8proc.h>
#include <stdio.h>
#define _XOPEN_SOURCE
#include <wchar.h>

int main(int argc, char *argv[])
{
    printf("%d\n", utf8proc_charwidth(0x99));
    printf("%d\n", wcwidth(0x99));
}
```
produces this output:
```
0
-1
```
on my system, so i don't think this is an underlying library issue, although i'm not entirely sure how tmux uses this information. looking through the logs, i see this:
```
1638778097.748305 UTF-8 \302\231, wcwidth() -1
1638778097.748312 input_top_bit_set 2 '\302\231' (width 1)
```
which sounds like it's getting -1 from wcwidth, but using a width of 1 anyway? not at all familiar with the tmux codebase though.

### Required information

* tmux version (`tmux -V`).
  * `tmux next-3.4`
* Platform (`uname -sp`).
  * `Linux unknown` (arch linux)
* $TERM inside and outside of tmux (`echo $TERM`).
  * inside tmux is `screen`, outside is either `alacritty-direct` under alacritty or `xterm` under xterm
  * attached
[tmux-client-2741948.log](https://github.com/tmux/tmux/files/7659122/tmux-client-2741948.log)
[tmux-out-2741950.log](https://github.com/tmux/tmux/files/7659123/tmux-out-2741950.log)
[tmux-server-2741950.log](https://github.com/tmux/tmux/files/7659124/tmux-server-2741950.log)

doy:
wcwidth uses -1 to mean both "this character is not printable" and "this
character is unknown", which means that always using a default of 1
gives incorrect results for control characters (they are not printable,
so they should have a width of 0, not 1). we don't have to care about C0
control characters here, because they are never treated as text, but C1
control characters are, and so we should use the correct values for
them.

this is only an issue with wcwidth - utf8proc returns useful values here
already.

this fixes #3002.

another option here could be to drop C1 control characters entirely, so that they never show up as text in the first place? that feels more actually correct, but also more invasive, and i'm not sure what the implications of that might be.

### DIFF
--- a/utf8.c
+++ b/utf8.c
@@ -248,7 +248,14 @@ utf8_width(struct utf8_data *ud, int *width)
 	 * to the terminal without tmux.
 	 */
 	if (*width < 0) {
+#ifdef HAVE_UTF8PROC
 		*width = 1;
+#else
+		/* C1 control characters are non-printable, so they are always
+		 * zero width. (This is true of C0 control characters as well,
+		 * but they are handled separately from text already.) */
+		*width = (wc >= 0x80 && wc <= 0x9f) ? 0 : 1;
+#endif
 		return (UTF8_DONE);
 	}
 #endif


nicm:
Thanks, this looks good, but I've decided to just remove this `#ifdef` entirely and see how we get on with that. The only problems I remember were with macOS and I am not happy enough to tell anyone using macOS that they must use utf8proc.

If we have to put it back I will include your change. Thanks!

PR #3012, opened 2021-12-21
tmux.h: Don't pack pointer-containing struct grid_line

jrtc27:
On CHERI, and thus Arm's prototype Morello architecture, C language
pointers are represented as hardware capabilities, which have bounds,
permissions and other metadata to enforce spatial memory safety and,
importantly, have an associated non-addressable tag bit to enforce the
validity of every capability (that is, the only way to construct a
capability with the tag bit set is by deriving a new capability as a
subset of an existing valid capability, and any action such as
overwriting the capability in memory with an arbitrary bit pattern will
clear the tag, rendering it no longer valid). There is one tag bit per
capability-sized word of memory, and thus capabilities must be stored at
aligned locations otherwise they cannot be tagged (load/store capability
instructions will take an alignment fault, and memcpy will silently drop
the tag). This therefore means that capabilities cannot be stored in
structs that use the GNU packed attribute (unless care is taken to
ensure they will be aligned at run time), which is intended to be used
when mapping hardware data structure layouts to C structs, not in
general purpose code.

Thus, since struct grid_line contains pointers, remove __packed from its
definition, increasing the portability of the code whilst adding no
space overhead on 32-bit architectures and only 8 bytes of overhead on
64-bit architectures (which, if that is genuinely a concern, can be
entirely mitigated by reordering the fields). This will also make the
code faster on architectures that don't natively support unaligned
acceses and currently have to use byte loads/stores to reconstruct the
multi-byte values.

### DIFF
--- a/tmux.h
+++ b/tmux.h
@@ -677,7 +677,7 @@ struct grid_line {
 	struct grid_extd_entry	*extddata;
 
 	int			 flags;
-} __packed;
+};
 
 /* Entire grid of cells. */
 struct grid {


nicm:
It is not OK to make this struct bigger on any platform.

nicm:
I don't see much - if any - hardware available to the general public that uses this architecture, so I suspect this change is slightly premature.

But anyway we can do something... tmux will have a great many of these structs so it is important that they not use more space, the `__packed` is not there for entertainment :-). It does look like now reordering the struct would be enough.

I think for the moment it would be best to reorder the struct and perhaps `#ifdef` the `__packed` away on your architecture in `compat.h` rather than removing it completely.

jrtc27:
About 1000 boards will be made available to industry and universities over the coming year and, if they're anything like me, one of the first things they'll want to install is tmux so they have a sane way to manage multiple terminal sessions. For those without real hardware, we have a fork of QEMU (you're not going to want to compile code on it, but it's fast enough for testing something like tmux), and Arm have a slower full system model ("FVP").

I'm unconvinced the space saving matters, tmux uses so little memory as it is, and this is only for grid_line, it's not for every cell. But fine we can do something uglier. The alternative would be something like:
```
diff --git a/tmux.h b/tmux.h
index 9cf60ecb..9ccbab55 100644
--- a/tmux.h
+++ b/tmux.h
@@ -677,7 +677,11 @@ struct grid_line {
 	struct grid_extd_entry	*extddata;
 
 	int			 flags;
-} __packed;
+}
+#ifndef __CHERI_PURE_CAPABILITY__
+__packed
+#endif
+;
 
 /* Entire grid of cells. */
 struct grid {
```
or you could introduce some kind of __packed_if_not_cheri if you really wanted to but that seems overkill for a single __packed. Note that __packed _is_ fine on the other structs for CHERI, they only contain integers.

Oh and we can't just ifdef out `__packed`, FreeBSD defines that in sys/cdefs.h. On Linux maybe you could, but we're a fork of FreeBSD.

nicm:
`compat.h` is included after system headers so it should be fine to `#undef` it away, this is what we do for `__dead`. Although let's also reorder the struct so compilers are inclined to omit padding. Does this work? [x.diff.txt](https://github.com/tmux/tmux/files/7755281/x.diff.txt)

The whole thing seems a little dodgy to me all round TBH. Not that I think this is a good idea, but what would happen if I cast the pointers to and from `intptr_t`?

nicm:
Maybe we could get away with removing `__packed` for `grid_line` if it is manually packed, although I am concerned because people routinely do set `history-limit` to enormous values... probably they shouldn't be doing that on boxes with low mem, the default is pretty small. I'll think about it a bit more.

jrtc27:
> `compat.h` is included after system headers so it should be fine to `#undef` it away, this is what we do for `__dead`. Although let's also reorder the struct so compilers are inclined to omit padding. Does this work? [x.diff.txt](https://github.com/tmux/tmux/files/7755281/x.diff.txt)

That's not true; compat/getopt.c includes it before standard headers and compat/imsg(-buffer).c include it before imsg.h which includes stdint.h. A few other places also include it before xmalloc.h and tmux-protocol.h but neither of those include any further headers. It probably works in practice but I cannot condone that approach.

> The whole thing seems a little dodgy to me all round TBH. Not that I think this is a good idea, but what would happen if I cast the pointers to and from `intptr_t`?

`intptr_t` is not implemented as a long but a full capability under the hood, and despite it being optional in C you very quickly find it needs to work (as do bitwise operations) if you want to have any hope of compiling real software (the current implementation for anything other than normal pointer arithmetic, which has dedicated instructions to add an offset, is "read integer address, perform arbitrary arithmetic, set address on original capability to new value", which ensures the metadata, including tag, is preserved; if that's too inefficient for certain common intptr_t idioms, dedicated instructions can be added, and in fact Arm already have align up/down as dedicated instructions for Morello).

Whilst I'd like to be able to support `__packed`, and you can _almost_ make it work if you're willing to waste ~4.09x (log2(17), since each capability-sized word has 16 possible starting locations for a capability, plus no capability, though there'd be some redundancy with adjacent locations) the amount of space on tag bits, it's then impossible to have an efficient memcpy implementation without a dedicated memcpy instruction in hardware (which is maybe a good idea anyway, Arm seem to be wanting to add it and memset in some future revision of AArch64, but it's sad to _require_ it, we want CHERI to be portable across architectures).

nicm:
TBH it doesn't really matter that much what `compat/*` do since they will never be using `struct grid_line`.

It seems like if `intptr_t` has to work then pointers in packed structs are pretty much the same problem in a slightly different way since I could as well replace these pointers with `intptr_t` and add a couple of casts. But what do I know, and if nobody does it in practice...

Anyway, I think I will actually just remove the `__packed`. We do want the compiler to make the compiler as small as possible (which is the intent of `__packed`), but it is not as critical as `struct grid_cell`, and if it is manually ordered in a better way I think it will be fine.

nicm:
I have done this in OpenBSD now, it will be in GitHub as soon as it syncs. Thanks!

jrtc27:
Great, thank you

PR #3019, opened 2021-12-27
Add access control list

dlas-snoyl:
To test:

- Build and run a shared tmux session: `make && ./tmux -S /tmp/tmux.socket -vv -f/dev/null -Ltest`
- Verify another user can't join: `./tmux -S /tmp/tmux.socket attach`
- Host allows guest to join: `allow-whitelist`
- Host allows guest to write: `allow-write`
- Host remove guest write: `deny-write`
- Host removes guest: `deny-whitelist`

This PR adds an access control list to reduce the possibility of a random individual joining a shared tmux session. 

It also adds commands for adding/removing people from the access control list as well as limiting their ability to write to the command line when joining. The owner of the session is ability to give/revoke write permissions through commands.


nicm:
Thanks - I am not opposed to doing this. I understand this is a work in progress, but here are some initial comments:

- I don't think we should limit each socket to one session, this needs to work with multiple sessions. I don't know whether you want to allow access to all sessions or try to whitelist per-session, either is fine. The former is probably much easier and would do for a start.

- Do not `chmod` the socket, leave that to the user to do manually. This should be very hard to turn on by accident.

- I don't see any point in making this optional, so we can do without `TMUX_ACL`.

- I do not want four new commands for this, please roll them all into one command. I would call it `server-access` or `user-access` or something like that and then have flags for add/remove/enabled write/disable write.

- Usage strings do not use `<>` around arguments.

- You need to update the header on new files to add your copyright.

- Code style does not match, please look at the existing code and https://man.openbsd.org/OpenBSD-current/man9/style.9. For example, please use 8-character tabs for indentation; limit lines to 80 columns; do not use C++-style `//` comments; etc. I will fix any small issues before commit but please try to get it closer.

- You are missing a few bits of error checking, for example `getpwuid` can fail.

- Is `SO_PEERCRED` available on Solaris, HP-UX, Cygwin? Do we need it to be wrapped in something in `osdep-*` (`osdep_so_peercred`?), even if it just fails and ACLs are not supported (it will still need to build)?

- Please use `TAILQ_*` rather than adding new uses of `SLIST_*`.

- Please do not `fgets`, use `getline` instead. Never use `strcat` or `strcpy` under any circumstances, use `strlcat` or `strlcpy`.

- Hacks and sleeps and crashes will all need to be fixed and removed.

- Please just use `strerror` directly intead of rolling your own wrapper.

- I don't think tmux should automatically write the whitelist to a file, the user can configure it themselves with commands in a config file.

nicm:
TBH an RB tree (`RB_HEAD` etc) keyed by UID would seem much better than an `SLIST` *or* `TAILQ`.

It might be simpler not to bother with `is_owner` but just special-case the owner UID in the access check functions as always permitted, but maybe I am missing something about how it works.

nicm:
This is looking much closer now. A few comments:

- The usage for `server-access` should probably look more like:
~~~~
[-adrw] user
~~~~
The rules (at least one must be given, -a/-d and -r/-w are mutually exclusive) can be documented in the man page.

- Does the user really need to be expanded as a format in `cmd_server_access_exec`? If so, could probably use `format_single` here.

- I don't know why you have a space at the start of all `cmdq_error` messages but it should not be needed.

- I think `uid_cmp` can be static since you are using `RB_GENERATE_STATIC`.

- Perhaps `struct acl_user` could be local to `server-acl.c` and just have a forward declaration in `tmux.h`. Anything not used outside the file should be static - `tmux.h` should only declare the minimum needed by other files.

- But don't include a forward declaration for system structs like `struct passwd`. Include `pwd.h` instead.

- `proc_acl_get_ucred` needs to go in the `proc.c` block in `tmux.h`, although maybe you are intending to move it around anyway.

- Style is still a bit all over the place :-).

Thanks!

nicm:
OK I have taken a look at this again.

*BSD actually has a function called `getpeereid` which has a much more sensible prototype than pissing around with `getsockopt`, so I think a better option is to use that and add a `compat/getpeereid.c` for platforms without it. `SO_PEERCRED` is a bit haphazardly supported - OpenBSD has it, but it returns `struct sockpeercred` and `struct ucred` doesn't exist; FreeBSD calls it `LOCAL_PEERCRED` instead; and macOS has it somewhere but I am not sure what it returns (not `struct ucred`). So hiding all that inside a single compat file seems a better way to go.

Please take a look at this diff against your current code which does this and also some other cleanup outside the main files. I do not have a Linux box around at the moment so I have not been able to test the actual compat code, but it works on macOS which does have `getpeereid`.

[3019_0.diff.txt](https://github.com/tmux/tmux/files/8120472/3019_0.diff.txt)

Some other comments, apologies if I said these before and you have not got to them yet:

- `cmd-server-access.c` could perhaps benefit from the exec calling one of two functions `cmd_server_access_deny` or `cmd_server_access_allow` which can do error checking up front and return early in error cases.

- Also you should return `CMD_RETURN_ERROR` for errors (if you call `cmdq_error`), so that command sequences are aborted.

- `cmd_server_access_entry` says 0 arguments is allowed, but then `cmd_server_access_exec` uses argument 0 without checking if it is `NULL`. I think it should be `1, 1` in `cmd_server_access_entry`, not `0, 0`.

- I do not see any need for `owner` in `server-acl.c`? Why not just call `getuid` when needed?

Generally it looks much better, the bulk of the needed changes now will be cleanup and style.

nicm:
I wrote the version of `compat/getpeerid.c` in the above diff, but it turns out OpenSSH actually has a version already (here: https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/bsd-getpeereid.c). I am not sure about their choice to return `getuid` for the unsupported case (I think failing would be a better choice for tmux), but they do have code to use `getpeerucred` which appears to be what we will need for Solaris. Anyway, if we get it working for Linux and other systems with only `SO_PEERCRED` then adding `getpeerucred` will be easy enough later.

dlas-snoyl:
Pushed the merged changes, unfortunately I haven't had a lot of time to look into the getpeereid side of things.

nicm:
thanks i will look at it again next week probably

nicm:
OK here is another diff, including the last one: [access3.diff.txt](https://github.com/tmux/tmux/files/8199356/access3.diff.txt)

I decided not to rename the file in the end since `server_access_` seems too long for all the tree bits, `server_acl_` is fine. I also made these changes:

- Add a read only flag to the user entry so that users do not always start read only.

- Change some prefixes, `acl_` to `server_acl_`.

- Shorten `user_id` to `uid`.

- Use `RB_FIND` instead of walking the tree.

- Remove logging and `status_message_set`.

- Flatten a lot of functions to return early instead of nested `if`s.

- Tidy up/add comments and other style bits.

Please look at/test and let me know of any problems.

It still needs man page bits added to tmux.1.

There are three other things I think we should include, but they can all be done later:

- When you attach but are not allowed, you get `server exited unexpectedly` instead of a useful message. I think just doing `close(newfd)` in `server_accept` is not enough - it needs to follow the normal client exit path so it can set `exit_message`.

- I think that messages when someone else attaches would be nice but 1) they can only apply when someone actually attaches, not when they use any tmux command 2) using `status_message_set` like this not the right way to do it. Probably the right thing to do is to add some hooks so the user can show messages if they want them. 

- Formats (`client_user`/`client_uid`) for the user to whom the client belongs would be useful. It would probably be best to save client UID/user in `struct client *` on accept and then they can be used both for formats and by `server-acl.c`.

nicm:
I have committed the compat `getpeereid` bits and also added formats for `client_user`/`client_uid` and server `uid`/`user`. I'll give you an updated diff once GitHub syncs these changes up.

nicm:
Here are the latest changes, this is on top of master, not on top of your branch: [accessX.diff.txt](https://github.com/tmux/tmux/files/8206888/accessX.diff.txt)

I think this looks almost ready to go, thanks for your work so far. Let me know if you see any problems.

This is what I think needs to be changed before it is done, mostly very small changes:

- I think `server-access -r` and `-w` should imply `-a` if `-a` is not given.

- We need a way (`server-access -l`?) to list the current access list (entirely or just the state for one user).

- The client should be told to exit more nicely when rejected so there is a better message.

- It should fire a hook when a user attaches to a session.

- I would like to see the user in the message list, I am not sure exactly how best to do this, maybe just add it as [user] after the client name if the user is not the owner of the server.

Let me know if you want to look at all or some of these.

nicm:
Great stuff, the third one is probably the trickiest, let me know if you have any questions.

In case you are not  aware by "message list" I mean `C-b ~` (`server_set_message` IIRC).

nicm:
https://github.com/tmux/tmux/commit/68002d309641edf84bf409afa71dc8998e491220 looks good but some minor comments:

* ` .args = { "adrwl", 0, 1, NULL },` <-- since you change the minimum arguments to 0, `args_get(args, 0)` can now return `NULL`, so you will need to check if this is the case and report and error before looking up the user or tmux could crash.

* `/* displays current access list */` <-- comments on their own on a line should be fully formed sentences, so capitalize "Displays" and add a period at the end. But in this case I'd probably just do without the comment and remove it.

* `int count; /* skip root being printed */` <-- this seems like an odd way to skip root. Why not just check if `loop->uid` is zero? Root will always have UID 0.

* It would be nice also to show the readonly state? Maybe add "(R)" or "(W)", or "(readonly)".

Thanks!

dlas-snoyl:
I am not quite sure how to show the readonly state, I assume it would be on the status bar?  I haven't gotten around to messing with the message list yet or firing a hook when a user attaches.

I tried taking a look into the client exiting more cleanly but haven't made any meaningful progress, still a little confused on how that would take place.

nicm:
I meant in the list of users with server-access -l  it can show it

nicm:
These changes look good now. I'll take a look at the client exit on `accept`.

The message list (`C-b ~`) I think it will be enough just to change these lines:
~~~~
cmd-queue.c:569:                        server_add_message("%s key %s: %s", name, key, tmp);
cmd-queue.c:571:                        server_add_message("%s command: %s", name, tmp);
~~~~
To lookup the user and do something like:
~~~~
user = blahblah;
if (user != NULL)
    server_add_message("%s[%s] key %s: %s", name, user, key, tmp);
else
    server_add_message("%s key %s: %s", name, key, tmp);
~~~~
Just to give some kind of audit trail of who has been doing what.

Adding a hook on attach should be pretty easy, look at this change 818b2176ef52a13867905d3d3e0269dc1b4d6fb4, except it will be `OPTIONS_TABLE_HOOK` not `OPTIONS_TABLE_WINDOW_HOOK` and fired with `notify_session`.

nicm:
If possible can you pull the PR up to master so the diff includes changes I have already applied?

nicm:
(It may be worth for the message list adding a helper like `static void cmd_queue_add_message(struct client *, const char *, ...)` which adds the client+user as prefix as necessary, but see what you think.)

nicm:
Here are two diffs, the first is everything in this branch but against master and with some trivial changes (mostly indentation). The second is on top of that and sets the client exit message to "access not allowed" instead of just closing the socket, it is also adds code to flush the imsg immediately after sending the identify messages, otherwise pledge can kill the process because we have dropped "sendfd" but the outgoing file descriptor is still sitting in the imsg at exit.

[access-all.diff.txt](https://github.com/tmux/tmux/files/8284970/access-all.diff.txt)

[access-client-exit.diff.txt](https://github.com/tmux/tmux/files/8284971/access-client-exit.diff.txt)

nicm:
Here is a complete diff against master which may be easier to work with at this point: [accessX.diff.txt](https://github.com/tmux/tmux/files/8287237/accessX.diff.txt)

nicm:
Can you merge master into this PR? It is very difficult to see what the changes are alone now.

nicm:
Thanks, here are the remaining differences between my tree and the PR at the moment: [x.diff.txt](https://github.com/tmux/tmux/files/8306761/x.diff.txt)

- Since we enforce read-only on accept, we no longer need to check in `attach-session`/`new-session`, so `server_acl_join` is not needed.

- Indentation in `cmd-server-access.c` is all spaces and should be tabs; also some other minor spacing nits.

- Don't need `pwd.h` in `tmux.h`.

nicm:
This works well for me now.

Even with this code, I don't think it is a good idea for people to make a socket mode 666 and let everyone have a go, I think it would still be wise to make use of group permissions.

Although I'm not sure it is possible to stop people shooting themselves in the foot if they really want, but I wonder if there is any way to make it difficult. It seems like there isn't much we can do about it if the user changes the socket permissions after tmux starts... even if we checked the permissions when someone connects, it wouldn't be any different from checking the UID like we now do, and there is no way to do it that isn't going to suffer from a TOUTOC race.

dlas-snoyl:
Sounds good.

With that in mind are there currently any further changes you can see this benefiting from?

nicm:
I'd like to see the message list (`C-b ~`) show the user for non-owner commands/key presses.

I think we can do without hooks/notify for the moment, I think people can use the existing hooks.

nicm:
Oh I forgot, the other thing we need is man page changes! Probably just an entry for `server-access`.

dlas-snoyl:
I added the entry for server-access, I tried to follow the conventions and formatting in the rest of the man page but let me know if there any issues.

nicm:
This looks good, here is a diff with some changes on top of yours:

- The flags should be sorted.

- New sentences get a new line (this is standard for OpenBSD man pages as it helps to make diffs easier to read).
 
- I think `user` is better than `name` for the argument.

- It is OK to refer to the argument by name in the text so use `.Ar user` instead of `the user` where appropriate.

- Refer to "the server" not "the session" since that is what we are granting access to.

- Add a note about file system permissions.

- Some minor language tweaks.

This also adds the user to `show-messages` and fixes a check for `getpwuid`: [access-changes.diff.txt](https://github.com/tmux/tmux/files/8378105/access-changes.diff.txt)

Screenshot 2022-03-30 at 08 32 01

Here is the full diff with all changes: [access-all.diff.txt](https://github.com/tmux/tmux/files/8378111/access-all.diff.txt)

I think this is good to go now unless you find any issues. OpenBSD is going to lock for release soon so I think this too big a change to apply now, but I will put it in GitHub soon and apply to OpenBSD after unlock.

Thanks!

nicm:
I have applied this to GitHub now, thanks!

nicm:
Just realized when I was merging this to OpenBSD that I forgot to credit you in the commit message, sorry about that! I've added it in the OpenBSD commit message so it will be in the merge commit over to GitHub.

dlas-snoyl:
Thank you!

PR #3022, opened 2021-12-31
Border indicators

eMPee584:
Salut..
after eight years, this is another attempt to implement a toggle for the split-colour drawing of borders and a feature to display arrow markers to "quickly find the pane that has input focus with low cognitive overhead." Although the end result is quite concise it was tricky to implement. Tmux and especially the core terminal plumbing is compact and efficient C code and with its low comment ratio is all but easy to grok for a non-tmux-core-developer. The functions could benefit from more verbose description of their black magic.. 🔮😅

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -63,6 +63,9 @@ static const char *options_table_cursor_style_list[] = {
 static const char *options_table_pane_status_list[] = {
 	"off", "top", "bottom", NULL
 };
+static const char *options_table_pane_border_indicators_list[] = {
+	"off", "colour", "arrows", "both", NULL
+};
 static const char *options_table_pane_border_lines_list[] = {
 	"single", "double", "heavy", "simple", "number", NULL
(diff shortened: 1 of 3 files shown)


nicm:
I think this should be one option (or possibly rolled into `pane-border-style`, although that might be annoying), something like `pane-border-indicators` with values `colour`, `arrows`, `both`.

ACS `+-,.` are not widely supported by the look of it so I wonder if this should only try to use the UTF-8 characters.

nicm:

eMPee584:
.. so I merged the two options.. `both` doesn't work too well actually though because the arrows are not drawn in the part not drawn with the active style..

nicm:
I think `both` looks fine TBH. I have applied this now with a couple of relatively minor changes (I added `#define`s for the option values and also tweaked some if statements). Thanks!

PR #3023, opened 2022-01-01
updates configuration files docs

dandev486:
## Summary

As can be noticed from the changelog for [version 3.1](https://raw.githubusercontent.com/tmux/tmux/3.1/CHANGES) and by comparing [version 3.1 with version 3.0](https://github.com/tmux/tmux/compare/3.0...3.1#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR16) a configuration file located at `~/.config/tmux/tmux.conf` is supported, but this option is not documented on the [man page](https://man.openbsd.org/tmux.1#FILES). I took a look at the [master](https://github.com/tmux/tmux/blob/master/tmux.1#L6627) and [3.3rc](https://github.com/tmux/tmux/blob/3.3-rc/tmux.1#L6437) versions of the file to make sure it wasn't already included on later versions.

This pull request is a proposal for inclusion of the aforementioned file. Thank you in advance!

## Test Evidence

Documentation changes were tested with `nroff -mdoc tmux.1 | less`.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -140,7 +140,9 @@ By default,
 loads the system configuration file from
 .Pa @SYSCONFDIR@/tmux.conf ,
 if present, then looks for a user configuration file at
-.Pa ~/.tmux.conf .
+.Pa ~/.tmux.conf 
+or
+.Pa ~/.config/tmux/tmux.conf .
 .Pp
 The configuration file is a set of
 .Nm
@@ -6628,6 +6630,10 @@ options.
 .Bl -tag -width "@SYSCONFDIR@/tmux.confXXX" -compact
 .It Pa ~/.tmux.conf
 Default
+.Nm
+configuration file.
+.It Pa ~/.config/tmux/tmux.conf
+User-specific
 .Nm
 configuration file.
 .It Pa @SYSCONFDIR@/tmux.conf


dandev486:
In time, prior to opening the pull request, I took a look at the contribution guidelines and I found a list of ideas for contributors and guidelines for opening issues, but not for pull requests. Please, let me know if there is something that should be changed or can be improved.

nicm:
Thanks but TBH I don't know about listing nonstandard paths here, I think it is enough to list the conventional paths, particularly since these locations are only supported in portable tmux.

dandev486:
Do you think that adding and additional note about the portability caveat would make the information worth of being documented? The intent was to make explicit that tmux also supported this path, since currently the information requires looking online at either the changelog file or the source code itself, both of which an user may not have at hand.

In case of a negative answer, feel free to decide upon closing this pull request, no problem at all!
Thank you for sparing time to look into this.

nicm:
You're right, it is best to mention it somewhere - and actually we can put it *only* in the portable tmux.1. So I have applied your diff with some minor changes. Thanks!

dandev486:
Thanks @nicm!

PR #3025, opened 2022-01-03
Add prefix3 and prefix4

aidanhs:
https://github.com/tmux/tmux/commit/535286c05aee55c0127cdabada8e54582905ef8e says "People who want three prefix keys are out of luck :-)."

I want four prefix keys!
 - `C-\` and `M-\` on UK English keyboards (`\` is between left shift and 'z' https://en.wikipedia.org/wiki/File:KB_United_Kingdom.svg)
 - `C-z` and `M-z` on US English keyboards which I use much more rarely (to retain the muscle memory for prefix/prefix2 https://en.wikipedia.org/wiki/File:KB_United_States-NoAltGr.svg)

I've lightly tested and both the send-prefix arguments and using them as prefixes seems to work.

Note: I didn't add error handling for when people try and pass multiple arguments to `send-prefix`, instead it'll just silently use the highest prefix number passed (e.g. `send-prefix -2 -3` will just send `prefix4`).

### DIFF
--- a/cmd-send-keys.c
+++ b/cmd-send-keys.c
@@ -47,8 +47,8 @@ const struct cmd_entry cmd_send_prefix_entry = {
 	.name = "send-prefix",
 	.alias = NULL,
 
-	.args = { "2t:", 0, 0, NULL },
-	.usage = "[-2] " CMD_TARGET_PANE_USAGE,
+	.args = { "234t:", 0, 0, NULL },
+	.usage = "[-2|-3|-4] " CMD_TARGET_PANE_USAGE,
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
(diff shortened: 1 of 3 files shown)


nicm:
Bind `switch-client -Troot`.

aidanhs:
It looks like this works:
```
bind-key -Troot C-z switch-client -Tprefix
bind-key -Troot M-z switch-client -Tprefix
```

Is prefix2 just there for historical reasons then? Would you be open to a PR expanding the docs for `prefix2` to point people in the right direction if they want more prefix keys?

nicm:
Sure but I'd keep it to one or two lines under switch-client or prefix2. I
don't remember anyone else asking about more than two prefixes so probably
not worth too much.

aidanhs:
Closing this PR, will open a new one

PR #3038, opened 2022-01-15
Since xstrdup() calls fatalx() if strdup returns NULL, it will crash …

4D617278:
…if str is NULL. The call to popup_display sets the title parameter to NULL and causes tmux to crash after xstrdup is called on title in popup_display. The bug was added in commit acba07629ebf2dc2f0c316f110493e720b30757c in GitHub issue 2941. The intention was to have the owner of the title string free the string while keeping a copy in popup, but it was likely assumed that title was not a NULL string as cmd_display_exec_cmd in cmd-display-menu.c calls popup_display with title as a non-NULL string.


4D617278:
This bug is triggered by editing a buffer by pressing e in buffer mode.

nicm:
`pd->title` is probably better as NULL in this case, so I made a slightly different fix. Thanks!

Issue #3115, opened 2022-03-12
When `tmux kill-server` returns, are we sure the server is fully killed?


alexisfrjp:
I have a bash script killing an existing server and creating a new one right away.

```
tmux -L mysrv kill-server
tmux -L mysrv new-session -d
```

The `new-session` command doesn't always return 0, especially when my tmux has been running for a while. (lots of stdout I guess)
Also, the file `/tmp/XXX/mysrv` isn't automatically deleted.

Is that normal?

nicm:
No there is no synchronization so the server may exist for a short while
after the command returns.

alexisfrjp:
Is it a command working as a daemon? Or is it unrelated to tmux?
I couldn't find any mention of it in the manpage, I'll add a check after the kill-server.
Thank you.

nicm:
I'm going to close this since it is working as expected but feel free to comment if you have questions or if you want to open another PR to improve this. Thanks!

PR #3116, opened 2022-03-13
Wait for child to be killed: kill-server

alexisfrjp:
Wait for the process to be killed, it must be done here, not in user's scripts using tmux.

Issue #3115

### DIFF
--- a/cmd-kill-server.c
+++ b/cmd-kill-server.c
@@ -17,6 +17,7 @@
  */
 
 #include <sys/types.h>
+#include <sys/wait.h>
 
 #include <signal.h>
 #include <unistd.h>
@@ -54,8 +55,11 @@ const struct cmd_entry cmd_start_server_entry = {
 static enum cmd_retval
 cmd_kill_server_exec(struct cmd *self, __unused struct cmdq_item *item)
 {
-	if (cmd_get_entry(self) == &cmd_kill_server_entry)
+	if (cmd_get_entry(self) == &cmd_kill_server_entry) {
+		int status;
 		kill(getpid(), SIGTERM);
+		wait(&status);
+	}
 
 	return (CMD_RETURN_NORMAL);
 }


cmd-kill-server.c
```diff
@@ -54,8 +55,11 @@ const struct cmd_entry cmd_start_server_entry = {
 static enum cmd_retval
 cmd_kill_server_exec(struct cmd *self, __unused struct cmdq_item *item)
 {
-	if (cmd_get_entry(self) == &cmd_kill_server_entry)
+	if (cmd_get_entry(self) == &cmd_kill_server_entry) {
+		int status;
 		kill(getpid(), SIGTERM);
+		wait(&status);
```
krader1961, line 61: This doesn't do what you think it does. You can't wait for the current process to terminate. The `wait` function waits for a child process to exit. And it will return when any, random, child process terminates. Not when all of them have terminated.

alexisfrjp: I see, maybe using `job_check_died(pid_t pid, int status)`? It'd be useful to make sure all the children are dead before returning `kill-server`.
But it uses a pid that is supposed to be dead and kernel can reassign it, leading to a false negative.

krader1961: The `kill(getpid(), SIGTERM)` is telling the current tmux daemon to terminate. See the logic in `server_signal()` and `server_send_exit()`. You're approaching this the wrong way. The check for whether the server is dead belongs in the tmux client code, not the server code. The tmux client that forwards the `kill-server` to the daemon should, arguably, check that the daemon is actually dead and only finish (i.e., exit) when that is true.

alexisfrjp: Indeed, I'll re-open a merge request once it's done and tested. Thank you!

nicm:
The client does not know the server PID, you would need to extend one of the messages (maybe `MSG_VERSION`?) to tell it. You cannot wait on it since the parent is not a child process, so you would need to poll in a loop with something like:
~~~~
while (kill (0, client_server_pid) == 0)
    usleep(100000);
~~~~

PR #3119, opened 2022-03-16
Add support for systemd socket activation

### LINKED ISSUE #3117: [Feature request] Support systemd socket activation
### Issue description

Reopening https://github.com/tmux/tmux/issues/2398 because socket activation would enable to launch on-demand a [fully hardened and sand-boxed](https://wiki.archlinux.org/title/User:NetSysFire/systemd_sandboxing#Sandboxing_systemd_service_units) `tmux` (eg. to run a persisting `irssi`).
This is currently not possible with the usual `systemd-socket-proxyd` palliative [because](https://github.com/systemd/systemd/issues/22744) `tmux` uses `SCM_RIGHTS` [to get the file descriptor of the client's terminal](https://superuser.com/questions/358943/is-sharing-a-tmux-sockets-between-hosts-possible).

### Required information

* tmux version (`tmux -V`).
tmux 3.2a
* Platform (`uname -sp`).
Linux

ju1m:
See http://0pointer.de/blog/projects/socket-activation.html
and https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html

Fixes #3117

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -204,6 +204,11 @@ if NEED_FORKPTY
 nodist_tmux_SOURCES += compat/forkpty-@PLATFORM@.c
 endif
 
+# Add compat file for systemd.
+if HAVE_SYSTEMD
+nodist_tmux_SOURCES += compat/systemd.c
+endif
+
 # Add compat file for utf8proc.
 if HAVE_UTF8PROC
 nodist_tmux_SOURCES += compat/utf8proc.c
--- a/compat.h
+++ b/compat.h
@@ -421,6 +421,11 @@ void		*reallocarray(void *, size_t, size_t);
 void		*recallocarray(void *, size_t, size_t, size_t);
 #endif
 
+#ifdef HAVE_SYSTEMD
+/* systemd.c */
+int		 systemd_create_socket(int, char **);
+#endif
+
 #ifdef HAVE_UTF8PROC
 /* utf8proc.c */
 int		 utf8proc_wcwidth(wchar_t);
--- a/compat/systemd.c
+++ b/compat/systemd.c
@@ -1,0 +1,60 @@
+/* $OpenBSD$ */
+
+/*
+ * Copyright (c) 2022 Nicholas Marriott <nicholas.marriott@gmail.com>
+ * Copyright (c) 2022 Julien Moutinho <julm+tmux@sourcephile.fr>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF MIND, USE, DATA OR PROFITS, WHETHER
+ * IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
+ * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <sys/types.h>
+#include <sys/un.h>
+
+#include <systemd/sd-daemon.h>
+
+#include "tmux.h"
+
+int
+systemd_create_socket(int flags, char **cause)
+{
+	int			fds;
+	int			fd;
+	struct sockaddr_un	sa;
+	int			addrlen = sizeof sa;
+
+	fds = sd_listen_fds(0);
+	if (fds > 1) { /* too many file descriptors */
+		errno = E2BIG;
+		goto fail;
+	}
+
+	if (fds == 1) { /* socket-activated */
+		fd = SD_LISTEN_FDS_START;
+		if (!sd_is_socket_unix(fd, SOCK_STREAM, 1, NULL, 0)) {
+			errno = EPFNOSUPPORT;
+			goto fail;
+		}
+		if (getsockname(fd, (struct sockaddr *)&sa, &addrlen) == -1)
+			goto fail;
+		socket_path = xstrdup(sa.sun_path);
+		socket_can_be_created_again = 0;
+		return (fd);
+	}
+
+	return (server_create_socket(flags, cause));
+
+fail:
+	if (cause != NULL)
+		xasprintf(cause, "systemd socket error (%s)", strerror(errno));
+	return (-1);
+}
--- a/configure.ac
+++ b/configure.ac
@@ -390,6 +390,30 @@ if test "x$enable_utf8proc" = xyes; then
 fi
 AM_CONDITIONAL(HAVE_UTF8PROC, [test "x$enable_utf8proc" = xyes])
 
+# Check for systemd support.
+AC_ARG_ENABLE(
+	systemd,
+	AS_HELP_STRING(--enable-systemd, enable systemd integration)
+)
+if test x"$enable_systemd" = xyes; then
+	PKG_CHECK_MODULES(
+		SYSTEMD,
+		libsystemd,
+		[
+			AM_CPPFLAGS="$SYSTEMD_CFLAGS $AM_CPPFLAGS"
+			CPPFLAGS="$AM_CPPFLAGS $SAVED_CPPFLAGS"
+			LIBS="$SYSTEMD_LIBS $LIBS"
+			AC_DEFINE(HAVE_SYSTEMD)
+			found_systemd=yes
+		],
+		found_systemd=no
+	)
+	if test "x$found_systemd" = xno; then
+		AC_MSG_ERROR("systemd not found")
+	fi
+fi
+AM_CONDITIONAL(HAVE_SYSTEMD, [test "x$found_systemd" = xyes])
+
 # Check for b64_ntop. If we have b64_ntop, we assume b64_pton as well.
 AC_MSG_CHECKING(for b64_ntop)
 	AC_LINK_IFELSE([AC_LANG_PROGRAM(
--- a/server.c
+++ b/server.c
@@ -100,7 +100,7 @@ server_check_marked(void)
 }
 
 /* Create server socket. */
-static int
+int
 server_create_socket(int flags, char **cause)
 {
 	struct sockaddr_un	sa;
@@ -214,7 +214,11 @@ server_start(struct tmuxproc *client, int flags, struct event_base *base,
 
 	gettimeofday(&start_time, NULL);
 
+#ifdef HAVE_SYSTEMD
+	server_fd = systemd_create_socket(flags, &cause);
+#else
 	server_fd = server_create_socket(flags, &cause);
+#endif
 	if (server_fd != -1)
 		server_update_socket();
 	if (~flags & CLIENT_NOFORK)
@@ -424,14 +428,16 @@ server_signal(int sig)
 		server_child_signal();
 		break;
 	case SIGUSR1:
-		event_del(&server_ev_accept);
-		fd = server_create_socket(server_client_flags, NULL);
-		if (fd != -1) {
-			close(server_fd);
-			server_fd = fd;
-			server_update_socket();
+		if (socket_can_be_created_again) {
+			event_del(&server_ev_accept);
+			fd = server_create_socket(server_client_flags, NULL);
+			if (fd != -1) {
+				close(server_fd);
+				server_fd = fd;
+				server_update_socket();
+			}
+			server_add_accept(0);
 		}
-		server_add_accept(0);
 		break;
 	case SIGUSR2:
 		proc_toggle_log(server_proc);
--- a/tmux.c
+++ b/tmux.c
@@ -40,6 +40,7 @@ struct environ	*global_environ;
 
 struct timeval	 start_time;
 const char	*socket_path;
+int		 socket_can_be_created_again = 1;
 int		 ptm_fd = -1;
 const char	*shell_command;
 
--- a/tmux.h
+++ b/tmux.h
@@ -2001,6 +2001,7 @@ extern struct options	*global_s_options;
 extern struct options	*global_w_options;
 extern struct environ	*global_environ;
 extern struct timeval	 start_time;
+extern int		 socket_can_be_created_again;
 extern const char	*socket_path;
 extern const char	*shell_command;
 extern int		 ptm_fd;
@@ -2583,6 +2584,7 @@ int	 server_start(struct tmuxproc *, int, struct event_base *, int, char *);
 void	 server_update_socket(void);
 void	 server_add_accept(int);
 void printflike(1, 2) server_add_message(const char *, ...);
+int	 server_create_socket(int, char **);
 
 /* server-client.c */
 RB_PROTOTYPE(client_windows, client_window, entry, server_client_window_cmp);


ju1m:
@nicm, this draft presents a working socket activation in `tmux`, I'm not used to C nor `systemd` nor `tmux`, so it's hacky and minimalist, but please advise changes you need for it to be mergeable and I'll see what I can do.
Note that I'm not sure what to do in the case of `SIGUSR1`, I don't think it makes sense to call `server_create_socket()` again in the case of a socket activated `tmux`.

As a test example, here's a dump of my `systemd` configuration running a `tmux` server with [DynamicUser=](https://0pointer.net/blog/dynamic-users-with-systemd.html):

The activating socket:
```shell
$ systemctl cat tmux-julm.socket
# /etc/systemd/system/tmux-julm.socket
[Unit]

[Socket]
Service=tmux@julm.service
SocketMode=600
SocketUser=julm
ListenStream=/home/julm/tmux.sock
```

The activated service:
```shell
$ systemctl cat tmux@julm
# /etc/systemd/system/tmux@.service
[Unit]
Description=tmux for user %i

[Service]
Environment="HOME=/var/lib/tmux/%i"
Environment="LD_LIBRARY_PATH=/nix/store/ifdz82izmfda5kbyw33jzinq5lnbw5mn-systemd-249.7/lib:/nix/store/fk8y620g6bnxcja67avd1f6a1vgcwfpv-nss-mdns-0.10/lib"
Environment="LOCALE_ARCHIVE=/nix/store/1paysdpkhz9qr1pni022x1sgdln9j5g5-glibc-locales-2.33-62/lib/locale/locale-archive"
Environment="PATH=/nix/store/vizjhz04x6xl57x2vrpqa52j8q6rkjfh-coreutils-9.0/bin:/nix/store/y7iizrf51pgi8vjhiwndajz4ihwm416f-findutils-4.8.0/bin:/nix/store/s6sb3m2312hjw78zg5gyjpky5w0z1cgw-gnugrep-3.7/bin:/nix/st>
Environment="TZDIR=/nix/store/msar9dx7saak0fsj22sk7fzsb3zbifd5-tzdata-2021e/share/zoneinfo"

DynamicUser=true
ExecStart=/nix/store/v9lllwdq9mhp13gcn7kjw4k4qc0mkmx6-tmux-3.2a/bin/tmux -D -f /nix/store/bfbayzr77ab29n59sy46cbmyi05xmfmj-tmux.conf
NotifyAccess=all
StateDirectory=tmux/%i
Type=notify
WorkingDirectory=/var/lib/tmux/%i
```

A little `tmux.conf` replacing the `nologin` shell of the `DynamicUser=` with `bash`, and notifying `systemd` that `tmux` has started (this is not strictly necessary AFAIK, `Type=simple` would work as well)
```shell
$ cat /nix/store/bfbayzr77ab29n59sy46cbmyi05xmfmj-tmux.conf
set-option -g default-shell /nix/store/2kh3c4v2vf6d6xg6c9n8zvfpwf3zzwca-bash-5.1-p12/bin/bash
new-session -d
run-shell -b "systemd-notify --ready"
```

Then, once `tmux-julm.socket` has been started:
```
$ systemctl status tmux-julm.socket
● tmux-julm.socket
     Loaded: loaded (/etc/systemd/system/tmux-julm.socket; enabled; vendor preset: enabled)
     Active: active (listening) since Wed 2022-03-16 04:27:20 CET; 2s ago
   Triggers: ● tmux@julm.service
     Listen: /home/julm/tmux.sock (Stream)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
      Tasks: 0 (limit: 9270)
     Memory: 0B
        CPU: 1ms
     CGroup: /system.slice/tmux-julm.socket

mars 16 04:27:20 oignon systemd[1]: Starting tmux-julm.socket...
mars 16 04:27:20 oignon systemd[1]: Listening on tmux-julm.socket.
```

Accessing the socket, spawn a `tmux`, shared between clients here because `Accept=` defaults to `no`:
```shell
$ tmux -S ~/tmux.sock run id
uid=63027(tmux) gid=63027(tmux) groups=63027(tmux)
```

```shell
$ systemctl status tmux@julm
● tmux@julm.service - tmux for user julm
     Loaded: loaded (/etc/systemd/system/tmux@.service; static)
     Active: active (running) since Wed 2022-03-16 04:28:38 CET; 6s ago
TriggeredBy: ● tmux-julm.socket
   Main PID: 952784 (tmux: server)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
      Tasks: 2 (limit: 9270)
     Memory: 2.0M
        CPU: 154ms
     CGroup: /system.slice/system-tmux.slice/tmux@julm.service
             ├─952784 /nix/store/v9lllwdq9mhp13gcn7kjw4k4qc0mkmx6-tmux-3.2a/bin/tmux -D -f /nix/store/bfbayzr77ab29n59sy46cbmyi05xmfmj-tmux.conf
             └─952785 -bash

mars 16 04:28:38 oignon systemd[1]: Starting tmux for user julm...
mars 16 04:28:38 oignon systemd[1]: Started tmux for user julm.
```

Most importantly, passing the file descriptor of the client's terminal now works well:
```shell
$ tmux -S ~/tmux.sock attach
```

Of course the hardening and sandboxing can be improved beyond `DynamicUser=` later on:
```shell
$ systemd-analyze security tmux@julm | tail -1
→ Overall exposure level for tmux@julm.service: 8.4 EXPOSED 🙁
```

nicm:
Great, this code looks fine but I think you should put this in `compat/systemd.c` (declare any functions in `compat.h`) and then the change to `server.c` can become simply:

````
#ifdef HAVE_SYSTEMD
server_fd = systemd_create_socket();
#else
server_fd = server_create_socket()
#endif
~~~~

This is to keep the diff between OpenBSD and portable tmux as small as possible, otherwise there are perpetually merge problems.

Thanks!

**Comment by nicm** [comment]
If it does not make sense to support recreating the socket, best just disable SIGUSR1 with `HAVE_SYSTEMD`?

**Comment by nicm** [comment]
I think you will also need to do something about `socket_path`, or the `TMUX` environment variable will be incorrect. You could pass it into `systemd_create_socket` and update it by calling `getsockname`?

**Comment by nicm** [comment]

**Comment by ju1m** [comment]
@nicm, yes, sorry for the delay, I only have a few hours per week to work on this. I'll do my best to propose something before April.

**Comment by ju1m** [comment]
- [X] Put `systemd_create_socket` in `compat/systemd.c`
- [X] Use `sd_is_socket_unix` to check the socket type.
- [X] Use `getsockname` to set `socket_path`.

About disabling SIGUSR1, I don't know how to do it at runtime only when `tmux` has been socket activated. Here is the current observable behavior :
```shell
$ systemctl stop tmux@julm
Warning: Stopping tmux@julm.service, but it can still be activated by:
  tmux-julm.socket
tmux $ tmux -S ~/tmux.sock ls
0: 1 windows (created Fri Mar 25 01:30:31 2022)
tmux $ tmux -S ~/tmux.sock run env | grep TMUX
TMUX=/home/julm/tmux.sock,487042,0

$ systemctl status tmux@julm
● tmux@julm.service - tmux for user julm
     Loaded: loaded (/etc/systemd/system/tmux@.service; static)
     Active: active (running) since Fri 2022-03-25 01:30:31 CET; 12s ago
TriggeredBy: ● tmux-julm.socket
   Main PID: 487042 (tmux: server)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
      Tasks: 2 (limit: 9270)
     Memory: 2.0M
        CPU: 160ms
     CGroup: /system.slice/system-tmux.slice/tmux@julm.service
             ├─487042 /home/julm/src/tmux/result/bin/tmux -D -f /var/lib/tmux/julm/.config/tmux/tmux.conf -f /nix/store/bfbayzr77ab29n59sy46cbmyi05xmfmj-tmux.conf
             └─487079 -bash

mars 25 01:30:31 oignon systemd[1]: Starting tmux for user julm...
mars 25 01:30:31 oignon systemd[1]: Started tmux for user julm.

$ systemctl kill -s USR1 tmux@julm.service

$ systemctl status tmux@julm
● tmux@julm.service - tmux for user julm
     Loaded: loaded (/etc/systemd/system/tmux@.service; static)
     Active: active (running) since Fri 2022-03-25 01:30:31 CET; 22s ago
TriggeredBy: ● tmux-julm.socket
   Main PID: 487042 (tmux: server)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
      Tasks: 1 (limit: 9270)
     Memory: 1.2M
        CPU: 161ms
     CGroup: /system.slice/system-tmux.slice/tmux@julm.service
             └─487042 /home/julm/src/tmux/result/bin/tmux -D -f /var/lib/tmux/julm/.config/tmux/tmux.conf -f /nix/store/bfbayzr77ab29n59sy46cbmyi05xmfmj-tmux.conf

mars 25 01:30:31 oignon systemd[1]: Starting tmux for user julm...
mars 25 01:30:31 oignon systemd[1]: Started tmux for user julm.
mars 25 01:30:49 oignon systemd[1]: tmux@julm.service: Sent signal SIGUSR1 to main process 487042 (tmux: server) on client request.
mars 25 01:30:49 oignon systemd[1]: tmux@julm.service: Sending signal SIGUSR1 to process 487079 (bash) on client request.

$ tmux -S ~/tmux.sock ls

$ tmux -S ~/tmux.sock run env | grep TMUX
TMUX=/home/julm/tmux.sock,487042,-1

$ tmux -S ~/tmux.sock new -d
$ tmux -S ~/tmux.sock ls
1: 1 windows (created Fri Mar 25 01:31:17 2022)

$ tmux -S ~/tmux.sock run env | grep TMUX
TMUX=/home/julm/tmux.sock,487042,1


```

ju1m:
I must stress that I'm not sure at all I'm doing things correctly, especially on error code paths. For instance, using `ListenStream=127.0.0.1:4242` instead of `ListenStream=/home/julm/tmux.sock` to make `sd_is_socket_unix` fail, just hangs the `tmux` client, with no error message printed on the stderr of the `tmux` daemon.

nicm:
I'll take a look, thanks

nicm:
This looks mostly good, but using `strlcpy` to copy the path is wrong (what if the existing `socket_path` is too short?), it'll have to use `xstrdup`.

Here is an updated diff with this fixed, with the code style changed to match the rest of tmux and to make it a bit more compact, and with the configure.ac bits changed to match how the other checks work.

Please let me know if this still builds and works: [systemd.diff.txt](https://github.com/tmux/tmux/files/8348031/systemd.diff.txt)

I don't see tmux hang when `server_create_socket` returns an error but it does not correctly report the error, I have fixed this in OpenBSD so it should be in GitHub in a few hours if you want to try it again.

Can you add your copyright to `compat/systemd.c`, or tell me what to put and I will add it?

nicm:
Oh, I forgot about SIGUSR1, how about this (on top of last diff)? [x.diff.txt](https://github.com/tmux/tmux/files/8348106/x.diff.txt)

This part can go into OpenBSD tmux.

ju1m:
@nicm, thanks for the corrections!
I would have preferred to keep the auto detection of `systemd`, but deviating from other tests could indeed introduce confusion.
Anyway, please let me know if anything else is wrong/missing, otherwise I'm good with the last patch. Below are my notes when testing corner cases.

#### Testing "systemd socket error"
Even with 422fcd294a6de4ba940144ede9ebb86a9ab6e41f I'm unable to trigger/see the "systemd socket error" message anywhere (neither on `stderr` nor in `tmux-server-*.log`), best I can do is to hang the `tmux` client when activating the server with an IPv4 socket instead of a Unix socket:
```shell
$ systemctl cat tmux-julm.socket
# /etc/systemd/system/tmux-julm.socket
[Unit]

[Socket]
Service=tmux@julm.service
ListenStream=127.0.0.1:4242


$ socat -ddd UNIX-LISTEN:socat-tmux.sock,fork TCP:127.0.0.1:4242,nodelay &
2022/03/26 07:13:00 socat[21791] I socat by Gerhard Rieger and contributors - see www.dest-unreach.org
2022/03/26 07:13:00 socat[21791] I This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit. (http://www.openssl.org/)
2022/03/26 07:13:00 socat[21791] I This product includes software written by Tim Hudson (tjh@cryptsoft.com)
2022/03/26 07:13:00 socat[21791] I setting option "fork" to 1
2022/03/26 07:13:00 socat[21791] I socket(1, 1, 0) -> 5
2022/03/26 07:13:00 socat[21791] I starting accept loop
2022/03/26 07:13:00 socat[21791] N listening on AF=1 "socat-tmux.sock"

$ tmux -S socat-tmux.sock ls
2022/03/26 07:13:25 socat[21791] I accept(5, {1, AF=1 "<anon>"}, 2) -> 6
2022/03/26 07:13:25 socat[21791] N accepting connection from AF=1 "<anon>" on AF=1 "socat-tmux.sock"
2022/03/26 07:13:25 socat[21791] I permitting connection from AF=1 "<anon>"
2022/03/26 07:13:25 socat[21791] N forked off child process 21843
2022/03/26 07:13:25 socat[21791] I close(6)
2022/03/26 07:13:25 socat[21791] I still listening
2022/03/26 07:13:25 socat[21791] N listening on AF=1 "socat-tmux.sock"
2022/03/26 07:13:25 socat[21843] I just born: child process 21843
2022/03/26 07:13:25 socat[21843] I close(4)
2022/03/26 07:13:25 socat[21843] I close(3)
2022/03/26 07:13:25 socat[21843] I just born: child process 21843
2022/03/26 07:13:25 socat[21843] I close(5)
2022/03/26 07:13:25 socat[21843] I setting option "tcp-nodelay" to 1
2022/03/26 07:13:25 socat[21843] N opening connection to AF=2 127.0.0.1:4242
2022/03/26 07:13:25 socat[21843] I starting connect loop
2022/03/26 07:13:25 socat[21843] I socket(2, 1, 6) -> 5
2022/03/26 07:13:25 socat[21843] N successfully connected from local address AF=2 127.0.0.1:60036
2022/03/26 07:13:25 socat[21843] I resolved and opened all sock addresses
2022/03/26 07:13:25 socat[21843] N starting data transfer loop with FDs [6,6] and [5,5]
2022/03/26 07:13:25 socat[21843] I transferred 2413 bytes from 6 to 5
2022/03/26 07:13:25 socat[21843] I transferred 16 bytes from 6 to 5
2022/03/26 07:13:25 socat[21843] I transferred 8192 bytes from 6 to 5
2022/03/26 07:13:25 socat[21843] I transferred 243 bytes from 6 to 5
[and here the tmux client hangs]

$ systemctl status tmux@julm.service
● tmux@julm.service - tmux for user julm
     Loaded: loaded (/etc/systemd/system/tmux@.service; static)
     Active: activating (start) since Sat 2022-03-26 07:16:25 CET; 54s ago
TriggeredBy: ● tmux-julm.socket
   Main PID: 22053 (tmux: server)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
      Tasks: 1 (limit: 9270)
     Memory: 1.1M
        CPU: 229ms
     CGroup: /system.slice/system-tmux.slice/tmux@julm.service
             └─22053 /home/julm/src/tmux/result/bin/tmux -v -D -f /var/lib/tmux/julm/.config/tmux/tmux.conf -f /nix/store/nm2xqvwy6wzxwhgxzxxl2wawjc4j1bqs-tmux.conf

mars 26 07:16:25 oignon systemd[1]: Starting tmux for user julm...
```
This `tmux -S socat-tmux.sock ls` (which does not need a terminal) does work if I remove the `sd_is_socket_unix` test, so here I should indeed test the `fail` path with `errno = EPFNOSUPPORT`. Same behavior when testing for `errno = E2BIG` by using multiple `ListenStream=`.

`tmux-server-*.log` stays on:
```
1648276229.006883 server loop enter
```

Could it be linked to my using `tmux -D`, IIUC, `cause` seems to be discarded in this case in `server.c`?

#### Declaring `systemd_create_socket`
To make the socket activation work I just had to reintroduce:
```
AC_DEFINE(HAVE_SYSTEMD)
```
otherwise I get this warning:
```
tmux> compat/systemd.c:27:1: warning: no previous prototype for 'systemd_create_socket' [8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wmissing-prototypes-Wmissing-prototypes8;;]
tmux>    27 | systemd_create_socket(int flags, char **cause)
tmux>
```
and the `tmux` client hangs (even though the `tmux` server is started).

### Handling `SIGUSR1`
Regarding `SIGUSR1`, I was testing it incorrectly because as shown in my previous quotes `bash` was also receiving the signal:
```
$ systemctl kill -s USR1 tmux@julm.service
[...]
mars 25 01:30:49 oignon systemd[1]: tmux@julm.service: Sent signal SIGUSR1 to main process 487042 (tmux: server) on client request.
mars 25 01:30:49 oignon systemd[1]: tmux@julm.service: Sending signal SIGUSR1 to process 487079 (bash) on client request.
```

Here's the correct command line to send `SIGUSR1` only to the `tmux` process:
```
$ systemctl kill -s USR1 --kill-who=main tmux@julm.service
```
Using it I can't actually observe any difference with or without introducing `socket_can_be_created_again`, in both cases sending `SIGUSR1` does not kill the `tmux` server and does not prevent reconnecting to it through `systemd`'s socket.

With one session attached, here's the log excerpt, the same in both cases:
```
1648251979.299509 server_signal: User defined signal 1                                                    
1648251979.299679 cmdq_next <global>: empty                                                                                                                                                                         
1648251979.299718 cmdq_next </dev/pts/12>: empty                                                                                                                                                                    
1648251979.299748 server_client_reset_state: client /dev/pts/12 mode CURSOR,WRAP                                                                                                                                    
1648251979.299773 server_client_reset_state: cursor to 68,6                                                                                                                                                         
1648251979.299801 server_client_check_pane_buffer: pane %0 is on
```

#### Multiple `ListenStream=`
For the sake of understanding, I've tried to modify the lookup of the activating socket to pick the first configured one which is a Unix socket, based upon [netsnmp_sd_find_unix_socket](https://github.com/haad/net-snmp/blob/c629882ba31aaf27c859de2d47a6401849661ccd/snmplib/sd-daemon.c#L517-L527).

The [systemd.socket manual](https://www.freedesktop.org/software/systemd/man/systemd.socket.html) says:
> [`ListenStream=`] may be specified more than once, in which case incoming traffic on any of the sockets will trigger service activation, and all listed sockets will be passed to the service, regardless of whether there is incoming traffic on them or not.

However I don't know if we can differentiate between multiple Unix sockets to get the one used to activate the service (maybe using `getsockopt`?).
Eg. if I configure:
```
ListenStream=/home/julm/tmux.sock
ListenStream=/home/julm/tmux2.sock
```
`tmux` will always select `tmux.sock`, even when it's `tmux2.sock` which is used to activate the service. This is a really contrived use case, so I've kept the `E2BIG` errno.

#### `FD_CLOEXEC`
Also, I as wondering whether we had to close the fds, but this is already done in [sd_listen_fds](https://github.com/systemd/systemd/blob/017228934833d8618774353dcccbc5ec9f78ce98/src/libsystemd/sd-daemon/sd-daemon.c#L79-L80)

nicm:
I've added the `AC_DEFINE` back.

It sounds like we don't need to disable `SIGUSR1`. If `socket_path` is set correctly and the socket does still exist, it shouldn't cause any problems.

 I'll take a look at failure with `-D`. I don't recommend using `-D` if you can avoid it.

nicm:
I have applied this now and also made a change to fix `-D` (in OpenBSD, so wait a  bit for it to make it to GitHub).

Please let me know if you see any problems. Thanks for your work!

IIRC at least one other person has asked about this before but not been interested in doing it, so I'm sure they will be happy to see you have!

PR #3121, opened 2022-03-16
Add option to set whether to scroll on display clear

robolange:
Github issue 3090. When clearing the whole display, tmux scrolls the
lines into the history. In some cases (e.g., zsh) this may result in
many extraneous lines being added to the history, so allow the user to
disable this using the new option 'scroll-on-clear'.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -1080,6 +1080,14 @@ const struct options_table_entry options_table[] = {
 	          "remain-on-exit is enabled."
 	},
 
+	{ .name = "scroll-on-clear",
+	  .type = OPTIONS_TABLE_FLAG,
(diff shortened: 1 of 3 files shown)


nicm:
Great stuff, applied with a couple of minor changes (notably - `wp` can be `NULL` in `screen_write_clearendofscreen`, for example in a popup). Will be in GitHub in a few hours when it syncs up. Thanks!

nicm:
Oh just realised - don't you want to do the same in `screen_write_clearscreen`?

nicm:
I think it would confuse people if it didn't, so I'll make it work for ED2
also.

Issue #3085, opened 2022-02-24
`tmux -CC` control mode should not require a controlling terminal?


revintec:
### Issue description

`tmux -CC attach` requires a controlling terminal, using a pipe for `stdin` doesn't work
(a really old version worked, I can't remember which, maybe 1.8?
```
ssh xxx 'tmux -CC attach'
tcgetattr failed: Inappropriate ioctl for device
```
then it hangs

this makes control mode difficult to use, starting it from other program(like java's ProcessBuilder https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html) won't work.
but controlling mode should not require a controlling terminal, as all it's work should be done through stdio.

one can currently workaround this issue by using `ssh -t localhost` or `forkpty` then `exec`(as for java's case, JNI is needed), but this is really inconvenient and leaves useless process wasting resources, clustering `ps` output etc.

cc @gnachman

### Required information

* tmux next-3.2
* linux debian 4.4.0-33
* $TERM is screen

nicm:
Use `-C` if you don't want it to require a tty.

revintec:
@nicm thanks for the swift reply

I'm also using iTerm2, running `tmux -CC attach` will trigger iterm's tmux integration(AFAIK tmux control mode was originally added for iTerm2 specifically for this case), but `tmux -C attach` won't.

I'll try to find out why

nicm:
tmux won't send the escape sequence with `-CC`... you need to use `-CC` with iTerm2, only use `-C` when you want to talk to it over stdin.

revintec:
Hi, I try to make `-CC` work with or without a tty, basically make `ssh xxx 'tmux -CC attach'` work(I know I can use `ssh -t`).

I start testing by searching all `CLIENT_CONTROLCONTROL` occurrences and replace them if appropriate, for example
```
	/* Set up control mode. */
	if (/* isatty(0) */ 0 && client_flags & CLIENT_CONTROLCONTROL) {
		if (tcgetattr(STDIN_FILENO, &saved_tio) != 0) {
			fprintf(stderr, "tcgetattr failed: %s\n",
			    strerror(errno));
			return (1);
```
I've replaced all 6 ones(except the one at line 815, that one shouldn't matter)
<img width="1204" alt="image" src="https://user-images.githubusercontent.com/8952734/159224314-c07c175c-3d07-4d0d-828f-2d6df4544d2b.png">
but it won't work, hangs here
```
[...]

PR #3129, opened 2022-03-21
Introducing `-CCC` commandline option #3085

revintec:
https://github.com/tmux/tmux/issues/3085


nicm:
I think this should be a named flag rather than adding more `-C` (what's next, `-CCCCC`?)

revintec:
ok I'll make it a flag, it seems that I should add them in `server_client_control_flags(...)`?
and I want add another flag e.g. `-f client_cmdline=tmux_cmd_name` that would change the commandline in output like `ps` or `top` from `tmux` to `tmux_cmd_name`, making it easier to manage multiple tmux processes, do you think this feature can be merged?

nicm:
tmux already changes the command name on platforms that support it, you probably just need to add whatever mechanism macOS supports to compat/setproctitle.c.

revintec:
I mean allow user to customize the tmux commandline displayed to the value they choose, instead of the currently hard coded one

nicm:

nicm:
GitHub lost my previous reply to this too. A flag to set the proc title is fine with me, but separate PR please.

revintec:
sorry, since there is a long time without reply, I thought this is not of much interest or is the wrong way to approach this. I've since moved away from this and solved the problem in other ways. thanks for your time, I'll close this issue for now

nicm:
how would that work if tmux cannot change it on macos?

PR #3148, opened 2022-04-05
Notify control clients of swap-pane

gnachman:
`swap-pane` should notify control-mode clients of a window layout change.

A user reported that the shortcut for swapping panes didn't work in iTerm2's tmux integration here: https://gitlab.com/gnachman/iterm2/-/issues/10301#note_894044693


nicm:
Applied, thanks!

PR #3160, opened 2022-04-14
Add osc7 to iTerm2 default features

gpanders:


nicm:
Are you sure this is supported? I looked at the code and didn't see it. What does iTerm2 do with it?

gpanders:
> Are you sure this is supported? I looked at the code and didn't see it. What does iTerm2 do with it?

It's part of iTerm2's "shell integration". Each time OSC 7 is emitted iTerm2 adds the directory a list of recent directories, which you can view with the (default) keymapping `Cmd+Opt+/`.

Screenshot:

Screen Shot 2022-04-14 at 2 24 12 PM

nicm:
OK I'll add it in the next day or so, thanks.

nicm:
Applied now, thanks!

PR #3165, opened 2022-04-23
#3164 [input-keys] input_key_get_mouse: saturate coordinates at max supported value

### LINKED ISSUE #3164: tmux is killing an application in x11 mouse mode because of the coordinate overflow
The x11 mouse mode with `kmous=\E[M` has a limitation for the maximum coordinates it can encode. Not only the button (first param) but also the Cx and Cy chars will only carry up to 222 maximum offset.

This is a sad limitation but wouldn't be fatal if not a bug in tmux. If tmux receives coordinates in the SGR mode, they can exceed 222; then tmux will [blindly translate those into x+33](https://github.com/tmux/tmux/blob/bf14bf6259a03d83502e281a34d88296faaa02dc/input-keys.c#L620) and happily send the result to the application.

The overflow of coordinates will inevitably get to sending \002 or \003 in place of the Cx or Cy, which will suspend or kill the application.

leonid-s-usov:
This fixes #3164 
I was going to make a unit test for this but couldn't figure out any existing framework for that in the project.

### DIFF
--- a/input-keys.c
+++ b/input-keys.c
@@ -613,12 +613,23 @@ input_key_get_mouse(struct screen *s, struct mouse_event *m, u_int x, u_int y,
 		len += input_key_split2(x + 33, &buf[len]);
 		len += input_key_split2(y + 33, &buf[len]);
 	} else {
-		if (m->b > 223)
+		if (m->b + X10_MOUSE_BTN_OFFSET >= X10_MOUSE_PARAM_CAP) {
+			// no meaningful mitigation to this, drop the event.
 			return (0);
+		}
+
 		len = xsnprintf(buf, sizeof buf, "\033[M");
-		buf[len++] = m->b + 32;
-		buf[len++] = x + 33;
-		buf[len++] = y + 33;
+		buf[len++] = m->b + X10_MOUSE_BTN_OFFSET;
+
+		/*
+		 * Since the SGR mouse protocol support has been added to tmux, the incoming
+		 * x and y may be out of range which can be supported by the original X10 protocol.
+		 * The coordinates should be clamped to the supported range so as not to upset clients
+		 *
+		 * Please note that `tty_keys_mouse` uses an "overflow" behavior when decoding
+		 */
+		buf[len++] = X10_MOUSE_CLAMP_TO_PARAM(m->x + X10_MOUSE_POS_OFFSET);
+		buf[len++] = X10_MOUSE_CLAMP_TO_PARAM(m->y + X10_MOUSE_POS_OFFSET);
 	}
 
 	*rbuf = buf;
--- a/tmux.h
+++ b/tmux.h
@@ -581,6 +581,12 @@ enum tty_code_code {
 #define MOTION_MOUSE_MODES (MODE_MOUSE_BUTTON|MODE_MOUSE_ALL)
 #define CURSOR_MODES (MODE_CURSOR|MODE_CURSOR_BLINKING|MODE_CURSOR_VERY_VISIBLE)
 
+/* X10 Mouse protocol support */
+#define X10_MOUSE_PARAM_CAP 0x100
+#define X10_MOUSE_BTN_OFFSET 0x20
+#define X10_MOUSE_POS_OFFSET 0x21
+#define X10_MOUSE_CLAMP_TO_PARAM(p) ((((uint)p) < (uint)X10_MOUSE_PARAM_CAP) ? (p) : (X10_MOUSE_PARAM_CAP-1))
+
 /* A single UTF-8 character. */
 typedef u_int utf8_char;
 
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -1061,17 +1061,17 @@ tty_keys_mouse(struct tty *tty, const char *buf, size_t len, size_t *size,
 		log_debug("%s: mouse input: %.*s", c->name, (int)*size, buf);
 
 		/* Check and return the mouse input. */
-		if (b < 32)
+		if (b < X10_MOUSE_BTN_OFFSET)
 			return (-1);
-		b -= 32;
-		if (x >= 33)
-			x -= 33;
-		else
-			x = 256 - x;
-		if (y >= 33)
-			y -= 33;
-		else
-			y = 256 - y;
+
+		b = b - X10_MOUSE_BTN_OFFSET;
+
+		/*
+		 * This is deliberately different from our behavior in `input_key_get_mouse`
+		 * The overflow behavior is kept for backward compatibility with 2756d12
+		 */
+		x = (x + X10_MOUSE_PARAM_CAP - X10_MOUSE_POS_OFFSET) % X10_MOUSE_PARAM_CAP;
+		x = (y + X10_MOUSE_PARAM_CAP - X10_MOUSE_POS_OFFSET) % X10_MOUSE_PARAM_CAP;
 	} else if (buf[2] == '<') {
 		/* Read the three inputs. */
 		*size = 3;


nicm:
Where is `min` coming from here? tmux does not define it (I'm not suggesting it should).

leonid-s-usov:
Good question, I haven't really thought of this. I must admit that I haven't attempted the build just yet.

leonid-s-usov:
Indeed, the compilation fails with

```
Undefined symbols for architecture x86_64:
  "_min", referenced from:
      _input_key_get_mouse in input-keys.o
```

I'm fixing it

nicm:
I see what you're going for here but this seems a bit overkill to me, I don't see any need for defining constants or temporaries, or such an extensive comment - anyone reading this code will need to understand the protocol already for it to make any sense, so a comment can just be a reminder. How about this instead: [x.diff.txt](https://github.com/tmux/tmux/files/8552965/x.diff.txt)

~~~~
--- a/input-keys.c
+++ b/input-keys.c
@@ -613,12 +613,18 @@ input_key_get_mouse(struct screen *s, struct mouse_event *m, u_int x, u_int y,
                len += input_key_split2(x + 33, &buf[len]);
                len += input_key_split2(y + 33, &buf[len]);
        } else {
-               if (m->b > 223)
+               /*
+                * "Normal" mouse mode uses a single 8-bit ASCII character for
+                * each of buttons and position; discard if buttons is too big
+                * (there is no meaningful way to fix this) and clamp the
+                * position.
+                */
+               if (m->b + 32 > 255)
                        return (0);
                len = xsnprintf(buf, sizeof buf, "\033[M");
                buf[len++] = m->b + 32;
-               buf[len++] = x + 33;
-               buf[len++] = y + 33;
+               buf[len++] = x + 33 > 255 ? 255 : x + 33;
+               buf[len++] = y + 33 > 255 ? 255 : y + 33;
        }

        *rbuf = buf;
~~~~

I kind of feel like `buf` should be `u_char` but I suppose it doesn't really matter. Separate change in any case.

leonid-s-usov:
I'm not really keen on doing the math twice, so would you at least let me define the temporary for x and y? and 255, I'd use (unit)UCHAR_MAX instead. I believe this make the code self-explanatory and may save us the comment.

leonid-s-usov:
I hope we can agree that defining consts and temporaries doesn't bloat the compiled result, and sometimes can actually result in an optimized version.

nicm:
I don't think it is necessary to have a temporary. I think we can rely on the compiler to optimize away the duplication and even if it doesn't, this is not code that will be run very often since mouse clicks are triggered by people. I think UCHAR_MAX is worse, by definition this protocol is dealing with ASCII and the upper limit will be 255 even if we ran on a platform where char was not 1 byte (which will never happen anyway).

leonid-s-usov:
If no temporary is needed, may I at least define a MAX macro to avoid the duplication in code? 😅

nicm:
Looks like with `-O1` or above, clang will optimize away the duplicate addition:
~~~~
       int b = a + 33 > 255 ? 255 : a + 33;
    19db:       83 c0 21                add    $0x21,%eax
    19de:       3d ff 00 00 00          cmp    $0xff,%eax
    19e3:       be ff 00 00 00          mov    $0xff,%esi
    19e8:       0f 4c f0                cmovl  %eax,%esi
~~~~
I'm not sure why the `mov` wouldn't also be conditional but honestly the time when I was willing to second guess compilers is 25 years ago now...

nicm:
I don't mind macros but it would need to go in `tmux.h` or at the top of the file. There are already a bunch of mouse macros (eg `MOUSE_MASK_BUTTONS`) in `tmux.h`, so best put it there maybe?

`tty_keys_mouse` parses these sequences so maybe the same macros could be used there too (it takes off 32/33 etc, look at `tty-keys.c` ll 1047 to 1075)?

leonid-s-usov:
OK looking at the code in `tty-key.c` I realize where the issue is coming from: 2756d12750
So.. do we want to do something about this code?

```
		if (x >= 33)
			x -= 33;
		else
			x = 256 - x;
		if (y >= 33)
			y -= 33;
		else
			y = 256 - y;
```

I mean.. are there features depending on this? Or is it only a tmux-in-tmux kind of story?

nicm:
I think someone asked for it, so I guess at least one terminal does do it,
but I don't remember which one.

leonid-s-usov:
But does it mean then our change in the input-keys will also break someone's expectations?

nicm:
No offence, but it is not an improvement to hide a relatively simple calculation behind 12 (!) macros, including some that are nested. Everyone will understand that `x > max ? max : x` is clamping a value, and understand that `max` is the limit, even if it is just a plain number and not a define.

Also I don't think we should call it X10 mouse mode, it is the same format as X10 mouse mode but this is what xterm calls "normal" mode (SM ?1000), not actual X10 mode (SM ?9) which tmux does not support.

I suggest doing one of the two things your macros are doing (see below), but not both. I don't really mind which.

----

So either define the limits:

~~~~
/*
 * Mouse parameter limits for normal mode (SM ?1000).
 */
#define MOUSE_PARAM_MAX 0xff
#define MOUSE_PARAM_POSITION_OFFSET 0x21
#define MOUSE_PARAM_BUTTON_OFFSET 0x20
~~~~
Then the code can become maybe something like:
~~~~
        b = m->b + MOUSE_PARAM_BUTTON_OFFSET;
        if (b > MOUSE_PARAM_MAX)
                return (0);
        len = xsnprintf(buf, sizeof buf, "\033[M");
        buf[len++] = b;

        x += MOUSE_PARAM_POSITION_OFFSET;
        buf[len++] = x > MOUSE_PARAM_MAX ? MOUSE_PARAM_MAX : x;

        ...
~~~~

----

Or the other option is to define functions:
~~~~
#define MOUSE_BUTTONS_TO_PARAM(b) ((b) > 255 - 32) ? -1 : (b) + 32)
#define MOUSE_BUTTONS_FROM_PARAM(b) ((b) < 32 ? -1 : (b) - 32)
#define MOUSE_POSITION_TO_PARAM(p) ((p) > 255 - 33) ? 255 : (p) + 33)
#define MOUSE_POSITION_FROM_PARAM(p) ((p) < 33 ? -1 : (p) - 33)
~~~~
And the `input-keys.c` code becomes:
~~~~
b = MOUSE_BUTTONS_TO_PARAM(m->b); // (note b will need to be int not u_int)
if (b == -1)
    return (0);
len = xsnprintf(buf, sizeof buf, "\033[M");
buf[len++] = b;
buf[len++] = MOUSE_POSITION_TO_PARAM(x);
buf[len++] = MOUSE_POSITION_TO_PARAM(y);
~~~~
`tty-keys.c` can use the "from" macros similarly.

leonid-s-usov:
No offense, you're the boss here.
I think I prefer the option which is closer to my original submission, that is, defining named constants and then using them in the code. I will use a temp var for the encoding bit though, hopefully, that's going to be acceptable.

Regarding the naming, following the [documentation](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking), the manipulations are required by the original X10 definition: the 32 offset for every parameter AND (1,1) coordinates of the upper left corner, requiring +1 for the position.

> Normally, parameters (such as pointer position and button number) for
all mouse tracking escape sequences generated by xterm encode numeric
parameters in a single character as value+32.  For example, !  specifies
the value 1.  The upper left character position on the terminal is
denoted as 1,1.  This scheme dates back to X10, though the normal mouse-
tracking (from X11) is more elaborate.

leonid-s-usov:
I wonder if anyone has ever tested https://github.com/tmux/tmux/commit/2756d127507cb939582adf2bbaf7be4cf5b711cd. What would prevent tmux from interpreting the `\003` char as ctrl-c when receiving an overflow pos in the X10 message format, just like my program did when receiving this kind of mouse event?

nicm:
tmux does not process control characters inside key sequences.

nicm:
Is your proram keeping the tty in cooked mode so the kernel tty driver steals the `C-c`?

leonid-s-usov:
> oked mode so the kernel tty driver steals the `C-c`?

Most probably that is the case!

nicm:
It is not a good idea for programs to do this (what if I set my kill or intr character to `\033`?), so you should probably report this to the application developers.

leonid-s-usov:
Well, the application is kind of in-house and I can edit the code which I do - in this case, was hacking the mouse support.
But I was sure that ncurses sets the tty to raw mode at the start, isn't that the case?

nicm:
That depends if you call raw(3), but I'd say that is recommended if you are handling mouse input, look at the BUGS section of getmouse(3).

I don't mind changing tmux not to do this anyway.

nicm:
OK I have applied this with some minor changes:

- I don't like calling the defines after the X10 protocol, these are for what xterm calls the "normal" protocol. tmux does not support the X10 protocol, so I renamed them to remove `X10_`.

- I don't see a need for `X10_MOUSE_CLAMP_TO_PARAM` when it is used in just one place, I think it is simpler and more obvious just to do it inline.

- I think a "max" value of 0xff is more useful than a "cap" of 0x100, so I changed the define to `_MAX` instead.

- It seems to me that the previous logic in `input-keys.c` was wrong - a 0 should have been 223 but was ending up as 256. So it seems likely nobody ever used this and it is probably better just to remove it.

- The UTF-8 mouse output could also benefit from using the same constants, so I have added `MOUSE_PARAM_UTF8_MAX` and changed that code to use the constants as well.

Please take a look at and test c6b51cea and let me know if there are any problems.

Thanks!

leonid-s-usov:
OK happy to contribute even if it's been a bumpy road and thanks for taking the time to review.

Just mentioning the X10 prefix think I have rationalized that in https://github.com/tmux/tmux/pull/3165#issuecomment-1109636604 which I don't think you've explicitly commented on, but again, won't fight over it.

I'll give that commit a test and let you know. I'm also happy you've removed the asymmetry between the input and output handling of the mouse events, and yes, given that overflow code is not needed I agree that MAX is more useful than the CAP. The only reason for the CAP was to support that optimized overflow calaculation.

nicm:
Thanks. Yes I saw that, I know that it is the same as the X10 protocol but the fact is it is not the X10 protocol - we don't support that, so I don't want to give the impression that's what these are for.

nicm:
maybe but now we have sgr mouse mode and terminals should support that
rather than relying on this behaviour

PR #3196, opened 2022-05-26
Update tmux.1

sergei-dyshel:
Fix `message-time` => `display-time` in the help of `display-message` command.


nicm:
Fixed, thanks!

PR #3209, opened 2022-06-02
Use only the definition of getpeerucred if available

ericvw:
f97d784f172a10e0444ad93bb536ee0428613aa7 did not compile because of the
redeclaration of `getpeereid()` when `HAVE_GETPEERUCRED` is defined.
f97d784f172a10e0444ad93bb536ee0428613aa7 can now be considered tested on
Solaris :).


nicm:
Applied, thanks!

Issue #3205, opened 2022-06-02
Issues with buffer handling (set-buffer and delete-buffer)


systemmonkey42:
I've split this from https://github.com/tmux/tmux/issues/3203 because it applies to master, and possibly goes back a while. 

Hey, I've had a problem with buffers. It appears that

delete-buffer -b name
will delete buffer "name" if it exists, otherwise it will delete another seemingly random buffer. (maybe the next in the list)

Is this correct behaviour?

Update: Looking at the source... cmd_set_buffer_exec()

      bufname = args_get(args, 'b');
      if (bufname == NULL)
          pb = NULL;
      else
          pb = paste_get_name(bufname);
The 'else' should probably check if "paste_get_name()" returns NULL and abort the buffer deletion..

Update2: This patch seems to do the right thing for the delete case,
```
    if (cmd_get_entry(self) == &cmd_delete_buffer_entry) {
+       if( pb == NULL && bufname != NULL )
+           return (CMD_RETURN_ERROR);
        if (pb == NULL)
            pb = paste_get_top(&bufname);
        if (pb == NULL) {
```
Though as mentioned by @krader1961,
```delete-buffer```
without parameters will incorrectly grab a named buffer, and delete it, and, as I've since found; 
```set-buffer -n newname```
will rename the most recent buffer, even if it isn't automatic, and finally
```set-buffer -b oldname -n newname```
when `oldname` does not exists, will rename the most recent buffer (even if its already a named buffer).

Thanks to @krader1961 for pointing out this issue is broader than first expected.

**Comment by systemmonkey42** [comment]
In addition to the above, can I ask for the following,  when selecting an automatic buffer, _always_ skip named buffers.

Eg, 
```
tmux show-buffer
```
Shows the most recent buffer, even if it is a named buffer.  It makes sense to me that named buffers should only be shown
when asked for by name

I'm sure thats how it used to work, but with 3.3/master there seems to be a strange interaction. Scripts (and hooks) I've used for ages no longer behave.

Thanks

systemmonkey42:
Ok, This is fascinating...
It appears the commit which introduced the feature which deleted a recent buffer if the specified one wasn't found was

62bb6e37e0bdc2de2a20f7adc38993d2ccfd8c11

which is 7 years old.  somewhere between release 2.0 and 2.1...
Which means as far am I'm concerned, it *always* worked this way.

I've wasted too much time on this, so I'll likely just find another way to do what I need to do.

krader1961:
[...]

PR #3212, opened 2022-06-04
Updated paste_buffer automatic buffer handling.

systemmonkey42:
Suggested updates based on #3205

- Implement paste_is_empty() boolean
- paste_get_top() always returns an automatic buffer.


systemmonkey42:
I've created a pull request based on you suggestions, but I have a concern about how buffers
are deleted when the maximum number of buffers is reached.

I will have to find and review the code, but I suspect it will discard named buffers in that scenario. I'm not sure if it should
consider consider named buffers to be part of the buffer liimit count, or only consider automatic buffers to be part of that.

If it only considers automatic buffers, it would be safe to have a much smaller limit, since its rare to look back beyond the first two or three automatic buffers.

nicm:
Applied now, thanks!

PR #3223, opened 2022-06-14
fixed some spelling mistakes

venthur:
that's all :)


nicm:
I've applied this except the regress changes, thanks!

PR #3240, opened 2022-06-26
support osc-8 hyperlinks

tizee:
# Background
Recently I was using using CLI apps based on [rich](https://github.com/Textualize/rich). Its support for hyperlinks are based on the OSC 8. But this feature is missing in the latest version of tmux.

- OSC 8 spec: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

# Steps to reproduce this bug

1. use tmux without conf
```
tmux -Ltest kill-sever
tmux -Ltest -vv new
```
2. In tmux:
```
printf '\e]8;id=1235;https://github.com\e\\This is a link\e]8;;\e\\\n'
```

# Feature Design

The terminal emulator cannot recognize this because tmux stripped out its unrecognized OSC escape sequences.

Usage:
```
set -as terminal-overrides ',*:Hls=\E]8;id=%p1%s;%p2%s\E\\:Hlr=\E]8;;\E\\'
```

It's a long missing feature that has been implemented by @nicm before but not yet merged into the master branch. 
> so I might leave it for another while. -- https://github.com/tmux/tmux/issues/911#issuecomment-427944765

Related issue [#911](https://github.com/tmux/tmux/issues/911)

So I pick up the outdated patch and managed to update it against the latest commit.  

## Reuse the id

To make this feature complete, I added a simple logic for reusing the hyperlink with the same id, which is unimplemented in original patch. 

If there is a existing entry with the same id, then reuse the existing entry. Otherwise creating a new entry and insert it into the tree and increment the id by 1.

# Code format

Th code style of this PR is partly inconsistent with the codebase's. AFAIK, there is not a formatter related config like `.clang-format`.  I wonder whether it's okay because I've found some other code mixing tabs and spaces in the codebase.


input.c
```diff
@@ -2562,6 +2566,55 @@ input_osc_4(struct input_ctx *ictx, const char *p)
...
+  log_debug("%s: OSC 8 id=%d", __func__, id);
+  hl = server_get_hyperlink(id);
+  if(hl==NULL) {
+    hl = xcalloc(1, sizeof *hl);
+    if(id!=0){
+      hl->id = id;
+    }else{
+      hl->id = ++next_hyperlink;
+    }
+    hl->link = xstrdup(end);
```
tizee, line 2609: Is it a good idea to update the link of existing entry to the latest one with the same id?

input.c
```diff
@@ -2562,6 +2566,55 @@ input_osc_4(struct input_ctx *ictx, const char *p)
...
+  }
+  /* parse ID and check pane tree, reuse entry if present */
+  id_=strstr(p,"id=");
+  id=0;
+  if(id_!=NULL&&id_<end){
+    id_+=3;
+    while(*id_>='0' && *id_<='9'){
+      id = id * 10 + *id_++ - '0';
+    }
+  }
```
tizee, line 2592: It's uggly for the hard-coded `id=` and it is error-prone for input like `id = `. Any suggestions are welcome on handling this kind of OSC 8 parameters.

input.c
tizee, line 2609: Should we update the link of existing entry with the same id?

nicm:
Does this also include the changes and comments from https://github.com/tmux/tmux/pull/2890?

tizee:
> Does this also include the changes and comments from #2890?

No, I reused the existing entry with the same id in `input_osc_8`.

The code design from #2890 is more robust. If needed, I'd like to pick up the work on #2890 to consummate this long missing feature.

nicm:
That is probably best but please try to conform to tmux's code style.

nicm:
Thanks!

nicm:
https://man.openbsd.org/OpenBSD-current/man9/style.9

nicm:
There are a few differences, eg we line up types by name not by "first word". But if you get 90% of the way there I will fix the remainder before commit.

tizee:
Thanks.

tizee:
@nicm 

I've fixed some minor issues mentioned on #2890 and added on usage relating to hyperlink in `tmux.1`. 

Here is my test script along with a minimal config. Some test cases for bad OSC 8 strings that ends without a `ST` have been ignored deliberately for the sake of simplicity (it is related to `input_osc_exit`). Though the server would exit when the osc 8 string didn't end with a `ST`, it is not caused by the hyperlink related functions. 

1. `tmux.conf`
```
set -sa terminal-overrides ',*:Hla=\033]8;;%p1%s\033\\:Hlr=\033]8;;\033\\'
set -sa terminal-overrides ',*:Hls=\033]8;id=%p1%s;%p2%s\033\\:Hlr=\033]8;;\033\\'
```
2. test script 
```
printf '\e]8;;https://github.com\e\\test0\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\test1\e]8;;\e\\\n'
printf '\e]8;id=1;https://github.com\e\\test2\e]8;;\e\\\n'
printf '\e]8;;https://github.com/tmux/tmux\e\\test3\e]8;;\e\\\n'
printf '\e]8;id=abcd;https://github.com\e\\test4\e]8;;\e\\\n'
printf '\e]8;id=;https://github.com\e\\test5\e]8;;\e\\\n'
printf '\e]8;;\e\\1\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\2\e\\\n'
```

It works as a charm to me. 

The `TODO` in `tty.c` is unsolved because its use case is not so clear to me.

nicm:
Only had time for a quick look, but I don't think we need three terminfo capabilities, we could get away with just one, something like:
~~~~
Hls=\\E]8;%?%p1%l%tid=%;%p1%s;%p2%s\\E\\
~~~~
Then p1 could be the param and p2 the URI. Both empty to clear (instead of `Hlr`), p1 empty to have no ID (like `Hla`).

nicm:
`Hls` now needs two parameters so you must always send it with `tty_putcode2`, whichever of the two parameters you are not using should be empty (`""`).

nicm:
Some other small comments:

- In `tty_hyperlink`, it is correct to do nothing if `hl == NULL` but if `gc->link == 0`, you need to turn off the hyperlink, so do `tty_putcode2(tty, TTYC_HLS, "", "")`.

- In `hyperlink.c`, you need `stdlib.h` for `free`, also please always include `sys/types.h` first (see the other files).

- In `input_osc_8`, the `log_debug` lines are too long (>80 columns). This would be fixed and I think the code woul be nicer if you `break` from the loop when `*end == ';'`and do both found/not found afterwards by checking if `end == NULL`.

- In `tmux.1`, we also need an entry in the `tty-features` table for the `hyperlinks` feature.

- In `tty-features.c.`, please rename `tty_feature_hyperlink` to match the feature name, so `tty_feature_hyperlinks`. Also the indentation is off in `tty_feature_hyperlink_capabilities`.

Need to look at `hyperlink.c` a bit more.

tizee:
I've tried to use `Hls=\\E]8;%?%p1%l%tid=%;%p1%s;%p2%s\\E\\` in both of terminfo feature and `terminal-overrides` but only found out that there is a bug on handling terminfo conditional parameters.  So I have to remove the conditional parameters in terminfo string.

The thing bothered me for a while is that following terminfo parameter string doesn't work when using the hyperlinks feature via `set -sa terminal-features "hyperlinks`.

```c
static const char *tty_feature_hyperlink_capabilities[] = {
	"Hls=\\E]8;id=%p1%s;%p2%s;\\E\\\\",
	NULL
};
```

However, by overriding it in `tmux.conf`, the OSC 8 hyperlinks could being printed out successfully instead of stripping out their control sequences. 

```
set -sa terminal-overrides ',*:Hls=\033]8;id=%p1%s;%p2%s\033\\'
```

Either of them could handle the correct OSC 8 strings via `input_exit_osc` and `input_osc_8` according to the server log.

Since there is a inconsistent behavior between the feature and the `terminal-overrides`, I'm trying to resolve this by some related functions:
- `tty_term_apply`
- `tty_term_apply_overrides`
- `tty_term_create`

nicm:
We need the conditional. What is the bug?

tizee:
> We need the conditional. What is the bug?

- terminfo string:
```c
static const char *tty_feature_hyperlink_capabilities[] = {
	"*:Hls=\\E]8;%?%p1%l%tid=%p1%s%;;%p2%s\\E\\\\",
	NULL
};
```

- Input: 
```
printf '\e]8;id=1;https://github.com\e\\test2\e]8;;\e\\\n'
```

- Log:
```
....
input_exit_osc: "8;id=1;https://github.com" (end ST)
....
/dev/ttys000: \033]8;3;https://github.com\033\\
/dev/ttys000: test2
...
```

The conditional block `%?%p1%l%tid=%p1%s%;` became the condition `%p1%l` instead of resulting `id=%p1%s`, which makes the whole OSC 8 string a invalid format.

nicm:
Are you on OpenBSD? You need lib/libcurses/tinfo/lib_tparm.c r1.10: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libcurses/tinfo/lib_tparm.c.diff?r1=1.9&r2=1.10

tizee:
According to the code in `input_osc_8`, the `\033]8;id=;\033\\` is the same as `\033]8;;\033\\`, which marks the end of a hyperlink.

```c
		if (end - start >= (ssize_t)(sizeof "id=") && /* id mustn't be empty */
		    strncmp(start, "id=", (sizeof "id=") - 1) == 0) {
			start += (sizeof "id=") - 1;
			param_id = hyperlink_write_namespaced(hl, param_id,
			    start, end - start);
		}

		/* The first ; is the end of parameters and start of the URI. */
		if (*end == ';') {
			uri = end + 1;

			if (*uri == '\0') {
				gc->link = 0;
				log_debug("close hyperlink");
				return;
			}
...
```




> 

No, I am using mac. The `ncurses` version is `6.3`.

tizee:
`ncurses` implementation for `lib_tparm.c`
- `tparm_internal`:
```
	    case 'l':
		s = spop(tps);
		npush(tps, (int) strlen(s));
		break;
```

nicm:
No, ncurses 6.3 works. This was fixed in ncurses 5.9. You must be building tmux against the system ncurses which is 5.7. Try building tmux against ncurses 6.3 instead.

I don't mind if this feature does not work with the system ncurses on macOS.

tizee:
> No, ncurses 6.3 works. This was fixed in ncurses 5.9. You must be building tmux against the system ncurses which is 5.7. Try building tmux against ncurses 6.3 instead.
> 
> I don't mind if this feature does not work with the system ncurses on macOS.

Now it works. Thanks.

tizee:
Updated script for unit testing. The previous one misses some closing sequences.

```sh
#!/usr/bin/env bash

# valid osc8
printf '\e]8;;https://github.com\e\\test0\e]8;;\e\\\n'
printf '\e]8;id=1;https://github.com\e\\test2\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\test1\e]8;;\e\\\n'
printf '\e]8;;https://github.com/tmux/tmux\e\\test3\e]8;;\e\\\n'
printf '\e]8;id=abcd;https://github.com\e\\test4\e]8;;\e\\\n'
printf '\e]8;id=;https://github.com\e\\test5\e]8;;\e\\\n'
printf '\e]8;;\e\\1\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\2\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\测试\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\試験\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\экспертиза\e]8;;\e\\\n'
printf '\e]8;;https://github.com\e\\εξέταση\e]8;;\e\\\n'
```

nicm:
Here is a diff with some changes on top of yours: [x.diff.txt](https://github.com/tmux/tmux/files/9000862/x.diff.txt)

- `display-panes` does not need to use the hyperlinks since there can't be any in the `grid_cell` it sets up.

- Rename `hyperlink_write_namespaced` to `hyperlink_add_namespace`.

- I do not see any need to use `const struct hyperlinks`, so don't bother. We do `const grid_cell` to avoid errors where we think we are changing the grid but are actually changing an extracted cell.

- In `tty_hyperlink`, don't do anything if no hyperlinks at all.

- Tweak the text in `tmux.1` and remove the example - people should use `terminal-features` not `terminal-overrides`.

- Rename `param_id` just to `id` outside `hyperlink.c` (for now).

These are a few other changes I think we should make but I don't have time to do now:

- I don't see a need for `hyperlink_add_namespace` to be in `hyperlink.c`, I think it should be inlined or a static helper function in `input.c`. Since it builds a buffer on the stack in `input_osc_8` it would be better closer to that function.

- The variable names for everything in `hyperlink.c` are a little verbose which makes the code a little hard to read. I don't think it is necessary to have both `param_id` and `attr_id`. I think we should rename `param_id` just to `id` and call `attr_id` something else, maybe just `link`.

tizee:
Now I'm tidying up the `hyperlink.c` to improve its readability as well as moving `hyperlink_add_namespace` to `input.c`.

nicm:
Great stuff.

tizee:
Changes:
1. Rename `param_id` to `id`
2. Rename `attr_id` to `inner` instead of `link`, which can be confusing when seen with `hyperlink`
3. Move definition of `struct hyperlinks` to `tmux.h`
4. Use pointer types for fields `forward_mapping` and `backward_mapping` in `struct hyperlinks`

nicm:
I would prefer not to move `struct hyperlinks` entirely into `tmux.h`. It would be better to add a `hyperlinks_get_namespace` accessor function.

I don't see any reason to limit the namespace to exactly three digits.

`hyperlink_add_namespace` is actually incorrect because `utf8_strvis` may increase the length of the string (the reserved size needs to be 4 * input string size). But I would be inclined to play it safe and use `xasprintf` and `utf8_stravis`, something like:

~~~~
static char *
input_make_hyperlink(struct hyperlinks *hl, const char *id, size_t idlen)
{
    char *tmp, *s;

    if (idlen > INT_MAX)
        return NULL;
    xasprintf(&tmp, "%x_%.*s", hyperlink_get_namespace(hl), (int)idlen, id);
    utf8_stravis(&s, tmp, VIS_OCTAL|VIS_CSTYLE);
    free(tmp);
    return (s);
}
~~~~

Then in `input_osc_8`:

~~~~
for (start = p; (end = strpbrk(start, ":;")) != NULL; start = end + 1) {
    if (end - start >= 4 && strncmp(start, "id=", 3) == 0) {
        start += 3;
        if (id != NULL) /* multiple id= */
            goto bad;
        id = input_make_hyperlink(hl, start, end - start);
        if (id == NULL)
            goto bad;
...
bad:
    log_debug("bad OSC 8 %s", p);
    free(id);
}
~~~~

Hope this makes sense!

nicm:
Actually, the `INT_MAX` check is unnecessary paranoia since we already limit the size of the input buffer, so you could do without that and make `input_make_hyperlink` unable to fail.

nicm:
I did wonder if we could use anything we already have as the namespace and avoid having a separate number altogether - but the problem is that although we could use the pane ID for panes, popups don't have a pane ID...

nicm:
Other than that this looks good apart from minor style bits.

tizee:
> I did wonder if we could use anything we already have as the namespace and avoid having a separate number altogether - but the problem is that although we could use the pane ID for panes, popups don't have a pane ID...

The hyperlink's namespace(its inner id) is only related to `screen`.  The popup windows initialize `input_ctx`using a null `wp`.

```c
	pd->ictx = input_init(NULL, job_get_event(pd->job), &pd->palette);
```

But we can always get the screen of the input context with `ictx->ctx.s`, from which we can get the hyperlinks. 

```c
	struct hyperlinks	*hl = ictx->ctx.s->hyperlinks;
```

- `input.c`:

```
input.c
780:    struct screen           *s = sctx->s;
1206:   struct screen           *s = sctx->s;
1264:   struct screen                   *s = sctx->s;
1340:   struct screen                  *s = sctx->s;
1800:                   if (sctx->s->mode & MODE_FOCUSON)
1840:   struct screen           *s = sctx->s;
1881:                           screen_push_title(sctx->s);
1892:                           screen_pop_title(sctx->s);
2304:           if (screen_set_title(sctx->s, p) && wp != NULL) {
2315:                   screen_set_path(sctx->s, p);
2377:   if (screen_set_title(sctx->s, ictx->input_buf) && wp != NULL) {
```

nicm:
Right, that works, but I meant to not have the hyperlink namespace and just use an ID we already have. But don't worry it is OK as it is.

nicm:
OK I have had another look at the spec and at hyperlink.c and I think we (I) have been overthinking this.

The spec says:

> The same id is only used for connecting character cells whose URIs is also the same. Character cells pointing to different URIs should never be underlined together when hovering over.

So we don't *want* to use the same ID for the same URI unless both of the URI and internal ID are the same - the hyperlink is keyed by *both* ID and URI.

So I don't think we need all this stuff with three trees, we only need two:

- (URI, internal ID) -> link number;

- link number -> (URI, external ID).

So I think we could just have one struct for each hyperlink that goes into two trees, something like:

~~~~
struct hyperlink_uri {
    u_int           inner;
    const char*     internal_id;
    const char*     external_id;
    const char*     uri;

    RB_ENTRY (hyperlink_uri) by_inner_entry;
    RB_ENTRY (hyperlink_uri) by_uri_entry; /* by internal ID and URI */
};
~~~~

I don't think we need to make the external ID have any relationship to the internal ID, so instead of prefixing the namespace, we can just have a global `u_int next_hyperlink_id` and generate the external ID as maybe something like `xasprintf(&external_id, "tmux%X", ++next_hyperlink_id);`.

This means URIs would be duplicated, but does that really matter? How many URIs are applications typically using? Should we instead have a mechanism to garbage collect old ones?

Does this make sense or am I missing something critical? You don't need to make any code changes right now.

tizee:
In the original design, the GC is performed when the `screen` is released and the prefix namespace sets to 1 . Once involved a global `next_hyperlink_id`, we can not reset it to 1 and it's okay until it exceeds the maximum.

nicm:
Can you please take a look at this on top of this PR and see if it 1) works and 2) looks sensible wrt the spec/how it should work? [hl2.diff.txt](https://github.com/tmux/tmux/files/9009162/hl2.diff.txt)

nicm:
This is the same but corrects the comment at the start of the file: [hl3.diff.txt](https://github.com/tmux/tmux/files/9009189/hl3.diff.txt)

nicm:
The only thing with this is that tmux will generate this as three separate URIs:
~~~~
printf '\e]8;;https://github.com\e\\1\e]8;;\e\\\e]8;;https://github.com\e\\2\e]8;;\e\\\e]8;;https://github.com\e\\3\e]8;;\e\\\n'
~~~~
But iTerm2 without tmux will join them together as one.

By the sound of the spec, VTE will keep them as three (consistency is great) so I guess tmux doing so is harmless...

nicm:
Probably want this as well so tmux-in-tmux works:
~~~~
@@ -444,14 +460,14 @@ tty_default_features(int *feat, const char *name, u_int version)
                },
                { .name = "tmux",
                  .features = TTY_FEATURES_BASE_MODERN_XTERM
-                             ",ccolour,cstyle,focus,overline,usstyle"
+                             ",ccolour,cstyle,focus,overline,usstyle,hyperlinks"
                },

nicm:
OK my only concern now is that there is no mechanism to remove hyperlinks even when they scroll away from the history.

The easiest options here would be to do one of:

1. Don't worry about it, they will be freed when the pane closes. I'm a bit worried about this because with very long running panes and intensive or stupid use of hyperlinks the memory could grow a lot.

2. Have a fixed maximum of allowed hyperlinks per pane (200 or so?) and just delete the oldest when it is hit.

3. Have a *global* limit of allowed hyperlinks and delete the oldest when it is hit. This would have the advantage of a memory ceiling no matter how many panes are open.

Either of 2 or 3 would be easy to do and probably fine.

It would be possible to track whether a line contained a hyperlink, reference count each hyperlink and remove it when the line was scrolled away, but I think it would probably be too complicated to be worth it.

Another option might be to timestamp the hyperlinks each time they are used and to throw away any that hadn't been used for a day or so. But I'm not sure that is better than just a limit.

tizee:
- A simple script for demonstration:

```sh
# 1<<32
for (( i = 0; i < 4294967296; i++ )); do
  printf '\e]8;;https://github.com\e\\test\e]8;;\e\\\n'
done
```

As a result, it is really picked up in memory usage quite dramatically.

2022-06-test-tmux

A LRU cache is overkill for this. Set a global limit to it sounds promising.

nicm:
Yes I think a global limit will be fine, I think if we said a max of 5000 or so hyperlinks and threw away the oldest nobody would really notice. It would also be nice to add a `-H` flag to `clear-history` to destroy all hyperlinks in a pane so users have the option of freeing them manually if they need to.

nicm:
Yes I know they are not passed on, but if tmux collected them we can do other things with them, and if we are collecting anything that is basically a link from a cell to a bit of data like a hyperlink it makes sense to use the same code and UI. But it isn't important for now, if anything else comes up it will be easy enough to add a "type" to each hyperlink.

tizee:
Now I see. We could reuse the data structures of `hyperlink.c` for later development like supporting annotations, or building up tmux UI components with this kind of information especially for terminals that do not support it. Right now it data structure only limited to `hyperlink`.
BTW, there lacks the logic on verifying the correctness of URL. Right now tmux merely delegates this to terminals.

nicm:
Exactly, so if there were other types of annotation they would work the same way. And for example tmux could provide commands to list hyperlinks/annotations, to jump between them in copy mode, to open a program in a popup with the hyperlink. These would not be terribly hard to implement (for example it would be relatively easy to add a format like `mouse_word` but for the hyperlink under the mouse, so it could be used in a menu or by double clicking).

I don't think tmux needs to verify the correctness of the URI since it doesn't use it, it is OK to rely on the receiving program to work out how to handle it. IIRC when this feature was discussed by the (now pretty much moribund) terminal standards group it was decided not to specify the format of the URI anyway, it is just a string.

nicm:
But all that can be left for later, the only thing I think we need to do now is to add a global limit on number of hyperlinks. Are you going to look at that?

tizee:
Yes. But first I'd like to share some of my thoughts on the code design.

nicm:
OK no worries, I will not do anything else with this until tomorrow anyway.

tizee:
- 1. add a global hyperlink counter and a global queue of hyperlinks
- 2. use this counter to  init the inner id of hyperlink and increment it by 1 each time a new hyperlink is created and decrement it by 1 each time a hyperlink is popped out from the queue
- 3. Now there is only one global set of hyperlinks 
 - Remove the hyperlink set from screen structure because we can create hyperlink's id based on the screen information from input context.

nicm:
It might be better to keep a set of hyperlinks per screen since that will keep the tree much smaller, it will be slow to search a few thousand hyperlinks and if there are multiple on a large screen we could be doing it for a few hundred cells.

tizee:
When each screen has its own set of hyperlinks,  then if the number reached the limit, we have to decide which trees to prune. It's kinda of tricky.
On the one hand, the hyperlink is associated with the set in a screen. 
On the other hand, when the screen is freed, the set is also freed. 
So the global queue store a new structure of information for this.

nicm:
> The tricky part is when the tree was destroyed but hyperlinks of the tree remained in the global queue.
> Or we could iterate over the queue and remove these deleted hyperlinks. So a clean up function is needed for this.

No, you know every hyperlink must be on the global queue, so you remove it from the global queue when you destroy the `struct hyperlinks`.

So you would have something like:
~~~~
void
hyperlink_remove(struct hyperlink_uri *hlu)
{
    struct hyperlinks *hl = *hlu->owner;
 
    TAILQ_REMOVE(&global_hyperlinks, hlu, list_entry);
    global_hyperlink_count--;

    RB_REMOVE(..., &hl->by_uid ...);
    RB_REMOVE(..., &hl->by_inner, ...);

    ...
    free(hlu);
}

tizee:
Thanks for suggestions. Now it's much clear for implementation.

nicm:
Looks good except for a  few comments:

- You can use `TAILQ_HEAD_INITIALIZER` instead of adding an initializer function.

- In `hyperlink_put` you know the list will not be empty, so I don't know why you need to loop around `TAILQ_FIRST`?

- In `hyperlink_put` you remove the item from the list, but then call `hyperlink_remove` which removes it again - this is likely to crash eventually.

tizee:
> Looks good except for a few comments:
> 
>     * You can use `TAILQ_HEAD_INITIALIZER` instead of adding an initializer function.
> 
>     * In `hyperlink_put` you know the list will not be empty, so I don't know why you need to loop around `TAILQ_FIRST`?
> 
>     * In `hyperlink_put` you remove the item from the list, but then call `hyperlink_remove` which removes it again - this is likely to crash eventually.

```c
#define	TAILQ_FIRST(head)		((head)->tqh_first)
```
So I have to use a dummy head to hold this pointer.

nicm:
I don't know what you are saying? What's wrong with just doing?
~~~~
    if (global_hyperlinks_count == MAX_HYPERLINKS)
        hyperlinks_remove(TAILQ_FIRST(&global_hyperlinks));

nicm:
I have applied this to OpenBSD now with some tweaks, most notably I renamed `hyperlink.c` to `hyperlinks.c` and the various functions similarly and also made a few minor fixes to your recent changes.

This will be merged into GitHub later and will probably need a few manual changes to get it to build which I will do after commit.

If you want to do anything further, it would be nice to:

- Add support to `capture-pane -e`.

- Add a `mouse_hyperlink` format like `mouse_word` which will work out if there is a hyperlink at the mouse position and return it (look at `format_cb_mouse_word` and `format_grid_word` in `format.c` - this will be much simpler because it will only need to look at one `grid_cell`).

Best wait until it has merged and open a new PR for these.

Thanks!

nicm:
Oh another thing that would be good is to convert your test script into an actual regression test in `regress`, but I think you will need `capture-pane -e` support first (the test would run tmux inside tmux and using `capture-pane -e` to verify the outer tmux is receiving the correct OSC 8 sequences from the inner, there are other tests that do similar).

chrysn:
Adding config lines from https://github.com/tmux/tmux/pull/3240#issuecomment-1167450899 enabled hyperlinks for me with a build from the current master; just setting `set -ga terminal-features "*:hyperlinks"` (as suggested in https://github.com/tmux/tmux/pull/3240#issuecomment-1168442071) did not enable hyperlinks.

What is the latest way to enable this feature? And why does it need to be enabled at all? (I see why it needs explicit support in tmux given wrapping etc, but I don't see why it needs to be configured, given that OSC don't need runtime probing).

tizee:
If you are using macOS, you need to link against with `ncurses > 5.8` since macOS's system library `ncurses` is 5.7 which is buggy for handling terminfo parameter string. 

See https://github.com/tmux/tmux/blob/dc6bc0e95acc04cdf43e869294ecba897a11d850/tty-features.c#L94

BTW, the OSC-8 hyperlinks feature is enabled by default right now in current master branch so you do not need to configure it.

chrysn:
I'm running the latest tmux master on Debian sid (with libncurses-dev around 6.3) -- and have so far always needed to use the two config lines (actually just the `*:Hls=` one, but I suppose that depends on whether I test with or without IDs) to get hyperlinks passed through tmux.

I tried:
* terminals: alacritty (recent master branch, and my terminfo might be inaccurate, not that it should matter) and gnome-terminal (which has had it for years)
* completely empty config file
* I made sure to use an independent tmux session from my regular one (`-S new`) and even shut down my regular tmuxes completely.

tizee:
It's unnecessary to use `*:Hls=`. You should use `set -ga terminal-features "*:hyperlinks"` if the tmux does not recognize your terminal emulator in a minimal config. 
Then you could use `printf '\033]8;id=tmux1;https://github.com\033\\test1\033]8;;\033\\\n'` to test it by clicking on the output (enable mouse mode if your terminal emulator cannot handle OSC 8 hyperlinks).

Here is a minimal config I used before to test this feature.
```
set-option -g default-terminal "xterm-256color"
set -ga terminal-features "*:hyperlinks"
set-option -g mouse on
```

rumpelsepp:
> terminals: alacritty (recent master branch, and my terminfo might be inaccurate, not that it should matter) and gnome-terminal (which has had it for years)

Can confirm. `alacritty` master shows hyperlinks, but with `tmux` master it doesn't.

chrysn:
Sorry, that doesn't change anything -- neither gnome-terminal's nor alacritty's link highlighting / recognition is kicking in.

I now also tested in xterm (as you mentioned that tmux would do something on its own), but didn't see any action happening. (Then again, I'm not sure what *should* happen there -- I wouldn't have expected tmux to fill in the gap). I started setting suitable PATHs just in case `./tmux` would start a server with the system-wide binary.

To rule out setup errors or misunderstandings, here's what I've done for the last run:
* Logged in as a user with no relevant configuration
* Opened gnome-terminal
* Ran the test line you provided: The `test1` text has a dotted underline, and moving the mouse over it gives a hand cursor and tooltip indicating github
* In /tmp/tmux (a current clone), ran `export PATH=/tmp/tmux:$PATH`; `./autogen.sh; ./configure; make; tmux`
* Ran the test line again: No underline, no hand, no tooltip.
* Set the configuration lines you suggested, closed tmux and started it again: Same behavior.

tizee:
> Ran the test line you provided: The test1 text has a dotted underline, and moving the mouse over it gives a hand cursor and tooltip indicating github

That's it.

Screen Shot2

Screen Shot1

I just tested it on ArchLinux using Alacritty with following config:
```
set-option -g default-terminal "xterm-256color"
set -ga terminal-features "*:hyperlinks"

set-option -s terminal-overrides ""
unbind-key C-b
unbind-key C-a
set-option -g prefix C-a
# disable prefix2
set-option -g prefix2 None
bind-key -r C-a send-prefix
set-option -g mouse on
```

chrysn:
OK, then we were just looking at completely different features:

I expected tmux to just transparently forward OSC8 (to a terminal that handles it on its own), and you expected and implemented tmux to turn it into an OSC52 (copying to clipboard), local typing, or at any rate handle it inside tmux.

I guess both have their justifications depending on how tmux is used. (Personally I use it mostly because it survives X server crashes or SSH outages, and because its tab switching and history management is superior to to the terminal emulator's -- but didn't even know that there was any kind of context menu). I'll take this as an occasion to re-evaluate whether maybe having tmux handle it is the better choice for my case (dumb terminal emulators are good terminal emulators, so why not...), but in case I stick with my workflow: Is setting `set -sa terminal-overrides ',*:Hla=...` etc an expected-to-stay-working solution for those who just want to keep using their terminal emulator's OSC8 handling?

tizee:
You should not use `set -sa terminal-overrides ',*:Hla=...` to override the terminal features provided by tmux. Instead, you should use `set -ga terminal-features "*:hyperlinks"` to force the tmux turn on this feature regardless of your current `default-terminal`.

Sorry, I forgot that this feature is not turned on by default right now.

see https://github.com/tmux/tmux/blob/dc6bc0e95acc04cdf43e869294ecba897a11d850/options-table.c#L364

tizee:
> I expected tmux to just transparently forward OSC8 (to a terminal that handles it on its own), and you expected and implemented tmux to turn it into an OSC52 (copying to clipboard), local typing, or at any rate handle it inside tmux.

If you expect the terminal emulator should handle the OSC 8 hyperlinks correctly, then you find tmux could handle it instead of the terminal emulator then probably the terminal does not fully support it. Or you forget to set the action for the terminal to tell how it should handle the OSC 8 hyperlinks. For example, this action is not configured with `xdgopen`.

FYI, tmux's context menu `mouse_hyperlink` is designed to be a fallback solution for terminals without OSC 8 hyperlinks support.

chrysn:
Sorry, I don't get it to work that way. In neither hyperlink supporting terminal I have at hand, setting the terminal-features to wildcard-hyperlinks changes anything. Is there any reference setting you could recommend? (The one I just tried is nested tmux, but I'm not sure I got mouse enablement and disablement right everywhere)

tizee:
How about `set-option -s terminal-overrides ""`? 
The hyperlink supporting terminals I use for testing are Kitty and Wezterm. Both of them could handle the OSC 8 sequences generated by tmux with following config. As I said before, you could read tmux's log file `tmux-server-xxxx.log` to find whether it's caused by tmux or by the terminal. 

```
set-option -g default-terminal "xterm-256color"
set -ga terminal-features "*:hyperlinks"

set-option -s terminal-overrides ""
unbind-key C-b
unbind-key C-a
set-option -g prefix C-a
# disable prefix2
set-option -g prefix2 None
bind-key -r C-a send-prefix
set-option -g mouse on
```

chrysn:
Using exactly that config I'm still in a situation where hyperlinks are visible to gnome-terminal.

Ignoring yylex_token, _parse and key_bindings_add log lines, I see these lines in tmux-server-....log that appear relevant to hyperlinks:

```
1657998293.676568 adding term xterm-256color
1657998293.676769 adding terminal features clipboard:ccolour:cstyle:focus:title
1657998293.676775 adding terminal feature: clipboard
1657998293.676779 adding terminal feature: ccolour
1657998293.676784 adding terminal feature: cstyle
1657998293.676788 adding terminal feature: focus
1657998293.676792 adding terminal feature: title
1657998293.676796 adding terminal features hyperlinks
1657998293.676800 adding terminal feature: hyperlinks
1657998293.676809 RGBCOLOURS flag is 0
1657998293.676814 DECSLRM flag is 0
1657998293.676818 DECFRA flag is 0
1657998293.676822 NOAM flag is 0
1657998293.676826 adding terminal features bpaste,focus,title
1657998293.676830 adding terminal feature: bpaste
1657998293.676835 applying terminal features: bpaste,ccolour,clipboard,hyperlinks,cstyle,focus,title
1657998293.676839 applying terminal feature: bpaste
1657998293.676843 adding capability: Enbp=\\E[?2004h
1657998293.676849 adding capability: Dsbp=\\E[?2004l
1657998293.676854 applying terminal feature: ccolour
1657998293.676864 adding capability: Cs=\\E]12;%p1%s\\a
1657998293.676869 adding capability: Cr=\\E]112\\a
1657998293.676875 applying terminal feature: clipboard
1657998293.676879 adding capability: Ms=\\E]52;%p1%s;%p2%s\\a
1657998293.676884 applying terminal feature: hyperlinks
1657998293.676888 applying terminal feature: cstyle
1657998293.676892 adding capability: Ss=\\E[%p1%d q
1657998293.676897 adding capability: Se=\\E[2 q
...
1657998293.677133 xterm-256color  44: Enfcs: (string) \\033[?1004h
1657998293.677137 xterm-256color  45: Enmg: [missing]
1657998293.677142 xterm-256color  46: fsl: (string) \\a
1657998293.677146 xterm-256color  47: Hls: [missing]
1657998293.677150 xterm-256color  48: home: (string) \\033[H
1657998293.677154 xterm-256color  49: hpa: (string) \\033[%i%p1%dG
...
1657998293.690221 /dev/pts/6: received secondary DA \033[>65;6800;1c
1657998293.690230 applying terminal features: bpaste,ccolour,clipboard,hyperlinks,cstyle,focus,title
1657998293.690235 /dev/pts/6: \033[?7727h
1657998293.690240 /dev/pts/6: complete key \033[>65;6800;1c 0xfe000000000
...
1657998295.388339 screen_write_start_pane: size 86x55, pane %0 (at 0,0)
1657998295.388411 input_enter_osc
1657998295.388470 input_exit_osc: "8;id=tmux1;https://github.com" (end ST)
1657998295.388506 hyperlink (id=tmux1) https://github.com = 1
1657998295.388528 input_esc_dispatch: '\\', 
1657998295.388552 screen_write_collect_end: 5 test1 (at 0,1)
1657998295.388573 utf8_from_data: (1 1 t) -> 41000074
1657998295.388592 utf8_from_data: (1 1 t) -> 41000074
1657998295.388610 utf8_from_data: (1 1 t) -> 41000074
1657998295.388627 utf8_from_data: (1 1 t) -> 41000074
```

later followed by processing of a right-click (but that's already after the OSC8 should have been forwarded to the terminal).

My interpretation of these is that while the hyperlink feature is set (at latest by the explicit configuration), that is insufficient to make the hyperlinks actually forwarded: No `Hls` is set, and thus tmux would not know how to tell the outer terminal that there is a hyperlink (even if it tried to). If that is accurate, that'd indicate that Hls needs to be given explicitly unless the termcaps are updated. (Though I wonder why this needs to come from termcap -- unlike the escape codes of old, can these really vary by terminal?).

nicm:
What ncurses version does it say in the first few lines of the log?

chrysn:
```
1657998293.601273 server started (299696): version next-3.4, socket /tmp/tmux-1001/default, protocol 8
1657998293.601308 on Linux 5.18.0-2-amd64 #1 SMP PREEMPT_DYNAMIC Debian 5.18.5-1 (2022-06-16)
1657998293.601327 using libevent 2.1.12-stable (poll); ncurses 6.3
1657998293.601400 input_key_build: 0x10e003 (PasteStart) is \033[200~
```

nicm:
The check for ncurses >= 5.8 in tty-features.c must be wrong, try removing it and see if it works.

chrysn:
Indeed removing the test makes things work -- NCURSES_VERSION_MAJOR/MINOR is not defined; minimal I had to do is:

```patch
diff --git a/tty-features.c b/tty-features.c
index e387f823..eb2b6c72 100644
--- a/tty-features.c
+++ b/tty-features.c
@@ -21,6 +21,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <curses.h>
+
 #include "tmux.h"
 
 /*
```

Probably on the tested platforms, curses.h got included transitively, and as none of it is used ifndef, the compiler did not complain.

nicm:
It'll need to be the curses include bit from the top of tty-term.c instead.

PR #3247, opened 2022-07-01
capture-pane -e support hyperlink

tizee:
#3240 follow-up feature
Now `capture-pane -e` support OSC 8 hyperlinks and paste these hyperlinks via `paste-buffer` correctly.

### DIFF
--- a/cmd-capture-pane.c
+++ b/cmd-capture-pane.c
@@ -177,7 +177,7 @@ cmd_capture_pane_history(struct args *args, struct cmdq_item *item,
 	buf = NULL;
 	for (i = top; i <= bottom; i++) {
 		line = grid_string_cells(gd, 0, i, sx, &gc, with_codes,
-		    escape_c0, !join_lines && !no_trim);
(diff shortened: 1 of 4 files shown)


tizee:
@nicm From my view, I only care about the URL part of OSC 8 hyperlink regardless of its id.

nicm:
Looks good, thanks, except:

- In `format_grid_hyperlink`, the URI has already been run through `utf8_stravis` so there is no point in doing so again, you can just `xstrdup` it.

- I don't see any need to keep these `log_debug` if you no longer need them for testing.

- The code style needs to be changed:
    - `{` goes on a line of its own after function arguments;
    - No `{}` around single line `if`/`while`/etc;
    - Space between `if` and `(` and around `==` or `!=`;
    - Line up variable declarations;
    - Indent with tabs;
    - Don't use `//` comments;
    - Keep lines under 80 columns.

nicm:
Good stuff, applied now with some minor fixes (use internal not external ID for capture-pane, end hyperlinks correctly, and some style bits). Thanks!

PR #3263, opened 2022-07-16
Include (n)curses for workaround detection

chrysn:
The relevant #if in tty_feature_hyperlinks_capabilities would otherwise fail silently.

This is a follow-up fix for https://github.com/tmux/tmux/pull/3240 prompted by @nicm's [comment](https://github.com/tmux/tmux/pull/3240#issuecomment-1186274909).

### DIFF
--- a/tty-features.c
+++ b/tty-features.c
@@ -21,6 +21,12 @@
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


rumpelsepp:
can confirm this works for alacritty master.

nicm:
Applied, thanks!

PR #3267, opened 2022-07-18
Add checking for yacc existence

willie5588912:
Hi there, I've found that if there is no bison/byacc installed, it still successes without errors in the configuration stage, and only in make stage that the errors related to something like **unknown command yacc** show.
I think it's better to check yacc existence in the configuration stage? 

I've tested on my machine which is Ubuntu 20.04. If there is no bison/byacc, an error shows in the configuration stage, otherwise it passes if bison or byacc installed.

Thanks for your efforts.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -268,10 +268,8 @@ if test "x$found_libevent" = xno; then
 fi
 
 # Look for yacc.
-AC_CHECK_PROG(found_yacc, yacc, yes, no)
-if test "x$found_yacc" = xno; then
-	AC_MSG_ERROR("yacc not found")
-fi
+$YACC &>/dev/null
+AS_IF([ test $? == 127 ], [AC_MSG_ERROR("yacc not found")])
 
 # Look for ncurses or curses. Try pkg-config first then directly for the
 # library.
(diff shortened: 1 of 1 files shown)


nicm:
It seems rather like this should be part of autoconf's `AC_PROG_YACC`, so it seems like a bug in autoconf really.

But there is no harm in tmux checking too. However, this check should make sure `$YACC` works, not look for `yacc`.

willshih92:
Agree. Try to check whether the exit code of `$YACC` is equal to 127 (means command not found). It works, but not sure if this is a good method.

willie5588912:
Agree. Try to check whether the exit code of `$YACC` is equal to 127 (means command not found). It works, but not sure if this is a good method.

nicm:
What about using `AC_CHECK_PROG` with `$YACC`? Does that work?

willie5588912:
Yes, it works! I didn't do this because I thought it may not work when `$YACC` = `bison -y` (it's an incomplete command or a flag).
Well, I've tested it and confirmed that `AC_CHECK_PROG` can handle this situation correctly, so I pushed a new commit, thanks!

nicm:
Applied, thanks!

PR #3268, opened 2022-07-18
Fix memory leak in format_find

gbsf:
`format_quote_{shell,style}` already `xmalloc`s a new string, so the `xstrdup` isn't necessary.

The leak is small, but after a couple of months tmux had the biggest memory consumption on my server.


nicm:

PR #3269, opened 2022-07-19
Fix minor typos in regress configs

afh:
found by [codespell](https://github.com/codespell-project/codespell).


nicm:
Applied, thanks!

PR #3274, opened 2022-07-23
Allow passthrough from invisible panes when allow-passthrough is 'all'

sergei-grechanik:


nicm:
This looks good, applied to OpenBSD now with some very minor changes. The only bit I was unsure about was in `tty_write` we could probably use `tty_client_ready` for both cases (passing the invisible flag in if necessary), but I think it is OK as it is for the moment - if we want to use this for any other commands, then we will probably need to pass the flag into `set_client_cb` anyway. Thanks!

PR #3286, opened 2022-07-31
Change 'g' and 'G' to go to top and bottom of menu respectively.

ndsizeif:
### Proposed changes
Menu navigation with 'g','G', PageUp and PageDown will go to the top or bottom
of a menu.

### Current behavior
Currently these keys increment or decrement menu selections by five.
These movements do not handle navigating over menu separators in every use case. 
This can result in navigation by values other than five, or a menu with seemingly 
nothing selected.

https://user-images.githubusercontent.com/84802637/182046894-e742ef22-fd00-422e-b83b-8eb2d99626d3.mp4

### New behavior
Navigation with the aforementioned keys will simply either go to the top of the menu or the bottom.
We also now account for all cases where the choices are separators and not actual options (even if the last
item in a menu is a separator).

https://user-images.githubusercontent.com/84802637/182046907-fdb10703-50af-4eb0-876a-cafc0210f0d1.mp4

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -326,24 +326,24 @@ menu_key_cb(struct client *c, void *data, struct key_event *event)
 	case 'g':
 	case KEYC_PPAGE:
 	case '\002': /* C-b */
-		if (md->choice > 5)
-			md->choice -= 5;
-		else
-			md->choice = 0;
-		while (md->choice != count && (name == NULL || *name == '-'))
+		md->choice = 0;
+		name = menu->items[md->choice].name;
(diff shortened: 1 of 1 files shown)


nicm:
Thanks. I don't really mind what `g` and `G` do but I think `NPage`/`PPage` should continue to go up and down by 5 entries (they should be fixed to skip separators though). It would be best to add two new cases for `g`/`Home` and `G`/`End` to go to start and end of the menu.

nicm:

PR #3297, opened 2022-08-10
Fix a logic error when setting $PWD in spawn.c

markus-oberhumer-forks:
This was introduced in commit 6a5d210 .


nicm:

PR #3298, opened 2022-08-10
Add missing copy-mode-vi key bindings for Home and End keys.

markus-oberhumer-forks:


nicm:

PR #3300, opened 2022-08-12
Improve const-correctness.

markus-oberhumer-forks:
Constify a number of static pointer arrays and static structs which results
in ~1000 bytes moved from the data (rw) to the text (ro) section.

Before:
 $ size tmux
   text	   data	    bss	    dec	    hex	filename
 853211	   8984	  30256	 892451	  d9e23	tmux

After:
 $ size tmux
   text	   data	    bss	    dec	    hex	filename
 854235	   7944	  30256	 892435	  d9e13	tmux

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -344,7 +344,7 @@ key_bindings_init_done(__unused struct cmdq_item *item, __unused void *data)
 void
 key_bindings_init(void)
 {
-	static const char *defaults[] = {
+	static const char * const defaults[] = {
 		/* Prefix keys. */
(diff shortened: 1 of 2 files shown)


nicm:
I don't usually bother too much about this because the syntax is quite ugly for the sake of a few hundred bytes of text, but I've applied this anyway. Thanks!

markus-oberhumer-forks:
Yes, I agree that it's ugly in C/C++.

But then I view moving as much data as possible to the text (read-only) section as both a correctness and security improvement that should be done if easily possible.

PR #3301, opened 2022-08-14
Introduce Nobr terminfo extension

### LINKED ISSUE #3275: Using OSC4 sequence changes bold color to bright one
OS: Arch Linux
tmux: 3.3a
terminal emulator: foot (TERM=foot outside, TERM=tmux-256color inside)

Here is an initial screenshot of all the configured colors when using bold and normal display
screenshot_20220724121726

And after "changing" the colors via the OSC 4 sequence (actually using the same colors already defined as theme), here's the screenshot of the result
screenshot_20220724121739

So you can see that the colors (30-37)  have been modified for the bold display to use the "bright" (90-97) colors

Here are the attached logs:
[tmux-client-282325.log](https://github.com/tmux/tmux/files/9175573/tmux-client-282325.log)
[tmux-out-282327.log](https://github.com/tmux/tmux/files/9175575/tmux-out-282327.log)
[tmux-server-282327.log](https://github.com/tmux/tmux/files/9175576/tmux-server-282327.log)

Some extra things:
- to confirm that this issue comes from tmux, I've also tested this on foot itself without tmux and the issue is then not present
- I've also tested the background colors (SGR codes 40-47, 100-107) and there's no issue there too
- as is visible from the first screenshot, I do not have the foot `bold-text-in-bright` config enabled

tardypad:
To match terminal not using bright colors for bold display.
Fix #3275

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -6433,6 +6433,10 @@ These are set automatically if the
 capability is present.
 .It Em \&Hls
 Set or clear a hyperlink annotation.
+.It Em \&Nobr
+Tell
+.Nm
+that the terminal does not use bright colors for bold display.
 .It Em \&Rect
 Tell
 .Nm
--- a/tmux.h
+++ b/tmux.h
@@ -515,6 +515,7 @@ enum tty_code_code {
 	TTYC_KUP6,
 	TTYC_KUP7,
 	TTYC_MS,
+	TTYC_NOBR,
 	TTYC_OL,
 	TTYC_OP,
 	TTYC_RECT,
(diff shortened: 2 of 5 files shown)


tardypad:
Hm... I think I got confused and I now see this should be listed as a "terminfo extension" and not as a "terminal feature".
It was indeed a bit weird to have as "terminal feature" because this is more of config "is it on or off".
I'll look at it again

nicm:
Great, I have applied this but without the tty-features.c bits (so you will need to use terminal-overrides). It'll be in GitHub as soon as it syncs up. Thanks!

PR #3302, opened 2022-08-14
Notify control-mode clients of changes to paste buffers.

gnachman:
Sends %paste-changed <name> when a paste buffer is modified.

This is motivated by the desire to support neovim's copy-to-tmux feature in iTerm2's tmux integration. More background here: https://gitlab.com/gnachman/iterm2/-/issues/10530

### DIFF
--- a/control-notify.c
+++ b/control-notify.c
@@ -234,3 +234,16 @@ control_notify_session_window_changed(struct session *s)
 		    s->curw->window->id);
 	}
 }
+
+void
+control_notify_paste_changed(const char *name)
+{
+	struct client	*c;
+
+	TAILQ_FOREACH(c, &clients, entry) {
+		if (!CONTROL_SHOULD_NOTIFY_CLIENT(c))
+			continue;
+
+		control_write(c, "%%paste-changed %s", name);
+	}
+}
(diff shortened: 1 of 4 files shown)


nicm:
I've applied this but I changed `%paste-changed` to `%paste-buffer-changed` and similarly changed the function names since I think that is a bit more consistent with the other commands and whatnot.

I also added a notify call to `paste_free` so it also notifies on `delete-buffer` (and when automatic buffers are rotated away if `buffer-limit` is hit).

I don't know if that will cause trouble - if so then don't mind if that becomes `%paste-buffer-deleted` instead, or a flag argument added to show the buffer was being deleted?

gnachman:
> I don't know if that will cause trouble - if so then don't mind if that becomes %paste-buffer-deleted instead, or a flag argument added to show the buffer was being deleted?

Yeah, a flag would be good. I don't think I'll track deletions since my goal is to sync the system clipboard with the latest call to `load-buffer`.

nicm:
There isn't a simple mechanism to add a flag, so I have changed deletion to `%paste-buffer-deleted`. I also noticed I missed actually changing the control mode notification to `%paste-buffer-changed` so I have done that also. Please let me know of any problems. Thanks!

PR #3307, opened 2022-08-20
add ability to center cursor vertically in copy mode. Github issue 3159.

varunkumare99:
cursor_center
in vi copy mode 'z' used to center the cursor vertically.
in normal copy mode 'C-m' used to center the cursor vertically.


nicm:
I can't get this to work, when I press the keys nothing happens, but I assume it works for you so I must be doing something wrong?

Also I noticed we already have `middle-line` (`M-r` in emacs) - does this do something different? Sorry I had completely forgotten about it!

varunkumare99:
Maybe,
- can you try clearing tmp/ then starting tmux?
- Have you tried it with a big enough file. Because if the file is smaller than the size of the screen, and the cursor is at the bottom. When you press `z` in vi-copy-mode it won't move the screen and cursor up as is it at the bottom.

 `middle-line` only moves the cursor to the middle. It does not move the screen along with it. 
cursor_middle_line

nicm:
Yes of course I was using the right socket.

I'll try it again later in the week but I'm sure I tried it with a many lines in the history.

In the meantime a few comments:

- I don't think emacs has this so I don't think it needs a default binding with emacs keys.

- The indentation is off, it should be 8 character tabs.

- This needs to be added to tmux.1, there is a table of the commands.

Thanks!

varunkumare99:
can you share your tmux.conf maybe? I can try to test it out.
I have added this feature only in the default mode and vim mode. (emacs not added)

_The indentation is off, it should be 8 character tabs._  Could you explain this?

Thank you.

nicm:
emacs is the default, so only add a key binding for copy-mode-vi.

> The indentation is off, it should be 8 character tabs. Could you explain this?

Your new code should be indented with a single 8 character tab, not spaces.

nicm:
OK I've got it to work now. I have applied this to OpenBSD with some fixes and also renamed the command to `scroll-middle` rather than `cursor-middle` since it better describes what it does. Thanks!

nicm:
This will be in GitHub later when it syncs up.

nicm:
No it's OK, I did them. Thanks

varunkumare99:
thank you so much.

PR #3308, opened 2022-08-20
Fix own fallback implementation of `getpeereid()`

pinotree:
Use `geteuid()` & `getegid()` (including `<unistd.h>`) to set uid & gid in
the out parameters, so they are set to callers. Also return 0, since
those functions do not fail.

Fixes commit 8bcd392ee79996f828fd40c52198071ec0f273dd.


nicm:
Fixed, thanks!

PR #3313, opened 2022-08-25
tmux.1: tweak docs for new-session's -d and -D params to match behaviour

atomicstack:
The `-d` and `-D` params to `new-session`, as described in the tmux man page, seem to conflict with the observed behaviour (and with itself).

They are described as following:

> The new session is attached to the current terminal unless -d is given. window-name and shell-command are the name of and shell command to execute in the initial window.  With -d, the initial size comes from the global default-size option; -x and -y can be used to specify a different size.

[later on in the same file]

> The -A flag makes new-session behave like attach-session if session-name already exists; in this case, -D behaves like -d to attach-session, and -X behaves like -x to attach-session.

In practice, the `-d` param creates a detached session:

```
photon ~ % tmux new-session -d -s detached
photon ~ % tmux list-sessions | grep detached
detached: 1 windows (created Thu Aug 25 14:00:24 2022)
photon ~ % tmux kill-session -t detached
```

(and it appears that `-D` will use `default-size`.)

OTOH, it looks like the other attach-session-ish parameters are in capitals (i.e. `-A` and `-X`), so it's possible the docs may illustrate the intended behaviour after all?

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -1235,14 +1235,14 @@ Lock all clients attached to
 Create a new session with name
 .Ar session-name .
 .Pp
-The new session is attached to the current terminal unless
+The new session is attached to the current terminal, unless
 .Fl d
 is given.
 .Ar window-name
 and
 .Ar shell-command
 are the name of and shell command to execute in the initial window.
-With
+When creating a detached session with
 .Fl d ,
 the initial size comes from the global
 .Ic default-size


nicm:
This seems perfectly correct to me, `-d` creates a detached session (that is, it does not attach, as the man page says). `-D` behaves like `attach -d` and detaches other sessions, if used with `-A`.

atomicstack:
Okay, thanks for the clarification. In that case, I adjusted the docs slightly to make this point clearer. Feel free to merge, if you feel it's useful.

nicm:
I don't think this is better TBH. Maybe we can change the first bit to mention detached, something like this:
~~~~
@@ -1234,7 +1234,9 @@ Create a new session with name
 .Pp
 The new session is attached to the current terminal unless
 .Fl d
-is given.
+is given (with
+.Fl d
+it is detached).
 .Ar window-name
 and
 .Ar shell-command
~~~~

PR #3319, opened 2022-08-30
Fix answer to CSI winops sequence to report window size

vincentbernat:
The answer should report window height, then window width. This is a
regression introduced in 3bbd66c0137f.


nicm:
Hah, fixed now, thanks!

PR #3324, opened 2022-09-04
Support moving the prompt for command on different status line. Githu…

varunkumare99:
- added an option 'status-prompt-line' (when status is set to 5, status-prompt-line can range from 1-5 where 5 is the bottom of the screen. 
- default value of status-prompt-line is 1

```
 set-option -g status 5
 set-option -g status-prompt-line 3
```
puts status prompt on the 3rd line from the bottom.

prompt

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -42,7 +42,7 @@ static const char *options_table_status_list[] = {
 	"off", "on", "2", "3", "4", "5", NULL
 };
 static const char *options_table_status_prompt_line_list[] = {
-	"0", "1", "2", "3", "4", "5", NULL
+	"0", "1", "2", "3", "4", NULL
 };
 static const char *options_table_status_keys_list[] = {
 	"emacs", "vi", NULL
--- a/options-table.c
+++ b/options-table.c
@@ -628,7 +628,7 @@ const struct options_table_entry options_table[] = {
 	  .type = OPTIONS_TABLE_CHOICE,
 	  .scope = OPTIONS_TABLE_SESSION,
 	  .choices = options_table_status_prompt_line_list,
-	  .default_num = 1,
+	  .default_num = 0,
 	  .text = "Line number of status prompt"
 	},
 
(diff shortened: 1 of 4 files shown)


nicm:
The code looks great but two things:

- In the `status-format` option, lines are numbered from zero where zero is the top line (`status-format[0]` is the top line). I think this should be the same.

- You need to add this option to tmux.1.

Thanks!

varunkumare99:
how do you check the formatting once I have made my changes in tmux.1. Any way to build the docs and check it?
Also what formatting/language is used here(in tmux.1)?

Thanks!

nicm:
It is called mdoc, if you are on Linux, IIRC tou can do `nroff -mdoc ./tmux.1`. But if you do what looks right I can check it.

varunkumare99:
Oh, nice interesting.

nicm:
I have made a few changes to your code - we have `message-style` which controls style of both command prompt and message line, so I have renamed your option to `message-line` and made it set the line for both command prompt and message. Please look at this and let me know what you think: [x.diff.txt](https://github.com/tmux/tmux/files/9503292/x.diff.txt)

I never liked the way `message-style` is named and that it does both message and command prompt but so far it has not been a problem, so I think it is OK. If we wanted to split them up and rename it we could do it and keep the old ones in `options_other_names`. But that would be a separate change...

varunkumare99:
I feel the styles and the line number for both `prompt` and `message` should be separated.

```
Line options:
status-prompt-line
message-line

```
```
Styles:
message-style
command-prompt-style
```

If its fine with you I will implement with this.
thanks

varunkumare99:
in the function status_prompt_redraw() in status.c. 
We choose between two styles` message-style` and `message-command-style`?
Doesn't this separate the styles for `prompt` and `message`?

nicm:
No, `message-command-style` is when it is in command mode with status-keys=vi. It will also need to be renamed.

If we separate them, it will break all configs using `message-style` and so far nobody has requested it. Are you going to set them to different things?

varunkumare99:
Then its fine.
the x.diff.txt looks good.
I made a mistake. The default prompt was set to `1` instead of `0`(ie topmost line of the status).
Add this change to x.diff.txt

thanks.

nicm:
Great, I have applied this to OpenBSD now, it will be in GitHub later. Thanks!

varunkumare99:
Thank You!

PR #3368, opened 2022-10-29
tty-keys and input-keys testing and new extended keys

### LINKED ISSUE #3366: C-S-Tab handling with extended keys
### Issue description

Tmux does not appear to detect `M-[ 1 ; 5 Z` as C-S-tab, which is what iterm2 sends (and I believe is what the spec specifies for C-S-tab w/ CSI). C-tab works fine. I'd be happy to add support, but I don't know where the CSI handling is in the code.

### Required information

* tmux version (`tmux -V`): 3.3a (also tested with master)
* Platform (`uname -sp`): Darmin arm
* $TERM inside and outside of tmux (`echo $TERM`).: outside: xterm-256color-italic, inside: tmux-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`).: I did this, but I'm not sure where logs go

aaronjensen:
Fixes #3366

Also ensures only Tab is sent for C-Tab and C-S-Tab if extended keys is off.

### DIFF
--- a/input-keys.c
+++ b/input-keys.c
@@ -316,7 +316,7 @@ static struct input_key_entry input_key_defaults[] = {
 	  .data = "\033[9;5u"
 	},
 	{ .key = '\011'|KEYC_CTRL|KEYC_SHIFT,
-	  .data = "\011"
+	  .data = "\033[Z"
 	},
 	{ .key = '\011'|KEYC_CTRL|KEYC_SHIFT|KEYC_EXTENDED,
 	  .data = "\033[1;5Z"
--- a/regress/input-keys.sh
+++ b/regress/input-keys.sh
@@ -1,0 +1,311 @@
+#!/bin/sh
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -Ltest"
+$TMUX kill-server 2>/dev/null
+
+SCRIPT=$(mktemp)
+trap "rm -f $SCRIPT" 0 1 15
+
+cat <<EOF >"$SCRIPT"
+for control in \$(stty -a | grep -oa '\w* =' | cut -d' ' -f1); do
+  stty \$control undef
+done
+
+stty -a
+
+cat -tve
+EOF
+
+$TMUX -f/dev/null new -x20 -y2 -d -- /bin/sh "$SCRIPT" || exit 1
+
+sleep 0.1
+
+exit_status=0
+
+assert_key () {
+	key=$1
+	expected_code=$2
+
+	$TMUX send-keys "$key" Enter
+
+	actual_code=$($TMUX capturep -p | head -1 | sed -e 's/\$//')
+
+	if [ "$actual_code" = "$expected_code" ]; then
+		if [ -n "$VERBOSE" ]; then
+			echo "[PASS] $key -> $actual_code"
+		fi
+	else
+		echo "[FAIL] $key -> $expected_code (Got: $actual_code)"
+		exit_status=1
+	fi
+
+	shift
+	shift
+
+	if [ "$1" = "--" ]; then
+		shift
+		assert_key "$@"
+	fi
+}
+
+assert_key 'C-Space' '^@'
+assert_key 'C-a' '^A'	 -- 'M-C-a' '^[^A'
+assert_key 'C-b' '^B'	 -- 'M-C-b' '^[^B'
+assert_key 'C-c' '^C'	 -- 'M-C-c' '^[^C'
+assert_key 'C-d' '^D'	 -- 'M-C-d' '^[^D'
+assert_key 'C-e' '^E'	 -- 'M-C-e' '^[^E'
+assert_key 'C-f' '^F'	 -- 'M-C-f' '^[^F'
+assert_key 'C-g' '^G'	 -- 'M-C-g' '^[^G'
+assert_key 'C-h' '^H'	 -- 'M-C-h' '^[^H'
+assert_key 'C-i' '^I'	 -- 'M-C-i' '^[^I'
+# assert_key 'C-j' ''	 -- 'M-C-j' '' # NL
+assert_key 'C-k' '^K'	 -- 'M-C-k' '^[^K'
+assert_key 'C-l' '^L'	 -- 'M-C-l' '^[^L'
+# assert_key 'C-m' ''	 -- 'M-C-m' '' # CR
+assert_key 'C-n' '^N'	 -- 'M-C-n' '^[^N'
+assert_key 'C-o' '^O'	 -- 'M-C-o' '^[^O'
+assert_key 'C-p' '^P'	 -- 'M-C-p' '^[^P'
+assert_key 'C-q' '^Q'	 -- 'M-C-q' '^[^Q'
+assert_key 'C-r' '^R'	 -- 'M-C-r' '^[^R'
+assert_key 'C-s' '^S'	 -- 'M-C-s' '^[^S'
+assert_key 'C-t' '^T'	 -- 'M-C-t' '^[^T'
+assert_key 'C-u' '^U'	 -- 'M-C-u' '^[^U'
+assert_key 'C-v' '^V'	 -- 'M-C-v' '^[^V'
+assert_key 'C-w' '^W'	 -- 'M-C-w' '^[^W'
+assert_key 'C-x' '^X'	 -- 'M-C-x' '^[^X'
+assert_key 'C-y' '^Y'	 -- 'M-C-y' '^[^Y'
+assert_key 'C-z' '^Z'	 -- 'M-C-z' '^[^Z'
+assert_key 'Escape' '^[' -- 'M-Escape' '^[^['
+assert_key "C-\\" "^\\"	 -- "M-C-\\" "^[^\\"
+assert_key 'C-]' '^]'	 -- 'M-C-]' '^[^]'
+assert_key 'C-^' '^^'	 -- 'M-C-^' '^[^^'
+assert_key 'C-_' '^_'	 -- 'M-C-_' '^[^_'
+assert_key 'Space' ' '	 -- 'M-Space' '^[ '
+assert_key '!' '!'	 -- 'M-!' '^[!'
+assert_key '"' '"'	 -- 'M-"' '^["'
+assert_key '#' '#'	 -- 'M-#' '^[#'
+assert_key '$' '$'	 -- 'M-$' '^[$'
+assert_key '%' '%'	 -- 'M-%' '^[%'
+assert_key '&' '&'	 -- 'M-&' '^[&'
+assert_key "'" "'"	 -- "M-'" "^['"
+assert_key '(' '('	 -- 'M-(' '^[('
+assert_key ')' ')'	 -- 'M-)' '^[)'
+assert_key '*' '*'	 -- 'M-*' '^[*'
+assert_key '+' '+'	 -- 'M-+' '^[+'
+assert_key ',' ','	 -- 'M-,' '^[,'
+assert_key '-' '-'	 -- 'M--' '^[-'
+assert_key '.' '.'	 -- 'M-.' '^[.'
+assert_key '/' '/'	 -- 'M-/' '^[/'
+assert_key '0' '0'	 -- 'M-0' '^[0'
+assert_key '1' '1'	 -- 'M-1' '^[1'
+assert_key '2' '2'	 -- 'M-2' '^[2'
+assert_key '3' '3'	 -- 'M-3' '^[3'
+assert_key '4' '4'	 -- 'M-4' '^[4'
+assert_key '5' '5'	 -- 'M-5' '^[5'
+assert_key '6' '6'	 -- 'M-6' '^[6'
+assert_key '7' '7'	 -- 'M-7' '^[7'
+assert_key '8' '8'	 -- 'M-8' '^[8'
+assert_key '9' '9'	 -- 'M-9' '^[9'
+assert_key ':' ':'	 -- 'M-:' '^[:'
+assert_key '\;' ';'	 -- 'M-\;' '^[;'
+assert_key '<' '<'	 -- 'M-<' '^[<'
+assert_key '=' '='	 -- 'M-=' '^[='
+assert_key '>' '>'	 -- 'M->' '^[>'
+assert_key '?' '?'	 -- 'M-?' '^[?'
+assert_key '@' '@'	 -- 'M-@' '^[@'
+assert_key 'A' 'A'	 -- 'M-A' '^[A'
+assert_key 'B' 'B'	 -- 'M-B' '^[B'
+assert_key 'C' 'C'	 -- 'M-C' '^[C'
+assert_key 'D' 'D'	 -- 'M-D' '^[D'
+assert_key 'E' 'E'	 -- 'M-E' '^[E'
+assert_key 'F' 'F'	 -- 'M-F' '^[F'
+assert_key 'G' 'G'	 -- 'M-G' '^[G'
+assert_key 'H' 'H'	 -- 'M-H' '^[H'
+assert_key 'I' 'I'	 -- 'M-I' '^[I'
+assert_key 'J' 'J'	 -- 'M-J' '^[J'
+assert_key 'K' 'K'	 -- 'M-K' '^[K'
+assert_key 'L' 'L'	 -- 'M-L' '^[L'
+assert_key 'M' 'M'	 -- 'M-M' '^[M'
+assert_key 'N' 'N'	 -- 'M-N' '^[N'
+assert_key 'O' 'O'	 -- 'M-O' '^[O'
+assert_key 'P' 'P'	 -- 'M-P' '^[P'
+assert_key 'Q' 'Q'	 -- 'M-Q' '^[Q'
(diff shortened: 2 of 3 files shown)


aaronjensen:
@nicm Moving the conversation to here.

This adds support for tab varieties as well as a few other keys. I haven't tested all of them yet, but I figured you could take a look at this so far and see if you have any concerns.

One thing I don't understand is why tty-keys converts `C-X` to `C-x`. input-keys puts it back, but it seems limiting. Why not allow tmux to bind to `C-X` and all the others? In order to get `C-M` and `C-m` to work, I had to pass them through, so `C-b / C-M` actually shows `C-M`, rather than `C-m`, which varies from all letters but `i`.

aaronjensen:
I was able to get NUL to send for `C-S-Space`, but it required an explicit case.

nicm:
Thanks. Some questions:

- Correct me if I am wrong here, but this looks like it will emit the extended  key (such as `C-M`) to the pane if the original key received was an extended key (because the `KEYC_EXTENDED` flag was set when the key was received in `input_key`). But this is not correct, it should not emit the extended key unless the `KEXTENDED` flag is set on the pane. So I think instead it should either:

   1) Set `KEYC_EXTENDED` on extended keys that come in (having this information stored is not a bad idea), but remove it before searching the tree if `KEXTENDED` is not set.

   2) Do not set `KEYC_EXTENDED` when the key comes in, but add it if `KEXTENDED` is set before searching the tree (then if that finds no key or if `KEXTENDED` is not set, fall back to searching without `KEYC_EXTENDED`).

   I would probably go for (1) but please update `key_string_lookup_key` to include `KEXTENDED` if `with_flags` is true.

- Why did you only add `C-m` and `C-S-m` and `C-i` and `C-S-i` to the tree? What about all the other keys and modifiers like `C-S-q`? Do they work as it is? If that is the case, why do `m` and `i` not work?

aaronjensen:
> Set KEYC_EXTENDED on extended keys that come in (having this information stored is not a bad idea), but remove it before searching the tree if KEXTENDED is not set.

Yes, that's what it's doing now, I modeled it after the other flags: https://github.com/tmux/tmux/pull/3368/files#diff-83d6d9b0dade6503cea511e72ff7e4ace8c126b5e743f071916045f0f2804e65R511

> Why did you only add C-m and C-S-m and C-i and C-S-i to the tree? What about all the other keys and modifiers like C-S-q? Do they work as it is? If that is the case, why do m and i not work?

Those work as they did before, but it's because those are reduced to their ctrl codes by tty-keys [here](https://github.com/tmux/tmux/pull/3368/files#diff-fd5d00eeffbec5a58a55c15bca5c5a0677d099df7999049afe30d42604228bb5R972) and [here](https://github.com/tmux/tmux/pull/3368/files#diff-fd5d00eeffbec5a58a55c15bca5c5a0677d099df7999049afe30d42604228bb5R978). I excluded i and m in order to allow them to be rendered properly by tmux and then converted back properly. My question is, can I just do that for all letters? i.e. allow tty-keys to leave ctrl letters unaffected. This would mean that if a terminal has extended key support and a person binds in tmux something to `<prefix> C-q`, then when they hit `<prefix> C-S-q` then it would not trigger. I personally think this is reasonable, but I was assuming that [that behavior was put in place](https://github.com/tmux/tmux/pull/3368/files#diff-fd5d00eeffbec5a58a55c15bca5c5a0677d099df7999049afe30d42604228bb5L950-L951) explicitly for a reason. Perhaps it was just because what this PR is doing couldn't be done yet?

nicm:
> Yes, that's what it's doing now, I modeled it after the other flags

OK but the other flags don't have a corresponding screen flag like `KEXTENDED`, so you need to make sure it does not send extended keys to panes without `KEXTENDED` even if `KEYC_EXTENDED` is set.

nicm:
> This would mean that if a terminal has extended key support and a person binds in tmux something to <prefix> C-q, then when they hit <prefix> C-S-q then it would not trigger.

I think that is fine initially - if people complain we may want to change to only do it if the `extended-keys` option is off.

But I'm not wild about the idea of adding every single key to the table, I think it should generate the sequence instead like it does for some of them now.

aaronjensen:
It's very possible I'm missing something, but is that what this does?

```c
	if (~s->mode & MODE_KEXTENDED)
		key &= ~KEYC_EXTENDED;
```

> But I'm not wild about the idea of adding every single key to the table, I think it should generate the sequence instead like it does for some of them now.

Agreed, I would be able to change the code that converts it back to be more straight forward (essentially moving the code from ttykeys to input-keys. I'll play with it.

aaronjensen:
Sure, I updated this PR to just be the tab support. I'll follow up with the rest.

aaronjensen:
Can you think of a way to write a test for this? Ideally I'd like to test sending tmux a key, find out what tmux thinks it is (similar to `tmux command-prompt -kp: "display-message -p %%"`, but automated). Then also a way to send a key to tmux and capture what it sends to the application.

nicm:
You can run tmux inside tmux then use `send-keys` to send keys to the inner tmux and `cat` to see what the program inside tmux receives. Seeing what tmux itself receives you will need to use `command-prompt -k` but you may want to add a flag to it to show the key including flags.

nicm:
I have applied this now to OpenBSD, it will be in GitHub later on. You can reuse this PR or open a new one for the other keys, I don't mind.

Look at `regress/am-terminal.sh` for example of running tmux inside tmux in a test.

aaronjensen:
Thank you for the merge and the help.

> You can run tmux inside tmux then use send-keys...

Hm, that's what I tried, though I ran into some trouble. Attempting to send a CSI sequence like this sends them individually:

```sh
tmux -Louter send-keys Escape ] 1 '\;' 5 Z
```

Sending something like `C-S-Tab` ends up arriving as just `Tab`. (this is all with extended-keys always)

nicm:
It is [ not ]?
~~~~
send Escape '[1;5Z'
~~~~

nicm:
~~~~
1667297808.084567 /dev/ttypa: read 6 bytes (already 0)
1667297808.084613 /dev/ttypa: keys are 6 (\033[1;5Z)
1667297808.084637 /dev/ttypa: next key is 6 (\033[1;5Z) (expired=0)
1667297808.084670 /dev/ttypa: keys in list: 0x40600000000009
1667297808.084720 /dev/ttypa: complete key \033[1;5Z 0x40600000000009
...
1667297808.085054 writing key 0x40600000000009 (C-S-Tab[E]) to %0
1667297808.085075 found key 0x40600000000009: "\033[1;5Z"
1667297808.085093 input_key: \033[1;5Z

aaronjensen:
Well that's embarrassing. I guess I should go back to sleep 😄 . Thanks, I can work w/ that. I found the problem w/ trying to send `C-S-Tab` as well, it builds the key but the extended flag isn't set so my code downgrades it to just Tab. I can reverse the usage of the extended flag and solve the issue, plus, I think it makes more sense.

tty-keys interprets the key to whatever it is. Then, input keys will translate that key as it is to either a regular key or an extended key based on mode.

```diff
diff --git a/input-keys.c b/input-keys.c
index 54047e5c..dc330878 100644
--- a/input-keys.c
+++ b/input-keys.c
@@ -508,9 +508,14 @@ input_key(struct screen *s, struct bufferevent *bev, key_code key)
 		key &= ~KEYC_KEYPAD;
 	if (~s->mode & MODE_KCURSOR)
 		key &= ~KEYC_CURSOR;
-	if (~s->mode & MODE_KEXTENDED)
-		key &= ~KEYC_EXTENDED;
-	ike = input_key_get(key);
+
+	if (s->mode & MODE_KEXTENDED)
+		ike = input_key_get(key|KEYC_EXTENDED);
+	else
+		ike = NULL;
+
+	if (ike == NULL)
+		ike = input_key_get(key);
 	if (ike == NULL && (key & KEYC_META) && (~key & KEYC_IMPLIED_META))
 		ike = input_key_get(key & ~KEYC_META);
 	if (ike == NULL && (key & KEYC_CURSOR))
diff --git a/tty-keys.c b/tty-keys.c
index b43329e1..38a5c2c2 100644
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -210,12 +210,12 @@ static const struct tty_default_key_raw tty_default_raw_keys[] = {
 	{ "\033[201~", KEYC_PASTE_END },
 
 	/* Extended keys (fixterms). */
-	{ "\033[1;5Z", '\011'|KEYC_CTRL|KEYC_SHIFT|KEYC_EXTENDED },
+	{ "\033[1;5Z", '\011'|KEYC_CTRL|KEYC_SHIFT },
 };
 
 /* Default xterm keys. */
@@ -957,7 +957,7 @@ tty_keys_extended_key(struct tty *tty, const char *buf, size_t len,
 		log_debug("%s: extended key %.*s is %llx (%s)", c->name,
 		    (int)*size, buf, nkey, key_string_lookup_key(nkey, 1));
 	}
-	*key = nkey|KEYC_EXTENDED;
+	*key = nkey;
 	return (0);
 }
```

aaronjensen:
Btw, stylistically, is something like this ok/preferred?

```c
	if (s->mode & MODE_KEXTENDED)
		ike = input_key_get(key|KEYC_EXTENDED) || input_key_get(key);
	else
		ike = input_key_get(key);
```

nicm:
That won't work because the result of an expression with `||` will be 0 or 1. I think your original change is fine but I will look at it again later.

aaronjensen:
Is there a way to make cat capture non printable characters like esc or nul? I couldn’t figure it out.

nicm:
`cat -v`?

aaronjensen:
I wanted to show you a preview of the first test file I'm putting together to get your thoughts:

```sh
#!/bin/sh

PATH=/bin:/usr/bin
TERM=screen

[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
TMUX="$TEST_TMUX -Ltest"
$TMUX kill-server 2>/dev/null
TMUX2="$TEST_TMUX -Ltest2"
$TMUX2 kill-server 2>/dev/null

TMP=$(mktemp)
trap "rm -f $TMP" 0 1 15

$TMUX2 -f/dev/null new -d || exit 1
$TMUX -f/dev/null new -d "$TMUX2 attach" || exit 1
sleep 0.1

# $TMUX send-keys Escape '[1;5Z'


assert_key () {
  key_base16="$1"
  control_name="$2"

  $TMUX2 command-prompt -k "display-message -p %%" > "$TMP" &
  $TMUX send-keys -H "$key_base16"

  wait

  name=$(cat "$TMP")

  key="0x$key_base16"

  if [ "$name" = "$control_name" ]; then
    echo "[PASS] $key -> $name"
  else
    echo "[FAIL] $key -> $control_name (Got: $name)"
    exit 1
  fi
}

assert_key 00 C-Space
assert_key 01 C-a
assert_key 02 C-b
assert_key 03 C-c
assert_key 04 C-d
assert_key 05 C-e
assert_key 06 C-f
assert_key 07 C-g
assert_key 08 C-h
assert_key 09 Tab
assert_key 0A C-j
assert_key 0B C-k
assert_key 0C C-l
assert_key 0D Enter
assert_key 0E C-n
assert_key 0F C-o
assert_key 10 C-p

$TMUX kill-server 2>/dev/null
$TMUX2 kill-server 2>/dev/null

exit 0
```

output:

```
❯ ./keys.sh
[PASS] 0x00 -> C-Space
[PASS] 0x01 -> C-a
[PASS] 0x02 -> C-b
[PASS] 0x03 -> C-c
[PASS] 0x04 -> C-d
[PASS] 0x05 -> C-e
[PASS] 0x06 -> C-f
[PASS] 0x07 -> C-g
[PASS] 0x08 -> C-h
[PASS] 0x09 -> Tab
[PASS] 0x0A -> C-j
[PASS] 0x0B -> C-k
[PASS] 0x0C -> C-l
[PASS] 0x0D -> Enter
[PASS] 0x0E -> C-n
[PASS] 0x0F -> C-o
[PASS] 0x10 -> C-p
```

Not sure what the norms are for the tests, if verbose output is OK. If not, I can hide it behind a flag. It's certainly useful for me. Thoughts?

nicm:
I have applied this other change now.

The test looks good, although when you are ready please make it silent for success.

aaronjensen:
```sh
## Double check this, tty-keys.c has these associated with S-F1 and S-F2, but
## tmux sees them as F11 and F12 - Aaron, Wed Nov 02 2022
assert_key_sequence 'Escape [23~' 'F11'
assert_key_sequence 'Escape [24~' 'F12'

## Double check this, tty-keys.c has these associated with S-F1 and S-F2, but
## tmux sees them as F11 and F12 - Aaron, Wed Nov 02 2022
assert_key_sequence 'Escape [23~' 'S-F1'
assert_key_sequence 'Escape [24~' 'S-F2'
```

tty-keys.c declares this in `tty_default_raw_keys`:

```c
	{ "\033[23~", KEYC_F1|KEYC_SHIFT },
	{ "\033[24~", KEYC_F2|KEYC_SHIFT },
```

But they are seen as F11 and F12. I can't find where that's happening. Any ideas? I know the two overlap in rxvt, but how should tmux treat them?

nicm:
It's probably kf11 from terminfo.

aaronjensen:
Ok, so that's this one:

```
	{ TTYC_KF11, KEYC_F11 },
```

Given that term is set to screen in my test, can I assume that that will hold and remove the S-F1?

nicm:
Yes if tmux is running in tmux or screen with TERM=screen it will expect it
to be F11.

aaronjensen:
Another question, most 

```
[PASS] 0x00 -> C-Space
[FAIL] \033 0x00 -> M-C-Space (Got: 'M-C-@')
[PASS] 0x01 -> C-a
[PASS] \033 0x01 -> M-C-a
```

What do you think we should expect C-M-Space to be? It sends `\033 0x00`, which seems to be both C-M-@ and C-M-Space depending on who you ask. It's not specifically addressed in fixterms, so I'm inclined to leave it to be considered `M-C-@`, even though that leaves us w/o a `M-C-Space`

nicm:
`key-string.c` has this code:
~~~~
        /* Display C-@ as C-Space. */
        if ((key & (KEYC_MASK_KEY|KEYC_MASK_MODIFIERS)) == 0)
                key = ' '|KEYC_CTRL;
~~~~
You could extend this to all modifiers.

nicm:
Is this ready to go?

aaronjensen:
I think so, it covers what I know about so far and is ready for the extended key work I’m going to do next. I also intend to add a similar one for input keys for both extended always and not. 

If you want to run this one to see if it works and apply it, please feel free. I have more coming, but it’s slow going.

aaronjensen:
Can you please take a look at `input-keys.sh` when you get a chance and let me know what I am doing wrong to get tmux into a mode where extended-keys is working? Also, if you could try running that test on your machine just to make sure I don't have any local quirks in it, I would appreciate it.

aaronjensen:
Figured it out, I needed to create a new window to make extended-keys take effect.

input-keys.c
```diff
@@ -316,7 +316,7 @@ static struct input_key_entry input_key_defaults[] = {
 	  .data = "\033[9;5u"
 	},
 	{ .key = '\011'|KEYC_CTRL|KEYC_SHIFT,
-	  .data = "\011"
```
aaronjensen, line 319: This was a mistake in my previous commit

aaronjensen:
Ok, both the input key and tty key tests are ready for a review. They are passing on my machine. There are some questions inline (`##` comments). This should be a good basis for additional work in the extended key realm, like adding support for `C-A` and such. I'll work on that next.

There's also a correction to the non-extended-key behavior of C-S-Tab in 28c81dd

aaronjensen:
Should we standardize on an order for the modifiers when printed? e.g. `C-M-S`? Right now, that's the order they are rendered in, but for meta + a standard control key, it's `M-C-a`, rather than `C-M-a`.

nicm:
Yes I think `C` first so `C-M-a`. I don't really mind whether we display `C-S-a` or `C-A` so long as it is consistent (accepting both might be nice at least for the obvious ones).

aaronjensen:
> but why don't you use stty raw instead?

Thanks for the pointer. I pushed a commit that uses it instead.

> Yes I think C first so C-M-a. I don't really mind whether we display C-S-a or C-A so long as it is consistent (accepting both might be nice at least for the obvious ones).

Copy that. I'll see what I can do about it accepting both when parsing key codes. 

I imagine we don't want to show `S-a` ever, but I think that would be represented as an unmodified key. For C-S-a, it probably makes sense to represent that internally as a `'a'|KEYC_CTRL|KEYC_SHIFT`, I think. 

Last time I was working on it, the problem I was running into is where exactly to translate things like 0x0f to C-o in tty-keys. My thinking is that tty-keys should emit that as `'o'|KEYC_CTRL`, rather than `0xf`, and then rely on input-keys to translate that back into `0xf`. I think that reduces the amount of special case code. Does that sound reasonable? If so, where would you recommend doing that translation in tty-keys?

nicm:
>  My thinking is that tty-keys should emit that as 'o'|KEYC_CTRL, rather than 0xf, and then rely on input-keys to translate that back into 0xf.

That is fine but bear in mind you will need to change `status.c`, `mode-tree.c`, `window-*.c`, `menu.c` and anywhere else that uses control keys.

You could probably map it at the end of the `first_key` section of `tty_keys_next` where the meta keys are handled (`if (*buf == '\033' && len >= 2) {`)

nicm:
I have applied these tests now, thanks! I will close this, can you open a new PR for any new changes? Thanks

Issue #3371, opened 2022-11-02
`tmux display-message -p '#{l:#}'` prints nothing


aaronjensen:
```sh
tmux display-message -p '#{l:#}'
```

As I understand it, it should print `#`, but it prints nothing. Doubling the `##` prints two of them. I don't know of a way to print just a `#`. This seems like a bug, so reporting just in case. 

Just found another case:

```sh
tmux display-message -p '#{l:%}'
```

Am I doing something wrong w/ the literal marker?

nicm:
`tmux display-message -p '##'`

aaronjensen:
Right, but in my instance I am dynamically printing single key presses, and ideally I can do that without having to write escape code. Is there a reason literal doesn’t work for this?

nicm:
I don't know why it doesn't print `#`s, needs debugging...

nicm:
Try this please: 
[x.diff.txt](https://github.com/tmux/tmux/files/9921167/x.diff.txt)

nicm:
Applied this now, thanks!

PR #3372, opened 2022-11-02
Add literal flag to display-message

aaronjensen:
Addresses #3371 

Adds a `-l` flag to `display-message` that causes it to print what is supplied literally. I think need this for my tests that use `display-message` to print key bindings.

This doesn't contain documentation as I'm not really familiar w/ the syntax of the docs or how to test them, so I didn't want to mess that up.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -5890,7 +5890,7 @@ The following keys are also available:
 .El
 .Tg display
 .It Xo Ic display-message
-.Op Fl aINpv
+.Op Fl aIlNpv
 .Op Fl c Ar target-client
 .Op Fl d Ar delay
 .Op Fl t Ar target-pane
@@ -5912,7 +5912,11 @@ is not given, the
 option is used; a delay of zero waits for a key press.
 .Ql N
 ignores key presses and closes only after the delay expires.
-The format of
+If
+.Fl l
+is given, the
+.Ar message
+is printed literally. Otherwise, the format of
 .Ar message
 is described in the
 .Sx FORMATS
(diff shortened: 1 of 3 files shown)


aaronjensen:
Also, please let me know if any of this can be improved... it's definitely getting out of my C comfort zone.

nicm:
This looks OK but I would make it much smaller and just do something like:
~~~~
-       msg = format_expand_time(ft, template);
+       if (args_has(args, 'l'))
+               msg = xstrdup(template);
+       else
+               msg = format_expand_time(ft, template);
+
~~~~
Also please add this to `tmux.1` and also the usage is sorted so `l` comes before `N`.

aaronjensen:
Thanks, I took a stab at the documentation, let me know if I can improve it. This is your changes broken into two commits. The literal flag works for me. The literal format block still doesn't, but it's improved.

tmux.1
```diff
@@ -5912,7 +5912,11 @@ is not given, the
 option is used; a delay of zero waits for a key press.
 .Ql N
 ignores key presses and closes only after the delay expires.
-The format of
+If
+.Fl l
+is given, the
```
aaronjensen, line 5917: I went back and forth on this "the".

nicm:
I have applied the `display-message` change now, thanks!

aaronjensen:
Thanks. Out of curiosity, do you intend to apply the format change as well?

nicm:
Did you test it?

aaronjensen:
Yes, `#{l:#}` still does not work, but I can understand why. I think `%` alone didn't work either (I'm guessing it's because date formatting may happen first?) But `##` at least got replaced w/ `#`, which is an improvement.

PR #3408, opened 2022-12-06
Add an auto split direction option for `split-window` command

key-moon:
Add the option to determine the direction to split based on the current aspect ratio of the pane. The aspect ratio can be configurated by `pane-autosplit-threshold` option, with a default value of `2:1`.
screenshot

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -39,8 +39,8 @@ const struct cmd_entry cmd_split_window_entry = {
 	.name = "split-window",
 	.alias = "splitw",
 
-	.args = { "bc:de:fF:hIl:p:Pt:vZ", 0, -1, NULL },
-	.usage = "[-bdefhIPvZ] [-c start-directory] [-e environment] "
+	.args = { "abc:de:fF:hIl:p:Pt:vZ", 0, -1, NULL },
+	.usage = "[-abdefhIPvZ] [-c start-directory] [-e environment] "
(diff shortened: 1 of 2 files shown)


nicm:
Thanks. I don't think this should have a new option, instead either default to some sensible value, or add an argument to `-a` which gives the ratio.

key-moon:
I believe that `2:1` would be a sensible value in most cases, but I thought it would be better to allow for flexibility as it varies depending on the font and line spacing settings. Since the sensible ratio rarely changes from command to command, I thought that specifying it as an option would avoid the redundancy of specifying the same value multiple times in the argument. This is the reason why I implemented in this way.

Below is the pros/cons of the each method:

- Add option
  - pros: concise command
  - cons: unnecessary option
- Fix at a sensible value
  - pros: concise command and option
  - cons: lack flexibility
- Take from argument
  - pros: most flexible
  - cons: verbose command

I don't have a strong opinion about this, so I will implement it in as you like.

nicm:
There are too many options already and I don't want to add ones that hardly anyone will use.

TBH does this really need to be builtin? What about doing this:
~~~~
run -C 'splitw #{?#{e|<=:#{e|*:#{pane_height},2},#{pane_width}},-h,-v}'

key-moon:
Actually why I submit this PR is that I didn't manage to implement this behavior in the existing feature. So, this PR is no longer needed now. Thanks for letting me know!

PR #3421, opened 2023-01-04
Add a simple `echo` command to tmux

tydar:
This patch adds a simple command `echo` to tmux. The new command is intended to demonstrate the basic steps in adding a new command to tmux.


tydar:
Woops, meant to create the PR on my local fork :)

PR #3431, opened 2023-01-10
Do not interpret keys when bracket pasting

oandrew:
Do not interpret input keys during a bracketed-paste sequence and simply forward them instead.

Time based `server_client_assume_paste` is not guaranteed to work in presence of connection issues, etc. It also **requires minimum two characters**. 

For example, I use tmux with `set -g prefix2 '\'`  and bash with bracketed paste enabled (`set enable-bracketed-paste on` in `~/.inputrc`).
Starting with empty bash prompt `$ `:
Without this patch: paste literal `\abc` -> bash prompt is `$ abc`. tmux **is left in prefix mode**.
With this patch: paste literal `\abc` -> bash prompt is `$ \abc` (as expected).   tmux **is not** left in the prefix mode


nicm:
Looks good, but I think this needs a time limit in case the terminal never sends an end key. Maybe 10 seconds?

nicm:
Also the flag should be on the client not the session or it will get confused if two people paste into tmux at the same time.

oandrew:
> Looks good, but I think this needs a time limit in case the terminal never sends an end key. Maybe 10 seconds?

I think timeout on tmux side is tricky as it can leave applications (e.g. bash) in inconsistent state, unless we also inject end key from tmux which I don't think is a good idea either.

In addition, bash buffers all input (without printing anything to screen) until end key is seen. 

> Also the flag should be on the client not the session or it will get confused if two people paste into tmux at the same time.

Fixed

nicm:
We need a timeout so that if the terminal doesn't send an end for whatever reason, people are not stuck without being able to issue tmux commands. I don't think it really matters than much if programs inside tmux could become inconsistent (they can have their own timeout).

oandrew:
Sounds good, added timeout (10s).

nicm:
Thanks but I have thought about it again and I think we can leave out the timeout - the user can always kill the client.

However, tmux only turns on bracketed pasting when the application inside does, which means you don't always get the benefit. Please take a look at this change which includes yours but also makes tmux always turn on bracketed paste when the terminal supports it (it also removes the timeout again): [paste.diff.txt](https://github.com/tmux/tmux/files/10401189/paste.diff.txt)

oandrew:
Yeah I was thinking about the same fix too. 

To clarify I understand it correctly:
- tmux always requests terminal to enable Paste Bracketing (if supported) regardless of current running app
- when pasting, tmux will forward KEYC_PASTE_START and KEYC_PASTE_END only if current application has requested Paste Bracketing

This should be even better.

I tested it and it seems to work - I can paste 
```
\c
\c
``` 
into `/bin/cat` (no paste-bracketing) or bash/vim (paste-bracketing) without tmux interpreting `\c`.

oandrew:
Updated. 

Includes a tiny change in `server_client_is_bracket_pasting` to `return (1)` after turning it off so that `PasteEnd` is forwarded immediately.

nicm:
Great - I have applied this to OpenBSD now, it should be in GitHub soon when it syncs. Thanks!

Issue #3456, opened 2023-01-29
[3.3a] Bind `?` to `lsk -Na` by default rather than `lsk -N`?


clarkwang:
### Issue description

The default `bind ? lsk -N` only shows key bindings with notes. Does it make sense to `bind ? lsk -Na` by default?

nicm:
No. It is intended to show useful help and not bewilder new users with commands.

clarkwang:
I myself is a new tmux user and this is how the current behavior bewidered me:

* I press `C-b ?` and it tells me `C-b ?` is for "List key bindings" but I cannot find my own key bindings.
* `C-b ?` also indicates there's `C-b /` to "Describe key binding". This sounds cool but it does not work either (issue #3438).

So to not bewilder new users, we better update the `C-b ?` help messages:

* "List key bindings" -> "List key bindings which are defined with `bind -N`"
* "Describe key binding" -> "Describe key binding which is defined with `bind -N`"

PR #3459, opened 2023-02-01
Updated notes for key bindings '?', '/' and 'D'

clarkwang:
For https://github.com/tmux/tmux/issues/3456


nicm:
Thanks, I took the `D` one but I don't think the other ones are an improvement.

Issue #3449, opened 2023-01-25
Control mode signalling


flowchartsman:
### Issue description

I'm writing a (Go) persistent plugin framework for Tmux using control mode with the following workflow:

1. command in tmux.conf runs `run -b <framework binary> spawn`
2. binary checks for session `__framework`
3. if not found, calls `tmux new -d -E -c $TMPDIR -s __framework -n __framework <framework binary> control`
4. new binary instance is now in its own session and spawns control-mode listeners for every existing session along with every new session (cleaning then up as necessary). It sends events from each session back to the central `__framework` session by using a shared channel. This allows a tiered approach where each session gets a listener that can dispatch tasks back to the central listener.

The idea being that the framework binary remains resident and can gather state and external information and explicitly create popups containing tools or menus in response to user requests.It works great, actually, except that there doesn't seem to be a way for the control mode client to receive any kind of meaningful message from the user client that it can use to determine what to display and where.

What I would like to be able to do is have a keybinding in the user sessions that signals the persistent control worker somehow so that it knows they want a popup for searching history versus, say a popup menu to configure some value. But I can't figure out any way for any client to reliably send a message or cause a control mode event that is meaningful. Is there any way whatsoever to do this? If not, could we get one? I feel like control mode has a lot of possibility for nifty stuff like this.

Popups also require referencing the client directly, and getting the origin client/session/window/pane is actually quite difficult, though I believe I can work around that. I just need an event that can somehow tell the controller what functionality to invoke.

But perhaps I'm missing something?

### Required information

Please provide the following information:

* tmux version (`tmux -V`). **3.3a**
* Platform (`uname -sp`).  **Darwin-Arm**
* $TERM inside and outside of tmux (`echo $TERM`)- outside: **xterm-256color**  inside: **screen-256color**

nicm:
You could try using a format subscription. So have your key bindings modify a user option and then the control client subscribe to for example `#{@myvar}`.

nicm:
https://github.com/tmux/tmux/wiki/Control-Mode#format-subscriptions

nicm:
You may find format subscriptions are too slow or that you need a global one (which we don't have but could possibly add). The other option without changing tmux would be to use an out-of-band mechanism to signal, so have your program also listen on a named pipe, Unix domain socket or poll a file.

flowchartsman:
[...]

PR #3465, opened 2023-02-05
Add filter (-f) capability to list-clients

flowchartsman:
Ref: #3449
Ref: 756591b


nicm:

Issue #3464, opened 2023-02-05
`set-buffer -n` should conveniently override the buffer if it already exists, same as without `-n`


Hexstream:
* tmux version (`tmux -V`): `tmux next-3.4` (built from latest commit `493922dc4b150218ab35fbbb8c5cf6a90fc4f24c` from 2 february 2023 with no patches applied)
* Platform (`uname -sp`): `Linux unknown` (Debian 11)
* $TERM inside and outside of tmux (`echo $TERM`): `tmux-256color` inside, `xterm-256color` outside.
* Logs from tmux (`tmux kill-server; tmux -vv new`): Please use provided `menu.conf`.

Hello,

I am using a [scheme](https://github.com/Hexstream/status-quo.hexstream.expert/blob/91ad42bb04cf872074fba5b099318259090fd241/assets/tools/tmux/config/main.conf#L51) whereby instead of using "automatic buffers", I basically use emacs-style "registers" (buffers named by one letter).

I need to easily override these registers/buffers with new values from existing (automatic and named) buffers, but unfortunately, `set-buffer -n` does not just conveniently override the existing value as `set-buffer` without `-n` does, and instead complains that the buffer already exists.

The [workaround](https://github.com/Hexstream/status-quo.hexstream.expert/commit/e586b634fb13b378ec9d63cf066f0cfd66bcf4eb) is to first explicitly delete the buffer if it already exists, but this is a bit painful.
(Perhaps `delete-buffer` should accept a `-q` flag for just doing nothing if the buffer does not exist?)

Please try this `menu.conf` with `tmux -f menu.conf`:

```
set -s @mnemonic-style 'bold'

bind m menu -T 'Test menu' -x P -y P -S - \
'#[#{@mnemonic-style}]S#[default]et multiple times' s {
    set-buffer -b 'b' 'Named 1'
    set-buffer -b 'b' 'Named 2'
    paste-buffer -b 'b'
} \
'#[#{@mnemonic-style}]R#[default]ename multiple times' r {
    set-buffer 'Auto 1'
    set-buffer -n 'b'
    set-buffer 'Auto 2'
    set-buffer -n 'b'
    paste-buffer -b 'b'
}
```

"Set multiple times" (`C-b m s`) works, but "Rename multiple times" (`C-b m r`) does not.
An error message says that buffer `b` already exists.

From the user's perspective, renaming buffer `source` to `destination` should be treated the same as setting the value of buffer `destination` to `source`'s value and then deleting buffer `source`.

[...]

PR #3467, opened 2023-02-07
Make "set-buffer" with "-n" also conveniently override existing buffer, if any

Hexstream:
Per #3464.

This trivial fix seems to work, although I am not confident about correctness.


Hexstream:
Oh, and congratulations on 10000 commits! :)

nicm:

Hexstream:
Awesome, thanks!

PR #3471, opened 2023-02-11
✨(format): Add `pane_current_command_pid` format variable for macOS

vegerot:
✨(format): Add `pane_current_command_pid` format variable for macOS

Summary: Tmux currently has format variables for:
- `pane_pid`: _pid_ of the _first_ pane process
- `current_command_name`: _name_ of the _current_ pane process

But there's no way to get the _pid_ of the _current_ pane process.  This would
be helpful for something like getting the full name of the current process:

```
#(ps -p #{???} -o command=)
```

This commit adds the `pane_current_command_pid` format variable to do exactly
that.

Test Plan:
```sh
$ tmux
$ nvim tmux.c # or any other long-running command of your choosing
# run tmux exec command
:display-message '#(ps -p #{pane_current_command_pid} -o command=)'
nvim tmux.c
```

Question for reviewers:
- should we add an additional format specifier called
`pane_current_command_full_name` or something that does what I just described?
- I know this patch will likely get rejected due to only supporting macOS.  I
will accept patches to add support for other platforms, and will add the
patches myself as I use other platforms in my day-to-day

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/tmux/tmux/pull/3471).
* __->__ #3471

### DIFF
--- a/format.c
+++ b/format.c
@@ -840,6 +840,24 @@ format_cb_current_command(struct format_tree *ft)
 	return (value);
 }
 
+// relies on osdep_get_pid which is only implemented for macOS
+#ifdef __APPLE__
+/* Callback for pane_current_command_pid */
+static void *
+format_cb_current_command_pid(struct format_tree *ft)
+{
+	struct window_pane	*wp = ft->wp;
+	char			 *value;
+
+	if (wp == NULL || wp->shell == NULL)
+		return (NULL);
+
+	pid_t pid = osdep_get_pid(wp->fd);
+	asprintf(&value, "%d", pid);
+	return value;
+}
+#endif
+
 /* Callback for pane_current_path. */
 static void *
 format_cb_current_path(struct format_tree *ft)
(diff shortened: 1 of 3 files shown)


nicm:
Thanks, but it seems to me that this is only useful to pass to a shell script and for that you could use `pane_tty` instead and give it to `ps -t`, no?

Also you are right this will need support for at least Linux and OpenBSD as well.

vegerot:
Thanks for the suggestion!  I didn't notice `pane_tty` 😁 

So I can simply replace

```tmux
set -g window-status-current-format '#I:#(ps -p #{pane_current_command_pid} -o command=)#{?window_flags,#{window_flags}, }'
```

with

```tmux
set -g window-status-current-format '#I:#(ps -t #{pane_tty} -o command= | tail -n1)#{?window_flags,#{window_flags}, }'
```
?

The problem with this is that my `ps -t` output looks like

```sh
❯ ps -t /dev/ttys000  -o "command="
-zsh
-zsh
-zsh
-zsh
/Users/m0c0j7y/.cache/gitstatus/gitstatusd-darwin-x86_64 -G v1.5.1 -s -1 -u -1 -d -1 -c -1 -m -1 -v FATAL -t 32
ps -t /dev/ttys000 -o command=
```

so my status bar doesn't look right when I'm just sitting in my shell.

image

it does do what I want for programs like vim 

image
.

Of course, I can continue to use `pane_tty` and make my logic more complicated, but `pane_current_command_pid` let me keep my logic simpler

nicm:
Try doing `pgid,command` and look for the command with `#{pane_pid}` as the process group ID, that's basically what tmux does.

PR #3496, opened 2023-03-15
Add Enter-to-confirm option (-e) to confirm-before command

martian17:
I wanted `confirm-before` command to accept other list of characters than the default 'y'. I thought of adding an option that can specify an arbitrary set of characters, but that would probably add unnecessary complexity to the code, and I didn't want to risk adding too much into an unfamiliar codebase.
So I resorted to adding an option to accept [Enter] key as confirmation instead, since that was the effect that I personally wanted to achieve.
As for the things I modified, I modified _cmd-confirm-before.c_ to accept the new (-e) argument option, and a new entry in the `cmd_confirm_before_data` struct so the callback will have access to what the new argument option.  
I also had to modify _status.c_ because it was sending `"\0"` to the `prompt_inputcb` when [Enter] key was pressed. Now, given the `key` is `'\r'` or `'\n'`, and the `prompt_buffer` string is null, it will always send `"\n"`.

### DIFF
--- a/cmd-confirm-before.c
+++ b/cmd-confirm-before.c
@@ -41,8 +41,8 @@ const struct cmd_entry cmd_confirm_before_entry = {
 	.name = "confirm-before",
 	.alias = "confirm",
 
-	.args = { "bp:t:", 1, 1, cmd_confirm_before_args_parse },
-	.usage = "[-b] [-p prompt] " CMD_TARGET_CLIENT_USAGE " command",
+	.args = { "bep:t:", 1, 1, cmd_confirm_before_args_parse },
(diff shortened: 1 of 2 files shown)


nicm:
Thanks but I would prefer a more general solution where you give an argument to `-e` with the key which should confirm. I don't see why it would be much different from yours?

I also don't think you should make `status_prompt_key` fire the callback with a modified string because this could break uses of `command-prompt`. `confirm-before` should treat the empty string as `Enter` itself.

You could pass the argument through `key_string_lookup_string` so you could use for example `-e Enter`.

PR #3498, opened 2023-03-16
Add 'pane_unseen_input' format

da-x:
This commit introduces a new pane_unseen_input format. When a pane has seen input while it was in some mode, it sets the flags to PANE_UNSEEN_INPUT. It is cleared when exiting the mode.

### DIFF
--- a/format.c
+++ b/format.c
@@ -1885,6 +1885,18 @@ format_cb_pane_input_off(struct format_tree *ft)
 	return (NULL);
 }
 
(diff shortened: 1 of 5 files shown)


nicm:
Code looks good, but I think this would be better described as output not input. Or maybe just "unseen_changes" or "changes_in_mode"?

da-x:
Riding with `unseen_changes`

Issue #3505, opened 2023-03-23
Feature request: copy-* commands raw/html mode


FractalBoy:
### Feature request

Currently all of the copy-* commands copy just the text that is highlighted, stripping any attributes. I am requesting a feature of a new argument to those commands that allows you to pass a format to the command. The ideas I have right now are:

* plain or text (functions like today)
* raw (prints ANSI control sequences)
* html (formats as HTML, using either a stylesheet or inline styles)
* rtf (formats as RTF)

The RTF one is probably unnecessary but I thought I'd include it. At the very least I would like the "raw" format.

This would allow easy copy/paste of data including any colors or other attributes.

I took a look at window-copy.c, but it seems like I would need to create some sort of virtual tty to write the grid cells to, and then somehow copy off of that in order to implement "raw" format.

For HTML, it could open a tag (such as &lt;span&gt;) every time a grid cell has a different attribute than the previous one, and apply a different style. However, this probably requires some sort of stack to ensure we correctly open and close tags, and may also be complicated.

Thoughts?

nicm:
Not keen on HTML or RTF.

You can use `grid_stringify_cells` to generate ANSI in the same way as `capture-pane` does. It'd probably be best to add new commands (`copy-pipe-annotated`? `formatted`? not sure) since flags to copy mode commands don't work that well.

FractalBoy:
Oh, cool, I thought everything had to go through `tty_attributes`.

`copy-pipe-raw`?

FractalBoy:
Did you mean `window_copy_stringify`? I can't find `grid_stringify_cells`.

nicm:
No, you are just going from grid -> string. `window_copy_copy_line` does not currently use `grid_string_cells` but I don't see why it couldn't use it (maybe with some changes).

`-raw` suffix would be fine.

nicm:
Sorry it is `grid_string_cells`.

FractalBoy:
Thanks. I'll take a stab at this then.

PR #3507, opened 2023-03-24
Create new raw copy/pipe/append commands

FractalBoy:
Addresses #3505.

This is not yet complete, but I am creating a pull request just to make sure I'm staying on the right track.

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -5040,24 +5040,12 @@ window_copy_copy_line(struct window_mode_entry *wme, char **buf, size_t *off,
 	struct grid_cell        *lastgc = NULL;
 	struct grid_line		*gl;
 	struct utf8_data		 ud;
-	u_int				 i, xx, wrapped = 0;
-	const char			*s;
-	char				*raws;
-	u_int				 rawlen;
+	u_int				 i, xx, wrapped = 0, rawlen;
+	const char			*s, *raws;
 
 
 	if (sx > ex)
 		return;
-
-	if (raw)
-	{
-		raws = grid_string_cells(gd, sx, sy, (ex - sx + 1), &lastgc, GRID_STRING_WITH_SEQUENCES, wme->screen);
-		rawlen = strlen(raws);
-		memcpy(*buf + *off, raws, rawlen);
-		*off += rawlen;
-		free(raws);
-		return;
-	}
 
 	/*
 	 * Work out if the line was wrapped at the screen edge and all of it is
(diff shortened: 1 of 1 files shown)


nicm:
Looks sensible to me so far.

FractalBoy:
The actual copying does not seem to be working - can you provide any guidance on the correct arguments to pass to `grid_string_cells`?

FractalBoy:
No, I get a buffer missing one character with no escape sequences, so it seems like maybe I need to add 1 somewhere.

The fact that I'm missing a flag explains the missing escape sequences.

nicm:
Well I don't mind having `copy-escape-sequences` option is that less useful than different commands?

nicm:
Are you still working on this? Thanks.

FractalBoy:
Yes, just haven't had a chance to work on it in a bit.

nicm:
OK no worries.

FractalBoy:
Okay, this seems to mostly work now, but I'm still having an issue getting the final character of my string.

For example, when I do `printf "\033[01;32mtest\033[00m\n"`, then highlight the resulting string, I get this on my clipboard:

`\033[1m\033[32mtes`

I would expect to get the `t` at the end of the word, and it would be nice to get the closing sequence but I don't think that's possible since I'm not highlighting past that.

nicm:
I am not seeing this problem, when I copy your example, I get:
```
$ tmux lsb
buffer3: 13 bytes: "\033[1m\033[32mtest"
```
Which looks fine to me. How are copying?

For the closing sequence you could add a flag to make `grid_string_cells` emit a reset at the end of the string.

It would be nice if `window_copy_copy_line` used `grid_string_cells` for the normal case as well (but that might need a flag to map ACS characters). You don't need to worry about that, we can do that separately.

FractalBoy:
I was getting it through OSC 52 - perhaps my terminal (Windows Terminal) isn't handling it well then.

nicm:
I think it is a bug in your terminal because when I try OSC 52 (although I need a fix to make it work at all with `send -X` rather than a key binding), it has the complete text.

nicm:
I think this is OK but I have been unwell so I haven't been able to look at it again, I'll try and get to it in the next couple of weeks.

FractalBoy:
I haven't quite finished it - still much documentation to do, and I would still like a way to change the default for all copy-* command to "raw" mode. Haven't had time to work on it lately.

CleanSl8 (approved):

nicm:
I'm going to close this but do comment again if you get back to it and I will reopen.

PR #3508, opened 2023-03-25
man: clarify the section on `new-session -A/-D/-X`

acid-bong:
The docs weren't clear, whether the -D and -X flags should follow -A or if they already imply that -A is given. This edit covers the real behaviour (that -A is still required)

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -1281,18 +1281,17 @@ behave like
 .Ic attach-session
 if
 .Ar session-name
-already exists; in this case,
+already exists; in this case, additional
 .Fl D
-behaves like
-.Fl d
-to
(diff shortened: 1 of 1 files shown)


nicm:
It seems obvious that "in this case" means "the case where the -A flag is used", what else could it mean? But maybe we could just do this:
~~~~
--- a/tmux.1
+++ b/tmux.1
@@ -1281,7 +1281,10 @@ behave like
 .Ic attach-session
 if
 .Ar session-name
-already exists; in this case,
+already exists;
+if
+.Fl A
+is given,
 .Fl D
 behaves like
 .Fl d

nicm:
Applied now, thanks!

PR #3509, opened 2023-03-26
man: clarify usage of '&&' and '||' operators

acid-bong:
The manual said that '&&' and '||' work with two arguments, but they can actually can take three and more

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -4889,7 +4889,7 @@ otherwise by
 .Ql ||
 and
 .Ql &&
-evaluate to true if either or both of two comma-separated alternatives are
+evaluate to true if either or all of the comma-separated alternatives are
 true, for example
 .Ql #{||:#{pane_in_mode},#{alternate_on}} .
 .Pp


nicm:
No... I'm not sure you tested this quite correctly... :-)
~~~~
$ tmux display -p '#{&&:1,1}'
1
$ tmux display -p '#{&&:1,1,0}'
1
$ tmux display -p '#{||:0,0}'
0
$ tmux display -p '#{||:0,0,0}'
1
~~~~

acid-bong:
I tested it differently: I wanted specific sessions to have their status bars off, so (with the help of the IRC members) I wrote a hook for it:
```tmux
set-hook -g after-new-session 'if -F "#{||:#{==:#S,aerc},#{==:#S,poezio}}" "set status off"'
```
I added the || operator, cuz if i had multiple lines like
```tmux
set-hook -g after-new-session 'if -F "#{==:#S,foo}" "set status off"'
set-hook -g after-new-session 'if -F "#{==:#S,bar}" "set status off"'
set-hook -g after-new-session 'if -F "#{==:#S,baz}" "set status off"'
```
it would only run the hook for the last declared session name (`baz` in this case).

Since the manual said that || works with only two expressions, I was confused about how to expand this (without nesting them like `(foo OR bar) OR baz`), but I tried it with three in a row: 
```tmux
set-hook -g after-new-session 'if -F "#{||:#{==:#S,aerc},#{==:#S,poezio},#{==:#S,bosh}}" "set status off"'
```
and it worked as I wanted: status was off for sessions called aerc, poezio and bosh and on (default value) for the rest

Issue #3183, opened 2022-05-09
tmux might try to put each pty/pane in a separate cgroup


mdengler:
From @benzea's comment in #3024:

> we want to separate each of the pty's into a separate cgroup. The purpose is so that when the OOM killer (i.e. systemd-oomd) kills an entire cgroup, only the processes for one pty are killed rather than also including tmux itself and all other pseudo terminals.
> [... it is actually solvable entirely without dbus ...]

_Originally posted by @benzea in https://github.com/tmux/tmux/issues/3024#issuecomment-1004166931_

I would love to help with this, as my tmux sessions are getting killed from time to time a la https://bugzilla.redhat.com/show_bug.cgi?id=1933494 .  I'm happy to help implement or just test anything.

I wanted to open this issue to continue the momentum from #3024.

mdengler:
I see the code in https://gitlab.gnome.org/GNOME/vte/-/blob/master/src/systemd.cc but not sure how to implement that, or if that's the right thing anyway, given @benzea mentioned "without dbus"...

nicm:
You can do pretty much whatever you like if you add a compat/forkpty-systemd.c or add appropriate minimal calls into functions in compat/systemd.c. tmux has a configure flag to link against systemd, but I don't want to link against dbus.

You can't copy VTE code because it is C++ and LGPL.

benzea:
Hm, don't remember what I was referring to with "without dbus". What comes to mind right now is:
 * Launching using systemd-run (which can also allocate a pty)
 * Using cgroup delegation and managing it yourself (requires being a systemd service in the first place though)

But, we are linking against systemd. So using dbus through the systemd API is trivial.

nicm:
Using dbus through systemd API is grand, I don't really want to add an explicit dependency on libdbus as well, but even then I don't mind too much so long as its still under `--enable-systemd`.

nicm:
Running `systemd-run` with fork/exec or system is gross and not OK though :-).

mdengler:
Understood about `systemd-run` via fork/exec; FWIW now I have something like this is my `.tmux.conf`:

```
...
new-window -d         -n mutt      "exec systemd-run --user  --pty --scope  mutt"
new-window -d         -n emacs     "exec systemd-run --user  --pty --scope  emacs -nw"
new-window -d         -n lab       "exec systemd-run --user  --pty --unit=jupyter-user --slice=jupyter.slice poetry run jupyter lab"
[...]

Issue #3024, opened 2022-01-03
systemd integration (for systemd-oomd primarily)


benzea:
Some distributions (e.g. Fedora) are starting to use systemd-oomd by default. Doing so means that the userspace OOM killer will always kill an entire cgroup rather than just a single process. This has many advantages, e.g.:
 * fork bombs can be easily detected and killed
 * resources can be distributed more fairly
 * better cleanup, as all processes belonging to the cgroup can be killed together

To better integrate this, tmux should launch each pseudo terminal in a separate cgroup. This can easily be done by moving the new PID into a separate transient systemd scope (requires a single DBus call to move the PID and ideally waiting for this to complete/error before `exec`'ing the child).

For reference, this is what VTE (the GNOME terminal library) does: https://gitlab.gnome.org/GNOME/vte/-/blob/master/src/systemd.cc

nicm:
I revisited this fairly recently, see this comment (including the diff) and the subsequent comments in the issue: https://github.com/tmux/tmux/issues/428#issuecomment-623491932

The consensus was that this was now pointless since either affected systems have by now worked around these problems, or it was impossible for tmux to sanely configure it anyway.

I don't use systemd at the moment, so I don't intend to do any further work on this myself for now, but if this kind of change (see the diff in the linked comment) now makes sense and you would like to open a PR then I'll take a look at it.

benzea:
Ah, hmm, maybe the whole systemd integration needs some more thought overall.

In general, the issue here is a different one though (and, it is actually solvable entirely without dbus). Here we want to separate each of the pty's into a separate cgroup. The purpose is so that when the OOM killer (i.e. systemd-oomd) kills an entire cgroup, only the processes for one pty are killed rather than also including tmux itself and all other pseudo terminals.

nicm:
OK we can do whatever you like in a new `compat/forkpty-linux.c` or `compat/forkpty-systemd.c`, so long as new dependencies on nonstandard libraries (ie, not libc) are optional.

benzea:
Cool, should be easy enough to do something within those constraints!

nicm:
I am closing this but please feel free to open another issue or a PR if you have anything further. Thanks!

PR #3514, opened 2023-03-30
Add support for spawning panes in separate cgroups

yut23:
This PR moves newly-spawned panes into their own cgroups via transient systemd transient scopes, as discussed in #3024 and #3183. It uses the dbus API provided by libsystemd, so no new libraries are needed.

Some questions that need to be addressed before merging:
* When should this be enabled: always, with a build flag, or with a server option?
* What should we do if the call fails? (by default, VTE just prints a debug message and continues as normal)
* If the child process exits before the dbus call goes through, the call will still succeed, but the unit will fail and won't be cleaned up by systemd. VTE avoids this by including a drop-in config file that says to remove the unit even if it fails (`CollectMode=inactive-or-failed` in `/usr/lib/systemd/user/vte-spawn-.scope.d/defaults.conf`). We could set that option directly in the dbus call, but putting it in a config file lets the end users/packagers override it if they want to inspect any errors. Do we want to include such a file with tmux?

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -425,6 +425,7 @@ void		*recallocarray(void *, size_t, size_t, size_t);
 /* systemd.c */
 int		 systemd_activated(void);
 int		 systemd_create_socket(int, char **);
+int		 systemd_move_pid_to_new_cgroup(pid_t, char **);
 #endif
 
 #ifdef HAVE_UTF8PROC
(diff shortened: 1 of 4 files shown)


nicm:
I can't confirm if the systemd bits work but it all looks good to me. Applied now with one minor fix (free *cause in spawn.c on failure) and some minor style nits (spacing and tidy comments). Thanks!

PR #3527, opened 2023-04-13
Default to vlock on Linux

jwboyer:
The default lock command is 'lock -np', which isn't available on Linux.  The standard answer is to set the lock command in a conf file to 'vlock', but that requires the user to do this manually.

This tests if vlock is available in the build environment so Linux distributions can optionally package tmux with a requirement on vlock and have the command work by default.


nicm:
Sorry forgot about this. I've applied it now, thanks!

PR #3548, opened 2023-04-26
Add a flag in confirm-before to select the confirmation key

EliasA5:
Hi, I wanted to be able to confirm the prompt of `confirm-before` without entering any character (only hitting enter), I then saw [pr #3496](https://github.com/tmux/tmux/pull/3496), that implements the behavior but only for enter key, so I implemented a more general solution per [nicm comment](https://github.com/tmux/tmux/pull/3496#issuecomment-1484736723).

I am still not sure of a couple of things:
- where should we place the `DEFAULT_CONFIRM_KEYCODE`, and if there is a macro for the keycode of 'y' that we can use instead.
- Do we still need the check of `s[1] != '\0'` in `cmd_confirm_before_callback`?

### DIFF
--- a/cmd-confirm-before.c
+++ b/cmd-confirm-before.c
@@ -52,8 +52,8 @@ const struct cmd_entry cmd_confirm_before_entry = {
 struct cmd_confirm_before_data {
 	struct cmdq_item	*item;
 	struct cmd_list		*cmdlist;
-    u_char              confirm_key;
-    int                 default_yes;
+        u_char                  confirm_key;
+        int                     default_yes;
 };
 
 static enum args_parse_type
@@ -71,7 +71,7 @@ cmd_confirm_before_exec(struct cmd *self, struct cmdq_item *item)
 	struct client			*tc = cmdq_get_target_client(item);
 	struct cmd_find_state		*target = cmdq_get_target(item);
 	char				*new_prompt;
-    const char          *confirm_key_ptr;
+        const char                      *confirm_key_ptr;
 	const char			*prompt, *cmd;
 	int				 wait = !args_has(args, 'b');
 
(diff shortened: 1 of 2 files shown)


nicm:
This is on the right lines, thanks. Some comments:

- I don't see any need for `DEFAULT_CONFIRM_KEYCODE`, just use the character constant `'y'` in the one place it is used.

- If the user gives an invalid key to `-c`, I think it should report an error. It should only fall back to `'y'` if the user does not give `-c` at all.

- This will not work the way you intend in `cmd_confirm_before_callback`. There is not much point in calling `key_string_lookup_string` on `s`. The only special key it can return is `Enter` and for that it will be an empty string. Instead you should do something like:
~~~~
if (s == NULL)
    goto out;
if (*s == '\0')
    key = KEYC_ENTER;
else if (*s > 31 && *s < 127)
     key = *s;
else
     goto out;
~~~~

- Please use tabs for indentation and keep lines under 80 columns.

nicm:
That would work but I think it should be: `if (*s != confirm_key && (*s != '\0' || !default_yes))`

EliasA5:
> That would work but I think it should be: `if (*s != confirm_key && (*s != '\0' || !default_yes))`

ah you are totally right :)
apparently tabbing settings aren't so straight forward in neovim.

the code should be done, man pages will need an updating.

nicm:
OK this looks and works fine apart from a few minor spacing things I can fix before commit.

Can you add the new flags to tmux.1?

EliasA5:
updated tmux.1 with the new flags.

nicm:
Looks good, applied to OpenBSD now with minor tidying, will be in GitHub later. Thanks!

PR #3586, opened 2023-06-03
fix configure error: utf8proc not found

dindinw:


nicm:
Applied now with some minor tidying. Thanks!

PR #3588, opened 2023-06-07
create list of last visited panes for each window

bertschingert:
This commit changes the "last-pane" feature to use a stack-like list of the visited panes in each window. This is so that it is easy to use the last-pane command to visit previously visited panes even when the most recently visited pane is deleted.

This is partially addressed by https://github.com/tmux/tmux/pull/1377 (0c94c3fbee3df39293de8b7fd957392251113a74).
That commit is helpful but I use layouts with 4+ panes frequently enough that I run into this when there are 3+ panes after deleting one.

A limitation of the solution here is that panes that are created without being visited are not added to the stack.
This is addressed for the 2-pane case by leaving in the solution from PR 1377.
While this commit does not address the issue for 3+ panes, this is not addressed for the last-window stack either so I think this acceptable.
A complete solution could add panes to the bottom of the stack if they are created without being visited (although that would not be very stacklike ;)). This could work for windows too.

### DIFF
--- a/cmd-find.c
+++ b/cmd-find.c
@@ -582,7 +582,7 @@ cmd_find_get_pane_with_window(struct cmd_find_state *fs, const char *pane)
 
 	/* Try special characters. */
 	if (strcmp(pane, "!") == 0) {
-		fs->wp = fs->w->last;
+		fs->wp = TAILQ_FIRST(&fs->w->last_visited_panes);
 		if (fs->wp == NULL)
 			return (-1);
 		return (0);
--- a/cmd-select-pane.c
+++ b/cmd-select-pane.c
@@ -98,7 +98,11 @@ cmd_select_pane_exec(struct cmd *self, struct cmdq_item *item)
 	struct options_entry	*o;
 
 	if (entry == &cmd_last_pane_entry || args_has(args, 'l')) {
-		lastwp = w->last;
+		lastwp = TAILQ_FIRST(&w->last_visited_panes);
+		/*
+		 * check for no last pane found in case the other pane
+		 * was spawned without being visited (e.g., split-window -d)
+		 */
 		if (lastwp == NULL && window_count_panes(w) == 2) {
 			lastwp = TAILQ_PREV(w->active, window_panes, entry);
 			if (lastwp == NULL)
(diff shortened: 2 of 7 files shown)


nicm:
Looks mostly good, but:

`window_pane_stack_remove` could be improved by having a flag saying if the pane is on the list or not instead of searching the list?

`window_destroy_panes` removes panes from the stack even if they are not on it, so it could crash.

bertschingert:
> `window_destroy_panes` removes panes from the stack even if they are not on it, so it could crash.

Thanks, I missed this. I resolved this by changing it to only remove panes that are on the stack.

While investigating that crash, I found another crash that I think is pre-existing (I could reproduce it with the current master branch). If you run the following sequence of commands:
```
split-window
respawn-window -k
last-pane
```
The server tries to set the last pane to one of the previously destroyed panes and crashes. This is addressed by the new change to set `w->active = NULL` in `spawn_window`.

> `window_pane_stack_remove` could be improved by having a flag saying if the pane is on the list or not instead of searching the list

I modeled this off of `winlink_stack_remove` and I think they both have the issue that since an entry can be removed from any location in the list, I don't think there's a way around looping over each element, since we can't assume that what we're removing the first or last entry. So I don't know if having a flag for the pane being in the list would help much. Or is there something I'm not seeing?

nicm:
> So I don't know if having a flag for the pane being in the list would help much. Or is there something I'm not seeing?

The TAILQ knows where they are in the list, and they can only be in the list once, so if you have a flag saying they are on the list (ie that the linked list pointers are valid) then you can just do TAILQ_REMOVE.

nicm:
I don't know why winlink_stack_remove works like this, either it is just an oversight or the code was changed from a different type of list.

bertschingert:
I've changed it to use a flag as you suggested. I put the flag at the end of the list in `struct window_pane`, but let me know if you'd like me to use one of the earlier unused flags instead (like 0x8).

Should the same change be made for `winlink_stack_remove`? If yes, would you like me to add that to this patch?

nicm:
Sorry for the delay, I was on holiday. This looks good now. Yes if you could do the same for `winlink_stack_remove` it would be good.

nicm:

bertschingert:
Sorry, I missed your last update. I will work on the winlink_stack_remove change now. Thanks for the ping!

bertschingert:
Latest revision makes the change for `winlink_stack_remove`.

nicm:
I have applied this with a few minor changes (most notable, I renamed `last_visited_panes` to simply `last_panes` which is shorter). It should be in GitHub later on. Thanks!

PR #3590, opened 2023-06-11
Check result of fdopen

Garfield96:
`fdopen` can fail for various reasons, e.g. if there is not enough memory. If `fdopen` fails, it returns `NULL`, which previously would have caused a segmentation fault in `fwrite`.

This issue was found by using the GCC static analyzer, which can be enabled using the `-fanalyzer` flag:
```
CFLAGS=-fanalyzer ./configure && make -j
```


nicm:

PR #3595, opened 2023-06-21
Add submodule and building script for macOS

jeff999955:
Hello,

I have added two dependencies as git submodules and the corresponding script.

Thank you


nicm:
Thanks but I do not want to have submodules, you could make your script check the dependencies out itself and then I could put it in `tools/`, or keep it in your own repository.

PR #3612, opened 2023-07-03
[WIP] [Help needed] Support terminals which have more than 256 colors in their colortable

grwlf:
In this draft PR we attempt to support color terminals which have more than 256 colors in their colortable. The plan is:
* [x] Support options `color256` .. `color65535`
* [ ] Allow tmux to issue `\033[38;5;XXXX` -like commands to terminals where XXXX is not limited to be within the 0..255 range.

The motivation for this work is the following:
1. We want to add support for custom terminal-side colors in order to refer to them from various client-side configs. 
2. In particular, we want to enable different per-client color themes to co-exist in the tmux runtime. In order to do this we plan to ask clients to configure their `color256`..`colorXXXX` to match the specific parts of the UI and then use these names in a tmux session configuration.

Now, Vim already supports this feature (e.g. we can freely use `:highlight LineNR ctermbg=260` where 260 is the index in the extended color table of a terminal) and we already have this working in Simple terminal (the changes are trivial, see this [patch](https://github.com/grwlf/st/commit/9b8a6c1c73fb3f93491d5d48fc8e4d77f1d4a473))

Unfortunately, the tmux part of the story is in half-working state. The current version is able to display the right colors in PS1 (if configured like `PS1="\n\[\033[38;5;260;1m\][\u@\h \w ]\\$\[\033[0m\] "`) but the statusline colors are displayed incorrectly.

Advice or comments are highly welcome!

### DIFF
--- a/colour.c
+++ b/colour.c
@@ -137,7 +137,7 @@ colour_tostring(int c)
 	}
 
 	if (c & COLOUR_FLAG_256) {
-		xsnprintf(s, sizeof s, "colour%u", c & 0xff);
+		xsnprintf(s, sizeof s, "colour%u", c & 0xffff);
 		return (s);
 	}
 
@@ -203,13 +203,13 @@ colour_fromstring(const char *s)
 	}
 
 	if (strncasecmp(s, "colour", (sizeof "colour") - 1) == 0) {
-		n = strtonum(s + (sizeof "colour") - 1, 0, 255, &errstr);
+		n = strtonum(s + (sizeof "colour") - 1, 0, 0xffff, &errstr);
 		if (errstr != NULL)
 			return (-1);
 		return (n | COLOUR_FLAG_256);
 	}
 	if (strncasecmp(s, "color", (sizeof "color") - 1) == 0) {
-		n = strtonum(s + (sizeof "color") - 1, 0, 255, &errstr);
+		n = strtonum(s + (sizeof "color") - 1, 0, 0xffff, &errstr);
 		if (errstr != NULL)
 			return (-1);
 		return (n | COLOUR_FLAG_256);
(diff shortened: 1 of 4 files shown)


nicm:
What is the point in this feature? RGB colour is already well-defined, widely supported and can express any colour that might be need. It is the path for terminals and applications which want more than 256 colours, not extending the colour table.

grwlf:
> What is the point in this feature? RGB colour is already well-defined, widely supported and can express any colour that might be needed. It is the path for terminals and applications which want more than 256 colours, not extending the colour table.

I tried to explain it in the paragraph following the word "motivation". Basically, I want to use relative addressing for colors. Saying that my background is "#262626" is a bad way. I want to say that my background is "color260" where the meaning of this is defined differently for every client.

nicm:
But you can already redefine the palette of 256 colours, is that not enough?

How widespread is this? What terminals support it?

grwlf:
> How widespread is this? What terminals support it?
As I already mentioned, Vim has support of this feature, so we should assume that it is not a news. I didn't study the situation on for terminal side in depth. The changes there seems to be trivial and I easily made it for my version of simple terminal.

> But you can already redefine the palette of 256 colours, is that not enough?
This would not be compatible because the meaning of the 256 standard colors are de-facto defined (16 basic colors; 6*6*6 color cube; finally 24 greyscale colors - no space left). See for example [here](https://talyian.github.io/ansicolors/). Lots of legacy software depend on these, the configurations were intended for making only minor color adjustments e.g. to tweak the display hardware.

The goal is to define entirely custom color themes as an extension of the table, while preserving the standard color semantics for apps that depend on it. Thus the PR.

nicm:
When I send `\033[38;5;998m` to iTerm2 it sets *some* colour but I can't work out what it is, there is nothing obvious in the code. What does xterm do?

I don't think tmux should be the driver of this. I would like to know this is out there and will be useful to other users before we add code to support it.

You may be right about the default 256 colour palette now being a de facto standard, but the reason for that is because we now have RGB colour. ncurses supports redefining the palette and until recently it was the only way to have RGB colour.

I don't see any terminfo entries where `colors` is more than 256 except those which are for RGB colour. How do we detect if the terminal supports this? Is there an existing capability?

nicm:
I haven't suggest alternatives because I'm trying to gauge whether this feature has enough adoption to make it the best approach and how difficult it would be to implement it in a way that would work on the various terminals people use. I'm not sure this is at a point now where that is possible to do.

Issue #3572, opened 2023-05-22
Assertion ret == 0 failed in sigfd_free_sigevent with libevent-2.2.1


mgorny:
### Issue description

After rebuilding tmux against libevent-2.2.1-alpha (we're testing to catch early issues), tmux is crashing immediately after opening a new session, with the following assertion:

```
/tmp/portage/dev-libs/libevent-2.2.1-r1/work/libevent-2.2.1-alpha-dev/signalfd.c:77: Assertion ret == 0 failed in sigfd_free_sigevent
```

Note that the assertion message appears on the new pty and disappears immediately as tmux exits. I have been able to capture it using asciinema. Curious enough, it doesn't appear at all if I pass `-v` (tmux just exits immediately with empty screen).

I can reproduce with tmux 3.3a and git (0eb5d2545367c4d051861c33452d33f25b427e89), against libevent 2.2.1 and git (libevent/libevent@2112c3969a63703662fa90796723ba50754f11c9).

### Required information

Please provide the following information:

* tmux version (`tmux -V`): `tmux next-3.4`
* Platform (`uname -sp`): `Linux AMD Ryzen 5 3600 6-Core Processor`
* $TERM inside and outside of tmux (`echo $TERM`): `xterm-256color` outside, can't get inside since it crashes ;-)
* Logs from tmux (`tmux kill-server; tmux -vv new`): follow

[tmux-client-150783.log](https://github.com/tmux/tmux/files/11530699/tmux-client-150783.log)
[tmux-out-150785.log](https://github.com/tmux/tmux/files/11530700/tmux-out-150785.log)
[tmux-server-150785.log](https://github.com/tmux/tmux/files/11530701/tmux-server-150785.log)

Backtrace:

```
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007fa50cf2200f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007fa50ced2ef2 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fa50cebc4ad in __GI_abort () at abort.c:79
#4  0x00007fa50d19cc2f in event_exit (errcode=-559030611) at /usr/src/debug/dev-libs/libevent-9999/libevent-9999/log.c:96
[...]

PR #3621, opened 2023-07-09
Fix spawn_pane() with newer libevent that uses signalfd

azat:
There is closefrom() and after it calls proc_clear_signals() which in turn tries to close the created signalfd and fails:

    Assertion ret == 0 failed in sigfd_free_sigevent

Fix this by reorder the calls.

Fixes: https://github.com/tmux/tmux/issues/3572
Fixes: https://github.com/libevent/libevent/issues/1460

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -427,9 +427,12 @@ spawn_pane(struct spawn_context *sc, char **cause)
 	if (tcsetattr(STDIN_FILENO, TCSANOW, &now) != 0)
 		_exit(1);
 
+	/* In newer version of libevent (since 2.2) it may use signalfd(), which
+	 * has real fds under it, so firstly you need to remove/close those fds and
+	 * only after execute closefrom() */
+	proc_clear_signals(server_proc, 1);
 	/* Clean up file descriptors and signals and update the environment. */
 	closefrom(STDERR_FILENO + 1);
-	proc_clear_signals(server_proc, 1);
 	sigprocmask(SIG_SETMASK, &oldset, NULL);
 	log_close();
 	environ_push(child);


nicm:
This looks OK, I will apply it tomorrow, thanks. It is not a great idea to assert if `close` fails however.

nicm:
I've applied this to OpenBSD now, it will be in GitHub later, thanks! I did name you in the commit but I forget to mention the issue number.

azat:
> It is not a great idea to assert if close fails however.

I thought about this, but decided to stick with assert for now, since this is for debug builds only, that should be used only by developers (who possibly could dig and fix such issues)

>I've applied this to OpenBSD now, it will be in GitHub later, thanks! I did name you in the commit but I forget to mention the issue number.

Thanks. Just out of curiosity why not simply merge/cherry-pick patches and instead re-apply them without original subject/author/description?

P.S. maybe you should put some name here? https://github.com/tmux/tmux/blob/e4c4ceb2865fd9d2006a85c7f48bb7b6dd2f5c49/COPYING#L6

nicm:
OK, fair enough.

There are three reasons not to merge using Git:

- I have to apply them to OpenBSD first and it uses CVS;
- I want to apply the patch in order to review and test it and once I do that it is rare that there is a patch (or commit message) that doesn't need some minor tweak, so rather than bouncing back to the submitter to ask them to make some little change, it is much quicker if I just do it;
- I don't like submitter information going into Git metadata rather than into the commit message.

COPYING is an example of the copyright header which appears at the top of each file, it doesn't need a name.

PR #3626, opened 2023-07-10
Fix SIGCHLD with libevent signalfd backend

azat:
Without this patch with signalfd() backend it works incorrectly, this is
what proc_clear_signals() does:

    6674  rt_sigprocmask(SIG_UNBLOCK, [CHLD], NULL, 8) = 0
    6674  rt_sigaction(SIGCHLD, {sa_handler=SIG_IGN, sa_mask=[CHLD], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7ffff7c78ab0}, NULL, 8) = 0
    6674  close(6<signalfd:[CHLD]>)         = 0

And this is what proc_set_signals() does:

    6674  rt_sigaction(SIGCHLD, NULL, {sa_handler=SIG_IGN, sa_mask=[CHLD], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7ffff7c78ab0}, 8) = 0
    6674  rt_sigprocmask(SIG_BLOCK, [CHLD], NULL, 8) = 0
    6674  signalfd4(-1, [CHLD], 8, SFD_CLOEXEC|SFD_NONBLOCK) = 6<signalfd:[CHLD]>

Pay attention to the SIG_IGN.

And also one more patch for the client, that calls waitpid in a loop to avoid loosing any childs.

**P.S. I'm going to disable signalfd backend in libevent by default, since there are some trickery, which will break some existing programs. Also note, that there was not stable release with signalfd enabled.**

### DIFF
--- a/client.c
+++ b/client.c
@@ -535,9 +535,15 @@ client_signal(int sig)
 	int		 status;
 
 	log_debug("%s: %s", __func__, strsignal(sig));
-	if (sig == SIGCHLD)
-		waitpid(WAIT_ANY, &status, WNOHANG);
-	else if (!client_attached) {
+	if (sig == SIGCHLD) {
+		for (;;) {
+			pid_t pid = waitpid(WAIT_ANY, &status, WNOHANG);
+			if (pid == 0)
+				break;
+			if (pid == -1)
+				log_debug("waitpid failed: %s", strerror(errno));
+		}
+	} else if (!client_attached) {
 		if (sig == SIGTERM || sig == SIGHUP)
 			proc_exit(client_proc);
 	} else {
--- a/client.c
+++ b/client.c
@@ -516,7 +516,7 @@ client_exec(const char *shell, const char *shellcmd)
 		xasprintf(&argv0, "%s", name);
 	setenv("SHELL", shell, 1);
 
-	proc_clear_signals(client_proc, 1);
+	proc_clear_signals(client_proc);
 
 	setblocking(STDIN_FILENO, 1);
 	setblocking(STDOUT_FILENO, 1);
--- a/cmd-pipe-pane.c
+++ b/cmd-pipe-pane.c
@@ -132,7 +132,7 @@ cmd_pipe_pane_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	case 0:
 		/* Child process. */
-		proc_clear_signals(server_proc, 1);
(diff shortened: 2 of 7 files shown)


nicm:
I'm not sure the client can actually have multiple children (maybe it can now), but even if it doesn't it is better to add a loop around `waitpid`, so I have applied your `client_signal` change, although you forgot to also exit the loop on `ECHILD`, so I added that.

I don't understand your other change. Why can libevent not restore the default signal behaviour for SIGCHLD?

azat:
>although you forgot to also exit the loop on ECHILD, so I added that.

Thanks

>
I'm not sure the client can actually have multiple children (maybe it can now), but even if it doesn't it is better to add a loop around waitpid, so I have applied your client_signal change, although you forgot to also exit the loop on ECHILD, so I added that.

> Why can libevent not restore the default signal behaviour for SIGCHLD?

It does not touches it, tmux does:

```c
$ gg SIG_IGN
client.c:       signal(SIGCHLD, SIG_IGN);
```

And so the problem here is if server had `SIGCHLD=SIG_IGN` before calling `proc_set_signals`, then for `execve`, when it calls `proc_clear_signals`, libevent will reset the `SIGCHLD` to `SIG_IGN`, since it remember what the signal handler was before and restore on event deletion.

nicm:
Hmm, yes...

That was added in 2015 but after that we blocked the signals around `daemon` (in 2017), so it seems like this should not happen - the client signal handler should be in place before we get to `server_start` and the server one should be in place before we unblock `SIGCHLD` at `server.c:199`, so it seems like one of them should pick up the extra `daemon` child.

I think we can probably just remove that, like this:
~~~~
--- client.c    10 Jul 2023 09:35:46 -0000      1.160
+++ client.c    10 Jul 2023 10:18:01 -0000
@@ -246,9 +246,6 @@ client_main(struct event_base *base, int
        u_int                    ncaps = 0;
        struct args_value       *values;

-       /* Ignore SIGCHLD now or daemon() in the server will leave a zombie. */
-       signal(SIGCHLD, SIG_IGN);
-
        /* Set up the initial command. */
        if (shell_command != NULL) {
                msg = MSG_SHELL;

azat:
Yes, this will be even better (the comment looked odd to me). I've tested this and can confirm that it works.

nicm:
Great, I have applied it now. Thanks!

PR #3631, opened 2023-07-13
Use silent makefile rules by default

proski:
This makes compiler warnings more visible. The original behavior can be enabled by adding V=1 to the make command line.


nicm:
I prefer it the way it is TBH.

krader1961:
I agree with @nicm. Silent makefile rules should not be the default. There are situations where inhibiting echoing the command being run makes sense but it should be done on a case by case basis. Also, this type of change requires a strong argument that the new behavior is better rather than just satisfying your personal preference.

proski:
If having warnings appear more prominently doesn't count as an argument, then let's keep it as is.

PR #3641, opened 2023-07-21
Set mouse_word on the status line

### LINKED ISSUE #3628: #{mouse_word} is blank from status bar click
### Issue description

This is almost a feature request:

I would like to access  #{mouse_word} from clicks in the status bar.

I have the following use case: I show the list of sessions in the status bar and I would like to attach to a different session by clicking on it.

I saw #2913 and could confirm that #{mouse_x} and #{mouse_y} are set, but #{mouse_word} remains blank. Here is a minimal tmux.conf

```tmux
# start with mouse mode enabled
set -g mouse on

set -g status-keys vi
set -g mode-keys vi

# add second line to status line
set -g status 2 
# format of second line: list of active sessions
set -g status-format[1] '#[align=centre]#{S:#{session_name} }'

# click on a session in the list to switch to it 
bind -Troot MouseUp1StatusDefault if -F '#{mouse_word}' {
    run-shell "tmux attach-session -t #{mouse_word}"
} { display -F 'mouse word: #{mouse_word} x: #{mouse_x} y: #{mouse_y}'}
```

My specific use case may be unpopular but I believe the 
ability to reference the #{mouse_word} from the status bar could enable more customizations.

### Required information

* tmux version (`tmux -V`).
  `tmux 3.3a`
   I also tried with c8494df

* Platform (`uname -sp`).
  `Linux x86_64`

* $TERM inside and outside of tmux (`echo $TERM`).
  Inside: `tmux-256color`
  Outside: `xterm-256color`

  [tmux-client-264627.log](https://github.com/tmux/tmux/files/12029491/tmux-client-264627.log)
  [tmux-out-264629.log](https://github.com/tmux/tmux/files/12029492/tmux-out-264629.log)
  [tmux-server-264629.log](https://github.com/tmux/tmux/files/12029494/tmux-server-264629.log)

faustind:
Closes #3628 

This PR attempts to make mouse_word available from clicks on the status_bar.

(Unfortunately, my cursor is not recorded)
render1689941946151

Here is the configuration used for the demo
```
set -g prefix C-b
bind C-b send-prefix

# start with mouse mode enabled
set -g mouse on

# add second line to status line
set -g status 2 
# format of second line
set -g status-format[1] '#[align=centre]first second third fourth'

bind -Troot MouseUp1StatusDefault if -F '#{mouse_word}' {
     display -d 0 -F 'mouse word: #{mouse_word} x: #{mouse_x} y: #{mouse_y}'
} 
```

If there is anything I missed, please let me know 🙏🏾

### DIFF
--- a/format.c
+++ b/format.c
@@ -1131,19 +1131,33 @@ format_cb_mouse_word(struct format_tree *ft)
 	if (!ft->m.valid)
 		return (NULL);
 	wp = cmd_mouse_pane(&ft->m, NULL, NULL);
-	if (wp == NULL)
-		return (NULL);
-	if (cmd_mouse_at(wp, &ft->m, &x, &y, 0) != 0)
-		return (NULL);
+	if (wp != NULL) {
+		if (cmd_mouse_at(wp, &ft->m, &x, &y, 0) != 0)
+			return (NULL);
 
-	if (!TAILQ_EMPTY(&wp->modes)) {
-		if (TAILQ_FIRST(&wp->modes)->mode == &window_copy_mode ||
-		    TAILQ_FIRST(&wp->modes)->mode == &window_view_mode)
-			return (s = window_copy_get_word(wp, x, y));
-		return (NULL);
+		if (!TAILQ_EMPTY(&wp->modes)) {
+			if (TAILQ_FIRST(&wp->modes)->mode == &window_copy_mode ||
+			    TAILQ_FIRST(&wp->modes)->mode == &window_view_mode)
+				return (s = window_copy_get_word(wp, x, y));
+			return (NULL);
+		}
+		gd = wp->base.grid;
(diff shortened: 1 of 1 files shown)


nicm:
This looks fine, but I don't know if this is the road we want to go down.

I think it might be better to change the `range` operator in the styles so you can specify a custom argument which will be provided in a format (probably changing `range_argument` to a string), so something like `range=custom|foo`. Then you are not bound by the text actually displayed.

Although I suppose doing that doesn't mean that `mouse_word` couldn't work as well...

faustind:
@nicm
It also works with the range operator (https://github.com/tmux/tmux/issues/3652#issuecomment-1682335268). So unless there is another reason `mouse_word` on the status line would be useful, it's okay to close this PR

nicm:
OK I will close this unless another use case comes up, thanks!

PR #2947, opened 2021-10-23
Add options and arguments for display-menu to set border, border style, and style

afh:
~~:warning: The merge base of this PR should be [`afh:afh-popup-style-arguments`](https://github.com/afh/tmux/tree/afh-popup-style-arguments) from #2931 instead of [`tmux:master`](https://github.com/tmux/tmux).
:information_source: To see only the relevant changes please compare [`afh:afh-popup-style-arguments` with`afh:afh-menu-style-opts-args`](https://github.com/afh/tmux/compare/afh-popup-style-arguments...afh:afh-menu-style-opts-args?expand=1).~~ Update: #2931 has been applied, so master as a merge base for this PR is now correct 🎉 

This PR brings the same styling capabilites from popup to menu.

To test:
* Build and run tmux `./tmux -vv -f/dev/null -Ltest`
* Set menu-border-lines, menu-style options and display menu (see `test_menu_style.sh` below)
* Use `-B`, `-b`, `-s, `-S` arguments when displaying a menu (see `test_menu_style.sh` below)

<details><summary><code>test_menu_style.sh</code></summary>

```
#!/bin/ksh

./tmux set -gwU menu-border-lines
./tmux set -gw menu-border-lines double
./tmux set -gw menu-border-style fg=black,bg=yellow
./tmux set -gw menu-style fg=magenta,bg=cyan
./tmux set -gw mode-style fg=yellow,bg=black
./tmux menu -T 'Styled via Options' '#[fg=red,bg=blue]foo' '' '' -bar '' '' '' baz '' ''

./tmux menu -B -T 'Styled via Options' '#[fg=red,bg=blue]foo' '' '' -bar '' '' '' baz '' ''

./tmux menu -b heavy -S fg=blue,bg=red -s fg=magenta,bg=cyan -T 'Styled via Arguments' foo '' '' -bar '' '' '' baz '' ''

./tmux menu -B -b heavy -S fg=blue,bg=red -s fg=magenta,bg=cyan -T 'Styled without Border' foo '' '' -bar '' '' '' baz '' ''
```

</details>

| | |
| --- | --- |
| Styled with Options | 2021-10-23-101801_905x562_scrot|
| Styled without Border | 2021-10-23-103011_905x562_scrot|
| Styled with Arguments | 2021-10-23-101809_905x562_scrot|
| Styled without Border | 2021-10-23-101815_905x562_scrot|

PR #3647, opened 2023-07-27
Minor clean-up around freeing a menu title

afh:
While working on follow up PRs for #2947 I noticed that the menu title could probably be safely freed right after `menu_create`.


nicm:

Issue #3608, opened 2023-07-01
[FR] Add `-c start-directory` option to `run-shell` command


pataquets:
### Issue description
As done for `new-window` command, for instance.
Being able to customize where commands are run has lots of scenarios where it is useful.

### Required information
``` console
$ uname -sp && tmux -V && echo $TERM
Linux x86_64
tmux 3.3a
tmux-256color
```

caiuswang:
Do you mean starting directory?  If it is, you can  run new-window with '-c' to specify it.

nicm:
I'm closing this because we do not need an issue and a PR, please see my comments on the PR, thanks.

PR #3661, opened 2023-08-09
Add -c option to specify starting directory in which run-shell command

caiuswang:
Add -c option to specify starting directory in which run-shell command.  
Associate issure is  [3608](https://github.com/tmux/tmux/issues/3608)

### DIFF
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -44,8 +44,8 @@ const struct cmd_entry cmd_run_shell_entry = {
 	.name = "run-shell",
 	.alias = "run",
 
-	.args = { "bd:Ct:", 0, 1, cmd_run_shell_args_parse },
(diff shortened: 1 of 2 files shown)


nicm:
Thanks. Looks good aside from two things:

- You need to copy both strings with `xstrdup` or the result of `args_get(args, 'c')` will be double freed.

- Can you please also update tmux.1 to add this flag?

caiuswang:
> Thanks. Looks good aside from two things:
> 
> * You need to copy both strings with `xstrdup` or the result of `args_get(args, 'c')` will be double freed.
> * Can you please also update tmux.1 to add this flag?

Thanks for pointing out problems. I have fixed it.

nicm:
Looks good, applied to OpenBSD now with minor fixes, will be in GitHub later. Thanks!

pataquets:
Thanks @nicm & @caiuswang!
Browsed around docs trying to figure out the release schedule with no luck.
When will this be available in a release?

Issue #3343, opened 2022-10-05
Feature request: switch to next or previous session after current session is closed


imcaizheng:
According to man page, by setting `detach-on-destroy` to `off` the client is switched to the most recently active of the remaining sessions after current session is closed.

I think the most recently active session is not always desired. It could be more predictable if the client is switched to previous session or next session just like how `switch-client -p` or `switch-client -n` does.

I suggest add a new option like `detach-on-destroy-behavior [most-recently-active | previous | next]` to par with the `detach-on-destroy` option.

`most-recently-active` refers to the most recently active of the remaining sessions. `previous`/`next` refers to the previous/next session in alphabetic order.

nicm:
I would be OK with that but instead of a new option change detach-on-destroy to be a choice option where the possible values are off/on/previous/next.

imcaizheng:
You're right. It's more reasonable to reuse the `detach-on-destroy` option.

nicm:
I have added this to the todo list.

PR #3665, opened 2023-08-17
New Feature: switch to next or previous session after current session is closed

afh:
While working in the feature I saw an opportunity to make the code more concise, by replacing the `server_next_session` and `server_next_detached_session` functions with a single `server_filtered_session` that allows to retrieve a session based on a filter function. If you believe this isn't in the spirit of tmux I'm happy to rework this PR according to your suggestions or "unfold" `server_filtered_session ` into `server_next_session`, `server_next_detached_session`, etc. functions.

Additionally in my mind the `preceding` and `following` option names for `detach-on-destroy` are more meaningful, but I'd be very much okay to have these named `previous` and `next` if this fits better with the existing wording in tmux.

Note that I did only some basic tests and would appreciate thorough testing from others, to ensure I've understood and implemented the feature request correctly.
@imcaizheng would you be willing to help out here?

- [ ] If merged remove the "_([#3343](https://github.com/tmux/tmux/issues/3343)) Extend detach-on-destroy to allow selecting which remaining session becomes attached._" entry from the the [Small things section on the Contributing wiki page](https://github.com/tmux/tmux/wiki/Contributing#small-things)

Implements #3343


nicm:
Thanks! The idea looks OK but yes please use `next` and `previous` or `before` and `after` both to match other places in tmux and because using simple words is better.

Why not put the attached check in the callback also?

Please give the callback functions a `server_` prefix like the others.

afh:
> Thanks! The idea looks OK […]

Glad to hear that 🙂 

> please use `next` and `previous` or `before` and `after` both to match other places in tmux and because using simple words is better.

✅ Done. While `before` and `after` also seem like good candidates I went with `next` and `previous` as requested in #3343.

> Why not put the attached check in the callback also?

Because 🤦😅. In all seriousness: What a great suggestion! I must've suffered from tunnel vision and operational blindness.

✅ Done.

> Please give the callback functions a `server_` prefix like the others.

✅ Done.


I took the liberty of renaming the changed/added functions, so they are a bit shorter and fit better onto a line according with fewer line breaks while styling adhering to [style(9)](https://man.openbsd.org/style).

nicm:
I don't think you tested these changes because it doesn't build now :-).

It looks OK apart from that.

The only thing is I think the sessions should wrap around. That is, if you have it set to `next` and you kill the last session, it should go to the first, and the same for `previous` (if on the first, go to the last).

afh:
Yikes, apologies for that.

Right, `previous` and `next` should wrap around. I'm not too familiar the `sessions` variable, RB trees and their implementation in tmux. Is there a `RB_` macro or something else that would be helpful to do the wrapping or getting the first / last element in the tree?

Also am I correct in assuming that iterating over `sessions` happens in an alphabetically sorted order on the session's `name`?

afh:
Reading through the code I see there is a [`session_previous_session`](https://github.com/tmux/tmux/blob/master/session.c#L317-L332) and [`session_next_session`](https://github.com/tmux/tmux/blob/master/session.c#L300-L315) function. Inserting them instead of `server_find_session(s, server_previous_session);` and `server_find_session(s, server_next_session);` compiles the code. Could these be re-used or can you think of any reason why that would be ill-advised/

nicm:
Yes that should work I think.

afh:
Great! I've update this PR accordingly, @nicm. Not sure if the introduction of `server_find_session` adds much value now that `server_(previous|next)_session` are used. I'll leave it at your discretion to include the `server_find_session` change or not and am happy to prepare this PR so it contains the smallest amount of changes in order to implement the feature request if you so desire.

nicm:
This needs some additional changes because currently when you exit the last pane in the last window in session, the session will be removed from the tree (in `session_destroy` called from `session_detach`) before `server_destroy_session` is called which means `session_alive` will be false and `session_next_session` will always return `NULL`.

So instead we need to move the call to `session_destroy` outside of `session_detach`. Please try this: [tmux-detach.diff.txt](https://github.com/tmux/tmux/files/12377744/tmux-detach.diff.txt)

afh:
Finally I'm getting around to this again, @nicm :)

I have (hopefully correctly) merged in the changes from the [tmux-detach.diff.txt](https://github.com/tmux/tmux/files/12377744/tmux-detach.diff.txt) you mentioned in your [previous comment](https://github.com/tmux/tmux/pull/3665#issuecomment-1683621925) into this PR so we work from the same code base.

Please find below `test_detach_on_destroy.sh`, which I use to test. If my understanding of [your last comment](https://github.com/tmux/tmux/pull/3665#issuecomment-1683621925) is correct, then when exiting _the last pane in the last window in session, the session_ tmux should switch to the next session—with my test script this should be sessions `abc`—instead of exiting, which it does for me currently. Is my understanding correct?

<details><summary><code>test_detach_on_destroy.sh</code></summary>

```sh
#!/bin/ksh

tmux=${1:-./tmux}

clear

echo "# Rename the current session to 'def'"
echo "% $tmux rename-session def"
$tmux rename-session def

echo "# Create detached session xyz"
echo "% $tmux new-session -d -s xyz"
$tmux new-session -d -s xyz

echo "# Create detached session abc"
echo "% $tmux new-session -d -s abc"
$tmux new-session -d -s abc

echo "# List tmux sessions"
echo "% $tmux ls -F '#{session_name}'"
$tmux ls -F '#{session_name}'

echo "# List tmux sessions"
echo "% $tmux set detach-on-destroy next"
$tmux set detach-on-destroy next

echo "# Switch to last session"
echo "% $tmux switch-client -t xyz"
$tmux switch-client -t xyz

echo "# Exit this window by typing 'exit' or pressing ^D (CTRL-D)"
echo "# tmux is expected to switch to the abc session"

```

</details>

nicm:
This is all working for me, including your test script, so I have applied this now. Are you sure you are exiting session `abc` where you set `detach-on-destroy`? Or try changing it to `set-g`. Let me know if you still can't get it to work or if you have other problems. Thanks!

PR #3650, opened 2023-07-27
Add options and arguments to display-menu to set menu border lines

afh:
This PR adds border line styling to `display-menu` and is a follow-up PR of #2947.

ℹ️ This PR builds upon the proposed changes from #3649; to only see the relevant changes please see https://github.com/afh/tmux/compare/afh-menu-border-style...afh:tmux:afh-menu-border-lines

To test:
* Build and run tmux `./tmux -vv -f/dev/null -Ltest`
* Set `menu-border-lines` options and display menu (see `test_menu_border_lines.sh` below)
* Use `-b` argument when displaying a menu (see `test_menu_border_lines.sh ` below)

👉 **NOTA BENE:** There are several `TODO:` comments in the code that should to be clarified prior applying the changes.

ℹ️ The `none` style is currently the same as `padded` due to difficulties in removing the border completely, for details see [this comment on #2947](https://github.com/tmux/tmux/pull/2947#issuecomment-1195668262)

ℹ️ The changes in [`tty-acs.c`](https://github.com/tmux/tmux/pull/3650#diff-0061931f0e579d624769a3e92854fbd0db819ee849c4c75d74928eb0114c2abb) fix an issue with the rounded characters which used the heavy ([U+2523](https://unicodeplus.com/U+2523) and [U+252B](https://unicodeplus.com/U+252B)) instead of the light ([U+251C](https://unicodeplus.com/U+251C) and [U+2524](https://unicodeplus.com/U+2524)) variant of the "Box Drawings * Vertical And (Right|Left)" character.  

<details><summary><code> test_menu_border_lines.sh </code></summary>

```sh
#!/bin/ksh

clear
echo "# Expect default tmux border lines menu when unstyled"
echo "% ./tmux set -gwU menu-border-lines"
./tmux set -gwU menu-border-lines
echo "% ./tmux menu -T Default lines '' '' -bar '' '' '' baz '' ''"
./tmux menu -T Default lines '' '' -bar '' '' '' baz '' ''

clear
lines=heavy
echo "# Expect tmux $lines menu border lines when set via tmux options, e.g.:"
echo "% ./tmux set -gw menu-border-lines $lines"
./tmux set -gw menu-border-lines $lines
echo "% ./tmux menu -T $lines lines '' '' -bar '' '' '' baz '' ''"
./tmux menu -T $lines lines '' '' -bar '' '' '' baz '' ''

for lines in double single rounded heavy ; do
clear
[...]

PR #3666, opened 2023-08-17
WIP: Add none border-lines style

afh:
🚧 This is a work-in-progress follow-up PR on #3650 that attempts to add the `none` border-lines style.

@nicm: I assume there are off-by-one errors when calculating the width and height of the menu and I may have missed some code that does `+ 4` for the width or `+ 2` for the height when figuring out the size and position of the menu.
Maybe the changes in this PR so far help you identify other sections where changes are needed or issues with the code I've written.
I think it might be helpful to get another pair of eyes on this 👀

<details><summary>Test using <code>test_menu_border_lines_none.sh</code></summary>

```sh
#!/bin/ksh

clear
echo "# Expect double tmux border lines menu when styled using arguments"
echo "% ./tmux menu -x 0 -y 1 -b padded -S bg=orange -T Padded border '' '' -bar '' '' '' baz '' ''"
./tmux menu -x 0 -y 1 -b padded -S bg=orange -T Padded border '' '' -bar '' '' '' baz '' ''

clear
set -A size $(stty size)
awk -vc=× -vr=${size[0]} -vl=${size[1]} 'BEGIN{for(i=0;i<r;i++) for(j=0;j<l;j++) printf c}'
echo -e "# Expect no tmux border lines menu with -B even when styled using arguments"
echo "% ./tmux menu -x 0 -y 1 -B -b double -S bg=green -T Menu Title no '' '' -border '' '' '' baz '' ''"
./tmux menu -x 0 -y 1 -B -b double -S bg=green -T Menu Title '' '' -'no border' '' '' '' baz '' ''

clear
set -A size $(stty size)
awk -vc=× -vr=${size[0]} -vl=${size[1]} 'BEGIN{for(i=0;i<r;i++) for(j=0;j<l;j++) printf c}'
echo -e "# Expect no tmux title or border lines menu with -B even when styled using arguments"
echo "% ./tmux menu -x 0 -y 1 -B -b double -S bg=green 'No title' '' '' -no '' '' '' border '' ''"
./tmux menu -x 0 -y 1 -B -b double -S bg=green 'No title' '' '' -no '' '' '' border '' ''
```

</details>

| Description | Screenshot |
| --- | --- |
| 👍 `padded` border-style | image
| 🙅 `none` with title: there seems to be one additional row at the very end of the menu and one additional column on the right | image
| 🙅 `none` without title: there seems to be two additional row at the very end of the menu and one additional column on the right | image

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -299,7 +299,7 @@ cmd_display_menu_exec(struct cmd *self, struct cmdq_item *item)
 	u_int			 px, py, i, count = args_count(args);
 	struct options		*o = target->s->curw->window->options;
 	struct options_entry	*oe;
-	u_short			 bw;
+	u_int			 bw;
 
 	if (tc->overlay_draw != NULL)
 		return (CMD_RETURN_NORMAL);
@@ -368,7 +368,7 @@ cmd_display_menu_exec(struct cmd *self, struct cmdq_item *item)
 		}
 	}
 
-	bw = ((lines != BOX_LINES_NONE) ? 2 : 2);
+	bw = ((lines != BOX_LINES_NONE) ? 2 : 0);
 	if (!cmd_display_menu_get_position(tc, item, args, &px, &py,
 	    menu->width + 2 + bw, menu->count + bw)) {
 		menu_free(menu);
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -38,8 +38,8 @@ const struct cmd_entry cmd_display_menu_entry = {
 	.name = "display-menu",
 	.alias = "menu",
 
-	.args = { "b:c:C:H:s:S:Ot:T:x:y:", 1, -1, cmd_display_menu_args_parse },
-	.usage = "[-O] [-b border-lines] [-c target-client] "
+	.args = { "Bb:c:C:H:s:S:Ot:T:x:y:", 1, -1, cmd_display_menu_args_parse },
+	.usage = "[-BO] [-b border-lines] [-c target-client] "
 		 "[-C starting-choice] [-H selected-style] [-s style] "
 		 "[-S border-style] " CMD_TARGET_PANE_USAGE "[-T title] "
 		 "[-x position] [-y position] name key command ...",
@@ -299,7 +299,7 @@ cmd_display_menu_exec(struct cmd *self, struct cmdq_item *item)
 	u_int			 px, py, i, count = args_count(args);
 	struct options		*o = target->s->curw->window->options;
 	struct options_entry	*oe;
-
+	u_short			 bw;
 
 	if (tc->overlay_draw != NULL)
 		return (CMD_RETURN_NORMAL);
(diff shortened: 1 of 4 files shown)


nicm:
> +       bw = ((lines != BOX_LINES_NONE) ? 2 : 2);

This is a no-op. Also please just use `u_int` here not `u_short`.

>+       u_short two = 2;

This is not needed? It should be `bw`?

afh:
Oh?! Then this appears to be an older state of the PR, let me investigate.

nicm:
You can probably just wrap around by doing the loop like this:
~~~~
start = s;
do {
     ...
     s = RB_NEXT(...);
     if (s == NULL)
         s = RB_MIN(...);
} while (s != start);
~~~~
But you will need to walk backwards (RB_PREV/RB_MAX) for `previous` so you'll need a flag saying whether the loop should go forwards or backwards.

nicm:
OK, this kind of seems like it works, but there is one bug, firstly:

- In `menu_prepare`, `lines` is checked before it is updated from `menu-border-lines`. The `options_get_number` bit needs to be moved up above the `bw` assignment.

Then there is some odd stuff...

The menu is too narrow by one character. Changing this from 2 to 3 fixes it, but I'm not sure why, it seems like `2 + bw` should be right:
~~~~
-       screen_init(&md->s, menu->width + 2 + bw, menu->count + 2, 0);
+       screen_init(&md->s, menu->width + 3 + bw, menu->count + bw, 0);
~~~~
Also the height here seems like it should be `+ bw` not a fixed `+ 2`, but changing it *without also changing the width from + 2 to + 3*, means the lines wrap...

nicm:
Removing `MODE_WRAP` from the screen stops it wrapping incorrectly when the screen width is `+ 2` but it is still too narrow...

afh:
Thanks for your comments, @nicm, much appreciated as always. For today I will focus my attention to other work on my agenda and will get back to this as soon as I can.

nicm:
OK I think I have fixed it, it was most of the way there just a couple of minor errors and it was not taking the title into account everywhere. It is complicated because it is easy to make an error in one place but then accidentally correct it in another...

Please try this: 
[tmux-menu-none.diff.txt](https://github.com/tmux/tmux/files/12377173/tmux-menu-none.diff.txt)

There are three cases which need to be handled:

1. Borders, with or without title:
2. No border without title
3. No border with title

So the width is the item width plus two spaces plus optionally two border lines. The height is the item width plus optionally two border lines or a line for the title. Then the item position similarly depends on if there is a border and a title.

So everywhere we need to do this:
~~~~
if (lines != BOX_LINES_NONE) {
       bw = 2;
       co = 0;
} else {
       bw = 0;
       co = (strlen(menu->title) == 0) ? 0 : 1;
} 
~~~~
Meaning `bw` is the width of the border and `co` is the extra lines for the title.

Then we can say:

width is `menu->width + 2 + bw`
height is `menu->count + bw + co`
the first item is at x position `bw / 2 + 1` and y position `bw / 2 + co`

nicm:
Does this look/work OK for you?

nicm:
I'm closing to close this now but if you or anyone else wants to pick it up, let me know. Thanks!

PR #3673, opened 2023-08-26
Fix SIXEL images not being processed

Jorengarenar:
A simple brace error. 
Code calling SIXEL parsing function was falling under an `if` statement discarding the input.


nicm:
Fixed, thanks!

Issue #3678, opened 2023-09-01
Memory leak


FractalBoy:
### Issue description

tmux appears to be leaking memory. I don't know how to reproduce, but I only have 2GB of memory on my system and tmux is currently using upwards of 700M.

I have history going back >90000 lines, but the memory is not released after closing panes with large amounts of history.

### Required information

Please provide the following information:

* tmux version (`tmux -V`).
tmux next-3.4
* Platform (`uname -sp`).
Linux x86_64
* $TERM inside and outside of tmux (`echo $TERM`).
Outside: xterm-256color
Inside: screen-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`).
I don't have logs because I don't know how to cause the memory leak, but I generated a core dump, if that is useful to you. Github won't let me attach it but I can email it if you think it would be helpful.

FractalBoy:
I started collecting a log file in the event that I can get the memory leak to happen again.

nicm:
Is this on Linux?

FractalBoy:
Yes

nicm:
Oh yeah it is, sorry, see that now.

glibc is spectacularly bad at returning memory to the kernel and so far nobody has given me any idea how to configure it to work properly. tmux will call `malloc_trim` to ask it to return memory once an hour (it is very slow to do this so once an hour is enough). If you wait for an hour after closing the panes, does the memory use go down?

FractalBoy:
I'll test that theory next time I see the memory leak, but this seemed to start happening only recently, as I've pulled from master.

FractalBoy:
Yeah, this does seem to happen when I have a lot of history.

I have my history-limit set at 100000; perhaps that's too high for my system.

FractalBoy:
Instead of running malloc_trim only once an hour, could it can be run after closing panes?

nicm:
I think I tried that before and it was too slow, but you could possibly try adding it like this:
~~~~
--- a/grid.c
+++ b/grid.c
@@ -308,6 +308,7 @@ grid_destroy(struct grid *gd)
        free(gd->linedata);

        free(gd);
+       malloc_trim(0);
 }

 /* Compare grids. */

FractalBoy:
Just tried that - instantly better. Putting malloc_trim in server_kill_pane actually didn't do anything.

FractalBoy:
Also putting it in grid_clear_history helps if you manually clear history.

PR #3679, opened 2023-09-01
Run trim_malloc after grid_destroy and grid_clear_history

FractalBoy:
Addresses issue #3678


nicm:
We're not going to do this again, sorry - see https://github.com/tmux/tmux/issues/2551 and ef9700816fd2bd521bab837ccd85e0a596aa3aa8.

FractalBoy:
Ah, okay, didn't know. Guess I'll just apply the patch myself unless I see the same issue.

PR #3709, opened 2023-10-05
don't error when '-h' is passed

cjbayliss:
I don't know if these changes the sort of thing tmux accepts, I just wanted to add '-h' support. :smile:

Few things:
 1) `--help` prints the help and exits with `1`. Even though this is non-standard, I didn't change this behaviour because IDK what scripts people may have that depend on that.
 1) Initially I used `EXIT_SUCCESS` and `EXIT_FAILURE`, but they didn't gain standardisation until C99, and I noticed in the generated make file that tmux supports numerous targets and compilers.

---

*Please don't hesitate to close this PR if it isn't wanted. I have no personal attachment to the code I write, and I don't want to add a maintenance burden.*

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -43,20 +43,20 @@ const char	*socket_path;
 int		 ptm_fd = -1;
 const char	*shell_command;
 
-static __dead void	 usage(void);
+static __dead void	 usage(const int);
 static char		*make_label(const char *, char **);
 
 static int		 areshell(const char *);
 static const char	*getshell(void);
 
 static __dead void
-usage(void)
+usage(const int exit_status)
 {
 	fprintf(stderr,
 	    "usage: %s [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]\n"
 	    "            [-S socket-path] [-T features] [command [flags]]\n",
 	    getprogname());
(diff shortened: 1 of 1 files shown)


nicm:
`-h` doesn't do anything so it should be the same as any other unsupported option, it isn't special.

cjbayliss:
Understood. Closing. Maybe I could write a help message in the future, but I'll create a new PR if I get around to that.

krader1961:
It is unclear what problem this change solves. It is usually a good idea to open an issue that describes a problem. Then a change can reference that issue and describe how the change resolves the problem. Speaking for myself, I think that explicitly invoking a program with its "help" flag should result in a zero exit status. So I like your proposed change. However, a lot of people think that a program should exit with a non-zero status in this situation. Which is why it is preferable to open an issue to discuss why this type of change is warranted before going to the trouble of creating a pull-request. Also, your opening comment about this change is unclear as to why `--help` is treated differently from `-h`.

krader1961:
Note also that Tmux does not currently have a `-h` flag to explicitly emit its CLI usage without doing anything else. Adding that flag might be a good idea but, again, is something that should be discussed before asking that a change be merged.

PR #3738, opened 2023-11-03
Use DEC mode 2026 for Sync terminal feature

gpanders:
The ESC P = %p1 s ST sequence is an iTerm2 extension to enable or
disable synchronized output. However, some effort has been made recently
to "standardize" (as much as anything can be standardized in this world)
on using DEC mode 2026 for this feature instead [1]. iTerm2 also
supports this mode and even recommends its use over its own extension
[2].

As far as I can tell the only other terminal emulator that uses iTerm2's
extension is kitty, but they copied it from tmux. So if tmux changes to
use DEC mode 2026 then that would make a strong case for kitty to change
as well.

[1]: https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036
[2]: https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec#decsetdecreset

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -6823,7 +6823,7 @@ capability.
 .It Em \&Sxl
 Indicates that the terminal supports SIXEL.
 .It Em \&Sync
-Start (parameter is 1) or end (parameter is 2) a synchronized update.
+Start (parameter is 1) or end (parameter is 0) a synchronized update.
 .It Em \&Tc
 Indicate that the terminal supports the
 .Ql direct colour
--- a/tty-features.c
+++ b/tty-features.c
@@ -226,7 +226,7 @@ static const struct tty_feature tty_feature_strikethrough = {
 
 /* Terminal supports synchronized updates. */
 static const char *const tty_feature_sync_capabilities[] = {
-	"Sync=\\EP=%p1%ds\\E\\\\",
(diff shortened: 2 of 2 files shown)


nicm:
I have a feeling this escape sequence came from Kitty in the first place. I do remember everyone saying how stupid it was to use DCS like this and how it should be an SM instead.

nicm:
Maybe it was iTerm2 originally, it doesn't matter anyway.

gpanders:
I stumbled upon [this thread](https://gitlab.freedesktop.org/terminal-wg/specifications/-/merge_requests/2#note_973680) and apparently _you_ were the one to originally propose this terminfo string in the first place! So we have really come full circle.

nicm:
I added the `Sync` capability, someone else suggested SM 2026.

nicm:
I have chaged this, but I have not included your tmux.1 change because `Sync` still needs 1 to start and 2 to end and the new capability changes that to SM or RM:

Sync=\\E[?2026%?%p1%{1}%-%tl%eh%;

`%p1 %{1} %-` is `p1 - 1`, so in the `%?`, 2 will be true (`l` = RM) and 1 will be false (`h` = SM).

nicm:
Er, I forgot to credit you or link this issue in the commit message, sorry!

PR #3747, opened 2023-11-15
Check for NULL window_pane

zsugabubus:
There are some missing checks for NULL `wp`.

Crash in `input_csi_dispatch_winops` can be reproduced with: `tmux display-popup printf '\033[t'`.


nicm:
This is right but I am away from home so I can't apply it, I'll do it when I am back in a few weeks.

nicm:
Applied now, thanks!

zsugabubus:
Do you intentionally omitted the changes in `input_csi_dispatch_winops`?

nicm:
They will be there when it syncs from OpenBSD, it doesn't have SIXEL so the SIXEL bits can go straight into GitHub.

PR #3750, opened 2023-11-26
Initialized screen mode

danix800:
Call from "popup.c:popup_draw_cb()" -> "screen.c:screen_init()" -> "screen_reinit()" shows 'screen->mode' is uninitialized (garbage value).


nicm:
Applied, thanks!

PR #3762, opened 2023-12-06
Fix common manpage glyph usage errors

afh:
by using correct special characters as defined in groff_char(7), e.g. using \[aq] instead of `'` , as `'` may get displayed as `’` (U+2019 Right Single Quotation Mark' instead of the desired U+0039 Apostrophe) by certain groff output devices, e.g. PDF or utf8.

For details see section _"When viewing man pages, some characters on my UTF-8 terminal emulator look funny or copy-and-paste wrong. Why?"_ in groff's [PROBLEMS file](http://git.savannah.gnu.org/cgit/groff.git/tree/PROBLEMS) and the grotty section in the [groff 1.23.0 announcement email](https://lists.gnu.org/archive/html/groff/2023-07/msg00051.html).

The following examples have been created using `groff -Tutf8 -mdoc tmux.1 | $PAGER`
**Before:**
```txt
Note that to bind the ‘"’ or ‘’’ keys, quotation marks are necessary, for example:

     bind-key ’"’ split-window
     bind-key "’" new-window
```

<details><summary>Screenshot emphasising the issue as GitHub's font may not convey the difference as well</summary>
image
</details>

**After:**
```txt
Note that to bind the ‘"’ or ‘'’ keys, quotation marks are necessary, for example:

     bind-key '"' split-window
     bind-key "'" new-window
```

<details><summary>Screenshot emphasising the issue as GitHub's font may not convey the difference as well</summary>
image
</details>

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -140,10 +140,10 @@ By default,
 loads the system configuration file from
 .Pa @SYSCONFDIR@/tmux.conf ,
 if present, then looks for a user configuration file at
-.Pa ~/.tmux.conf,
+.Pa \[ti]/.tmux.conf,
 .Pa $XDG_CONFIG_HOME/tmux/tmux.conf
 or
-.Pa ~/.config/tmux/tmux.conf .
+.Pa \[ti]/.config/tmux/tmux.conf .
 .Pp
 The configuration file is a set of
 .Nm
@@ -290,7 +290,7 @@ Rename the current session.
 Split the current pane into two, left and right.
 .It &
 Kill the current window.
-.It '
+.It \[aq]
 Prompt for a window index to select.
 .It \&(
 Switch the attached client to the previous session.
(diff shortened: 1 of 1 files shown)


afh:
Thanks for asking, @nicm. I've found a small mistake I made earlier, have force pushed the fix and tested this on macOS using nixpkgs using mandoc 1.14.6 and now it does generate the same output with mandoc (see below)! 🥳 

```sh
git show master:tmux.1 | mandoc -Tutf8 > tmux.1-master.txt
git show fix-man-glyphs:tmux.1 | mandoc -Tutf8 > tmux.1-fix-man-glyphs.txt
cmp tmux.1-*.txt
echo $?
0
```

nicm:

afh:
Thanks and happy new year, @nicm! 🎉

PR #3798, opened 2024-01-06
add `pane-border-indent` option

andreykaipov:
Hi there - this patch removes the hardcoded 2's in `screen-redraw.c` and makes them customizable through a new `pane-border-indent` option. I've been running this patch for a few years but have been a bit reluctant opening a PR with it because of this comment from a related issue https://github.com/tmux/tmux/issues/2028#issuecomment-569234413:

> We are moving away from having options for every little thing, so I don't want pane-border-indent, it needs to be expressed in the style, which would also allow it to be used in other contexts.

I'd love to know if your thoughts have since changed and if you'd be open to an option like this. I tried achieving border indentation using a style directive like mentioned in the comment, but I wasn't successful because the hardcoded 2's got in the way regardless.

Please let me know what you think!

---

Here's a demo using the following config:

```tmux
set -g pane-border-status bottom
set -g pane-border-style ""
set -g pane-active-border-style ""
set -g pane-border-format "#[#{?#{pane_active},bg=red fill=red,bg=blue fill=blue}]"
set -g pane-border-indent 0
```

With default indentation:

<br><br>

With zero indent:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # configure.ac
 
-AC_INIT([tmux], next-3.4)
+AC_INIT([tmux], andrey-3.4)
 AC_PREREQ([2.60])
 
 AC_CONFIG_AUX_DIR(etc)
--- a/format-draw.c
+++ b/format-draw.c
@@ -717,7 +717,7 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 					   RIGHT,
 					   ABSOLUTE_CENTRE };
 	int			 focus_start = -1, focus_end = -1;
-	int			 list_state = -1, fill = -1, even;
+	int			 list_state = -1, pad = 0, fill = -1, even;
 	enum style_align	 list_align = STYLE_ALIGN_DEFAULT;
 	struct grid_cell	 gc, current_default;
 	struct style		 sy, saved_sy;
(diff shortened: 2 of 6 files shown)


nicm:
I'd rather see this done through the style or format option, so remove the hardcoded 2 and then add some way to express this in the style. I think `pad` would be a better name than `indent` since this is on both sides.

andreykaipov:
I appreciate you taking a look! I pushed a WIP commit for the `pad` style option but I'm not entirely sure what I'd have to change in `format-draw.c` after passing it through. It doesn't feel quite as elegant so please let me know if it's not the right idea. I used the existing `fill` as an example to work from.

> remove the hardcoded 2 and then add some way to express this in the style

Also maybe I'm misunderstanding, but if we do this, will the default padding effectively be zero in `screen-redraw.c`? And preserving existing behaviour will come from a default `pad=2` in the default style? Wouldn't that cause any users that already have a style set to no longer have that padding, since they'll essentially be overwriting it?

nicm:
I'll take a look at your changes later but yes you will need to change the default to include `pad=2` and we will need to put it in the change log and release issue so people can update their configs if needed.

configure.ac
andreykaipov, line 1: Woops I'll remove this file. Had changed it to make sure my build process was actually picking up my changes.

nicm:
Are you still working on this? Is it finished?

configure.ac
bhalo216: but you forget even u deltd file or some think is nothing what u done now u will see

PR #3806, opened 2024-01-09
Add options `keep-last` and `keep-group` to `destroy-unattached` to keep the last session whether in a group

huangyxi:
This pull request introduces two new options to the `destroy-unattached` session parameter.  When set to `keep-last`, the session will be retained if it **does NOT belong to any group** or **is the last session within a group**. When set to `keep-group`, the session will only be retained if it **belongs to a group** and **is the last session within that group**. This feature proves beneficial for maintaining a group of sessions, keeping them active even when unattached to any clients.

|             Case \ Option            | `off` |   `on`  | `keep-last` | `keep-group` |
|:------------------------------------:|:-----:|:-------:|:-----------:|:------------:|
|             NOT in group             |  keep | destroy |     keep    |    destroy   |
| within a group, NOT the last session |  keep | destroy |   destroy   |    destroy   |
|   within a group, the last session   |  keep | destroy |     keep    |     keep     |

The modification is backward compatible, preserving the default behavior of not destroying sessions under any circumstances.

To conduct the test for `keep-last`:

1. Terminate the tmux server to clear all sessions, ensuring reproducibility:
```
tmux kill-server
```
2. Add / replace the following to the config file (e.g. `~/.config/tmux/tmux.conf`):
```
set-option -g destroy-unattached keep-last
```
3. Execute the following command:
```
tmux new-session -s example-session
```
4. Detach from the session.

Subsequently, the session should remain active, and the following command ought to display the session:
```
tmux ls
```

To conduct the test for `keep-group`:

1. Terminate the tmux server to clear all sessions, ensuring reproducibility:
```
tmux kill-server
```
2. Add / replace the following to the config file (e.g. `~/.config/tmux/tmux.conf`):
```
set-option -g destroy-unattached keep-group
```
3. Execute the following command:
```
tmux new-session -s example-session -t example-group
```
4. Detach from the session.

Subsequently, the session should remain active, and the following command ought to display the session:
```
tmux ls
```

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -85,7 +85,7 @@ static const char *options_table_remain_on_exit_list[] = {
 	"off", "on", "failed", NULL
 };
 static const char *options_table_destroy_unattached_list[] = {
-	"off", "on", "keep-group", NULL
+	"off", "on", "keep-last", "keep-group", NULL
 };
 static const char *options_table_detach_on_destroy_list[] = {
 	"off", "on", "no-detached", "previous", "next", NULL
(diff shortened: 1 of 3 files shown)


nicm:
Perfect, applied to OpenBSD now, will be in GitHub later. Thanks!

PR #3809, opened 2024-01-12
Increase the buffer size on style_tostring() to avoid truncation.

ricbit:
Line 278 performs an snprintf of "user|%s" to variable b, which currently is 16 bytes. However, the string being written is style.range_string, which is itself 16 bytes. Therefore, to avoid truncation, the type of b must be char b[21], to fit 16 from range_string plus 5 from "user|".

This is not a buffer overflow security bug since snprintf truncates instead of overflowing. Still nice to avoid truncation, though.


nicm:

PR #3847, opened 2024-02-16
Use bracketed paste for choose-buffer by default

gpanders:
The -p option only has any effect if the pane has enabled bracketed paste mode, so this should be safe to use unconditionally.


nicm:
I think this is fine, I will change it after the next OpenBSD release in a few weeks.

nicm:

Issue #3857, opened 2024-02-23
Zoomed windows crash tmux after killing session with `x` on tree mode


serranomorante:
### Required information

Please provide the following information:

* tmux version (`tmux -V`).
tmux next-3.5 (also on most recent release 3.4.4)
* Platform (`uname -sp`).
Linux unknown (arch linux)
* $TERM inside and outside of tmux (`echo $TERM`).
tmux-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`).

## Steps to replicate

1. Open 2 tmux sessions
2. Enter 1 of those 2 sessions and split the window with `C-b %` then zoom the window with `C-b z`
3. Switch to the other *tmux session* (the one without the zooming) and open tree mode (with `C-b s`)
4. Hover onto the session that has the zooming and press `x` to kill that session
5. Tmux will crash after you confirm with `y`

PR #3860, opened 2024-02-25
Don't notify when unzooming windows from window_destroy()

rfrancoise:
Since commit 36e1ac65560, window_destroy() calls window_unzoom() but the latter emits a 'window-layout-changed' notification with the window which is about to be freed. This results in the callback later dereferencing a freed window, or a double-free.

Closes: #3857.


nicm:
This looks fine, will apply it in a few weeks after the next OpenBSD release is complete.

nicm:

PR #3869, opened 2024-03-02
only set selection_present format variable if the selection size is non-zero

mgrant0:
Per discussion in https://github.com/orgs/tmux/discussions/3831

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -795,7 +795,7 @@ window_copy_formats(struct window_mode_entry *wme, struct format_tree *ft)
 	format_add(ft, "copy_cursor_x", "%d", data->cx);
 	format_add(ft, "copy_cursor_y", "%d", data->cy);
 
-	format_add(ft, "selection_present", "%d", data->screen.sel != NULL);
+	format_add(ft, "selection_present", "%d", (data->screen.sel != NULL) && (data->endselx-data->selx) && (data->endsely-data->sely));
 	if (data->screen.sel != NULL) {
 		format_add(ft, "selection_start_x", "%d", data->selx);
 		format_add(ft, "selection_start_y", "%d", data->sely);


mgrant0:
When running tmux remotely via putty, I noticed that when I brought the putty window back into focus by clicking in the window, putty sent a drag event of zero length.  A `MouseDrag1Pane` followed immediatly by a `MouseDragEnd1Pane`.  To get around this, I wanted to put the following into my .tmux-conf:
```
bind -T copy-mode MouseDragEnd1Pane if-shell -F "#{selection_present}" { send -X copy-selection-no-clear }
```
The idea here is that copy-selection-no-clear will only be sent if there's a real selection and not just a simple click in the window without a real drag.

This PR adds a check to see if the selection is non-zero before setting the format variable `selection_present`.  This allows me to know if this situation of a zero length drag occured so I bind the above.

nicm:
Is there any use case for creating empty paste buffers in copy mode? Would it not be better to make copy ignore empty buffers?

mgrant0:
I can't think of a use case for an empty paste buffer and `selection_present` being true.  I would have thought that a selection implied that at least one character was selected and in this case, none are, so having `selection_present` being 1 and no characters selectected seems wrong to me.

I also can't think of any human way of producing this condition, am I wrong?  When a human selects something, it's a mouse down, then a drag a distance, then a mouse up.  The cursor would have to move to be a drag event and in this case, it didn't.

My guess as to why Putty does this is that it trys too avoid moving the cursor to this location on window refocus unless the user is actually trying to select something in the other window.  But this is only a guess.

nicm:
You can't create an empty paste buffer (`paste_set` and `paste_add` will not allow zero size buffers), so I think we should not emit them with OSC 52 so `window_copy_append_selection` and `window_copy_copy_buffer` should skip that bit if `len == 0`.

mgrant0:
For me, I'm just trying to work around some behaviour in Putty.  Maybe (probably!) there's a better way to do or avoid this behaviour all together?  I only noticed `selection_present` being set with a zero-length buffer when I went to try and use that format variable to work around this issue with Putty.

I get it why this is happening.  MouseDrag1Pane causes it to go into copy-mode followed by MouseDragEnd1Pane and under normal circumstances MouseDragEnd1Pane would be used to extend a selection.  In my workaround, I avoid doing the OSC 52 because I just call copy-mode -q.

Is there any reason to need to clear the selection buffer like this?  I can't really do that by hand.  You can't select nothing and then hit copy in windows.  So what you say make sense.

But regardless, I think I need some way to either avoid going into copy-mode when Putty does this, or some way to check if something actually got selected.  If not `selection_present`, maybe a `selection_length` format variable?  I didn't a length variable anywhere.

mgrant0:
Technically, this PR doesn't actually stop going into copy-mode.  It allows me to write a key binding to drop out of copy-mode if I got in there because of Putty causing an empty string to be copied.  

Do you have a better fix to avoid going into copy-mode in this situation?  That would be good, perhaps better, I just didn't see how to do that myself!

nicm:
I think this change is fine. I'm not sure we can do anything else since we don't know where a drag is going to end when it is started. I will apply it after the next OpenBSD release in a few weeks.

nicm:

PR #3872, opened 2024-03-06
Stop translating CSI u sequences back to Ctrl-keys

### LINKED ISSUE #2705: Possible regression: tmux silently swallows extended keys
### Issue description

My terminal emulator has special extended keys for C-i and C-m distinct from the standard translation of these key sequences to TAB and RET respectively.

Up until recently I used to be able to bind TAB and C-i independently without issue. For example after running `tmux bind-key -T prefix / command-prompt -k -p key "list-keys -1N "%%%""`. I can do `<prefix> / C-i` and it shows C-i, whereas `<prefix> / TAB` showed TAB.

However now `C-i` is shown as TAB and C-m is shown as RET.

I tried installing an older build (not sure about the specific commit, the version still says next-3.3) and this wasn't an issue there dispite using the same config. I believe this is a regression.

### Required information

* tmux version (`tmux -V`). `tmux next-3.3`
* Platform (`uname -sp`). `Linux unknown`
* $TERM inside and outside of tmux (`echo $TERM`). `st-256color`, `tmux-24bit`
* Logs from tmux (`tmux kill-server; tmux -vv new`). [older-server-log](https://github.com/tmux/tmux/files/6459489/tmux-server-82895.log), [master-server-log](https://github.com/tmux/tmux/files/6459498/tmux-server-88764.log)

realSaltyFish:
Fixes #2705 - see discussion there
This PR is WIP and not meant to be merged right away - it needs more work to examine the entire function and fix other potential problems.

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -985,8 +985,6 @@ tty_keys_extended_key(struct tty *tty, const char *buf, size_t len,
 		    onlykey != 13 &&
 		    onlykey != 27)
 			/* nothing */;
-		else if (onlykey >= 97 && onlykey <= 122)
-			onlykey -= 96;
 		else if (onlykey >= 64 && onlykey <= 95)
 			onlykey -= 64;
 		else if (onlykey == 32)


nicm:
This is here for a reason. tmux does not turn on modifyOtherKey=2 (only 1), but if it is turned on or the terminal chooses to believe it is, the terminal will send Ctrl keys as lowercase, so for example `^[[27;5;97~` for `C-a`. Without these two lines, that will be mapped to `'a'|KEYC_CTRL` rather than `'\001'`, meaning that tmux itself will not be able to detect `C-a`, because all the key bindings expect that `C-a` is `'\001`'.

You can see this easily if you change `Eneks` to send `\E[4;2m` instead of `\E[4;1m` and look at the logs. With your change:
~~~~
1709889946.381447 /dev/ttys014: keys are 10 (\033[27;5;97~)
1709889946.381498 /dev/ttys014: extended key \033[27;5;97~ is 200000000061 (C-a)
1709889946.381595 /dev/ttys014: complete key \033[27;5;97~ 0x200000000061
~~~~
But without:
~~~~
1709889998.511750 /dev/ttys014: keys are 10 (\033[27;5;97~)
1709889998.511783 /dev/ttys014: extended key \033[27;5;97~ is 1 (C-a)
1709889998.511798 /dev/ttys014: complete key \033[27;5;97~ 0x1
~~~~
If you want to move forward with modifyOtherKeys=2 or other input protocols, I think the first thing to do is to decide:

- How keys should be represented internally in an unambiguous way - for example, should `C-a` be `'\001'` or `'a'|KEYC_CTRL`? What about `C-A` -  is it `'A'|KEYC_CTRL` or `'a'|KEYC_CTRL|KEYC_SHIFT`?
- What metadata (such as how `KEYC_XTERM`) needs to be preserved with the key so tmux can generate the correct key into the pane, depending what protocol the application in the pane has requested?

Once the internal representation of each possible key is clear, it should just be a matter of fixing up or writing new mapping code for each protocol to map to tmux's internal key, and to do the reverse when writing a key to a pane.

I would probably consider breaking the code for extended keys out of `tty-keys.c` and `input-keys.c` into a new file.

PR #3874, opened 2024-03-08
fix comment

wellweek:


nicm:
Fixed in OpenBSD, will be in GitHub later, thanks!

PR #3888, opened 2024-03-16
Remove ~/.tmux.conf repetition from -f option in manual

rylenko:
The manual of -f option describes the procedure for searching for the configuration file. The ~/.tmux.conf is there twice.


nicm:
Applied now, thanks!

PR #3893, opened 2024-03-18
Removed redundant config file location from man page

jcompose:
The ~/.tmux.conf location was listed twice, so I just removed the second one.


jcompose:
Edit: just saw this pull request was already made by someone else.

PR #3919, opened 2024-04-01
Properly detect mouse SGR button release events and rapid clicks of different buttons

dther:
I'm trying to implement mouse chording in another application, and stumbled upon two issues which resulted in valid mouse events being dropped by tmux.

- Mouse release events (SGR or otherwise) are only detected on the last clicked button, preventing underlying applications from maintaining accurate mouse state
- Clicking two different buttons in succession would sometimes result in the second button press being undetected due to the double-click timeout

These two commits fix those issues. In my testing, all mouse events emitted by suckless st 0.9.1 are now properly detected and passed to the underlying application. Theoretically, this improved state tracking means that mouse chording could be implemented within a tmux binding... But I haven't tested that.

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -644,7 +644,10 @@ server_client_check_mouse(struct client *c, struct key_event *event)
 				log_debug("triple-click at %u,%u", x, y);
 				goto have_event;
 			}
-		} else {
+		}
+
+		/* DOWN is the only remaining event type. */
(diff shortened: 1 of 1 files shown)


nicm:
Looks sensible enough to me, applied to OpenBSD now, will be in GitHub later. Thanks!

PR #3930, opened 2024-04-09
support ignore osc 2 after user specify pane title

cache-missing:


nicm:
This shouldn't be the default, I suggest adding an option `allow-title`. You can either make it `on`/`off` like `allow-rename` and just always forbid it, or have three states`on`/`off`/`unless-user-set` (not wild about the name of the third value).

rmGFW:
> This shouldn't be the default, I suggest adding an option `allow-title`. You can either make it `on`/`off` like `allow-rename` and just always forbid it, or have three states`on`/`off`/`unless-user-set` (not wild about the name of the third value).

ok, i add a new option named `allow-rename-pane`

nicm:
This is not the pane name it is it's title so use a different option name please.

cache-missing:
> This is not the pane name it is it's title so use a different option name please.

how about `allow-rename-title`

nicm:
Also please add the new option to tmux.1.

nicm:

Issue #3940, opened 2024-04-13
 There is a memory leak defect on line 233 of the file /tmux/layout-custom.c.


LuMingYinDetect:
Dear developers of the tmux project,

Hello! I have discovered a memory leak defect in the latest version of tmux. The defect is located on line 233 of the file /tmux/layout-custom.c. The specific path to the defect is as follows:

1.In the file /tmux/layout-custom.c, a pointer variable named 'lc' is defined at line 159. This pointer is allocated a block of dynamic memory through the function layout_construct at line 176. When the if statement at line 231 evaluates to true, the function returns at line 233. During this process, the dynamic memory area pointed to by 'lc' is neither freed nor released as it is done at line 259, thus resulting in a memory leak defect, as illustrated in the following diagram:
https://github.com/LuMingYinDetect/tmux_defects/blob/main/tmux_4.png

Issue #3939, opened 2024-04-13
There is a memory leak defect on line 80 of the file /tmux/cmd-confirm-before.c.


LuMingYinDetect:
Dear developers of the tmux project,

Hello! I have discovered a memory leak defect in the latest version of tmux. The defect is located on line 80 of the file /tmux/cmd-confirm-before.c. The specific path to the defect is as follows:

1.In the file /tmux/cmd-confirm-before.c, a pointer variable named 'cdata' is defined at line 70, and at line 77, the xcalloc function is used to allocate a block of dynamic memory for this pointer. When the if statement at line 79 evaluates to true (indicating the failure of the assignment to cdata->cmdlist at line 78, but since the code uses a member of the pointer cdata, it implies that the dynamic memory allocation for cdata was successful), the program returns at line 80. During this process, the dynamic memory area pointed to by 'cdata' is not freed, leading to a memory leak, as illustrated in the following diagram:
https://github.com/LuMingYinDetect/tmux_defects/blob/main/tmux_3.png

nicm:
Fixed by #3941, thanks!

Issue #3938, opened 2024-04-13
 There is a memory leak defect on line 146 of the file /tmux/cmd-command-prompt.c.


LuMingYinDetect:
Dear developers of the tmux project,

Hello! I have discovered a memory leak defect in the latest version of tmux. The defect is located on line 146of the file /tmux/cmd-command-prompt.c. The specific path to the defect is as follows:

1.In the file /tmux/cmd-command-prompt.c, a pointer variable named 'cdata' is defined at line 87. The program allocates a block of dynamic memory for the 'cdata' variable using the function xcalloc at line 98. When the if statement at line 144 evaluates to true, the program returns at line 146 without executing the release operation for the dynamic memory area pointed to by 'cdata' at line 159. This leads to a memory leak defect, as illustrated in the following diagram:
https://github.com/LuMingYinDetect/tmux_defects/blob/main/tmux_2.png

nicm:
Fixed by #3941, thanks!

Issue #3937, opened 2024-04-13
A memory leak defect exists on line 174 of the file /tmux/arguments.c.


LuMingYinDetect:
Dear developers of the tmux project,

Hello! I have discovered a memory leak defect in the latest version of tmux. The defect is located on line 174 of the file /tmux/arguments.c. The specific path to the defect is as follows:

1.In the file /tmux/arguments.c, a pointer variable named 'new' is defined at line 151. This variable allocates a block of dynamic memory using the function xcalloc at line 154. When the if statement at line 171 evaluates to true, the program returns at line 174. During this process, the dynamic memory area pointed to by the 'new' pointer is neither used nor freed, thus resulting in a memory leak, as illustrated in the following diagram:
https://github.com/LuMingYinDetect/tmux_defects/blob/main/tmux_1.png

nicm:
Fixed by #3941, thanks!

PR #3941, opened 2024-04-13
Fix memory leak

Sberm:
Fix memory leaks pointed out by issues: #3940 #3939 #3938 #3937


nicm:
Applied with two changes (can't use `cmd_confirm_before_free` because `cdata->cmdlist` is NULL, also `args_free_value` does not free the container so also need `free(new)`). Will be in GitHub later on. Thanks!

PR #3961, opened 2024-04-29
Fix copy-mode-test-emacs.sh

JoyceLin2020:
The commands in the test script of the form `echo -e "..." ` was treating -e as a string resulting in the test script to fail. This PR fixes this by replacing `echo -e` with `printf`


nicm:
What about regress/copy-mode-test-vi.sh as well?

JoyceLin2020:
> What about regress/copy-mode-test-vi.sh as well?

Fixed

nicm:
Applied now, thanks!

PR #3962, opened 2024-04-29
Add CI regression checks

JoyceLin2020:
Add CI checks to run the regression tests located in the regress folder on every pull request and push to master

Example run: https://github.com/JoyceLin2020/tmux/actions/runs/8885335408/job/24396464277


nicm:
Thanks but I don't think I want these tests run on every commit, they are not 100% reliable, it is better if they are run manually.

PR #3973, opened 2024-05-10
Add hook command-error

hughdavenport:
I found this useful to have when an error happened, but none of the other hooks fired (as an error happened). Potentially others will find this useful?

### DIFF
--- a/cmd-queue.c
+++ b/cmd-queue.c
@@ -664,9 +664,17 @@ cmdq_fire_command(struct cmdq_item *item)
 
 out:
 	item->client = saved;
-	if (retval == CMD_RETURN_ERROR)
+	if (retval == CMD_RETURN_ERROR) {
+		fsp = NULL;
+		if (cmd_find_valid_state(&item->target))
+			fsp = &item->target;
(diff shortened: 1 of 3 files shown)


nicm:
I'm not `after-error` is a good name because most of the `after-` are commands. Maybe `command-error`?

Can you please add to tmux.1 also?

hughdavenport:
Done and done. First time editing a man page, but I think I got it working right. Tested with `man -l tmux.1`. Put a wee example similar to the `after-` hooks. I didn't think it was similar to the general hooks to be included in that list so had a separate paragraph for it. Let me know if you intended it to be part of the list of general hooks (like `alert-activity`).

Thank you for the quick response. I really appreciate the work you do for the project :)

nicm:
Thanks! Applied now to OpenBSD with a few very minor changes below, will be in GitHub later when it syncs.

- Fixed a long line in cmd-queue.c

- Man page writing is impersonal, so change not to address the reader as "you".

- Switched the example from notify-send which is not portable to echo, which is not terribly useful but everyone has.

- There is a list of hooks near the end where the new one also needs to be added.

Issue #3675, opened 2023-08-29
Keyboard interrupt a running search


YodaEmbedding:
### Issue description

1. Output a ton of data in the terminal window.
2. Now, run a regex search, e.g. `^hel*o`
3. It may take quite some time. Tmux freezes until the search completes.

Possible proposed fixes:

1. Ability to cancel the search via e.g. Ctrl-C, or (less discoverable, but more robust) Prefix+Ctrl-C, or something like that.
2. Perform search on a different non-UI thread. Lock user input on only the current pane, but not the window/session/client.

### Required information

Please provide the following information:

* tmux version: tmux 3.3a
* Platform: Arch Linux
* $TERM inside: tmux-256color; outside of tmux: alacritty.
* Logs from tmux: N/A.

nicm:
I cannot make a search take long enough to even hit the timeout now, so I don't know that this is necessary. What are you searching for and in how much history?

YodaEmbedding:
Looks like it's a problem only if searching for contains certain regex characters (e.g. `^`).

Searching backwards for the regex `^hello` is problematic. In contrast, a non-regex `hello` searches nearly instantly.

Here's a reproducible example:

```
echo "set  -g history-limit 20000" > example.conf
tmux -f example.conf
python -c 's = "0123456789" * 100000; print(s); print("hello world"); print(s)'

<C-b>[?^hello<CR>
```

If that doesn't freeze for long enough (depends on your CPU), increase the magnitude by 10x or 100x. Note that sufficiently large numbers may be prone to crashes.

---

Forgot to mention earlier: I have also noticed that tmux is prone to crashing if the amount of text to search is very large (or is it because the text contains "binary output data"? not sure), and I try to do some actions while that search is running (C-c, switch panes, etc).

nicm:
Yes, we do far too much work when you have very log lines. Please try this: [x.diff.txt](https://github.com/tmux/tmux/files/12509193/x.diff.txt)

We should probably make this part of searching have the timeout as well (timeout is currently only applied to marking search results).

YodaEmbedding:
`WINDOW_COPY_SEARCH_MAX_LINE` seems to make everything much more tolerable, including my usual use-case of accidentally printing a ton of long lines and then running a regex.

It's a bit strange that a regex on a long line takes many hours to run, though. For comparison, python runs a similar regex search in 0.005s:

```python
import re
import timeit

[...]

Issue #3864, opened 2024-02-29
searching is broken


benoit-pierre:
### Issue description

43e5e80343185e69a1b864fc48095ede0b898180 broke search for wrapped lines: when the pattern occurs before the wrap, it's correctly highlighted, but ignored by the search.

Example, execute the following command:
```bash
▸ a=haystack b=needle; for n in $(seq 5); do printf '%*s%*s%*s\n' $((COLUMNS/2)) $a $((COLUMNS/2)) $b $((COLUMNS/2)) $a; echo; t=$a a=$b b=$t; done
```
Then try a backward search for `needle`: on lines that wrap, matches that occur before the wrap are ignored (even if correctly highlighted).

Reverting 43e5e80343185e69a1b864fc48095ede0b898180 fix the issue.

### Required information

* tmux version: `3.4`.
* Platform: Arch Linux.
* `$TERM` inside and outside of tmux: `tmux-256color` / `alacritty`.

yan12125:
Thank you for the investigation! I also get fully-working search back after reverting that commit.

As a short-term workaround, I propose to revert that commit in Arch Linux package: https://gitlab.archlinux.org/archlinux/packaging/packages/tmux/-/issues/2

moralground:
@ThomasAdam @nicm Any update on this?

ThomasAdam:
@moralground -- No, and like all other issues, patience is a virtue.  When there's an update, you'll know.

yan12125:
I took a look, and it seems 43e5e80343185e69a1b864fc48095ede0b898180 is not doing as intended.

In `window_copy_search_jump` function, grid lines with the flag `GRID_LINE_WRAPPED` are skipped. In a wrapped line, all grid lines other than the last one have this flag, so only the last grid line is passed to `window_copy_search_rl_regex` or a similar function. As a result, only the last grid line is searched.

I assume the intention was to passing the first grid line to `window_copy_search_rl_regex` or a similar function, so that the inner loop will search from the first grid line to the end of a wrapped line. I'm not sure how to fix it, as I didn't find a special flag to identify the first grid line.

nicm:
Yes, this was intended to skip lines that are already being visited as part of a wrapped line. So it needs to skip all wrapped lines except the first instead of skipping all but the last. I don't have time to get this to work as intended so I have reverted this part of the change for the moment. This will be in GitHub later on.

PR #3975, opened 2024-05-11
Allow searching wrapped lines to be enabled through configuration

MeanderingProgrammer:
## Details

This resolves issue [3864](https://github.com/tmux/tmux/issues/3864) by adding a new opt-in (off by default) flag `search-wrapped-lines`.

Enabling this flag will cause searches to scan the entirety of wrapped lines at the cost of slower speeds. This is especially true with regexes as described in the original issue [3675](https://github.com/tmux/tmux/issues/3675).

I figured it would make sense to keep the default behavior as is, rather than the other way around, but was 50/50.

If there is anything else you would like me to test / information to add please lmk! I've been using TMUX for most of my career, really love this tool, and appreciate all the effort being put in here!

If you had an entirely different approach involving optimizing performance rather than this kind of forked behavior, or something else entirely that's completely understandable.

## Testing

Was able to confirm locally that setting `set -g search-wrapped-lines on` slows down regex searches while finding all matches in wrapped text.

Removing this setting seems to lead to the current behavior, faster searches but missing matches in long lines of wrapped text.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -383,6 +383,14 @@ const struct options_table_entry options_table[] = {
 	  .text = "Maximum number of commands to keep in history."
 	},
 
+	{ .name = "search-wrapped-lines",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_SERVER,
+	  .default_num = 0,
+	  .text = "Whether to include full wrapped lines when searching for "
+		  "text in copy mode."
+	},
+
 	{ .name = "set-clipboard",
 	  .type = OPTIONS_TABLE_CHOICE,
 	  .scope = OPTIONS_TABLE_SERVER,
--- a/tmux.1
+++ b/tmux.1
@@ -3797,6 +3797,17 @@ each client.
 .It Ic prompt-history-limit Ar number
 Set the number of history items to save in the history file for each type of
 command prompt.
+.It Xo Ic search-wrapped-lines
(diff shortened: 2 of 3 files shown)


nicm:
Can you add to tmux.1 as well please?

MeanderingProgrammer:
> Can you add to tmux.1 as well please?

Done!

nicm:

nicm:
I'm actually going to revert this, because the check in `window_copy_search_jump` is going away and I don't have time to get it to work with the functions further down the call tree - also it should not highligh matches it is not going to visit when you search again. If you'd like to revisit it then please do.

MeanderingProgrammer:
Removing the check in `window_copy_search_jump` accomplishes exactly what I was looking for (being able to search wrapped lines), and makes it the default behavior which I think is great and would be my preference.

I thought the decision to add the skipping behavior was intentional, sacrificing accuracy in favor of performance, to resolve issue https://github.com/tmux/tmux/issues/3675. It seems like I misunderstood the expected behavior here, and this is definitely because of my lack of familiarity with the codebase. From my understanding now it seems like wrapped lines have some redundancy in how they are processed, the original change was meant to remove the redundancy, but instead ended up ignoring the lines.

I will see if I can implement the optimization to avoid the redundant processing on wrapped lines.

Thank you for looking over this PR, looking into the problem further, and your amazing contributions as a whole!

Issue #3866, opened 2024-03-01
Feature request: repeat search backward with N in choosers


lobre:
### Required information

Please provide the following information:

* tmux version (`tmux -V`): `tmux 3.4`.
* Platform (`uname -sp`): `Linux x86_64`.

### Issue

Currently, when triggering "choosers" (`choose-client`, `choose-tree`, `customize-mode`, `choose-buffer`), there is the possibility to search with `/` or `C-s`. After a search has been entered, it is repeatable with `n`. However, there does not seem to be an opposite motion to "repeat backwards".

Here is the extract of the man page.

```
C-s    Search by name or content
n      Repeat last search
r      Reverse sort order
```

In other places in tmux such as in copy-mode, there is the opposite motion to search backwards assigned to `N` (uppercase `n`).

> The search commands come in several varieties: `search-forward` and `search-backward` search for a regular expression; the `-text` variants search for a plain text string rather than a  regular  expression;  `-incremental` perform  an  incremental  search  and expect to be used with the -i flag to the command-prompt command.  `search-again` repeats the last search and `search-reverse` does the same but reverses the direction (forward becomes backward and backward becomes forward).

To stay consistent, my suggestion would be to also add a "Repeat last search backwards" in the choosers assigned to the `N` key.

Note: As a current workaround, it is possible to use `r` in the chooser to reverse the order. Then, the next `n` would repeat still forward, but as the order of the list has changed, it kind of has the effect of "searching backwards". However, I think the ergonomics are not as good as adding a `N` action.

Would the maintainers be interested in such a change? Thank you.

nicm:
Yes this is a good idea, if you want to do it you need to look at `mode-tree.c`.

FadiAfani:
I can try working on this :)

PR #3982, opened 2024-05-16
added an option to repeat the last search backwards, assigned to the …

FadiAfani:
this should address #3866, I've added an option to repeat the last search but backwards using the 'N' key as requested and for that I wrote a function that traverses the tree backwards, I've also added a new field to 'mode_tree_data' which controls the search direction (which is flipped when the user pushes the respective key).


nicm:
Look good, applied to OpenBSD now with some minor style fixes, will be in GitHub later. Thanks!

PR #3995, opened 2024-06-02
Adds Yacc as a dependency.

johnnymatthews:
As reference in https://github.com/tmux/tmux/issues/2396 (closed as stale), Yacc is a dependency for compiling Tmux. This PR: 

1. Adds `yacc` to the list of dependencies.
2. Organizes the dependencies in an unordered list.

The link to download yacc was taken from the [Wikipedia article](https://en.wikipedia.org/wiki/Yacc). There may be a better location to point people to, but I can't find anything else.


nicm:
It is already mentioned in the following paragraph. If we were going to link to one, bison would be best probably rather than some ancient version, but I don't think it's necessary, just saying yacc is enough like just saying "a C compiler" is enough.

nicm:
I think it is fine the way it is TBH. The README is meant to be compact, the longer form version is https://github.com/tmux/tmux/wiki/Installing.

PR #4014, opened 2024-06-22
Add refresh-client -r

gnachman:
I received an [issue report](https://gitlab.com/gnachman/iterm2/-/issues/11658) from a user stating that in control mode `OSC 11 ; ?` responds differently than outside control mode. This PR attempts to address this problem.

Outside of control mode, tmux queries the client with `OSC 10 ; ?` and `OSC 11 ; ?` and remembers the foreground and background colors of the client. When only a control mode client is attached, `struct tty` is not populated and a default value was used.

This PR adds `refresh-client -r` as a mechanism for control mode clients to preemptively tell tmux what the correct responses to `OSC 10 ; ?` and `OSC 11 ; ?` are. In the future it could be extended to other reports, such as `DA`.

If there are multiple concurrent control mode clients, the most recently set value is used. I believe a similar problem exists outside control mode, although I had trouble following the logic. The control-mode fg and bg colors are ignored if no control mode clients are attached.

I chose to make the values per-pane because in control mode a client could use a different color scheme for each window pane. If `refresh-client -r` were extended in the future to support non-per-pane values (e.g., `DA`) then the requirement to begin with a window pane ID could be relaxed or a different flag could be used.

### DIFF
--- a/cmd-refresh-client.c
+++ b/cmd-refresh-client.c
@@ -34,9 +34,9 @@ const struct cmd_entry cmd_refresh_client_entry = {
 	.name = "refresh-client",
 	.alias = "refresh",
 
-	.args = { "A:B:cC:Df:F:l::LRSt:U", 0, 1, NULL },
+	.args = { "A:B:cC:Df:r:F:l::LRSt:U", 0, 1, NULL },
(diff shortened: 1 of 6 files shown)


nicm:
Looks good to me, applied now with some trivial style changes. Thanks!

PR #4030, opened 2024-07-07
Install the mdoc format man page on illumos

citrus-it:
illumos uses mandoc to render man pages and therefore needs the mdoc format page installed as-is. As an OpenSolaris derivative it contains solaris in the host triple, so an additional check against `uname -o` is required to differentiate.


nicm:
Applied now, thanks!

PR #4036, opened 2024-07-10
Update tmux.1

sol:


sol:
@ncim `RGB` is not documented in `terminfo`, no?

sol:
Actually, you would get a follow up PR, because the `RGB` feature in `tmux` is very different from what `terminfo` assume.

nicm:
The capabilities are documented in terminfo(5).

sol:
I think `RGB` is an `ncurses` extension; it's documented in `user_caps(5)`.

However, I think the more significant issue here is that `RGB` not only indicates `truecolor` support, but also redefined the semantics of `setaf` / `setab` in a noncompatible way.

From what I understand `Tc` is just used to indicate `truecolor` support.  For that reason saying 

> This (== Tc) is equivalent to the RGB terminfo(5) capability.

is, from what I understand, an inaccuracy.

nicm:
Yes, I know but we are referring to a "terminfo capability" and the page is that describes terminfo is terminfo(5), "a user_caps(5) capability" does not make sense.

`Tc` and `RGB` are entirely equivalent for tmux, how other programs interpret them is not important for tmux(1).

sol:
> Yes, I know but we are referring to a "terminfo capability" and the page is that describes terminfo is terminfo(5), a "user_caps(5) capability" does not make sense.

In so far as that the old sentence is nicer to read, I agree with that. 👍 

> `Tc` and `RGB` are entirely equivalent for tmux, how other programs interpret them is not important for tmux(1).

Yes, that makes sense.

The thing that worries me is that `RGB` is frequently misunderstood.

I think a `tmux` user has basically three options to enable `truecolor` support (using `alacritty` as an example here, as `xterm` and `kitty` already seem to work out of the box):

1. `set-option -sa terminal-features ",alacritty*:RGB"` (preferred)
2. `set-option -sa terminal-overrides ",alacritty*:Tc"`
3. `set-option -sa terminal-overrides ",alacritty*:RGB"`

Technically there is a fourth option.  Give (3) an unexpecting user could be misled to "do things properly" and  select a `terminfo` description that sets `RGB` via `export TERM=alacritty-direct`.

While this works as long as you are only running `tmux`, it breaks 256-color applications run outside of `tmux`.

The man page already states that

> the `terminal-features` option should be used

which I think is is great 👍

Do you think it also makes sense to change

> This is equivalent to the RGB terminfo(5) capability.

To something along the lines of

> Note that while tmux treats `Tc` and the `RGB` terminfo(5) capability the same, selecting a `terminfo` description that sets `RGB` (such as `xterm-direct`) is not advisable as it breaks 16-color and 256-color applications when run outside of tmux.

?

ThomasDickey:
As Nicolas said, there's no need for this.  If you're using some application which doesn't check the number of colors anyway, that's definitely a bug -- in **that** application.

nicm:
But its not wrong to use `alacritty-direct` if that's what you want... applications that check `RGB` but still use `setaf`/`setab` with 256 colour numbers just have a bug, no?

ThomasDickey:
Actually, it's possible to have RGB for something other than 16 million colors.  The application has to check and decide how to handle the combination.  (ncurses probably will handle it).

sol:
> Actually, it's possible to have RGB for something other than 16 million colors. The application has to check and decide how to handle the combination.

That's actually even worse, no?  You wrote an application that checked for `tput colors` to return exactly 256, and suddenly, if a user runs your application with a 256-color direct `terminfo` description, `tput setaf 10` will instead of *green* select some shade of red maybe?

ThomasDickey:
Are you referring to an application that you wrote?  (If so, you ought to read the documentation and fix your bugs).

sol:
> Are you referring to an application that you wrote? (If so, you ought to read the documentation and fix your bugs).

To clarify, I'm not looking at this from the perspective of an application developer. I am looking at this from the perspective of a user who wants to decide on which `terminfo`description to select.

ThomasDickey:
That's actually a problem for Nicolas to deal with, because anyone using tmux is going to be using one of the terminal descriptions which it supports (`tmux` and `tmux-256color`).  The external value for `TERM` is outside that scope (not the user's problem).

nicm:
tmux should work because when sees `RGB` it uses its own copies of `setaf`/`setab` which still support 256 colours. I don't think it'd work with terminals which aren't 24-bit RGB but I don't know if any exist.

Whether the fact that `RGB` changes `setaf`/`setab` is handled properly by other applications outside tmux is not something we can really do anything about and I don't want to make recommendations in tmux(1) based on what they may or may not support. If you think this is something worth covering, a brief FAQ entry would be more appropriate.

PR #4037, opened 2024-07-10
Update example_tmux.conf

sol:
Use `terminal-features` instead of `terminal-overrides` to enable truecolor support.


nicm:
Applied, thanks!

PR #4038, opened 2024-07-10
Improve support for extended key reporting modes

crzcrz:
This PR is far from being ready to be merged, but is complete enough to be reviewed.

Fundamentally, this should fix the extended key handling, by fully implementing support for xterm modifyOtherKeys=1 and 2; and iTerm2 CSI u modes.

Key changes:

- Request `modifyOtherKeys=2` mode from the parent terminal (if configured),
- When receiving a key, always store it in the extended form, that includes fixing up C0 control codes that can’t be entered with a dedicated key to corresponding `Ctrl-` keys,
- When sending out a key, encode it according to the pane key mode,
- Store key bindings as extended keys, do not convert to C0 control codes.

On top of that I added support for iTerm 2 CSI u mode, which is very similar to xterm modifyOtherKeys=1 mode, but need not be explicitly requested by the client. To achieve that I have repurposed the `always` value of `extended-keys` configuration key. The exact logic is as follows:

- `extended-keys off` - do not request `modifyOtherKeys` from the parent terminal, output according to what the client has requested, (so standard, `modifyOtherKeys=1`, or `modifyOtherKeys=2`),
- `extended-keys on` - request `modifyOtherKeys=2` from the parent terminal, output according to what the client has requested,
- `extended-keys always` - do not request `modifyOtherKeys` from the parent terminal, output CSI u sequences if the client has not requested `modifyOtherKeys`, or one of the two extended modes otherwise.

Note that it may not be possible to form a correct `modifyOtherKeys=2` sequence when the input mode was anything else. In particular recovering the missing Shift modifier is not possible.

There are comments in the source code explaining the corner cases.

Some bugs were squashed in the process, in particular handling of modified UTF-8 keys was fixed.

A lot of things were probably broken as well, however the last time I ran the regression tests only the `TAB` key tests did not pass (because I have partly removed it from the key tree due to removal of `KEYC_EXTENDED` flag). `KEYC_IMPLIED_META` may also not be handled correctly in every added code path.

Key bindings still need work, but possibly mainly in the documentation area. In particular the subtleties of the `Shift-` modifier need to be explained, and why binding, say, `C-S-2` will never work.

This PR still needs a lot of work, and I’m determined to complete it, but I’m asking for feedback before continuing.

References:

1. https://invisible-island.net/xterm/modified-keys.html
2. http://www.leonerd.org.uk/hacks/fixterms/
3. https://github.com/tmux/tmux/issues/2705
4. https://github.com/tmux/tmux/pull/3872
5. https://en.wikipedia.org/wiki/ANSI_escape_code
6. https://invisible-island.net/xterm/ctlseqs/ctlseqs.html
7. https://unix.stackexchange.com/a/289871


nicm:
Thanks for working on this. Some initial comments:

- I don't see much point in supporting the xterm 27 form on output, most terminals and applications have converged on CSI u now. It is just adding complexity - I think it is OK to diverge from xterm here. It would be good to still accept it on input if possible, although I would be happy enough to discard that as well and just have tmux switch the outside terminal to CSI u by default.

- I think JUST_KEY is a bit pointless but if you must have it please make it KEYC_JUST_KEY.

- I think some of this should be broken off into separate PRs to make the diff smaller: utf8_fromwc, changes to KEYC_IS_UNICODE, and probably some other bits like the changes to the MODE_KEYS_* flags (even if some flags don't do anything to start with). You can do at any point you like.

The code itself largely looks good, but as you say there is much to do. Most obviously is to get keys handled by tmux itself working.

There are also some missing pieces in the emulation if an application does not request extended keys - for example `C-1` sends nothing when it should send `1`, similarly `C-S-3`, `C-'`, there are probably others.

crzcrz:
I did not have much time to work on this, but I still want to have some sort of cadence, so I addressed a quick and easy thing, which is `JUST_KEY` macro.

The reason I added it, was to avoid the following pattern, which I disliked:

```c
key_code onlykey;

/* A lot of code… */

onlykey = key & KEYC_MASK_KEY;
if (onlykey this || onlykey that) {
…
```

But this is not the hill I’m willing to die on, so I reverted it everywhere.

crzcrz:
@nicm when you say

> * I don't see much point in supporting the xterm 27 form on output, most terminals and applications have converged on CSI u now. It is just adding complexity - I think it is OK to diverge from xterm here. It would be good to still accept it on input if possible, although I would be happy enough to discard that as well and just have tmux switch the outside terminal to CSI u by default.

Do you mean simply dropping the `^[[27;m;n~` output form, so essentially eliminating the `switch` statement below, or do you mean more significant logic changes?

```c
switch (s->mode & EXTENDED_KEY_MODES) {
case MODE_KEYS_EXTENDED:
case MODE_KEYS_EXTENDED | MODE_KEYS_CSI_U:
case MODE_KEYS_EXTENDED_2:
case MODE_KEYS_EXTENDED_2 | MODE_KEYS_CSI_U:
	xsnprintf(tmp, sizeof tmp, "\033[27;%c;%llu~", modifier, key & KEYC_MASK_KEY);
	break;
case MODE_KEYS_CSI_U:
	xsnprintf(tmp, sizeof tmp, "\033[%llu;%cu", key & KEYC_MASK_KEY, modifier);
	break;
default:
	return (-1);
}
```

As for accepting it on input, I believe that's a must, just because of the robustness principle.

nicm:
Yes, can be removed there and also anything that lets the application select - CSI u would be the only option for extended keys.

crzcrz:
OK, just to make sure that I understand your proposal.

1. We don't react to `^[[>4;Xm` from clients at all, and don't send it out to the outside terminal either, essentially completely removing the support for `modifyOtherKeys` 1 and 2 modes,
2. We lose the ability to request the extended key reporting mode from the outside terminal, instead the user will have to enable the CSI u mode in their terminal settings,
4. If `extended-keys` configuration key is set, we output CSI u keys, otherwise VT10x keys.

In this case I believe we should drop support for `^[[27;m;n~` on input as well, due to differences in `Shift` modifier reporting compared to CSI u mode (and for which I still don't have an elegant solution in mind).

nicm:
No, that's not what I meant. We request `modifyOtherKeys` as now. If the application inside tmux requests extended keys, we send `CSI u`. If they don't, we send standard keys. We never send the function 27 form.

The big disadvantage of dropping 27 coming in as well is that xterm does not appear to have an escape sequence to switch `formatOtherKeys` so it won't work by default...

crzcrz:
I don't think that doing this will yield a good user experience, and I need to try to persuade you to reconsider.

The problem is that applications running in the terminal, and the same applications running in the same terminal, but inside a tmux session, will see different keys and may behave differently. I strongly believe that we should be as transparent as possible. Of course, full transparency is not possible in a multiplexor, as it would require somehow letting the client know that the outside terminal suddenly changed its capabilities, but at least we can faithfully emulate all key modes.

Let me make an example. Say I have configured `zle-line-init` and `zle-line-finish` Zsh hooks to emit `^[[>4;1m` and `^[[>4m` respectively, and then made fancy ZLE keybindings, distinguishing `C-W` and `C-S-W`, using `^[[27` form of course. Now if I start tmux, my keybindings will suddenly stop working.

There will be many other similar scenarios where things will break, resulting in people complaining and submitting issues.

I would appreciate if you let me implement this in its originally envisioned form, and the complexity is really not very high, compared to many other parts of tmux source.

nicm:
I don't feel strongly about it so keep it if you like, it was only to simplify things for a feature hardly any applications use.

tmux should absolutely not be as transparent as possible, entirely the opposite. tmux aims to present as consistent behaviour as possible to the application, no matter what the terminal outside does. If an application is started in tmux running in a terminal that sends CSI u and the same session is later attached to a terminal that sends the function 27 sequence, tmux should continue to send CSI u. If a session is attached simultaneously to 10 different terminals which all have different capabilities, the application should not see any change in tmux's behaviour, it is tmux's job to balance the differences between the outside terminals and tmux's own terminal capabilities. Consistency is important, not transparency.

crzcrz:
I agree completely, and this is how it will behave (within the limitations of remapping one key mode into another). Perhaps I used a wrong word. What I meant is that tmux should faithfully emulate a well-behaving terminal, and avoid doing some very heavily opinionated remapping even when the input mode matches the output mode.

Thank you for your feedback and support, I will carry on.

nicm:
Great, yes that makes sense to me. I'm OK with having both, it was just a thought since CSI u is more widely supported. Thanks for working on this!

crzcrz:
@nicm I believe that all keys are now handled correctly in all input and output modes. All tmux key bindings also work, except when a `Shift-` binding is made to an ASCII key, and the input mode is `CSI u`, which does not report the `Shift-` modifier for ASCII keys. I'm not yet sure how to best handle this.

I have reworked the `Tab` key handling as well. It was previously touched following #3366 and the associated PR, which may not have gotten it 100% right IMO. Distinction between `Tab` and `Backtab` was lost, modifiers were lost for `Backtab`, and neither was reported correctly with respect to output modes.

With my change we treat `S-Tab` as `Backtab` internally, and that sorts out the `^]]1;_Z` modifiers mystery, which I believe was discussed in #3366 as well.

Can you take a look please.

nicm:
Thanks. Here are a few comments:

- You also need this:
~~~~
--- a/options-table.c
+++ b/options-table.c
@@ -613,7 +613,7 @@ const struct options_table_entry options_table[] = {
        { .name = "prefix",
          .type = OPTIONS_TABLE_KEY,
          .scope = OPTIONS_TABLE_SESSION,
-         .default_num = '\002',
+         .default_num = KEYC_CTRL|'b',
          .text = "The prefix key."
        },
~~~~

- There are a good number of other Ctrl keys that look like they will need to change (something like `git grep -E "(=|case) '\\\([0-9]{3}|[a-z])'" *.c` looks like it will find most).

- If I press `M-§` then I get `^[[27;3;1107341762~` which doesn't seem right (iTerm2 send `^[§`).

- I think the `pane_key_mode` format variable would be more useful as flag names than a number.

-`MODE_KEYS_CSI_U` is always set when `extended-keys` is `always`, but I don't think that is right. I think `extended-keys on`, `off` and `always` should keep their existing meaning (so they only affect whether extended keys are on or off, not the format) and CSI u should either be additional values or a separate option (`extended-keys-format`?).
 
- I think this should be `p - standard_map` or `p - &standard_map[0]`:
~~~~
+                       key = standard_map[1][p - standard_map[0]];
~~~~

- `mode` here not `modea`:
~~~~
+               /* When in CSI u modea, add a flag to enable lookup of CSI u extensions. */
~~~~

nicm:
A minor thing, but how about making `key_string_lookup_key` treat ASCII < 32 differently since they are no longer how tmux represents Ctrl keys. Maybe show them as `\\%u` or `[NUL]` `[SOH]` etc or something? This means that if such keys do creep in anywhere, we can easily see they are different in the logs.

crzcrz:
> * If I press `M-§` then I get `^[[27;3;1107341762~` which doesn't seem right (iTerm2 send `^[§`).

A good catch. There are two issues here actually:

1. We output `M-` Unicode keys in the extended form in `modifyOtherKeys=1` mode. I have based this off xterm specs, where all `M-` ASCII keys are reported in the extended form (which differs from what iTerm2 does). I'll change that to send out `^[`-prefixed form (*).
2. All Unicode keys not reported correctly in the extended mode, as we don't convert from the internal representation into UTF-32. No idea how I could have missed that.

*) I'd be a bit careful about mimicking iTerm 2 behaviour exactly, in my opinion it's slightly buggy with respect to `modifyOtherKeys=1` mode, and definitely does not follow the xterm specs.

crzcrz:
> * You also need this:
> 
> ```
> --- a/options-table.c
> +++ b/options-table.c
> @@ -613,7 +613,7 @@ const struct options_table_entry options_table[] = {
>         { .name = "prefix",
>           .type = OPTIONS_TABLE_KEY,
>           .scope = OPTIONS_TABLE_SESSION,
> -         .default_num = '\002',
> +         .default_num = KEYC_CTRL|'b',
>           .text = "The prefix key."
>         },
> ```
> 
>  * There are a good number of other Ctrl keys that look like they will need to change (something like `git grep -E "(=|case) '\\\([0-9]{3}|[a-z])'" *.c` looks like it will find most).

Now I _finally_ understand what you meant by "to get keys handled by tmux itself working". I think I got all of them.

> * If I press `M-§` then I get `^[[27;3;1107341762~` which doesn't seem right (iTerm2 send `^[§`).

Fixed.

> * I think the `pane_key_mode` format variable would be more useful as flag names than a number.

Agree, fixed.

> * I think this should be `p - standard_map` or `p - &standard_map[0]`:
> 
> ```
> +                       key = standard_map[1][p - standard_map[0]];
> ```

No, the code is correct. `standard_map` is an array of two `char *`. We look up in `[0]`, then take the replacement char from `[1]` at the same position. I can split into two variables if you wish.

>  * `mode` here not `modea`:
> 
> ```
> +               /* When in CSI u modea, add a flag to enable lookup of CSI u extensions. */
> ```

Fixed.

crzcrz:
> * `MODE_KEYS_CSI_U` is always set when `extended-keys` is `always`, but I don't think that is right. I think `extended-keys on`, `off` and `always` should keep their existing meaning (so they only affect whether extended keys are on or off, not the format) and CSI u should either be additional values or a separate option (`extended-keys-format`?).

That was a temporary solution, to enable switching modes when testing it. I will now think of an ergonomic way to configure this.

nicm:
This looks good now apart from the last bits for choosing CSI u.

Can you please also review and include this diff as well which fixes some minor style issues (missing ()s around return, long lines, a couple of spelling mistakes in comments) and removes the ` quoting in comments (it looks odd and doesn't add anything): [x.diff.txt](https://github.com/user-attachments/files/16488589/x.diff.txt)

crzcrz:
> A minor thing, but how about making `key_string_lookup_key` treat ASCII < 32 differently since they are no longer how tmux represents Ctrl keys. Maybe show them as `\\%u` or `[NUL]` `[SOH]` etc or something? This means that if such keys do creep in anywhere, we can easily see they are different in the logs.

Would it be OK to use Unicode Control Pictures (U+2400 to U+241F) for that, so ␀, ␁, ␂, and so on; or should we limit it exclusively to ASCII rendering?

nicm:
Definitely stick to ASCII please.

crzcrz:
> * `MODE_KEYS_CSI_U` is always set when `extended-keys` is `always`, but I don't think that is right. I think `extended-keys on`, `off` and `always` should keep their existing meaning (so they only affect whether extended keys are on or off, not the format) and CSI u should either be additional values or a separate option (`extended-keys-format`?).

I think there are three ways to go about this.

1. Add more enumeration values to `extended-keys`,
2. Add another configuration key, modifying the effect of `extended-keys` (ugh..),
3. Keep `extended-keys` behaviour exactly as it was before this PR, but declare it deprecated, and add a new one, say `modern-keys`.

I feel that (3) is the most logical option. The values could be:

- `off` - Do not request anything from the parent, ignore `Eneks` from client and always output standard keys,
- `ext1` - Request mode1 from parent terminal, output according to the client request, so standard/mode1/mode2,
- `ext2` - Like `ext1`, but request mode2 from the parent terminal,
- `csiu` - Do not request anything from the parent terminal, output is always `CSI u` (the user is expected to enable `CSI u` mode in their terminal).

nicm:
I don't mind between 1 and 2 but I'm not happy with 3, it seems like it would confuse things unnecessarily. I think 2 makes the most sense TBH...

I don't see why CSI u should be restricted? It should probably be the default if extended keys are on, so the options should be any combination of:

1) ignore requests and never send extended keys (current off);

2) accept requests (current on, this can be the default);

3) ignore requests and always send extended keys mode 1 (current always);

4) ignore requests and always send extended keys mode 2.

And:

a) send CSI (default);

b) send fn key 27.

You can express all this with one option but it's going to have a lot of choices - I think two options makes more sense.

crzcrz:
@ncim the thing with `CSI u` is that it can't be requested. It's a configuration option in the terminal application to enable sending it. However any `Eneks` request will override it, and switch the terminal to sending mode1 or 2 keys. `Dseks` will enable it again. So in essence it's an extension to the standard mode.

As for always sending mode2 keys, I don't see much value in this, as it will break every "regular" terminal application, rendering all the usual Ctrl- keys unusable.

crzcrz:
I think adding a `csiu` enumeration value to `extended-keys` will be a reasonable compromise.

The behaviour is then to not request anything from the parent terminal, and send `CSI u` keys in the standard mode.

crzcrz:
> Can you please also review and include this diff as well which fixes some minor style issues (missing ()s around return, long lines, a couple of spelling mistakes in comments) and removes the ` quoting in comments (it looks odd and doesn't add anything): [x.diff.txt](https://github.com/user-attachments/files/16488589/x.diff.txt)

Done. I must add I'm a bit embarrassed by the sheer number of typos and formatting issues in my code...

nicm:
> the thing with CSI u is that it can't be requested.

Yes, but CSI u needs to be available for both `on` and `always`, not just `on`, and I think it should probably be the default. If we add to `extended-keys` then two new choices should enable the same (`on`/`always`) but with function key 27.

I'm a bit concerned that we are going to run into the same problem if someone adds another type of extended keys and we will end up with another two choices for `extended-keys` and another and so on.

I'm fine with no option to always have mode 2.

nicm:
We could always invent an escape sequence to request CSI u as well of course, maybe add a third argument to `CSI > m`. Or is there any prior art?

crzcrz:
I think the confusion here is that you treat `CSI u` as an output format, whereby I see it as a mode, distinct from modes 1 and 2. So there's no need for us to send out `CSI u`-formatted mode 1 or 2 keys.

This is exactly how iTerm2 behaves. If the client has not requested anything, it outputs standard, or `CSI u` keys, depending on the checkbox in the settings. Otherwise it outputs mode 1 or 2 keys, following the client application request.

We can invent an escape sequence of course, but then we will have to patch every terminal and every client application in the world to support it, which is probably unfeasible.

nicm:
OK I was just thinking we need a way for the user to be able to say "I want applications that request extended keys to receive CSI u", the same as changing `formatOtherKeys` to `1` in xterm. But maybe this is actually never useful.

Note however that at the moment with `extended-keys always`, tmux will send `CSI u`, and this should probably stay the same or it will break configs.

I don't suggest we change any applications, if we add an escape sequence then applications will use it if they need to, particularly if other terminals adopt it as well.

crzcrz:
OK. I see your point now. There are two decisions to make then.

1. Do we see value in having an equivalent of `formatOtherKeys=1` _as xterm implements it_, that is to purely switch the output format.
2. Do we see value in having an equivalent of `CSI u` mode _as iTerm2 implements it_, where all key sequences not representable in the standard mode are sent out in the extended form, without the application having to request anything.

I do have an argument for supporting (2), as it makes it possible to have advanced key bindings in Zsh without any extra configuration.

As for (1), I'm not sure.

nicm:
Yes, I'm not sure either. *Are* there applications out there that only understand CSI u? It is certainly my impression that all the people using custom bindings want CSI u, but I don't know about applications. Perhaps it is unlikely since there is no way for them to request it...

I think the iTerm2 way is probably fine, but we can't change the format that `always` sends by default. So I reckon we have:
- `off` - never send extended keys;
- `on` - only send extended keys if the application requests, in which case send function key 27 mode 1 or 2;
- `always` - send CSI u unless the application requests, in which case send function key 27 mode 1 or 2.

I imagine if anyone else comes along with yet another key format (someone opened a PR about the Kitty one recently but I didn't look at it yet), it will have its own escape sequence to turn it on and can just be rolled into `on` and `always` using that.

crzcrz:
> Definitely stick to ASCII please.

Done.

crzcrz:
@nicm I think I managed to come up with a reasonable implementation, while maintaining backwards compatibility with existing user configs.

Two changes were made.

1. A `csi-u` enumeration value was added to `extended-keys` option,
2. A new `extended-keys-request` option was added, with possible values of `off`, `1`, and `2`, defaulting to `1`.

All permutations of `extended-keys` and `extended-keys-request` will work, but not all are useful. I will describe the most common ones.

### `extended-keys off`

The standard output mode is forced for all clients, and requested from parent terminals. `Eneks` from clients is ignored. Value of `extended-keys-request` has no effect.

### `extended-keys on`

The default output mode for clients is standard, switchable to mode 1 or 2 as per client request. The mode requested from the parent terminal follows the value of `extended-keys-request`. The sensible value for the latter is `2`, enabling reporting of _all_ possible key sequences. The default value of `1` will result in 3.4-compatible behaviour.

### `extended keys always`

Mode 1 is forced for all clients, `Eneks` and `Dseks` from clients are ignored. The default value `1` of `extended-keys-request` will result in 3.4-compatible behaviour.

### `extended-keys csi-u`

The default output mode for clients is `CSI u`, switchable to mode 1 or 2 as per client request. The mode requested from the parent terminal follows the value of `extended-keys-request`. The sensible values are `2` or `off`. The latter would rely on the terminal being configured to always emit extended key sequences.

In summary, sensible settings in the "new world" are `extended-keys csi-u` and `extended-keys-request off`, or `extended-keys on` and `extended-keys-request 2`. Old configurations (with `extended-keys-request` unset), will be 3.4 compatible.

nicm:
I don't like this. What is the need to let the user set what mode tmux requests? tmux can always request mode 2 and if the user really doesn't like it they can use `terminal-overrides`, there is no need for a new option for this.

We should pick a scheme similar to one of the two existing implementations we have been talking about:

1) like iTerm2: if the application asks for extended keys it gets what it asks for (function key 27 and either mode 1 or 2); if it doesn't ask then tmux either defaults to CSI u mode 1 or to no extended keys.

2) (sort of) like xterm: extended keys can be on or off, although I would extend this to also include "only on if application asks for them"; format (function key 27/CSI u) is controlled by a separate option.

The first seems more sensible to me TBH, I do not see any realistic use case for the user turning on mode 2 themselves, and it seems better to let applications always get the same format instead of letting the user pick for them.

crzcrz:
No problem, I will change it.On 11 Aug 2024, at 17:44, Nicholas Marriott ***@***.***> wrote:﻿
I don't like this. What is the need to let the user set what mode tmux requests? tmux can always request mode 2 and if the user really doesn't like it they can use terminal-overrides, there is no need for a new option for this.
We should pick a scheme similar to one of the two existing implementations we have been talking about:


like iTerm2: if the application asks for extended keys it gets what it asks for (function key 27 and either mode 1 or 2); if it doesn't ask then tmux either defaults to CSI u mode 1 or to no extended keys.


(sort of) like xterm: extended keys can be on or off, although I would extend this to also include "only on if application asks for them"; format (function key 27/CSI u) is controlled by a separate option.


The first seems more sensible to me TBH, I do not see any realistic use case for the user turning on mode 2 themselves, and it seems better to let applications always get the same format instead of letting the user pick for them.

crzcrz:
@nicm I've removed the `extended-key-request` option, which is completely superseded by the existing `terminal-overrides` functionality.

I also went to study iTerm2 sources to better understand how the `CSI u` mode works. It turns out that it is in fact an implementation of the Kitty Key Protocol, and enabling "CSI u mode" in settings, turns on what the Kitty spec calls [Disambiguate escape codes](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#disambiguate).

To the best of my knowledge this is not documented anywhere, iTerm2 is not mentioned as a compliant terminal on the Kitty page, and the issue regarding the Kitty Key Protocol in iTerm2 Gitlab is still open. So I got confused.

I need to rethink what to do with this PR. I still think there's value in the work done though, in getting rid of C0 codes internally, in xterm mode 2 support, and so on. We also appear to now have the (partial) Kitty Key Protocol implementation, save for push, pop, set, and query ANSI sequences.

I gave a quick scan to the other PR which was recently submitted, which seems very complicated, and it's not entirely clear if the issues of unambiguous internal representation, key bindings, remapping between modes, and internal keys are addressed there.

In any case, I'll wait for your input before continuing.

nicm:
If you want to break out the bits of the PR you are sure are good we can apply them at any time.

I don't think iTerm2 can be too far from xterm so it seems like to shouldn't be too much work to make this compatible.

crzcrz:
Very well. I will break out all the internal key handling changes + full xterm mode2+mode1 support into phase 1. All the `CSI u` changes will come later, as phase 2. This will take a bit of surgery, but should be doable.

nicm:
Does xterm do anything for CSI u other than change the escape sequence?

crzcrz:
To the best of my knowledge, no. Quoting:

```
formatOtherKeys (class FormatOtherKeys)
               Overrides the format of the escape sequence used to report
               modified keys with the [modifyOtherKeys](https://invisible-island.net/xterm/manpage/xterm.html#VT100-Widget-Resources:modifyOtherKeys) resource.
               0  send modified keys as parameters for function-key 27
                  (default).
               1  send modified keys as parameters for CSI u.
```

nicm:
So if you have xterm mode 1 and 2 working then why can't you also have CSI u, just change the escape sequence?

I think we should probably be compatible with xterm and if there are other modes (Kitty, iTerm2, whatever) they should be configured differently.

crzcrz:
Yes, of course. What I wanted to say is that iTerm2/Kitty stuff will be phase 2.

nicm:
OK that sounds good.

crzcrz:
@nicm I did some git surgery and squashed too many commits into just two. I hope it makes more sense now.

dbaae17 introduces support for xterm `modifyOtherKeys=2`, disambiguation of C0 code and `Ctrl-` keys (Such as `Tab` vs. `^I`), remapping of any input into any output mode, and configurable output format. So all the good stuff.

86ed192, which I will ask you to ignore for now, is a WIP support for iTerm2 `CSI u` mode (which is in fact Kitty Key Protocol as we've learned).

nicm:
OK thanks, I am away until the 22nd but I will look when I'm back.

nicm:
https://github.com/tmux/tmux/commit/dbaae17a2d0abb817d1f2406344838e0b43611c2 looks OK to me, but can you also add `extended-keys-format` and `pane_key_mode` to `tmux.1`?

crzcrz:
Done. Please excuse me if I did not use all the correct macros, as I'm not intimately familiar with roff, and even less so with mandoc.
 
To ease proofreading, I'm including the formatted version below.

```
     extended-keys [on | off | always]
             Controls how modified keys (so keys pressed together with
             Control, Meta, or Shift) are reported.  This feature is also
             known as modifyOtherKeys protocol in xterm(1).

             With respect to client applications, when set to on, the output
             will follow what the client has requested, so the standard mode,
             mode 1, or mode 2.  When set to always, the client requests are
             ignored, and mode 1 output is forced.  When set to off, the
             client requests are ignored and the standard output mode is
             forced.  The selected output mode is reflected in pane_key_mode
             variable.

             With respect to parent terminals, when set to on or always, the
             escape sequence to enable extended keys is sent, if tmux knows
             that it is supported.

             Regardless of what is configured, tmux always recognises extended
             keys itself.

             If unsure, set to on.

     extended-keys-format [csi-u | xterm]
             Selects one of the two possible formats for reporting modified
             keys to client applications.  This is the equivalent of
             formatOtherKeys xterm(1) resource.  For example, C-S-a will be
             reported as ‘^[[27;6;65~’ when set to xterm, and as ‘^[[65;6u’
             when set to csi-u.

             If unsure, set to xterm.

...

 The following variables are available, where appropriate:

     Variable name          Alias    Replaced with
...
     pane_key_mode                   Extended key reporting mode in this pane
```

nicm:
OK I have applied this (the xterm commit plus the man page) to OpenBSD now, it will be in GitHub later when it syncs up. I made a couple of minor changes:

- I trimmed the man page text a bit, changed some of the terms and added a bit of explanation what mode 1 and mode 2 actually meaning. Although TBH it is not terribly clear unless you know what it does already, but we can't go on for too long in the man page.

- I changed the default for `extended-keys-format` to `xterm` and `extended-keys` to `on`. This seems more sensible and also meant the "If unsure" bits in the man page could go (which are too conversational anyway).

- I made a couple of minor spacing tweaks so you may have some small conflicts.

If you have time it would be worth you reviewing https://github.com/tmux/tmux/wiki/Modifier-Keys and updating it for these changes.

Thanks!

nicm:
OK I have noticed one problem. `emacs` sends `\033[>4;1m` in both tmux and iTerm2 but then `M-<` stops working in tmux but not iTerm2.

nicm:
OK `TERM=xterm emacs` fixes it so this is clearly a bug in emacs, not sure what to do here, maybe `extended-keys` would be better left defaulting to `off`.

nicm:
LOL `C-Space` is broken as well in emacs. OK I am going to change the default for `extended-keys` to `off` again.

nicm:
OK I turned the default back to `off` and also made a fix to `input.c` so that off actually works :-):
~~~~
Index: input.c
diff -u -p -r1.227 input.c
--- input.c     21 Aug 2024 04:17:09 -0000      1.227
+++ input.c     21 Aug 2024 04:36:52 -0000
@@ -1412,7 +1412,7 @@ input_csi_dispatch(struct input_ctx *ict
                 * Set the extended key reporting mode as per the client request,
                 * unless "extended-keys always" forces us into mode 1.
                 */
-               if (options_get_number(global_options, "extended-keys") == 2)
+               if (options_get_number(global_options, "extended-keys") != 1)
                        break;
                screen_write_mode_clear(sctx,
                    MODE_KEYS_EXTENDED|MODE_KEYS_EXTENDED_2);
Index: options-table.c
RCS file: /cvs/src/usr.bin/tmux/options-table.c,v
diff -u -p -r1.176 options-table.c
--- options-table.c     21 Aug 2024 04:17:09 -0000      1.176
+++ options-table.c     21 Aug 2024 04:36:52 -0000
@@ -313,7 +313,7 @@ const struct options_table_entry options
          .type = OPTIONS_TABLE_CHOICE,
          .scope = OPTIONS_TABLE_SERVER,
          .choices = options_table_extended_keys_list,
-         .default_num = 1,
+         .default_num = 0,
          .text = "Whether to request extended key sequences from terminals "
                  "that support it."
        },

nicm:
Actually, I think this is a problem in tmux now. `M-<` is not meant to sent a translated sequence but it is:
~~~~
$ tmux set -g extended-keys on
$ printf '\033[>4;1m'
$ cat # press M-<
^[[27;3;60~
~~~

nicm:
This fixes C-Space as well, let me know:
~~~~
--- input-keys.c	21 Aug 2024 04:17:09 -0000	1.95
+++ input-keys.c	21 Aug 2024 04:53:37 -0000
@@ -547,17 +547,19 @@ input_key_mode1(struct bufferevent *bev,
 	 */
 	onlykey = key & KEYC_MASK_KEY;
 	if ((key & (KEYC_META | KEYC_CTRL)) == KEYC_CTRL &&
-	    (onlykey == '/' || onlykey == '@' || onlykey == '^' ||
+	    (onlykey == ' ' ||
+	     onlykey == '/' ||
+	     onlykey == '@' ||
+	     onlykey == '^' ||
 	     (onlykey >= '2' && onlykey <= '8') ||
 	     (onlykey >= '@' && onlykey <= '~')))
 		return (input_key_vt10x(bev, key));

 	/*
-	 * A regular or shifted Unicode key + Meta. In the absence of a
-	 * standard to back this, we mimic what iTerm 2 does.
+	 * A regular key + Meta. In the absence of a standard to back this, we
+	 * mimic what iTerm 2 does.
 	 */
-	if ((key & (KEYC_CTRL | KEYC_META)) == KEYC_META &&
-	    KEYC_IS_UNICODE(key))
+	if ((key & (KEYC_CTRL | KEYC_META)) == KEYC_META)
 		return (input_key_vt10x(bev, key));

 	return (-1);
~~~~

nicm:
I have applied this now rather than leaving it not working but if you prefer a different fix, let me know when you are able to. Thanks!

crzcrz:
I will take a detailed look at your feedback in the evening, during the business hours I’m not allowed to deal with that due to contractual reasons.

Just a quick note about emacs. First, you do need to set TERM to xterm(-256), so emacs loads the matching xterm.el. Second, it still won’t work well, because what they have there is a bit broken and you still need to patch the respective keymap with more ansi sequences for keys to work.

Once we work through all issues, and tmux with mode 2 support is released I want to attempt to submit a patch to emacs as well, as I spend all my days in emacs in a tmux session. Which is why I started all that.

nicm:
> First, you do need to set TERM to xterm(-256), so emacs loads the matching xterm.el. 

I don't think this is true any more, look at `tmux.el`:
~~~~
(defcustom xterm-tmux-extra-capabilities '(modifyOtherKeys)
~~~~
It works with the old `extended-keys` (and with my fix above).

I don't really care if all the extended keys work with emacs, the important thing is that standard ones don't stop working :-).

nicm:
Changes are all synced to GitHub now.

crzcrz:
Thank you for merging it, for your patience through all the iterations this went through, and for your feedback.

I will update the Wiki.

As far as your changes, I have just reviewed the xterm documentation again, and there's a contradiction with regard to Meta- modifier in mode 1.

In _modifyOtherKeys_ section it states:

```
                    The Alt- and Meta- modifiers do not cause  xterm  to  send
		    escape  sequences.	 Those	modifier  keys are interpreted
		    according to other resources,  e.g.,  the  metaSendsEscape
		    resource.
```

But then in the layout examples (which I have followed to the letter) it shows something else:

image

The real applications clearly expect the former. So your fix makes things right.

nicm:
Great, thanks. I'm going to leave it default off for the moment but I'll change it to on after it gets a bit of mileage.

nicm:
I'm happy to add the other CSI u variant but I would prefer to follow iTerm2 for that rather than Kitty.

crzcrz:
I will return to `CSI u` support after a short break. I want to make a reciprocal change to `xterm.el` first, so `C-S-letter` keys are also recognised.

nicm:
OK no worries, thanks.

nicm:
There is still some weirdness here, I am not sure exactly what is happening. Try this:
~~~~
$ tmux -Lx -f/dev/null set -g extended-keys on \; new
~~~~
Then run `cat` and press `C-b C-b`. There is a noticeable delay before the `^B` appears, compared with `extended-keys off`.

It looks like it is getting stuck behind the escape timer because setting `escape-time 0` improves it. But it should be a complete key, and the same does not happen with other keys, only `send-prefix`.

It gets worse if you run a nested tmux because it is has to wait for the escape timer twice...

nicm:
(Ignore that last comment about `send-prefix`, I don't think it is right.)

crzcrz:
> There is one problem standing in the way of `extended-keys` being on by default, and that is that mode 2 is extremely inconvenient if it is left on when tmux is killed (for example `~.` in ssh or cu).
> 
> I don't think there is much we can do about this since tmux needs them to be enabled all the time to handle the keys itself. The only option might be to make it turn on mode 1 by default and only turn on mode 2 if the pane requires it, like we do with the "all" mouse reporting mode.
> 
> I think it will probably just need to remain off by default.

I don't know how useful this response is, but iTerm2 can detect when an ssh session died and something undesirable was left on. I suppose we can ask to include mode 2 in the list of "undesirables". Other popular terminals may also have a similar feature.

image

craigbarnes:
> (which is in fact Kitty Key Protocol as we've learned)

This isn't entirely accurate. iTerm2's `CSU u` protocol is based on the old ["fixterms"](https://www.leonerd.org.uk/hacks/fixterms/) proposal, whereas Kitty's Keyboard Protocol is a [much improved](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#bugs-in-fixterms) variant of it that's now supported by several [other terminals](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#:~:text=this%20protocol%20is%20also%20implemented%20in). iTerm2 also has an open [issue](https://gitlab.com/gnachman/iterm2/-/issues/10017) tracking support for the full thing.

crzcrz:
> > (which is in fact Kitty Key Protocol as we've learned)
> 
> This isn't entirely accurate. iTerm2's `CSU u` protocol is based on the old ["fixterms"](https://www.leonerd.org.uk/hacks/fixterms/) proposal, whereas Kitty's Keyboard Protocol is a [much improved](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#bugs-in-fixterms) variant of it that's now supported by several [other terminals](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#:~:text=this%20protocol%20is%20also%20implemented%20in). iTerm2 also has an open [issue](https://gitlab.com/gnachman/iterm2/-/issues/10017) tracking support for the full thing.

I haven't analysed it in great detail yet, but there are source code comments in iTerm that reference the Kitty spec.

craigbarnes:
> ... there are source code comments in iTerm that reference the Kitty spec.

I think those probably came after the initial `CSI u` support, e.g. after [this issue](https://gitlab.com/gnachman/iterm2/-/issues/9506) and [this subsequent commit](https://gitlab.com/gnachman/iterm2/-/commit/bd3865bd94ef7f34e45d412350333d56f881a4ee).

Quoting myself:

> iTerm2's CSU u protocol is based on the old ["fixterms"](https://www.leonerd.org.uk/hacks/fixterms/) proposal

Source: https://gitlab.com/craigbarnes/dte/-/issues/130#note_864453071

nicm:
I'm going to close this at the moment, the thread is a bit big now anyway. I know you're still active (thanks) so when you get back to this, please either open a new PR or let me know and I can reopen.

PR #4039, opened 2024-07-11
Fix handling of UTF-32 code points in extended key sequences

crzcrz:
As suggested, I'm breaking off pieces from #4038 to make it smaller. Once this is merged, I will rebase #4038.

This fix ensures that UTF-32 codepoints in incoming extended key sequences are converted into internal representation. `utf8_fromwc` is implemented for both `--with-utf8proc` and `--without-utf8proc` build variants.

### DIFF
--- a/tmux.h
+++ b/tmux.h
@@ -3316,6 +3316,7 @@ void		 session_renumber_windows(struct session *);
 
 /* utf8.c */
 enum utf8_state	 utf8_towc (const struct utf8_data *, wchar_t *);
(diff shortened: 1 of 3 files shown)


nicm:
Originally UTF-8 keys were stored as Unicode, I must have missed this when I implemented the alternative internal representation. This looks OK I'll try to apply it tomorrow if I have time.

nicm:
This is done now, thanks!

Issue #4029, opened 2024-07-04
Apparent memory leak with large history, part 2


intelfx:
### Issue description

Similar to #1973, I'm experiencing what _looks_ like a memory leak, i. e. high memory consumption when running with a very large history-limit (e.g. 1M) for some time that persists after clearing history.

I understand this is likely due to glibc allocator behavior. There is a number of alternative allocator implementations for Linux (e.g. [jemalloc](https://github.com/jemalloc/jemalloc), [mimalloc](https://github.com/microsoft/mimalloc)), however tmux is relatively unique in that it spawns user processes. Thus, doing the usual LD_PRELOAD trick to override the libc allocator would be pretty complicated because you don't want to pass it down to the payload.

Perhaps it could be worth it to offer an option to explicitly build tmux against a different allocator?

### Required information

Please provide the following information:

* tmux version (`tmux -V`): `tmux 3.4`
* Platform (`uname -sp`): `Linux unknown`
* $TERM inside and outside of tmux (`echo $TERM`): `tmux-256color` / `xterm-256color`
* Logs from tmux (`tmux kill-server; tmux -vv new`): N/A

rfrancoise:
Doesn't directly address your question, but I'll note that 3.4 has a regression that makes it use x4 to x5 more memory than 3.3a, which I fixed in db16658 (see #4013).

rfrancoise:
Yes, tmux is an OpenBSD project and this repo is the portable version which receives changes via patches synced from CVS, see https://github.com/tmux/tmux/wiki/Contributing

nicm:
I'm OK with adding an option to `configure.ac` to build against one alternative malloc implementation if you want to open a PR.

PR #4040, opened 2024-07-13
Support building with jemalloc memory allocator

rfrancoise:
When `history-limit` is set to a large value, tmux can use lots of resident memory (sometimes in the GBs). On glibc-based systems, memory is returned to the system mostly when tmux calls `malloc_trim()` from an hourly cleanup timer. By contrast, jemalloc returns memory to the system almost immediately after closing a large pane (or exiting its copy mode), and supports nice improvements like using THP on Linux, etc.

Closes: #4029


nicm:
Applied now, thanks!

PR #4041, opened 2024-07-15
Rename header guards on compat/queue.h

saagarjha:
Some platforms (…macOS) have a <sys/queue.h> but it isn't full-featured enough to have all the functionality we need (e.g. TAILQ_REPLACE). This means the configure check fails and we fall back to our compatibility header, but if *someone else* happens to include the system header first then it will activate the _SYS_QUEUE_H_ header guard and ours will not define anything. Since we can't really control whether the system header is included or not (it is pulled in transitively) we probably should just rename ours so it doesn't conflict.

### DIFF
--- a/compat/queue.h
+++ b/compat/queue.h
@@ -32,8 +32,8 @@
  *	@(#)queue.h	8.5 (Berkeley) 8/20/94
  */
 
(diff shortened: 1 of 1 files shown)


nicm:
Applied now, thanks!

Issue #1799, opened 2019-06-18
[suggestion] Use tabs and spaces consistently


BourgondAries:
The source code mixes tabs and spaces for alignment, as well as indentation. I think this should be changed as it looks odd in editors that treat tabs and spaces differently

Here a sample, → is a tab.
```
→struct window_copy_mode_data→*data = wme->data;↲                                                                                                                                                                                                                                                                                                                                                                                         
→struct window_copy_cmd_state→ cs;↲
→enum window_copy_cmd_action→ action;↲
→const char→→→*command;↲
→u_int→→→→ i;↲
```

Here's how spaces and tabs are misused for indentation:
```
→→→matched = window_copy_search_compare(gd, px, py, sgd,↲                                                                                                                                                                                                                                                                                                                                                                                 
→→→    bx, cis);↲
→→→if (!matched)↲ 
```

I'm open to fix this if the maintainer is willing to accept it.

nicm:
No. https://man.openbsd.org/style.9

"Indentation is an 8 character tab. Second level indents are four spaces. All code should fit in 80 columns."

It used to be a rule to line up variable names in blocks of declarations, but I like it so I'm not changing it now even if it isn't required.

tmux is extremely consistent with these. Make sure your editor has tabs of width 8.

PR #4042, opened 2024-07-15
Add EditorConfig file

yochem:
In #1799 it is made clear that contributors should make sure that their editor is configured to use a tab width of 8. This file can help with that; it configures supported editors to have just that.

It is also used by Github for displaying the appropriate width in the browser.

I tried to be conservative with the file patterns used (for now only `*.c`, `*.h`, `Makefile.am` and `configure.ac` are set to have an indent style of tabs and a width of 8. But this list can of course be extended, or other [properties](https://editorconfig.org/#file-format-details) can be added.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[{*.c,*.h,Makefile.am,configure.ac}]
+indent_style = tab
+indent_size = 8


nicm:
I really don't think we want to carry around all sorts of random config files in the source repo, sorry.

PR #4043, opened 2024-07-15
Preserve paste buffer selection when deleting last buffer

d-e-s-o:
When in buffer selection mode, selecting the last buffer and deleting it, the first buffer in the list is selected. This behavior seems counter intuitive. It also does not really appear to be intended: we deliberate invoke mode_tree_down() with the 'wrap' flag set to false so as to *not* wrap selection around to the beginning. Yet, that is ultimately happening because the selection effectively gets invalidated (after all, we delete the selected element).
Adjust the logic so that we preserve the selection: when we fail to select the element below the current one (because it is the last one already), select the one above it instead.

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -261,19 +261,21 @@ mode_tree_up(struct mode_tree_data *mtd, int wrap)
 	}
 }
 
-void
+int
 mode_tree_down(struct mode_tree_data *mtd, int wrap)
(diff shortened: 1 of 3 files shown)


nicm:
Looks good, applied to OpenBSD with some minor tweaks - will be in GitHub later. Thanks!

PR #4054, opened 2024-07-25
Pass terminal data on socket on Cygwin platforms

rkitover:
On Cygwin and MSYS2 use a controlling pty for the server client handler
instead of passing a file descriptor over the socket which is not
possible on Cygwin platforms.

Attach an event to STDIN on the client, passing data from it to the
server and attach an event to the pty out_fd on the server client
handler and pass data from it to the client.

This fixes this issue:

https://github.com/microsoft/terminal/issues/5132

which also applies to other Windows terminals.

I am also including some patches that are included in the MSYS2
and Cygwin tmux packages in this pull request.

### DIFF
--- a/client.c
+++ b/client.c
@@ -244,6 +244,9 @@ client_main(struct event_base *base, int argc, char **argv, uint64_t flags,
 	char			*line = NULL, **caps = NULL, *cause;
 	u_int			 ncaps = 0;
 	struct args_value	*values;
+#ifdef TTY_OVER_SOCKET
+	struct tty*		 tty = xmalloc(sizeof *tty);
+#endif
 
 	/* Set up the initial command. */
 	if (shell_command != NULL) {
@@ -397,6 +400,10 @@ client_main(struct event_base *base, int argc, char **argv, uint64_t flags,
 	} else if (msg == MSG_SHELL)
 		proc_send(client_peer, msg, -1, NULL, 0);
 
+#ifdef TTY_OVER_SOCKET
+	tty_attach_stdin_to_socket(tty, client_peer);
+#endif
+
 	/* Start main loop. */
 	proc_loop(client_proc, NULL);
 
@@ -474,12 +481,14 @@ client_send_identify(const char *ttynam, const char *termname, char **caps,
 		    caps[i], strlen(caps[i]) + 1);
 	}
 
+#ifndef TTY_OVER_SOCKET
 	if ((fd = dup(STDIN_FILENO)) == -1)
 		fatal("dup failed");
 	proc_send(client_peer, MSG_IDENTIFY_STDIN, fd, NULL, 0);
 	if ((fd = dup(STDOUT_FILENO)) == -1)
 		fatal("dup failed");
 	proc_send(client_peer, MSG_IDENTIFY_STDOUT, fd, NULL, 0);
+#endif
 
 	pid = getpid();
 	proc_send(client_peer, MSG_IDENTIFY_CLIENTPID, -1, &pid, sizeof pid);
@@ -805,5 +814,10 @@ client_dispatch_attached(struct imsg *imsg)
 		system(data);
 		proc_send(client_peer, MSG_UNLOCK, -1, NULL, 0);
 		break;
+#ifdef TTY_OVER_SOCKET
+	case MSG_TTY_READ:
+		write(STDOUT_FILENO, data, datalen);
+		break;
+#endif
 	}
 }
--- a/compat.h
+++ b/compat.h
@@ -69,6 +69,11 @@
 #define __weak __attribute__ ((__weak__))
 #endif
 
+#if defined(_WIN32) || defined(__CYGWIN__) || defined(__MSYS__)
+#define WIN32_PLATFORM
+#define TTY_OVER_SOCKET
+#endif
+
 #ifndef ECHOPRT
 #define ECHOPRT 0
 #endif
--- a/server-client.c
+++ b/server-client.c
@@ -22,6 +22,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
@@ -2898,6 +2899,11 @@ server_client_dispatch(struct imsg *imsg, void *arg)
 	case MSG_READ_DONE:
 		file_read_done(&c->files, imsg);
 		break;
+#ifdef TTY_OVER_SOCKET
+	case MSG_TTY_WRITE:
+		write(c->fd, imsg->data, datalen);
+		break;
+#endif
 	}
 }
 
@@ -3069,6 +3075,7 @@ server_client_dispatch_identify(struct client *c, struct imsg *imsg)
 			c->cwd = xstrdup("/");
 		log_debug("client %p IDENTIFY_CWD %s", c, data);
 		break;
+#ifndef TTY_OVER_SOCKET
 	case MSG_IDENTIFY_STDIN:
 		if (datalen != 0)
 			fatalx("bad MSG_IDENTIFY_STDIN size");
@@ -3081,6 +3088,7 @@ server_client_dispatch_identify(struct client *c, struct imsg *imsg)
 		c->out_fd = imsg_get_fd(imsg);
 		log_debug("client %p IDENTIFY_STDOUT %d", c, c->out_fd);
 		break;
+#endif
 	case MSG_IDENTIFY_ENVIRON:
 		if (datalen == 0 || data[datalen - 1] != '\0')
 			fatalx("bad MSG_IDENTIFY_ENVIRON string");
@@ -3109,8 +3117,8 @@ server_client_dispatch_identify(struct client *c, struct imsg *imsg)
 	c->name = name;
 	log_debug("client %p name is %s", c, c->name);
 
-#ifdef __CYGWIN__
-	c->fd = open(c->ttyname, O_RDWR|O_NOCTTY);
+#ifdef TTY_OVER_SOCKET
+	c->fd = open("/dev/ptmx", O_RDWR|O_NOCTTY);
 	c->out_fd = dup(c->fd);
 #endif
 
--- a/tmux-protocol.h
+++ b/tmux-protocol.h
@@ -68,6 +68,11 @@ enum msgtype {
 	MSG_WRITE_READY,
 	MSG_WRITE_CLOSE,
 	MSG_READ_CANCEL
+#ifdef TTY_OVER_SOCKET
+	,
+	MSG_TTY_READ,
+	MSG_TTY_WRITE
+#endif
 };
 
 /*
--- a/tmux.h
+++ b/tmux.h
@@ -2461,6 +2461,11 @@ void	tty_cmd_sixelimage(struct tty *, const struct tty_ctx *);
 void	tty_cmd_syncstart(struct tty *, const struct tty_ctx *);
 void	tty_default_colours(struct grid_cell *, struct window_pane *);
 
+#ifdef TTY_OVER_SOCKET
+int		tty_attach_stdin_to_socket(struct tty *, struct tmuxpeer *);
+int		tty_attach_out_to_socket(struct tty *, struct client *);
+#endif
+
 /* tty-term.c */
 extern struct tty_terms tty_terms;
 u_int		 tty_term_ncodes(void);
--- a/tty.c
+++ b/tty.c
@@ -250,6 +250,11 @@ tty_write_callback(__unused int fd, __unused short events, void *data)
 	struct client	*c = tty->client;
 	size_t		 size = EVBUFFER_LENGTH(tty->out);
 	int		 nwrite;
+#ifdef TTY_OVER_SOCKET
+	enum msgtype	 msg = MSG_TTY_READ;
+
+	proc_send(c->peer, msg, -1, EVBUFFER_DATA(tty->out), size);
+#endif
 
 	nwrite = evbuffer_write(tty->out, c->fd);
 	if (nwrite == -1)
@@ -303,6 +308,40 @@ tty_open(struct tty *tty, char **cause)
 
 	return (0);
 }
+
+#ifdef TTY_OVER_SOCKET
+static struct tmuxpeer *client_peer;
+
+static void
+tty_stdin_socket_callback(int fd, __unused short events, void *data)
+{
+	struct tty	*tty = data;
+	int		 nread;
+	enum msgtype	 msg = MSG_TTY_WRITE;
+
+	nread = evbuffer_read(tty->in, fd, -1);
+	if (nread == 0 || nread == -1) {
+		event_del(&tty->event_in);
+		return;
+	}
+
+	proc_send(client_peer, msg, -1, EVBUFFER_DATA(tty->in), nread);
+}
+
+int
+tty_attach_stdin_to_socket(struct tty *tty, struct tmuxpeer *peer)
+{
+	client_peer = peer;
+
+	event_set(&tty->event_in, STDIN_FILENO, EV_PERSIST|EV_READ,
+	    tty_stdin_socket_callback, tty);
+	tty->in = evbuffer_new();
+	if (tty->in == NULL)
+		fatal("out of memory");
+
+	return (0);
+}
+#endif
 
 static void
 tty_start_timer_callback(__unused int fd, __unused short events, void *data)


nicm:
Thanks. How does this change things over the old way where we reopened the pty device in the server for Cgywin? Does that no longer work?

The code looks OK but I'm not wild about how big the change is to workaround what seems like a bug on a closed source platform. I can't think of obvious ways to contain it in compat though...

rkitover:
The issue was that the server process could not access the special Windows terminal device from the client in a local session in a Windows terminal such as Microsoft Terminal, because the device is not accessible from other processes, see:

https://github.com/microsoft/terminal/issues/5132

. Apparently Cygwin will soon have a fix for this, see the discussion here:

https://github.com/msys2/MSYS2-packages/pull/4797

. I still think my solution is better, but I will let you decide that. I can drop this commit.

nicm:
I'll have to think about it.

nicm:
Here are a few comments on this:

- You can't just `write` to the tty in `client_dispatch_attached`, what if it does a partial write? It would need to be wrapped in a `bufferevent`. But tmux already has code to do this in `file.c`, which would avoid the need to add to the protocol. Why not reuse this for attached clients as well?

- Why do you still need to open `/dev/ptmx`? I thought the point was it couldn't be opened.

- There is no code reading from stdin, so how are key presses going to work?

- `tty_write_callback` is doing both `proc_send` and `evbuffer_write`, shouldn't it only do `proc_send`? Also in that case, doesn't it need an `evbuffer_drain`?

- Can you make the existing Cgywin patches a separate PR (egMSYSTEM) and they can probably go in?

rkitover:
> Here are a few comments on this:
> 
> * You can't just `write` to the tty in `client_dispatch_attached`, what if it does a partial write? It would need to be wrapped in a `bufferevent`. But tmux already has code to do this in `file.c`, which would avoid the need to add to the protocol. Why not reuse this for attached clients as well?

In this case the point is just that instead of passing the file descriptors from the client terminal, the input of the client terminal is connected to to a PTY on the server and the output of the PTY is written to the client terminal.

The terminals just do what terminals do with the bytes, what would a partial write be and why would that matter?

> * Why do you still need to open `/dev/ptmx`? I thought the point was it couldn't be opened.

That is a PTY on Linux and Cygwin. I should use the POSIX function for this instead.
The point of this is to have an unattached terminal device to actually do the terminal things in the tmux pane.

> * There is no code reading from stdin, so how are key presses going to work?

In `client.c`, `tty_attach_stdin_to_socket()` is called to send the client terminal input over the socket.

> * `tty_write_callback` is doing both `proc_send` and `evbuffer_write`, shouldn't it only do `proc_send`? Also in that case, doesn't it need an `evbuffer_drain`?

That event writes to the terminal fd. When `TTY_OVER_SOCKET` mode is enabled it **ALSO** writes the terminal data to the socket which is then written to the client terminal.

I'm not sure about `evbuffer_drain()`, but it writes the same data to both the terminal file descriptor as it did before, it just also writes it to the socket so it can be written to the actual client terminal.

> * Can you make the existing Cgywin patches a separate PR (egMSYSTEM) and they can probably go in?

Sure.

I should mention that some people have tested this patch already and tmux worked correctly for them.

rkitover:
You can actually enable `TTY_OVER_SOCKET` on any OS, it does not depend on that, if you want to see how this works.

nicm:
> > Here are a few comments on this:
> > 
> > * You can't just `write` to the tty in `client_dispatch_attached`, what if it does a partial write? It would need to be wrapped in a `bufferevent`. But tmux already has code to do this in `file.c`, which would avoid the need to add to the protocol. Why not reuse this for attached clients as well?
> 
> In this case the point is just that instead of passing the file descriptors from the client terminal, the input of the client terminal is connected to to a PTY on the server and the output of the PTY is written to the client terminal.
> 
> The terminals just do what terminals do with the bytes, what would a partial write be and why would that matter?

Actually it looks like in the clinet the fd is blocking so it will block instead on short writes, which I don't like much either. What if the user wants to detach? This should be nonblocking and use a bufferevent.

> > * Why do you still need to open `/dev/ptmx`? I thought the point was it couldn't be opened.
> 
> That is a PTY on Linux and Cygwin. I should use the POSIX function for this instead. The point of this is to have an unattached terminal device to actually do the terminal things in the tmux pane.

So you are getting a pty master fd for the client, but not opening the slave?

Then you are proxying the input and output from the client's stdin and stdout over the master fd?

> > * There is no code reading from stdin, so how are key presses going to work?
> 
> In `client.c`, `tty_attach_stdin_to_socket()` is called to send the client terminal input over the socket.

OK so you are reading from the imsg and writing into the pty master fd.

`c->fd` is definitely nonblocking (and has to be - the server process cannot block on a client) so you really can't just write into it, you need to handle EAGAIN which means a bufferevent.

> > * `tty_write_callback` is doing both `proc_send` and `evbuffer_write`, shouldn't it only do `proc_send`? Also in that case, doesn't it need an `evbuffer_drain`?
> 
> That event writes to the terminal fd. When `TTY_OVER_SOCKET` mode is enabled it **ALSO** writes the terminal data to the socket which is then written to the client terminal.
> 
> I'm not sure about `evbuffer_drain()`, but it writes the same data to both the terminal file descriptor as it did before, it just also writes it to the socket so it can be written to the actual client terminal.

This may be what I am missing, why does it need to write the data to the client stdin *and* to the terminal? There is only one terminal?

> > * Can you make the existing Cgywin patches a separate PR (egMSYSTEM) and they can probably go in?
> 
> Sure.
> 
> I should mention that some people have tested this patch already and tmux worked correctly for them.

rkitover:
> Actually it looks like in the clinet the fd is blocking so it will block instead on short writes, which I don't like much either. What if the user wants to detach? This should be nonblocking and use a bufferevent.

I'll take care of this.

> So you are getting a pty master fd for the client, but not opening the slave?
> 
> Then you are proxying the input and output from the client's stdin and stdout over the master fd?

Yes, and that seems to work.

> `c->fd` is definitely nonblocking (and has to be - the server process cannot block on a client) so you really can't just write into it, you need to handle EAGAIN which means a bufferevent.

I'll take care of this too.

> This may be what I am missing, why does it need to write the data to the client stdin _and_ to the terminal? There is only one terminal?

There are two terminals. The detached PTY used by the server for the stuff in the tmux pane to "happen" in, and the client terminal which sends input and receives output over the socket.

nicm:
> There are two terminals. The detached PTY used by the server for the stuff in the tmux pane to "happen" in, and the client terminal which sends input and receives output over the socket.

I don't really understand this, can you be more specific? What is the need to update the pty if you also send the data over the socket?

What is reading the data you write to the pty fd?

nicm:
I don't know if there is any point in doing more on this if it is going to be superseded by https://github.com/tmux/tmux/pull/4086.

I don't think it is likely that tmux will support any other platforms that don't have file descriptor passing.

rkitover:
> I don't really understand this, can you be more specific? What is the need to update the pty if you also send the data over the socket?
> 
> What is reading the data you write to the pty fd?

The `tti_write_callback()` event is used by tmux to write the actual tmux things to the terminal, status messages, window borders, control sequences, etc..

The PTY in which things "happen" has to be in the same state as the client terminal for this to work, so the data tmux writes has to go to both.

nicm:
>The tti_write_callback() event is used by tmux to write the actual tmux things to the terminal, status messages, window borders, control sequences, etc..

I still don't understand. Normally, tmux writes everything - pane output, status line, everything - to the client's file descriptor, which will mostly be the slave side of a pty. It is read by one terminal on the master side.

Why does it now need to go to two places? The client's stdout and a pty master, right? What process is reading the other side of each of these?

rkitover:
> I don't know if there is any point in doing more on this if it is going to be superseded by #4086.
> 

The two things are unrelated, even if they are for the same platform.

> I don't think it is likely that tmux will support any other platforms that don't have file descriptor passing.

Let me give you some background here.

So you are aware now that you are using this hack on the Cygwin runtime of reopening `cllient->fd` to `client->ttyname`, which is the terminal device node used by the client, because the messages to pass the file descriptors to `client->fd` and `client->out_fd` do not do anything on the Cygwin runtime.

There was a years' long outstanding issue in the Cygwin runtime with tmux not working in native Windows terminals, such as the Windows Terminal and some other ones, from a shell using the Cygwin runtime, e.g. Cygwin bash/zsh, MSYS2 bash/zsh, Git Bash with a tmux binary, etc..

The reason for this is that the underlying Windows console is a fake device node, as Windows does not actually have device nodes, and cannot be accessed from another process, such as the tmux server process. So the Cygwin hack fails in this case.

I decided to fix this issue by writing this patch. As far as I can tell, it works perfectly, and I am happy to resolve your objections to it to make it correct.

At the same time as I made this, some Cygwin developers implemented a separate Cygwin device node as a proxy for the underlying Windows console device, which makes the Cygwin hack here work in native terminals.

So using the latest Cygwin runtime in Cygwin or MSYS2 fixes this issue without my patch, but not on older runtimes, for which my patch also works.

I wrote this patch under the most horrifying conditions, with this government **** breaking my compiles and my binaries, even breaking into my friends' computers, both in the US and other countries, to sabotage the binary when they were testing it. Not to mention the **** they were doing to me.

However it was tested and confirmed to be fully functional, and you can test it as well, it's not specific to Cygwin other than defining the `TTY_OVER_SOCKET` macro.

So, I would like to finish this, and I believe it's a better solution than the current hack, and it's also a very small amount of code. Once I implement your changes it will still be a small amount of code.

What else this, in some form, would also be useful for I don't know. It's possible an actual native Windows port will happen some day, or functionality will be added for other forms communication between the client and server, for example with the server process being on another machine and connected with a TCP socket, which is how the Cygwin runtime works for this.

So up to you.

rkitover:
> > The tti_write_callback() event is used by tmux to write the actual tmux things to the terminal, status messages, window borders, control sequences, etc..
> 
> I still don't understand. Normally, tmux writes everything - pane output, status line, everything - to the client's file descriptor, which will mostly be the slave side of a pty. It is read by one terminal on the master side.

What I said is not entirely correct, and I don't know the whole tmux codebase very well, or these semantics of PTYs and TTYs.

So `tty_write_callback()` is the event that actually writes **EVERYTHING** to `client->fd`, the file descriptor passed in from the client. It should really be `client->out_fd`, but apparently that doesn't matter.

So presumably this also includes the read side of whatever PTY is the current tmux window, the program running in it, as well as the tmux-specific stuff.

So this makes it the event that would send the data to the client terminal.

> Why does it now need to go to two places? The client's stdout and a pty master, right? What process is reading the other side of each of these?

So I needed to set `client->fd` to something that is a TTY, and it is used in a lot of places.

The state of this TTY is queried and altered among many other things, so that both reads from it and writes to it.

In this case it is just a consistent representation of the client's TTY.

I think now that conceptually, this may not be the best way to do this, now that you have mentioned some more details about the way tmux works to me. However, it does work, with a very small amount of code.

If you can suggest a way to do this at a higher level without this PTY, using the other tmux PTYs, I can do that, but it may not be worthwhile.

nicm:
> So tty_write_callback() is the event that actually writes EVERYTHING to client->fd, the file descriptor passed in from the client.

Yes.

> It should really be client->out_fd, but apparently that doesn't matter.

It doesn't matter whether it is `c->fd` or `c->out_fd` because clients can only attach when `stdin` is a tty and ttys are read-write. `out_fd` is for non-attaching clients when the difference matters (because the user could do for example `tmux saveb - >file </dev/null` so writing to `stdin` would be wrong).

> The state of this TTY is queried and altered among many other things, so that both reads from it and writes to it.

It's not actually used all that many places. I don't think we should be writing everything to it, although I assume it is just throwing it away on Cgywin if it is just a fake node? It won't be throwing it away on other platforms though, it will just be sitting in the evbuffer or the kernel waiting for the slave to read it.

If stuff like `tcsetattr` is actually a noop on the `/dev/ptmx` fd. How does it avoid eg LF being translated incorrectly? Does the terminal handle them correctly without being told to?

rkitover:
> > The state of this TTY is queried and altered among many other things, so that both reads from it and writes to it.
> 
> It's not actually used all that many places. I don't think we should be writing everything to it, although I assume it is just throwing it away on Cgywin if it is just a fake node? It won't be throwing it away on other platforms though, it will just be sitting in the evbuffer or the kernel waiting for the slave to read it.

It is a real PTY on Cygwin.

The fake node I was talking about was the underlying Windows console before they made it a real node in the recent Cygwin runtime changes.

I also ran this on macOS and it worked perfectly. Somehow it does not matter that nothing is reading from the slave side of this PTY.

> If stuff like `tcsetattr` is actually a noop on the `/dev/ptmx` fd. How does it avoid eg LF being translated incorrectly? Does the terminal handle them correctly without being told to?

I'm guessing what is happening here is that functions like `tcsetattr()` change the state of this PTY, which affects the output of the programs running in the higher level PTYs which keeps things consistent, even if the actual `tcsetattr()` call is not itself passed through to the client terminal. That's probably not great, but it seems to work. Writing the control codes to the output buffer would be better.

nicm:
It soesn

nicm:
It sounds odd to me. Why would someone implement a working tcsetattr but not then bother to implement actual read/write to the pty. What's the point of tcsetattr on a pty that doesn't work?

How much data did you put through it on macOS? If you write a few 10s of MB, is it definitely being consumed by the pty? The evbuffer is draining and tmux isn't switching into redraw mode for the client?

nicm:
I'll try it on macOS if I have time tomorrow.

nicm:
OK I tested this on macOS. I had to make a few changes to get it to run:

- it will not start unless I restore the code to pass stdin/stdout over the fd so that `c->fd` is a valid tty fd, I assume this is because just opening `ptmx` is not enough, it is also necessary to open the slave and do the `grantpt` dance;

- if I do that, everything tmux sends is written to the tty twice because both `tty_write_callback` and `client_dispatch_attached` write it;

- to fix this I added a `return` and `evbuffer_drain` in `tty_write_callback` after the `proc_send` so only `client_dispatch_attached` is writing.

But then if I do `ls` for example, the output is coming over rather staggered, because the `write` in `client_dispatch_attached` is returning `EAGAIN` - this can happen because when the server sets the file descriptor to be nonblocking, that also applies to the client (because they are the same underlying file). It probably doesn't happen on Windows because they are not the same (stdin on the client is different from the `/dev/ptmx` opened in the server). So you definitely will need that bufferevent there, at least for it to work on non-Windows platforms.

Here are the changes I made: [macos.diff.txt](https://github.com/user-attachments/files/17240082/macos.diff.txt)

I've also noticed a problem - the maximum imsg size is 64K minus the header size, but the buffer could easily contain much more than that. So you will need to limit how much you send at a time to `proc_send`.

This isn't really a good test anyway because on Cgywin opening `ptmx` alone actually works, so I won't really be able tell if data is going anywhere on macOS because they are entirely different implementations... but TBH I don't see why you would write to `c->fd` at all instead of just calling `evbuffer_drain`, so the whole question is probably moot.

I'm more interested in if `tcsetattr` works but I'm not sure how you can test it on Windows... on Linux or macOS you can test by doing this:
````
TERM=vt100 tmux -Ltest -f/dev/null new
stty raw;printf 'abc\ndef\n';stty sane;echo
````
Which will cause tmux to emit `abc\ndef` and show this:
Screenshot 2024-10-03 at 07 17 00
If tmux is not able to correctly set up the terminal, it the `\n` will be translated to `\r\n`, so it will look like this:
Screenshot 2024-10-03 at 07 18 31
Are you able to try something similar to this in a known-working terminal like mintty or putty and see if it works correctly?

**Comment by rkitover** [comment]
> It sounds odd to me. Why would someone implement a working tcsetattr but not then bother to implement actual read/write to the pty. What's the point of tcsetattr on a pty that doesn't work?

The PTY works, that is how tmux itself and any terminal programs work. The device node is also opened with `O_NOCTTY`.

> 
> How much data did you put through it on macOS? If you write a few 10s of MB, is it definitely being consumed by the pty? The evbuffer is draining and tmux isn't switching into redraw mode for the client?

Nothing needs to read from it, it just has to behave like a TTY. I should have turned that off and I will fix that.

**Comment by rkitover** [comment]
> OK I tested this on macOS. I had to make a few changes to get it to run:
> 
> * it will not start unless I restore the code to pass stdin/stdout over the fd so that `c->fd` is a valid tty fd, I assume this is because just opening `ptmx` is not enough, it is also necessary to open the slave and do the `grantpt` dance;

I looked at the [manual](https://linux.die.net/man/7/pty) on this topic. I didn't understand the master/slave part of PTYs before.

Note that the `open()` call is with `O_NOCTTY` and it succeeds.

Also according to said manual, `unlockpt()` is only necessary for the SYSV API and SYSV nodes, can you try this please:

```c
#ifdef TTY_OVER_SOCKET
        int master = open("/dev/ptyzf", O_RDWR|O_NOCTTY);
        c->fd = open("/dev/ttyzf", O_RDWR|O_NOCTTY);
        c->out_fd = dup(c->fd);
#endif
```
, to use the BSD nodes.

rkitover:
> > How much data did you put through it on macOS? If you write a few 10s of MB, is it definitely being consumed by the pty? The evbuffer is draining and tmux isn't switching into redraw mode for the client?
> 
> Nothing needs to read from it, it just has to behave like a TTY. I should have turned that off and I will fix that.

Actually no, that is not true. Since tmux commands work, it must also need to read from the PTY.

nicm:
The dev nodes do not go up as far as `ttyzf` on macOS, but I tried `ttywf` and tmux hangs in the server process. If I do this instead to open a pty:
~~~~
     int master;
     openpty(&master, &c->fd, NULL, NULL, NULL);
     c->out_fd = dup(c->fd);
~~~~
tmux manages to write to the terminal, but key presses are not processed and instead are printed.

This approach is not going to be right on macOS whatever we do -  even if all reads and writes go over the socket, it wouldn't work - it has to call `tcsetattr` to put the terminal into the right state. It needs to call it on the client's tty not a new pty. So either `tcsetattr` needs go over the socket as well, or it needs access to the client's tty in the server.

I think you should probably forget about non-Windows platforms here and concentrate on Windows. There are two things that tmux needs to do with the client's tty:

1) Read and write to it;
2) Set terminal attributes with `tcsetattr`.

It seems like from what you said that the reason for this change here is:

- On Cgywin with a Cgywin terminal, tmux can open the client's tty in the server and do both (1) and (2) and all works;

- But Cygwin with a native Windows terminal (whatever that is) (1) does not work, so instead you want to send the data over the socket.

The reason for that makes sense, but my questions are about (2), that is - on a native Windows terminal, what does `open(c->ttyname)` *do*?
 
- Does it give access to the client's terminal for tmux to call `tcsetattr`? If so, why doesn't read/write work as well - is that a bug? Why do you want to write data both to the client's tty *and* over the socket and what happens to the data sent to the client's tty?

- Does it *not* give access to the client's terminal? If so, how does tmux set the terminal attributes? Is `tcsetattr` just a noop and setting attributes not necessary? That is, the terminal does not do any translation of the data tmux sends, it is always in the right mode?

Whatever the answers, I don't like this dual approach of using both a tty fd and writing over the socket. It should be one thing or the other:

- Either the client should take care of everything - read, write and terminal setup *if it is needed* - and the server just send it messages over the socket. `c->fd` should be -1 and that should be handled appropriately.

- Or the client should get its file descriptor into the server somehow and the server take care of everything, that is how tmux normally works.

rkitover:
I agree with your points about this being a bad implementation, whether it works or not, and I'm sure it does, in the normal use-cases.

I want to rewrite it correctly when I am able to. And I really appreciate your help with that.

Currently I can do nothing, because said government poos break literally anything I am doing as I am doing it.

But I want to talk to you about something else. And if you have contacts in the OpenBSD community, I'd like their help as well.

So, I'm sure I ran this patch on macOS and it worked perfectly (in the normal use cases, running tmux, opening new windows, running vim, etc..) And that was not a hallucination.

There are two things happening here:

1. The code in this Git commit has hidden code in it.
2. The government poos I mentioned get into anything instantly and change anything. When you did the above experiment, that is what may have happened, but it may also have been 1. They can get your IP address from GitHub etc..

In regards to the error you got on macOS initially, that is impossible, and the indication that this Git commit has hidden code in it. I got the same error when I ran it this time.

To prove that this is the case, I have a small C program:

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <termios.h>
#include <unistd.h>

void check_tty(int fd, const char *msg) {
    if (isatty(fd))
        printf("isatty() returned true for %s\n", msg);
    else
        printf("isatty() returned false for %s\n", msg);

    struct termios term;

    if (tcgetattr(fd, &term) == 0)
        printf("tcgetattr() succeeded for %s\n", msg);
    else {
        char buf[256];
        snprintf(buf, sizeof(buf), "tcgetattr() failed for %s", msg);
        perror(buf);
    }
}

int main(int argc, char *argv[]) {
    int master_fd = open("/dev/ptmx", O_RDWR|O_NOCTTY);

    check_tty(master_fd, "master_fd after open()");

    int slave_fd = open(ptsname(master_fd), O_RDWR|O_NOCTTY);

    check_tty(slave_fd, "slave_fd after open()");

    grantpt(master_fd);
    unlockpt(master_fd);

    check_tty(master_fd, "master_fd after grantpt() and unlockpt()");
    check_tty(slave_fd, "slave_fd after grantpt() and unlockpt()");

    return 0;
}
```
. This is the same behavior tmux uses to check that an fd is a tty.

The two syscalls should succeed on all platforms at all points. In fact, the first one succeeded on Linux for me when I tried it, and then said poos broke something else.

The code I pasted may have hidden code in it as well, so it would be best to type it out manually if testing, and compare the behavior with the pasted version. Unless said government poos are in that system at the time, and you have no way of knowing that or preventing that, even if the system is physically disconnected from a network.

You may have heard about "the compiler conspiracy." This is what is happening here. It may be across our whole computing reality, in everything, with us being completely blind to it.

Now, let me tell you a story about an ncurses patch. This is something I need for the other PR I am working on.

I was working on the patch in my fork of the GitHub ncurses mirror here:

https://github.com/rkitover/ncurses/commit/e88132e5db3debbf5520f983924d610c3ac70c96
.

It just rearranges some cpp macros to make some Win32 functionality available under Cygwin. So it is completely impossible that it doesn't work.

This commit has this hidden code in it that breaks it.

I sent the patch to the ncurses mailing list, which said government poos failed to stop me from doing, along with the hidden code.

The maintainer accepted my patch with a minor change and a new dev release was cut.

I updated the MSYS2 ncurses package to this dev release along with instructions on how to test the term-driver functionality. The package pulled the dev release archive which had the hidden code in it.

The MSYS2 maintainer said that it does not work.

At that point I discovered the functionality of the hidden code breaking the code path in gdb, you can see my posts about this here:

https://github.com/msys2/MSYS2-packages/pull/4886
.

Basically, it changes something in the runtime to make a memory allocation for a few bytes fail, which is completely impossible, leading the code path to the failure they wanted it to produce.

This includes simple instructions on how to verify this in gdb.

A month later, the term-driver feature was disabled due to a supposed nano regression, which I believe is very unlikely, but I haven't looked into that yet, if I am even able to do that.

A couple of weeks later, a doc patch was sent to the ncurses list, which is gernerally very quiet. I believe this doc patch had hidden code to remove the hidden code introduced by my patch, at least that may be possible.

Issue #3705, opened 2023-09-24
Suggestion: Disable entering copy mode by mouse wheel up when alternate screen is on


zalza:
### Issue description
Even when alternate screen is on, mouse wheel up enters copy mode and shows scrollback buffer, which might confuse some users using some terminal emulators which disables scrolling of scrollback buffer.
Is this default behavior intentional?

### Reproduction steps

1. `$ man man | cat`
2. `$ man man`
3. Mouse wheel up enters copy mode. Contents of both alternate screen and scrollback buffer are mixed.

### Modified key binding
```
bind-key -n WheelUpPane {
  if-shell -F '#{mouse_any_flag}' {
      send-keys -M
  } {
    if-shell -F '#{alternate_on}' {
    } {
      copy-mode -e
    }
  }
}
bind-key -n WheelDownPane {
  if-shell -F '#{mouse_any_flag}' {
      send-keys -M
  }
}
```
### Required information

* tmux version (`tmux -V`) : `next-3.4` (the lateset commit b77778072052c14d1450d2eb89542f2493121f84)
* Platform (`uname -sp`) : `Linux x86-64` (Ubuntu 22.04.3)
* $TERM inside and outside of tmux (`echo $TERM`): Outside = `xterm-256color` / Inside = `tmux-256color`

jyn514:
this should actually not just disable copy mode, but translate the scroll events into up and down arrow keys. otherwise things like `less` will break and https://github.com/Canop/bacon/issues/174 will break.

here is a keybind that does that:
```tmux
bind-key -n WheelUpPane {
  if-shell -F '#{mouse_any_flag}' {
      send-keys -M
  } {
    if-shell -F '#{alternate_on}' {
      send-keys Up Up Up
    } {
      copy-mode -e
    }
  }
}
bind-key -n WheelDownPane {
  if-shell -F '#{mouse_any_flag}' {
      send-keys -M
  } {
    if-shell -F '#{alternate_on}' {
      send-keys Down Down Down
    }
  }
}

nicm:
Yes, changed it now, thanks.
~~~~
--- key-bindings.c      28 Oct 2025 16:36:52 -0000      1.155
+++ key-bindings.c      29 Oct 2025 09:32:54 -0000
@@ -441,10 +441,10 @@ key_bindings_init(void)
                "bind -n MouseDown1Pane { select-pane -t=; send -M }",

                /* Mouse button 1 drag on pane. */
[...]

PR #4076, opened 2024-08-15
fix mouse scrolling in alternate mode

jyn514:
terminal emulators are supposed to translate mouse scroll events to arrow keys when the alternate screen buffer is active. you can verify this locally with `tput smcup` (*outside* tmux, in e.g. kitty, gnome-terminal, xterm). in this mode, scrolling will move up and down in the shell history.

tmux previously didn't do this, which broke programs like `less` which don't capture mouse events.

you can test this PR as follows:
1. start a new tmux server with this branch with a default config, e.g. `./tmux -L new-session -f /dev/null`
2. enable mouse support (`C-b :set mouse on`)
3. run any program which has more than the height of a pane in output and pipe it to less, e.g. `stty --help | less`
4. scroll down and then up with the mouse

before this PR, scrolling down will do nothing and scrolling up will enter copy mode. after this PR, scrolling up and down passes the up and down keys through to less.

fixes https://github.com/tmux/tmux/issues/3705

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -441,7 +441,10 @@ key_bindings_init(void)
 		"bind -n MouseDrag1Pane { if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' { send -M } { copy-mode -M } }",
 
 		/* Mouse wheel up on pane. */
-		"bind -n WheelUpPane { if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' { send -M } { copy-mode -e } }",
+		"bind -n WheelUpPane { if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' { send -M } { if -F '#{alternate_on}' { send-keys -N 3 Up } { copy-mode -e } } }",
+
+		/* Mouse wheel down on pane. */
+		"bind -n WheelDownPane { if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' { send -M } { if -F '#{alternate_on}' { send-keys -N 3 Down} } }",
 
 		/* Mouse button 2 down on pane. */
 		"bind -n MouseDown2Pane { select-pane -t=; if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' { send -M } { paste -p } }",


nicm:
This is an optional feature of some terminals and is trivial to configure
with a key binding in tmux.

jyn514:
> This is an optional feature of some terminals and is trivial to configure with a key binding in tmux.

it's true that it's possible to configure, but it's difficult to figure out what's going on (at least, it took me several hours), and i think it would be nice to enable it out of the box so that others don't have to go through the same effort.

the [xterm standard](https://invisible-island.net/xterm/manpage/xterm.html#VT100-Widget-Resources:alternateScroll) says that this should be an option. it doesn't enable it by default, but most other terminals i've tried have it on (kitty, Windows Terminal, gnome-terminal).

 note that for this to have an effect at all, the user needs to `set mouse on`; i would expect that to enable optional features. my PR only changes the default, it doesn't affect anyone who's explicitly bound a command to `MouseScrollUp`.

nicm:
xterm has this enabled by config, and so does tmux. I don't want it by default. It is already here https://github.com/tmux/tmux/wiki/Recipes#send-up-and-down-keys-for-the-mouse-wheel, but if you want to suggest some text for the FAQ instead or as well then that would be fine.

jyn514:
hmm, i worry we've had some miscommunication. the config you've linked to does something different than my PR - it turns scrolling into arrow keys in *all* modes, not just in the alternate mode, and so as a result you can't access the scrollback at all using the mouse. i agree that shouldn't be on by default. but my PR only has an effect in the alternate mode, and i would be kind of surprised to see people trying to access the scrollback in that mode? is that something you do?

here is a comparison of the three:
1. tmux with the default config (and `:set mouse on`) https://asciinema.org/a/IGNMPt3sNzQi9pvpYPv7fSdf9
2. tmux with the config in the wiki https://asciinema.org/a/O7uQis3NfSOkiIhHJY0NKLmmI
3. tmux with the config in this PR https://asciinema.org/a/rsrvGs8DbzEenmML3YG5WCezX

nicm:
You're right it isn't quite the same, we should add another entry which includes the check for the alternate screen.

PR #4080, opened 2024-08-16
character based scrollbars feature (was issue 3832)

mgrant0:
This adds scrollbars to tmux.  Scrollbars serve several purposes:

1. They give you a visual representation of what part of the screen / copy-buffer you are currently looking at.
2. They can be used to navigate (up/down) the copy buffer.  The scrollbar can be dragged or clicked above or below to page up or down.
3. Although you can wheel up the mouse or page up, being able to drag a scrollbar up or down a great distance is much quicker and more intuitive.

`tmux set pane-scrollbars [ off | modal | on ]`

modal makes them appear when a pane is in copy-mode or view-mode.  on means all the time.  I searched for a better them than 'modal' but this one stuck with me.  In a real sense, the window is in a mode and the pane itself is waiting for user input to be scrolled or copied or whatever.  copy-mode and view-mode are "modal" in this sense.  "terminal mode" might be the other mode.

Scrollbars can be set on left or right (default is right).  Though I didn't iitially set out to implement this, it became super useful in chasing down bugs and understanding the code, so it's an option:

`tmux set pane-vertical-scrollbars-position [ left | right ]`

The style can be set:

`tmux set pane-scrollbar-style "fg=WhiteSmoke,bg=DarkSlateGray"`

The width of the scrollbars (default=1) and a padding (default=0) are also options.  width=2 looks nice as well.  

fg is the scrollbar elevator, bg is the bg of the scrollbar.

I had an early version with a border between the scrollbar and the pane and decided it was a waste of space.  The padding can serve this purpose if desired and is visually easier on the eye.

I tried to match your short variable naming style but I needed some longer names in places to keep track of things in my head.  We can shorten the names of things if you prefer.  It may also be possible to factor some of the large if statements and simplify the code.

My .tmux.conf:
```
set-option -g history-limit 3000

# Permit PgUp and PgDn to work in an editor rather than scrolling copy-mode
bind-key -Troot PPage if-shell -F -t = "#{alternate_on}" { send-keys PPage } {set-option mouse on; copy-mode -eu}
bind-key -Troot NPage if-shell -F -t = "#{alternate_on}" { send-keys NPage } {set-option mouse off; copy-mode -eq }

# Enter key exits copy-mode
bind-key -T copy-mode Enter "copy-mode -q"

# Copy and clear the selection but do not exit copy mode
bind -T copy-mode MouseDragEnd1Pane if-shell -F "#{selection_present}" { send -X copy-selection-no-clear } { copy-mode -q }

# so that vi and emacs don't take the middle button in tmux
bind-key -T root MouseDown2Pane select-pane -t = \; if-shell -F "#{pane_in_mode}" { send-keys -M } { paste-buffer -p }

#osc52 copy-paste support from tmux local machine (if your terminal supports osc-52 codes)
set -s set-clipboard external
set -as terminal-features ',tmux-256color:clipboard'
set -as terminal-overrides ',*-256color:Ms=\E]52;%p1%s;%p2%s\E\\\\'

# enable mouse mode all the time
set-option -g mouse on

# enable scrollbars that show you the copy mode buffer in scrollback area of a pane
set-option -g pane-scrollbars on

```


mgrant0:
I'll fix all of these.  

On the pad and width, what if they are moved into the style?  I was going to do that anyway after seeeing the PR on the indent within the pane status line.  If yes, then, perhaps that PR should be split 2 PRs.  I can very likely do that work.

I do not mind at all setting the default colours to whatever you think.  Keep in mind though that green may bleed into the status line.  I'll try it though and see what it looks like.  Perhaps just default it to green/white and put an example in the man page for what I have which looks really good in my opinion?

I added copy-mode -d for consistency since there was a copy-mode -u.  It just looked odd using -X page-down for that one action when the other 2 use the copy-mode cmd (and scrollbars really works with copy-mode).  But send -X page-down does work as well.  Let me know if -d lives or dies.  It can be documented as a synonym for the page-down.

nicm:
Leaving `copy-mode -d` is fine.

See how green looks but it can't use 256 or RGB colour by default.

I don't really see any need for pad/width at all but if you want to leave them as 1 for now and later make a PR to put them in the style then that's fine but yes better do it separately.

mgrant0:
The first 4 bugs you reported, I can't exactly reproduce these:

> The window split line does not appear for me anymore with scrollbars off, and horizontal borders do not appear with them on.
> 
> With scrollbars on, adding text to a window (for example runnning ls -1) does not redraw the scrollbar, I need to do C-b r for it show correctly.
> 
> When pane-border-status is top as well as pane-scrollbars on, the pane status line is not drawn correctly.
> 
> Turning pane-scrollbars to off does not appear to correctly remove the scrollbars and adjust the pane sizes (the rightmost pane is left one character too narrow).

I think this may be related to the size of the window.  I did manage to create a situation where the scrollbar is only the bg and no scroll elevator, but C-b r does not redraw it.  I think I know what that is though.

Can you tell me what size window you were using and what terminal program, what platform?  I just need a way to reproduce what you're seeing.  Thanks.

nicm:
It happens in iTerm2 and in tmux itself in iTerm2. I'm not sure the exact terminal size but I'd guess about 100x40.

Maybe it is related to ACS rather than UTF-8 line drawing, or vice versa?

mgrant0:
The scrollbar code doesn't use any of those ACS characters.  It's just drawing it with a plain space characters.  Could it have something to do with needing to do put the START_ISOLATE and END_ISOLATE before and after drawing the scrollbars in each pane?  I wasn't sure what that was, looks like it has to do with bi-directional printing to the screen.

Drawing the scrollbars does a lot of tty_cursor movements and tty_cell calls.  Is there something that needs to be called before doing lots of those calls?  Or is there a better way to draw character by character than that?  See the screen_redraw_draw_scrollbar function at bottom of screen-redraw.c.

I'll have to find a mac to test iTerm2.

nicm:
Here is what it looks like (right) compared with what it should look like (left) - ignore the different background colour. This is with scrollbars and pane status line off:
Screenshot 2024-08-21 at 04 24 32

nicm:
It looks to me like in `screen_redraw_pane_border` you have added the scrollbars cases but removed or broken the existing  non-scrollbar cases?

nicm:
It looks OK though, not sure what is wrong...

nicm:
This fixes it for me. It is not obvious why so I assume it is the compiler playing silly buggers (`i` would have be uninitialized for the `log_debug` so it is UB and perhaps it thinks it can remove the whole block).
~~~~
--- a/screen-redraw.c
+++ b/screen-redraw.c
@@ -932,10 +932,8 @@ screen_redraw_draw_borders(struct screen_redraw_ctx *ctx)
                wp->border_gc_set = 0;

        for (j = 0; j < c->tty.sy - ctx->statuslines; j++) {
-                log_debug("%s: %s %u %u", __func__, c->name, i,j);
-               for (i = 0; i < c->tty.sx; i++) {
+               for (i = 0; i < c->tty.sx; i++)
                        screen_redraw_draw_borders_cell(ctx, i, j);
-                }
        }
 }

nicm:
You should make this change as well (`!x !=` -> `x ==`).
~~~~
--- a/screen-redraw.c
+++ b/screen-redraw.c
@@ -779,7 +779,8 @@ screen_redraw_pane(struct client *c, struct window_pane *wp)
        screen_redraw_draw_pane(&ctx, wp);

         pane_scrollbars = ctx.pane_scrollbars;
-        if (pane_scrollbars == PANE_SCROLLBARS_MODAL && ! window_pane_mode(wp) != WINDOW_PANE_TERMINAL_MODE) {
+        if (pane_scrollbars == PANE_SCROLLBARS_MODAL &&
+           window_pane_mode(wp) == WINDOW_PANE_TERMINAL_MODE) {
                 pane_scrollbars = 0;
                 wp->flags &= ~PANE_REDRAW_SCROLLBARS;
         }

mgrant0:
Awsome.  Did that little change fix the top 4 of your visual glitches?

nicm:
It fixes some of them - these ones still happen:

1. With scrollbars on, adding text to a window (for example runnning ls -1) does not redraw the scrollbar, I need to do C-b r for it show correctly.

2. If I set history-limit to say 10000000 and generate a bit of output, the scrollbar does not show the position indicator. There needs to be some limit here so the indicator is always at least one cell.

It would be very useful if you can break out the following into their own PRs (or just give me a .diff file, it doesn't need to be a PR):

- `copy-mode -d`;

- the `screen_redraw_ctx` and making `flags` `uint64_t` in `screen-redraw.c`;

- `window_pane_in_copy_or_view_mode` (it doesn't matter if nothing uses it yet);

- possibly also `layout_cell_is_left` and `layout_cell_is_right`?

I don't think this will be too hard to do and will make the diff much smaller and easier to review (particularly the second one).

mgrant0:
Yes I'll take care of these.

mgrant0:
I have now merged in the other PRs.  

I tried to split the long lines.  I managed to cleanly split most of them.  Some more will be split when I clean up the names.  There are still a handfull that I could split but become difficult to read.  

I chose fg=white,bg=black for the default style.  This looks pretty good to me, comments welcome though.  bg=terminal also viable, even more minimal as the scroll bar itself is the same colour as the terminal which makes it invisible except where the elevator is but this makes it less obvious where to click above/below the elevator to page.  I'll default the elevator to fg=green if you prefer.

Still to do:

- The bug which shows up so far only in iterm2 where the scrollbar doesn't refresh properly on scrolling
- I seem to have introduced a bug when I merged: clicking below the scroll elevator to page down scrolls incorrectly
- Move pad, width, and probably position into the style 
- Fix the naming inconsistency (this will shorten more lines).  Will probably do this with the pad/width/position item.
- Look at the long if statements to see if I can coalease the cases and simplify them -- they currently work though
- I'll give one more try to see if I can remove the 'force' redraw to redraw the scrollbars, then remove the dead code in comment
- The options in the options table need more sorting

nicm:
The reason the no update only happens in iTerm2 is because it supports margins so it doesn't need to do a full redraw when scrolling only a section of the terminal's horizontal width. If you turn margins off then it happens in iTerm2 also:
~~~~
set -as terminal-overrides '*:Enmg@:Dsmg@:Clmg@:Cmg@'
~~~~
You probably need to redraw the scrollbars in `screen_write_collect_flush` if scrolled != 0.

Other than that the behaviour looks OK on a quick go, will take a more thorough look and look at the code again once you have finished your changes. Thanks!

mgrant0:
Setting that as terminal-overrides has a strange effect for me.  It takes the terminal out of the alternate screen and the mouse wheel scrolls the terminal instead of the mouse wheel being captured by tmux.

Second, I had a look at `screen_write_collect_flush()` and I tried calling `screen_redraw_draw_pane_scrollbar(c, wp)` from there but I don't have either a *client or a *wp.  So I tried looping over all the clients and all the window panes like this:

```
c = TAILQ_FIRST(&clients);
w = c->session->curw->window;
if (options_get_number(w->options, "pane-scrollbars") != 0) {
      TAILQ_FOREACH(wp, &w->panes, entry) {
          screen_redraw_draw_pane_scrollbar(c, wp);
       } 
   }
}
```
This seems overkill, is there a better way to do this?

However, this does a sort of weird thing.  In iTerm2, when this code is hit, it draws the scrollbar in column 1 (left side).  In PuTTY, it merely overwrites the existing scrollbar that is already output to the screen, so extraneous in that case.  Tracing through it with gdb, it's definitely positioning the cursor using tty_cursor() in the correct x and y.  I don't see why it would be doing that.

How can I check if the margins are on as you say and only do this refresh if necessary?  And is there an easy solution to know which client and wp to use?

nicm:
There is a `wp` in `struct screen_write_ctx`, but it may be NULL.

You don't want to do the redraw inside this function, you want to trigger it at the end of the main loop, so you will need a flag you can set similar to `PANE_REDRAW` and have `server_client_check_redraw` deal with it in a similar way - redraw only scrollbars on panes with the flag set, unless the scrollbars will be redrawn anyway by a more comprehensive flag.

You don't want to be redrawing entire panes when only the scrollbar changes, or redrawing all panes when only one changes, or all clients when only one. So you will need flags to allow you to redraw the minimum need where you need it, whether that be one client (similar to `CLIENT_REDRAWPANES` or one pane on all clients similar to `PANE_REDRAW`).

mgrant0:
Nic, I'd love to know if this current version fixes your iTerm2 display issue.  I tested it as best I could on a mac and it seems to update fine for me the little I played with it.  

This version does as you recommend by setting a flag and only updating the scrollbar when triggered.

Still some cleanup of names and such to come that I promised to do.

mgrant0:
I have started to move the position (left/right), the width, and padding into a style.  I needed to add these 3 fields to the style struct.  There is an 'align' in the style but 'centre' and 'absolute-centre' have no meaning for a scrollbar.  I can reuse align if you prefer.

Should I create a separate PR for these new style attributes?

nicm:
This looks good and it works well for me so far, but I think this diff is big enough, we can get it in first and then add the style attributes on top. Here are some more comments on the current diff:

- In tmux.h please remove argument names from the declarations and wrap at 80 columns.

- I think `window_copy_mode_get_current_offset_and_size` could go in as their own PR, but I would shorten it to just `window_copy_get_current_offset`. Do you want to pull it out into a new PR? Can you also fix the code style to match the surrounding code and [style(9)](https://man.openbsd.org/style) (don't worry about every detail because I can fix it up before commit but if you can get close it would be good).

- `layout_add_vertical_border` is unused?

- I think there are too many options and their names are inconsistent:
* `pane-vertical-scrollbars-position` and `pane-scrollbars` can be rolled together into one option called `pane-scrollbar-position` (to match `pane-border-status` and `pane-scrollbar-style`) with values `off`, `left`, `right`, `modal-left`, `modal-right`.
* `pane-vertical-scrollbars-width` can be deleted, the width can be fixed at 1. You can make this a `#define` if you like.
* `pane-vertical-scrollbars-pad` can also be deleted and the padding fixed at 0. Can also be  `#define` if you want for now.

- Why do you set `PANE_REDRAWSCROLLBAR` when the pane title changes in `input.c`? Is that necessary?

- In `options.c` all the bodies of the `if` statements do the same thing so they can be rolled together with `pane-border-status`:
~~~~ 
if (strcmp(name, "pane-border-status") == 0) ||
    strcmp(name, ...) {
...
}
~~~~

- `screen-redraw.c` looks good apart from some style issues.

- I think I would probably break the mouse stuff into its own PR to go in after the scrollbars if that is not too hard to do.

- I'm not really wild about the name "elevator" - is this a common name for this part of the scrollbar? I'm not sure of any alternatives.

- tmux.1 is a good start but need a little bit of work, again if you want to get close I can edit it further:
* A man page is not the place for opinion or advice, so don't say something is useful or desirable, instead give concrete examples or mention the defaults.
* Each new sentence should get a new line, so a new line after every `.` and `;`, this is a convention to make it easy to diff.
* Do not use the first or second person, it should be purely descriptive.
* You need to add the new mouse locations to the "MOUSE SUPPORT" section.
* I think there is too much text - I don't think it is necessary to explain tmux history or it would be enough just to say something like:
~~~~
Enable pane scrollbars.
Scrollbars are shown as a single column on the left or right of the each pane; a filled section know as the
.Ql
elevator indicates the position and size of the visible part of the pane content. 
By default, clicking the mouse on a scrollbar will enter copy mode; in copy mode, dragging the elevator will scroll.
See also
.Xr pane-scrollbar-style
and the scrollbar mouse events in the
.Sx MOUSE SUPPORT
section.

mgrant0:
Ok thanks for those comments.  Working through these as my time permits, thanks for being so responsive and patient!

The term "elevator" was what I recall when I worked for Sun ages ago.  I did some searching for the anatomy of scrollbar and sometimes it's called a "slider" and other times a "thumb".  "thumb" seems to be more popular at the moment and it's used as part of webkit.  Figma refers to it as a "thumb" in their docs.  It seems like "thumb" is the current most popular term.  That ok with you?

Some of the large if statements definitely need to be folded together carefully, I will try to do this without breaking the already working code.

mgrant0:
I'm good with "slider".

mgrant0:
I've addressed most of your comments.  I'd like to discuss the options...

I would prefer not to overload the `pane-scrollbar` option with `left`, `right`, `modal-left`, `modal-right`, it feels wrong.  There is already a `pane-scrollbar-style` to set the colours.  Would it be so bad if I moved the left-right position into this style?  I propose using the existing `align` style but ignore `centre` and `absolute-centre` alignment styles for scrollbars.  

With respect to the width and pad, I would also like to move those into this same style variable.  Some people prefer a wider scrollbar, some are more minimalist.  On a 4k screen, the scrollbar can be really narrow.  One person prompted me to add the pad which becomes more useful if you have a wider scrollbar.  I think there was another person asking about some padding in the pane-status, this `pad` style could be used for that as well.

Moving these 3 (align, width, and pad) into the style would remove those options.

I can add these styles as a separate PR.

I will move `window_copy_get_current_offset` into a separate PR.  I had a quick look at doing a separate PR for the mouse stuff but this looks hard.  I may try though.

nicm:
I think pad and width can go into the style, but I think the position is better as an option to match how we configure other components like the status line. It can be left as two options (pane-scrollbar off/on/modal and pane-scrollbar-position left/right) if you want.

In either case I don't want to commit this with options that will be changed or removed a week later, so pad/width should be removed for now and the others renamed.

nicm:
I don't think left/right makes sense in the style - the style should configure the style of that component *inside the component in place* if that makes sense - actually moving it around makes more sense as an option. This is now the others work - `pane-border-status` sets the position and `pane-border-style`/`pane-border-format` the content.

mgrant0:
Cool. Agreed.

mgrant0:
I believe I've addressed all of your comments above except moving the mouse support to a separate PR.  I'm still looking at this.

I'm looking into a bug where sometimes the scrollbar doesn't disappear when the window is scrolled to the bottom.

I can clean up / coalease some of the large if statements in screen-redraw.c.

Anything else?

nicm:
Looks good:

- Can you rename the options to all be `pane-scrollbars` or `pane-scrollbar` rather than a mixture? Singular would be match `pane-border-*` but then `pane-scrollbar` looks a little odd... I don't mind whether it is scrollbar or scrollbars really so long as they are all the same.

- Can you look at/apply this diff which fixes some spaces->tabs, renames some variables to reduce line lengths, makes some changes to the man pages and fixes some other style issues? [x.diff.txt](https://github.com/user-attachments/files/17464288/x.diff.txt)

- There are a number of long lines and other style issues in screen-redraw.c and server-client.c but I'll look at them after you have finished tidying up.

nicm:
Oh, there is a warning - `window_copy_scroll` is missing a declaration in `tmux.h`.

nicm:
Oh I also saw one minor bug, the slider does not go right to the bottom in copy mode, I guess just miscalculates something slightly:
Screenshot 2024-10-21 at 22 26 51

This is with 1696 lines of history.

nicm:
How about making the default style `bg=default,fg=white` rather than black?

nicm:
I think `window_copy_scroll` should not use `data->scroll_exit` but always 0, that is it should not exit copy mode if you use the scrollbar to scroll to the bottom.

nicm:
OK sorry for the flood of messages, here is another problem. Try a large history:
~~~~
tmux set -g history-limit 1000000
tmux neww
seq 100000
tmux set -g pane-scrollbars modal
C-b [
~~~~
It takes around 25 seconds to enter copy mode, and similar to exit it when I press `q`.

With scrollbars off it is pretty much instant.

mgrant0:
> OK sorry for the flood of messages, here is another problem. Try a large history:
> 
> ```
> tmux set -g history-limit 1000000
> tmux neww
> seq 100000
> tmux set -g pane-scrollbars modal
> C-b [
> ```
> 
> It takes around 25 seconds to enter copy mode, and similar to exit it when I press `q`.
> 
> With scrollbars off it is pretty much instant.

Hmm ok I will get this.  I must be doing something horribly inefficiently.  Thanks for finding those!

mgrant0:
> How about making the default style `bg=default,fg=white` rather than black?

Here's the difference for me:

This is with bg=default:

image

and this is with bg=black:

image

The difference is bg=black shows a "gully" where the scrollbar is.  I did that intentionally because it just didn't look right having the slider floating in space like that.  If you wanted to click above or below the slider to page up/down, you kinda have to guess where the scrollbar is if it's not visible.  I don't really mind what the default is, if you have a strong opinion here, let me know.  I can always set bg=black in my own setup.

nicm:
OK let's leave it as black for the moment.

mgrant0:
> Oh, there is a warning - `window_copy_scroll` is missing a declaration in `tmux.h`.

Did this go away with my latest push?

mgrant0:
I am getting to your other commets and the diff.

mgrant0:
I applied the diff and just going through what you changed.

You changed the variable `mouse_scrollbar_slider_grip` to `mouse_sb_sl_pos`.  This isn't the position of the slider.  It's the vertical position within the slider that the mouse was over when it first clicked down.  In other words, the point of the slider that the mouse "grabbed" on to.  This is tracked so that the slider moves up and down with the mouse and the mouse stays roughly attached to that point in the slider.  I struggled to find a more descriptive term than 'grip' or maybe 'grab' or perhaps 'anchor'.  Do you have any ideas other than 'pos'?

mgrant0:
What about `sl_my`?

nicm:
That's fine or `mpos` if it doesn't push the lines over 80 columns again.

mgrant0:
I tried what you said with the long history and it was instant for me.  Is it possible that this is something iTerm2 is causing when it reflows the lines to narrow and widen the window?

I fixed the bug where if you page up to the top and then press page up again, then try paging down to the bottom, it wasn't redrawing the scrollbar.  I think this is the same bug you were seeing as well.

nicm:
It is not slow on macOS only on OpenBSD... it does seem like it is only `modal` so I expect it is reflow, this is not a fast machine. Don't worry about it.

mgrant0:
> I think `window_copy_scroll` should not use `data->scroll_exit` but always 0, that is it should not exit copy mode if you use the scrollbar to scroll to the bottom.

The key binding by default is:

`bind -n MouseDrag1ScrollbarSlider { copy-mode -Se }`

If you remove the `e` then it does not exit copy mode.  The behaviour is configurable.  If I pass 0, then there's no way to configure it. What is the default when you have mouse mode on and you use the mousewheel to scroll up and then down?  For me it exits copy mode.  It feels like for consistency if you drag the slider, it should default to the same behaviour, however, I do see why you would want it not to exit copy mode.  You could be trying to do something in copy mode and you scroll to the bottom and all of a sudden you're no longer in copy mode.  I wonder if there is something more nuanced here like only exit if there is no selected text? There are several ways to scroll down, how to decide which ones causes copy mode to exit when at the bottom by default?

Do you prefer I pass in the arg like this?
```
window_copy_scroll(wp, c->tty.mouse_slider_mpos, event->m.y, args_has(args, 'e')
```
I see I may have an off-by-1 error when the slider is at the bottom and still in copy mode.

mgrant0:
Try what I just pushed.  I think I fixed the off-by-one error.  

I defaulted it to stay in copy mode when the user uses the scrollbar. 

> Actually that won't work because then if you enter copy mode with scrollbar it won't exit if you go to the bottom with keys

Which keys?  The NPage key?

I tried binding the NPage key in copy-mode to `page-down-and-cancel`.  This definitely cancels copy mode when you use the page down key to go to the bottom. 

I do see where it's possible to bind things such that by getting into copy mode by dragging the slider causes a different copy mode cancel behaviour versus getting into copy mode a different way.  But I wasn't able to create a situation where I couldn't get out of copy mode. 

Can you give me an example of what doesn't work?

nicm:
Don't worry, it is OK to use `scroll_exit` but don't pass `-e` for the scrollbar by default, I don't think it should exit at the bottom automatically.

This works well but there is still something funny about the slider. If I have enough lines in the history, I can make it disappear:

https://github.com/user-attachments/assets/c6467e65-d5f3-4fec-b7c1-905e1fb8f759

mgrant0:
Ok I see, that helped.  I reproduced it like you did with 2 side-by-side windows.  I hope that little fix fixed it, it seemed to fix it for me.

nicm:
Yes, it works for me now, thanks. What have you still to do on this?

mgrant0:
I was going to try and simplify or fold together some of the large if clauses in `screen_redraw_pane_border`.  Unless you find it's easier to deal with as it is.

There are a lot of places I check `pane-scrollbars` option before setting the `PANE_REDRAWSCROLLBAR` bit.  I think I can clean some of that up and remove those checks since it will only redraw the scrollbar if the scrollbar is actually active.  I was going to look at this today.

Any other cleanup needed...

Also, shouldn't the scrollbar disappear if the pane is in the alternate screen, for example, if you are editing a file?  Any clues on where/how to do that would be helpful.

If you are eager to commit this without those things, go ahead and I can always do another PR for those.  I also owe you a PR for the pad and width as style attributes.

nicm:
OK I'll wait until you do the cleanup before I look at it again. I would leave alternate screen for a separate MR.

mgrant0:
that cleanup cleans up the extraneous options checking.

mgrant0:
I did a scary cleanup of the if clauses in `screen_redraw_pane_border()`.  I sure hope I didn't break working code there!  If you can please try it and let me know if you see something weird with respect to scroll bars and the pane-border-status.

mgrant0:
I re-tabified the files I touched.  I noticed tmux.h, options-table.c, and screen-write.c have whitespace in lines that I don't think I touched, so I didn't re-tabify them.

I have updated my .emacs now to include `(setq-default indent-tabs-mode t)`.

mgrant0:
I wrapped as many long lines as I could.  server-client.c is hard because of the deep indenting.

mgrant0:
I think I see where to manage the alternate screen, will do that as a later PR.  

What else would you like cleaned up?

mgrant0:
Thinking of how to address some of the long lines and deep indenting in server-client.c...

At server-client.c:758 (the if which tries to determine the mouse `where` is in the border, pane, or scrollbar), I can pull this out into a separate function (or even 3 separate functions if you prefer).  Checking the status line just above this, it too might also be a candidate for a separate function or putting that into the above function. 

Anyway, just a thought as I was staring at this code today.

nicm:
OK this looks pretty close - I have made a new `scrollbars` branch in which I have removed the mouse code for the moment and done some minor tidying up.

Please take a look: 78b9096c3828ebd66caa4eded32c035d3d43b5a6 is the bulk of the changes and then fa007e43d5fd66e43cce35b3364ecfbdfc1b442f the changes to `screen-redraw.c`. The notable changes are:

- Remove all the mouse stuff for now.
- Remove or edit some comments and debug logging that I think is either not necessary or not clear.
- Remove `PANE_REDRAWSCROLLBAR` from `window_copy_write_lines` since both the callers set it already.
- In `screen_redraw_pane_border`, get rid of `pane_scrollbars` since it only cares about on/off and just use `sb_w == 0`.
- Always call `layout_fix_panes` when entering and leaving modes - it can't hurt anything to do it unnecessarily and it is simpler than checking the option.
- Remove unnecessary spacing changes (some of them my previous changes).

There are three things that still need to be looked at I think for this part of the code to go in:

- You redraw the scrollbars in `cmd-select-pane.c`. I'm not saying this is wrong, but why is is necessary? What could happens when the active pane changes that needs the scrollbars to be redrawn? 

- In `server-client.c`, `c->redraw_panes` is reused for both `PANE_REDRAW` and `PANE_REDRAWSCROLLBAR`. This makes it impossible to redraw the scrollbars without redrawing the entire pane, which means that for example `window_copy_scroll_up` will end up redrawing the entire pane as well as the lines which have scrolled. I think this probably needs to be split into two bitmasks.

- There is a bug - oversize windows are not handled correctly. Try this (increase the 200 if your terminal is more than 200 columns) and you will see the right pane scrollbar is always visible even when it should be hidden, and if you split and move around the panes you will see it is not redrawn in the right position.
~~~~
tmux new \; set -g pane-scrollbars on \; splitw -dh
tmux resizew -x200
~~~~

It would probably be best to make a new PR on this branch for any more changes on this.

Breaking up `server_client_check_mouse` or adding helper functions is a good idea but it would be best to do it first before we apply your mouse changes.

mgrant0:
Ok, so that I don't make a git mess, you want me to clone this new `scrollbars` branch into my github and then fix the bug and couple things and do PRs to address those?  At this point, I guess I shouldn't touch this branch/PR anymore or I'll create divergent branches?  The mouse stuff, are you going to merge that in yourself after?

nicm:
Here is an idea - pass a format into the style to let the slider colour to be configured: [sl.diff.txt](https://github.com/user-attachments/files/17544427/sl.diff.txt)

mgrant0:
I see the width bug.  With a single window, the scrollbar seems to be displayed correctly with respect to the screen itself, but incorrectly with respect to the actual size of the window pane.  And furthermore in the mouse enabled branch, it seems like the clickable area is off the screen (where you think the scrollbar aught to be).

Trying to reason this through a bit  If there was a horizontal scrollbar in the future, it seems logical to me it would be scrolling across this pane which was wider than the visible area.  And to be clear here, if you make the window wider, the screen shifts to the right when you enter a long line.  For me, if there was a horizontal scrollbar, it would be displayed with respect to the screen and not have part of the scrollbar hidden off the screen.

So I don't think the scrollbar should be invisible when the window is larger than the screen.  But the mouse should also work.

But the vertical scrollbar works with the _saved lines_ and this is not the same thing as the height of the window!  If the window is taller than the screen, what is it scrolling?  only the copy buffer? both?  If the window was overly tall, maybe first you scroll up to see the extended height of the screen, then, above that it's the copy buffer.  But does scrolling up put you in copy mode if you are not displaying the saved lines yet?  When I think about it this way, it is definitely easier to say the scrollbar is for the copy buffer only and draw the scrollbar up beyond the visible area and not have to worry about this intermediate case where the scrollbar seems to be scrolling two separate things.  It seems like there should be a second set of scrollbars to deal solely with windows that are wider or taller than the visible screen.  I am not a fan of scrollbars within scrollbars.

If I split the window, really weird stuff happens.  

I am using the width and height and offset of the pane to draw the scrollbar.  Maybe I should be using some others which are relative to the visible area instead?  Hmm.

nicm:
The scrollbar is part of the pane and relates to the pane content so the size of the window only matters in that it needs to be drawn (or not drawn or partly drawn) in the right place, everything else behaves the same.

mgrant0:
Where do i get the current size of the visible window (versus the window pane)?  

I see what's going on, should be easy to do what you say.  It's using `tty_cursor` and it's trying to position the cursor in column 200 when the actual window is only say 100 columns wide.  `tty_cursor` seems to just be going to the maximum width of the screen rather than trying to draw off the screen.

mgrant0:
thanks, that defiintely looks like what I need but haven't gotten it to work yet.

nicm:
I have applied the mouse changes to the `scrollbars` branch now, could you please look over it and make sure I have not missed anything (or undid any previous changes)? The commit is 3eb7ecfd.

nicm:
If you can fix it then you can either make a PR or attach the diff here, otherwise I'll take a look later.

mgrant0:
I think I'm going nuts.  Maybe there isn't a problem.  I can't reproduce what I thought I saw.  What I thought I saw was when I enabled `pane-border-status` on either top or bottom, the scrollbar didn't go all the way to the bottom.  I dont know.  Maybe I ran an older version, then compiled it and the problem was gone.  This appears to be a false alarm.

nicm:
OK that's good, I'll take another look over the mouse stuff later if it is all working and get it in as well.

PR #4082, opened 2024-08-18
tty_keys_clipboard: Fix size calculation

holio0:
When parsing for both terminators '\007' and '\033\\', the index `end` consistently points to the final character of the terminator. Therefore, the size calculation should be simplified to `end + 1` instead of `end + size-of-terminators`.


nicm:

PR #4085, opened 2024-08-20
Include in the man page, that `save-buffer` and `load-buffer` are capable to read from `stdin`

### LINKED ISSUE #3341: The man page should include, that `save-buffer` and `load-buffer` are able to read from `stdin`
### Issue description
The man page does not tell the user, that `load-buffer` and `save-buffer` are able to read from `stdin`; only `path` is mentioned:
```bash
$ man 1 tmux
load-buffer [-w] [-b buffer-name] [-t target-client] path
                   (alias: loadb)
             Load the contents of the specified paste buffer from path.  If -w
             is given, the buffer is also sent to the clipboard for
             target-client using the xterm(1) escape sequence, if possible.
[...]
save-buffer [-a] [-b buffer-name] path
                   (alias: saveb)
             Save the contents of the specified paste buffer to path.  The -a
             option appends to rather than overwriting the file.
```

It should look like this, so it is obvious to choose between `path` and `stdin`:
```no-highlight
load-buffer [-w] [-b buffer-name] [-t target-client] [path | -]
[...]
             target-client using the xterm(1) escape sequence, if possible.
             If - is given, the contents are read from stdin.
[...]
save-buffer [-a] [-b buffer-name] [path | -]
[...]
             option appends to rather than overwriting the file.
             If - is given, the contents are read from stdin.
```

I am going to create a pull-request, once I have figured out how to edit the man page file [`tmux.1`](https://github.com/tmux/tmux/blob/master/tmux.1) properly.

### Required information
* `tmux` version

```bash
$ tmux -V
tmux 3.3a
```

* Platform
```bash
$ uname --kernel-name --processor
Linux Intel(R) Core(TM) i5-3450 CPU @ 3.10GHz
```

* `${TERM}` inside and outside of `tmux`
* Inside `tmux`
```bash
$ printenv TERM
tmux-256color
```
* Outside `tmux`
```bash
$ printenv TERM
st-256color
```
I am actually using `alacritty` as terminal emulator.

Not needed.

---

Side note:
The parameter is mentioned in the file [CHANGES](https://raw.githubusercontent.com/tmux/tmux/master/CHANGES) at `CHANGES FROM 1.2 TO 1.3, 18 July 2010`.

keks24:
Resolves #3341.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -6559,7 +6559,7 @@ section.
 .Op Fl w
 .Op Fl b Ar buffer-name
 .Op Fl t Ar target-client
-.Ar path
+.Op Ar path | -
 .Xc
 .Tg loadb
 .D1 Pq alias: Ic loadb
@@ -6572,6 +6572,9 @@ is given, the buffer is also sent to the clipboard for
 using the
 .Xr xterm 1
 escape sequence, if possible.
+If
+.Ar -
+is given, the contents are read from stdin.
 .Tg pasteb
 .It Xo Ic paste-buffer
 .Op Fl dpr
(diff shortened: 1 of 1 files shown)


nicm:
I have applied this (to OpenBSD, will be in GitHub later) although I changed the text slightly. Thanks!

PR #4088, opened 2024-08-22
Unset systemd socket environment variables for sessions

floppym:
These variables are only useful for the tmux server process when it is started via systemd socket activation.

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -377,6 +377,12 @@ main(int argc, char **argv)
 		environ_put(global_environ, *var, 0);
 	if ((cwd = find_cwd()) != NULL)
 		environ_set(global_environ, "PWD", 0, "%s", cwd);
+#ifdef HAVE_SYSTEMD
+	/* Avoid passing systemd socket info to sessions */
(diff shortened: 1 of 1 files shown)


floppym:
It might make sense to unset some other systemd-specific variables as well.

https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Environment%20Variables%20Set%20or%20Propagated%20by%20the%20Service%20Manager

If there's a better way to handle this, let me know.

nicm:
How is this different from https://github.com/tmux/tmux/commit/2ac0faf119525ef6d2c4f8f1875c3f51ec224d33?

floppym:
Apologies, I wasn't aware of 2ac0faf119525ef6d2c4f8f1875c3f51ec224d33. That looks like it will resolve the issue in a similar way.

The only difference is that 2ac0faf119525ef6d2c4f8f1875c3f51ec224d33 removes the variables for each session, whereas my PR removes them from the global environment a single time.

floppym:
Closing as this appears to be resolved on master already.

PR #4093, opened 2024-08-25
Uint64 client flags

mgrant0:
This cleans up several cases where the struct client flags or variables using this field were not uint64_t.  

In server_client_check_redraw in server_client.c, changed new_flags to new_cflags for clarity because the variable flags was also used in this function but for the tty->flags. 

I noticed in server_client_dispatch_identify, flags is signed int but didn't change it, it doesn't seem like it's used now that flags is passed in with MSG_IDENTIFY_LONGFLAGS instead of MSG_IDENTIFY_FLAGS.  I left that case and that signed integer var in the code along with the MSG_IDENTIFY_FLAGS case.  Should that case be deleted now completely?

### DIFF
--- a/client.c
+++ b/client.c
@@ -452,11 +452,12 @@ client_send_identify(const char *ttynam, const char *termname, char **caps,
 {
 	char	**ss;
 	size_t	  sslen;
-	int	  fd, flags = client_flags;
+	int	  fd;
+	uint64_t  flags = client_flags;
 	pid_t	  pid;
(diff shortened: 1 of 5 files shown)


nicm:
Applied now with some minor changes (I renamed new_cflags and client_flags and flags to tty_flags in server_client_check_redraw). MSG_IDENTIFY_FLAGS can stay since it is harmless and means we are still compatible with old clients. Thanks!

PR #4094, opened 2024-08-25
Remove the pane_status arg from many function calls and use it in screen_redraw_ctx instead

mgrant0:
Multiple functions in screen_redraw.c pass a pane_status flag variable around.  This patch cleans up this by passing around screen_redraw_ctx instead and using the pane_status within that struct.  This will allow adding other flags in the future.


nicm:

PR #4095, opened 2024-08-26
add copy-mode -d arg to page down

mgrant0:
Add a '-d' option to copy-mode for consistency.  Though this is not strictly necessary since you can use `send -X page-down`, this just adds consistency so you can bind NPage and PPage to copy-mode commands.

Also added clarification that the copy-mode command not only gets you into copy mode but it also functions in copy mode, as in, to scroll up and down.


nicm:

PR #4096, opened 2024-08-26
new window_pane_mode function

mgrant0:
New function to return which mode a window pane is in.


nicm:

PR #4097, opened 2024-08-26
new layout_cell_is_left and layout_cell_is_right functions

mgrant0:
New layout_cell_is _ functions to return if cell is on left or right for future feature.


nicm:
These look fine, thanks, but I will actually not apply it now because they are unused so it adds warnings, so leave it as part of your main diff. Thanks!

PR #4108, opened 2024-09-06
Add an option prefix-timeout to ignore prefix key after timeout

taylorconor:
Hi! This PR introduces the `prefix-timeout` option. When set to non-zero, it defines the amount of milliseconds after the `prefix` key is pressed before it is assumed to have "timed out".

Example config entry:
```
# Use a 5 second prefix key timeout.
set -g prefix-timeout 5000
```

This is my first tmux PR -- happy to iterate on it as much as necessary if you have feedback!

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -385,6 +385,17 @@ const struct options_table_entry options_table[] = {
 	  .text = "Maximum number of server messages to keep."
 	},
 
+	{ .name = "prefix-timeout",
+	  .type = OPTIONS_TABLE_NUMBER,
+	  .scope = OPTIONS_TABLE_SERVER,
+	  .minimum = 0,
+	  .maximum = INT_MAX,
+	  .default_num = 0,
+	  .unit = "milliseconds",
+	  .text = "The timeout after pressing the prefix key if no subsequent key is pressed. "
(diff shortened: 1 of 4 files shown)


nicm:
Looks OK thanks but what's the use case for this?

taylorconor:
My prefix key is used outside of tmux with subsequent key combinations, which are processed lazily. Each key in the combination is sent to tmux as it's pressed, and the combination is only recognized (by karabiner) when the following keys are pressed. This results in tmux receiving a prefix key that’s part of a combination outside tmux. I want tmux to "time out" the prefix after X seconds if unused, to avoid manual cancellation when back inside tmux.

I know this is a pretty niche use-case but I was surprised there wasn't already support for prefix timeouts given all the configurability already present in tmux!

nicm:
OK thanks - this looks fine except `server_client_key_table_activity_diff` - it seems like it could be a lot neater with a couple of temporaries or maybe using `timersub`, also it needs a newline before the `{` and should be `struct client *c` not `struct client* c`.

OpenBSD is currently preparing for release so I will probably wait for a while before I put this in.

taylorconor:
Thanks, that's much cleaner. I've updated it now :)

nicm:

Issue #3873, opened 2024-03-07
scroll-{bottom,middle,top} copy mode commands removed from tmux.1


borosai:
While looking at older tmux.1 commits, I came across the addition of the scroll-middle, scroll-bottom, and scroll-top copy mode commands. However, when I searched the man page, none of them were there. After looking into it, it seems that they were accidentally removed in the following commit:

https://github.com/tmux/tmux/commit/efded95ed7f5d206527bf517940f08a54f3c5930

I didn't check if any other commands were left out, but it wouldn't hurt to review and confirm that nothing else is missing.

PR #4121, opened 2024-09-15
Add missing copy mode commands to the manpage

druckdev:
When the list of copy mode commands in the manpage was rewritten to add descriptions in commit efded95ed7f5 ("Add descriptions of copy mode commands, from Michael Bianco."), some of the commands were forgotten. Add those.

I tried to stick as close as possible to already existing wording.

Fixes: #3873

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -1814,6 +1814,23 @@ is used to name the new paste buffer.
 .Xc
 Copy from the cursor position and exit copy mode.
 .It Xo
+.Ic copy-pipe-end-of-line
+.Op Ar command
+.Op Ar prefix
+.Xc
+Copy from the cursor position to the end of the line and pipe the text to
+.Ar command .
+.Ar prefix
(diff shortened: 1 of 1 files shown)


druckdev:
I know that you credit the commit authors via the message because you apply the patches first in CVS but would it be possible for you  - in case you merge this - to somehow keep the author data in the git commit? Thanks!

druckdev:
I added a Co-authored-by trailer to the commit message. If there is a way for you to keep the author data intact that'd be great and I would prefer that, but if not, it is preserved in somewhat of a git-ish way like this.

nicm:

Everyone who has something that at least looks like a real name or email gets credited in the commit message and I also mention the GitHub issue number which means anyone can look up the issue (at least so long as GitHub still exists and tmux is still using it). That is the best I can do for authorship in commit messages.

PR #4122, opened 2024-09-16
Remove spurious white space.

larunbe:
Just remove a couple of extraneous whitespaces.


nicm:
Thanks for this. It is in OpenBSD now and should be in GitHub later. Unfortunately because I committed it as part of another change I forgot to mention this GitHub issue as well, sorry :-/.

PR #4126, opened 2024-09-19
Add install step from source

supsup:
Perhaps this can be worded better, but it would be nice to have an install step from source as well as just build.


nicm:
I think we should just add "$ sudo make install" after the existing make step.

I don't think its necessary to give a separate example with utempter.

nicm:
Added the install line now, thanks.

Issue #2168, opened 2020-04-16
event.h is deprecated


lhoenig:
"The <event.h> header is deprecated in Libevent 2.0 and later; please
  use <event2/event.h> instead.  Depending on what functionality you
  need, you may also want to include more of the other event2/
  headers."

As tmux depends on libevent2 anyways I see no problems adhering to this suggestion, other than a few extra include lines.

nicm:
This is not going to happen for the moment because we still use libevent 1.4 on OpenBSD and it is not worth the portable goop until they actually remove the header.

lhoenig:
I see, thank you for the quick reply!

nicm:
I thought of doing it via -I in configure.ac but it would be a pain and probably be hard to make work properly, I'm not sure it is worth it for the moment. Thanks for mentioning it anyway :-).

PR #4138, opened 2024-09-27
Deprecate event_init() by event_base_new()

Freed-Wu:


nicm:
`event_base_new` is OK but `event_assign` is missing from libevent 1.4 which means the code would diverge between OpenBSD and portable, so I don't think we want to do this, at least in this way. Are libevent actually removing `event_set`/`event_init`?

Freed-Wu:
https://github.com/tmux/tmux/issues/2168#issuecomment-614593074
So this program cannot drop these deprecated functions, right?

Why it has to use libevent 1.4 in BSD?

PR #4148, opened 2024-10-02
Cygwin and MSYS2 package patches

rkitover:
These are the patches I applied from both the MSYS2 and Cygwin packages of tmux, with the authors set to their original authors.

If there are any issues with them I will be happy to fix them.


nicm:
Thanks.

What does `MSYSTEM` do and why does it need to be preserved? I can't find any obvious documentation on it.

Is there a difference between `__CYGWIN__` and `__MSYS__`? Can we pick one and use it consistently?

nicm:
I've applied the `out_fd` change.

nicm:
How about something like this instead of the `#ifdef` for the permissions check? Can you test? [x.diff.txt](https://github.com/user-attachments/files/17228585/x.diff.txt)
~~~~
diff --git a/Makefile.am b/Makefile.am
index 8e5f72b1..6c7d2d3f 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -66,6 +66,11 @@ if IS_HAIKU
 AM_CPPFLAGS += -D_BSD_SOURCE
 endif
 
+# Set flags for Cygwin.
+if IS_CYGWIN
+AM_CPPFLAGS += -DTMUX_SOCK_PERM=0
+endif
+
 # List of sources.
 dist_tmux_SOURCES = \
 	alerts.c \
diff --git a/configure.ac b/configure.ac
index b63d7e11..4f90fcd5 100644
--- a/configure.ac
+++ b/configure.ac
@@ -982,6 +982,7 @@ AM_CONDITIONAL(IS_NETBSD, test "x$PLATFORM" = xnetbsd)
 AM_CONDITIONAL(IS_OPENBSD, test "x$PLATFORM" = xopenbsd)
 AM_CONDITIONAL(IS_SUNOS, test "x$PLATFORM" = xsunos)
 AM_CONDITIONAL(IS_HPUX, test "x$PLATFORM" = xhpux)
+AM_CONDITIONAL(IS_CYGWIN, test "x$PLATFORM" = xcygwin)
 AM_CONDITIONAL(IS_HAIKU, test "x$PLATFORM" = xhaiku)
 AM_CONDITIONAL(IS_UNKNOWN, test "x$PLATFORM" = xunknown)
 
diff --git a/tmux.c b/tmux.c
index 30960245..6659e1c3 100644
--- a/tmux.c
+++ b/tmux.c
@@ -222,7 +222,7 @@ make_label(const char *label, char **cause)
 		xasprintf(cause, "%s is not a directory", base);
 		goto fail;
 	}
-	if (sb.st_uid != uid || (sb.st_mode & S_IRWXO) != 0) {
+	if (sb.st_uid != uid || (sb.st_mode & TMUX_SOCK_PERM) != 0) {
 		xasprintf(cause, "directory %s has unsafe permissions", base);
 		goto fail;
 	}
diff --git a/tmux.h b/tmux.h
index dfc81cee..4db4b670 100644
--- a/tmux.h
+++ b/tmux.h
@@ -84,6 +84,9 @@ struct winlink;
 #ifndef TMUX_SOCK
 #define TMUX_SOCK "$TMUX_TMPDIR:" _PATH_TMP
 #endif
+#ifndef TMUX_SOCK_PERM
+#define TMUX_SOCK_PERM (7 /* o+rwx */)
+#endif
 #ifndef TMUX_TERM
 #define TMUX_TERM "screen"
 #endif

rkitover:
Not really, the government ****s I've mentioned are preventing me from even running `./configure`.

But I'm sure this is fine.

I will ask someone else to test.

Issue #3879, opened 2024-03-10
sixel image bleed


joseluis:
This simple sixel image [map8_libsixel.six.zip](https://github.com/tmux/tmux/files/14551022/map8_libsixel.six.zip) created with libsixel from this png ![map8](https://github.com/tmux/tmux/assets/5832068/58b12c14-5f07-479e-93bd-9788c3f0b054) seemingly displays incorrectly in tmux, bleeding the first color onto the next line, in all terminals tested (alacritty graphics branch, wezterm and xterm -ti vt480.

This is all using: `tmux next-3.5` on `Linux x86_64`, invoked with `-f /dev/null -vv new`. $TERM in tmux is `tmux-256color`.

![alacritty](https://github.com/tmux/tmux/assets/5832068/d5e9a5ac-92e4-43c9-b714-2d974f492018) alacritty
![alacritty-tmux](https://github.com/tmux/tmux/assets/5832068/e72c3366-41ad-4929-9d09-deb85f775ec8) alacritty + tmux

![wezterm](https://github.com/tmux/tmux/assets/5832068/f6bdca66-5ead-421c-8856-657165a7dcfc) wezterm
![wezterm-tmux](https://github.com/tmux/tmux/assets/5832068/05b4db8e-f020-4d38-8453-83a88a6f8d56) wezterm + tmux

![xterm_vt340](https://github.com/tmux/tmux/assets/5832068/b1799693-5640-4fc9-a761-e32cef5f5fb1) xterm
![xterm_vt340-tmux](https://github.com/tmux/tmux/assets/5832068/ab16dfb7-a883-4d4a-81f2-0831ad4bfb77) xterm + tmux

[tmux-logs.zip](https://github.com/tmux/tmux/files/14551082/tmux-logs.zip)

nicm:
This seems to be fixed in master by #4149, let me know if you still see problems. Thanks!

PR #4149, opened 2024-10-02
Support Sixel transparency

bptato:
By default, unspecified pixels on Sixel images are opaque. However, when P2 is set to 1, they are left transparent by the terminal. Pass through this parameter.

Additionally, raster attribute dimensions smaller than the actual image size may be specified, which turns unset pixels *outside* the specified rectangle transparent. This is commonly used to crop out some rows of the last sixel of an image (which otherwise would have to align to a multiple of 6).

Instead of sending the actual dimensions of the bitmap, we now pass through the user-specified raster attribute dimensions.

Finally, to avoid collapsing empty sixel lines, we no longer ignore duplicate "-" newline specifiers.

Fixes GitHub issue #3879

### DIFF
--- a/image-sixel.c
+++ b/image-sixel.c
@@ -37,8 +37,13 @@ struct sixel_image {
 	u_int			 xpixel;
 	u_int			 ypixel;
 
+	u_int			 set_ra;
+	u_int			 ra_x;
+	u_int			 ra_y;
+
 	u_int			*colours;
 	u_int			 ncolours;
+	u_int			 p2;
 
 	u_int			 dx;
 	u_int			 dy;
@@ -166,6 +171,10 @@ sixel_parse_attributes(struct sixel_image *si, const char *cp, const char *end)
 	si->x = x;
 	sixel_parse_expand_lines(si, y);
 
+	si->set_ra = 1;
+	si->ra_x = x;
+	si->ra_y = y;
+
 	return (last);
 }
 
@@ -268,7 +277,7 @@ sixel_parse_repeat(struct sixel_image *si, const char *cp, const char *end)
 }
 
 struct sixel_image *
-sixel_parse(const char *buf, size_t len, u_int xpixel, u_int ypixel)
+sixel_parse(const char *buf, size_t len, u_int p2, u_int xpixel, u_int ypixel)
 {
 	struct sixel_image	*si;
 	const char		*cp = buf, *end = buf + len;
@@ -282,6 +291,7 @@ sixel_parse(const char *buf, size_t len, u_int xpixel, u_int ypixel)
 	si = xcalloc (1, sizeof *si);
 	si->xpixel = xpixel;
 	si->ypixel = ypixel;
+	si->p2 = p2;
 
 	while (cp != end) {
 		ch = *cp++;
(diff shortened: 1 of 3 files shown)


nicm:
This looks OK, thanks.

Which terminals did you test this on?

Can you please give me at least one test image showing these features?

nicm:
Your last change is wrong, you can't use the incoming size in `sixel_scale`.

bptato:
So I'm thinking of an approach like:

```
--- a/image-sixel.c
+++ b/image-sixel.c
@@ -434,9 +434,17 @@ sixel_scale(struct sixel_image *si, u_int xpixel, u_int ypixel, u_int ox,
        new->xpixel = xpixel;
        new->ypixel = ypixel;
        new->p2 = si->p2;
+
        new->set_ra = si->set_ra;
-       new->ra_x = si->ra_x;
-       new->ra_y = si->ra_y;
+       /* clamp to slice end */
+       new->ra_x = si->ra_x < psx ? si->ra_x : psx;
+       new->ra_y = si->ra_y < psy ? si->ra_y : psy;
+       /* subtract slice origin */
+       new->ra_x = new->ra_x > pox ? new->ra_x - pox : 0;
+       new->ra_y = new->ra_y > poy ? new->ra_y - poy : 0;
+       /* resize */
+       new->ra_x = new->ra_x * xpixel / si->xpixel;
+       new->ra_y = new->ra_y * ypixel / si->ypixel;
 
        for (y = 0; y < tsy; y++) {
                py = poy + ((double)y * psy / tsy);
```

but I can't figure out when sixels are actually scaled, so I don't know how to test it...
(First thing I tried was changing the XTerm font size, but that just clears the sixels. So does reattaching with a different font size, or splitting the window, etc. Is this intended?)

nicm:
They will be scaled if the xpixel and ypixel for the outside terminal is different from tmux's. Try attaching the same session from two terminals with different font sizes, or changing the font size after showing an image out.

IIRC it's also used if a SIXEL is partially off the terminal so scrolling or if the window size is bigger than the terminal (try something like `resize-window -x 200; splitw -h`), but I'm not sure if that will show what you want since the xpixel and ypixel will be the same.

bptato:
OK, it seems to work like with this patch; please let me know if there's anything else that looks incorrect.

nicm:
OK this seems good to me, I have applied it now. iTerm2 does not seem to support these features but the images still show up and redraw as I expect so I will assume that you know what you are doing when it comes to the rest :-).

Thanks!

bptato:
Thank you :)

> iTerm2 does not seem to support these features

Oh, I was not aware of that. Sixel emulation is quite widespread nowadays, but terminals still handle the details inconsistently... e.g. MLTerm, which had sixels quite early on, also only recently got P2=1 support.

Even though, as arakiken observes[^0], unset pixels had to be treated as "transparent" in the original Sixel dot-matrix printers, because it was physically impossible for them to control the "background color" (i.e. paper). So it's natural that VT340 terminals could also emulate transparency/layered images.

[^0]: https://github.com/isaki68k/sayaka/issues/36#issuecomment-2354378255

Issue #4123, opened 2024-09-17
Feature: Finer control over set-clipboard / copy-selection behaviour


hetmankp:
### Issue description

Current tmux behaviour means the set-clipboard option determines both the OSC 52 security level, and whether copying text into the tmux buffer also automatically sends it to the terminal emulator via OSC 52. This is a problem if one would like to use different shortcuts to copy into the tmux buffer and into the system clipboard by using OSC 52, without these two behaviours overlapping with each other.

### Solutions

Two possible solutions to tackle this issue:

1. Provide a different configuration options to manage OSC 52 security vs the option used to determine if copied data should be automatically sent via OSC 52. Unfortunately right now, setting `set-clipboard` to `off`, also means that `copy-pipe` is not able to send OSC 52 messages to the terminal emulator. The most backwards compatible solution would probably be to create a new config option, for example called `send-to-clipboard` and have it `on` by default. This setting would determine if tmux should send copied data to the clipboard regardless of security level. This would ensure existing configurations continue working as they currently do.
2. Provide an alternative to `copy-selection` and `copy-selection-and-cancel` etc (e.g. `copy-selection-no-send` ?). These new commands would copy a selection to the tmux buffer but not send it onto the clipboard.

### Required information

* tmux version (`tmux -V`): `tmux next-3.5`
* Platform (`uname -sp`): `Linux x86_64`
* $TERM inside and outside of tmux (`echo $TERM`): `tmux-256color`, `xterm-256color`
* Logs from tmux (`tmux kill-server; tmux -vv new`): N/A

nicm:
I'd say add a flag to `copy-selection` and friends (like `-o` for `next-prompt`) which skips setting the clipboard.

hetmankp:
Yes I like that better than option no. 2. Is this something you'd like to see a pull request for?

nicm:
It would certainly help :-).

hetmankp:
Great, I'll try to take a look at this soon.

PR #4153, opened 2024-10-03
Finer control over copy-selection behaviour

hetmankp:
So this is a somewhat ambitious attempt to resolve issue #4123. In order to make the code for this feature simpler there is an overhaul of how the window-copy.c command table deals with arguments so that individual commands don't have to reimplement argument parsing from scratch.

If this change is too broad, it's possible to keep the old approach to argument handling for existing command while only leveraging it for the copy commands, otherwise I can revert to doing something internal to the copy commands only, but this can get a bit ugly given some of them also want the prefix argument.

I've chosen to implement two flags for the copy commands. The original one discussed in #4123, but also one to suppress adding text to the paste buffers. This is both for symmetry but also because the functionality to paste only to the terminal clipboard would be useful to users (like my self) who wouldn't need to reimplement this with an external script given tmux can already do it. The choice of '-C' and '-P' is just what made sense to me but let me know if different letters would fit better here.

Finally, the code for the new feature could be neater if the `struct args` pointer was passed deeper into the call stack. For consistency with the other functions I haven't done that, doing all the argument handling at the same level, but that may have been partially an artefact of the older way argument handling was being done, and it results in some repetition. I think I'd like to push the `cs->wargs` pointer all the way into the window_copy_copy_buffer() function so there's fewer repetitions in the code (replacing the `int set_paste` and `int set_clip` function parameters), if you think that will fit in the style of the codebase.

I can make an attempt to tweak the wiki once this feature is finalised.


nicm:
Great stuff, this looks good. But incremental search doesn't work anymore. Try:
~~~~
tmux -Ltest -f/dev/null new -d \; set -g mode-keys emacs \; a
~~~~
Then search for something with `C-r` in copy mode - it will search once but pressing `C-r` again will not repeat.

Probably `args_parse` is eating it because of the leading `-` used as a repeat marker for `PROMPT_INCREMENTAL`? Not sure the best way to fix this...

hetmankp:
Oh boy, that's quite the side effect.

I think there is an argument to be made here for expanding the behaviour or `args_parse()` a little. Specifically it should be possible to tell it a particular command doesn't expect any flags so all arguments should be treated as ordinary values.

So instead of the usual empty template `""` (which says all flags should be rejected) we might specify something like `"-"` and give that string a special meaning (resulting in `.args = { "-", 1, 1, NULL }`). Alternatively we could declare a reference to a magic string in memory, so we would have something like this instead: `.args = { ARGS_PARSE_NO_FLAGS, 1, 1, NULL }`. Either way, then it would be easy for me to fix the commands affected by this in `window-copy.c`.

The other option would be to support both the old parsing scheme and the new parsing scheme along side each other. So some commands would be declared like this:

```
	{ .command = "search-backward-incremental",
	  .minargs = 1,
	  .maxargs = 1,
	  .clear = WINDOW_COPY_CMD_CLEAR_ALWAYS,
	  .f = window_copy_cmd_search_backward_incremental
	},
```

And some like this:

```
	{ .command = "copy-selection",
	  .args = { "CP", 0, 1, NULL },
	  .clear = WINDOW_COPY_CMD_CLEAR_ALWAYS,
	  .f = window_copy_cmd_copy_selection
	},
```

Relying on the fact the `struct args_parse` `template` field will be `NULL` if it isn't declared. But this feels a little uglier to me and might lead to a bit more inconsistency in the code.

nicm:
What if you change the `C-r` key binding to add `--`?

nicm:
I think that is OK, I can't see a lot of people having these as custom commands and it is an easy fix if they do.

hetmankp:
Ok I pushed a fix that should take care of all calls to affected commands. I wasn't sure if I should be updating the configs in the `regress/` directory, but it looks like at least one in there had been updated so I did those as well.

nicm:
This looks good to me now, thanks! I will apply it - if you want to be credited let me know your real name or an email address, otherwise I will just mention this issue number.

hetmankp:
Nah if you don't mention Github handles the issue number is fine, glad we got something to solve the issue.

nicm:
Great stuff, I have applied this to OpenBSD now, it will be in GitHub later. I made a couple of very minor changes for line length, let me know of any problems. 

Thanks!

Issue #3712, opened 2023-10-08
Question about repeat-time


systemmonkey42:
Under most Operating systems, the keyboard has two confiurable settings for key repeat;, Initial delay, and repeat speed.

I find the `repeat-time` setting has to be higher than the default to overcome the "initial delay", but once you have
it repeats fine.

The downside of having the repeat time set so high, is once you finish resizing a pane for example, you must wait for a second
before pressing the cursor keys for example.  I would often try to press <kbd>Up</kbd> to recall the last bash command, and inadvertently end up triggering the previous-pane binding.

I would like to request the ability to configure initial and repeat delays separately.

The initial delay would apply only after the initial key press, all subsequent presses would be the shorter delay.

A quick read of the code shows you might be able use `c->flags & CLIENT_REPEAT` to identify the initial repeat vs subsequent
repeats. 

eg,
```
set repeat-time 500
set repeat-time-inital 1000
```

I tried implementing a trivial example, but there was an issue where changing which key was being held (like from <kbd>Ctrl-Up</kbd> to <kbd>Ctrl-Down</kbd> wouldn't work unless the initial delay was applied again.  This required remembering the key-code of the last key to detect changes.  I managed this by adding a `keycode last_key` to the client struct, but I'm sure theres a much cleaner way.

Having a shorter repeat-time (300) and a longer repeat-time-initial (900), I've found usablility improved significantly.

I can create a PR for my initial attempt if that would be of value.

Regards

nicm:
I completely missed this issue for some reason, sorry... yes if you have something please open a PR. I think `initial-repeat-time` would be more readable than `repeat-time-initial`.

systemmonkey42:
As suggested, I've create pull request #4164 with the new option name.

I've used this on all my installations of tmux for about a year now, and while I have no problems with it, there is one other problem related to key-repeat handling I havent been able to solve...

Press <kbd>PREFIX</kbd>, press <kbd>UP</kbd>, then press <kbd>UP</kbd> again.  Unless you've waited a while, its going to trigger the repeat-time and move up another pane.
The only "correct" soluition for this is to have the terminal report key-release events.  No exactly a widely supported feature.. (only possible on the console on linux and only by taking over the keyboard)

PR #4164, opened 2024-10-05
Added support for initial-repeat-time #3712

systemmonkey42:
This patch adds a new global option called `initial-repeat-time`.  This provides the means to differentiate between the time taken for the first key repeat, and the time between subsequent key repeats.

This helps to reduce the instances where a repeating key, once released, triggers
additional repeats when it should be sent to the active pane.

A typical example is using <kbd>PREFIX-ALT_UP</kbd> to resize a pane.  After resizing a few lines, then releasing <kbd>ALT</kbd>, you would normally have to wait a second before pressing  <kbd>UP</kbd> to recall a shell command.  With this patch, the delay is tunable to a much smaller value.

I currently use the following values which work well on remote sessions
(local sessions can often have a smaller "repeat-time")

```
set-option -g repeat-time 350
set-option -g initial-repeat-time 900
```

Patch includes the addtional man page entries.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -571,6 +571,18 @@ const struct options_table_entry options_table[] = {
 		  "If changed, the new value applies only to new panes."
 	},
 
+	{ .name = "initial-repeat-time",
+	  .type = OPTIONS_TABLE_NUMBER,
+	  .scope = OPTIONS_TABLE_SESSION,
+	  .minimum = 0,
+	  .maximum = 10000,
+	  .default_num = 900,
+	  .unit = "milliseconds",
+	  .text = "Time to wait for a key binding to repeat the first time the "
+	          "key is pressed, if it is bound with the '-r' flag. "
+	          "Subsequent presses use the 'repeat-time' option."
+	},
+
 	{ .name = "key-table",
 	  .type = OPTIONS_TABLE_STRING,
 	  .scope = OPTIONS_TABLE_SESSION,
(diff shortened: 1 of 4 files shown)


nicm:
This looks good but I think either `initial-repeat-time` should default to zero *or* we should change to your 900/350 suggestion...

nicm:
I have made a few changes to your diff. I've rewritten the man page info and option description to be a bit clearer, and also broken the repeat check out into a helper function to avoid some long lines in `server-client.c`. Can you please give this a look/test and make sure I haven't broken it? [repeat.diff.txt](https://github.com/user-attachments/files/17266197/repeat.diff.txt) Thanks

systemmonkey42:
Testing now.

> I think either initial-repeat-time should default to zero

I tend to agree the initial should default to zero, with the repeat at the default value of 500,  Then users who need it can bump `initial-repeat-time` to something higher.  A good combination is having `initial` slightly higher than 500, and the repeat time slighly lower.  600/350 works well on the connection I'm currently testing on.

For me, the default "repeat-time" of "500" is too short to ever get standard key-repeat to kick in.  If I press `PREFIX` and hold `CTRL-UP`, the active pane is resized by one line, then the shell command history starts scrolling.

I suspect the 500 is short is because my initial key repeat is almost exactly 500 ms, however `CTRL-UP` for example is 6 characters, which seems to take almost 100ms to transmit and process.  I suspect if I were directly on a console, I would have a different experience.

systemmonkey42:
I've rebased on the latest build, and implemented your patches.

Seems to work perfectly.  I've reset the repeat-time/initial-repeat-time to 500/0 to see how that works.

nicm:
Great, I have applied this now - let me know if you think we should change the defaults. Thanks!

Issue #4117, opened 2024-09-10
[Feature proposal] Options to customise cursor behaviour in different modes


lexurco:
Hello!

At the moment cursor styling in tmux is not very consistent. I can set the `cursor-style` option, but it reverts to block in copy-mode regardless of what I set, in the command prompt the cursor is actually hidden, and is instead emulated with reverse-video, and in other modes it behaves in yet more mysterious ways that I'll have to investigate at some point.

I am not entirely sure how to approach the issue, or if there would be any interest in fixing it at all. I did, however, go ahead, and implement an option to switch between emulated and "hardware" cursor for the command prompt on my own branch for a start. The branch: https://github.com/lexurco/tmux/tree/emulate-cursor-option

Any comments would be appreciated. Thanks!

### Required information

tmux version: `tmux next-3.5`
Platform: `OpenBSD amd64`
`$TERM`: `tmux-256color`
Logs: N/A

lexurco:
Great! The diff does work for the copy mode. I'll return once I implement the remaining changes.

lexurco:
I am not exactly sure when I'll be able to continue working on that. So far I've added the `prompt-cursor-{colour,style}` options. Seems to work fine. So I'll see about the modes when I have some time next.

nicm:
I have applied my change (to OpenBSD, should be in GitHub later), if you make any further changes please open a PR. Thanks!

PR #4170, opened 2024-10-07
Add the prompt-cursor-{colour,style} options (issue #4117)

lexurco:
This implements the two options, as discussed in issue #4117:

- `prompt-cursor-colour` cursor colour in prompt, cf. `cursor-colour`
- `prompt-cursor-style` cursor style in prompt, cf. `cursor-style`

At the moment, upstream tmux hides "real" cursor when entering prompt, to instead emulate it. These options make the behaviour unnecessary, so that's been removed.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -207,6 +207,7 @@ const struct options_name_map options_other_names[] = {
 	{ "display-panes-active-color", "display-panes-active-colour" },
 	{ "clock-mode-color", "clock-mode-colour" },
 	{ "cursor-color", "cursor-colour" },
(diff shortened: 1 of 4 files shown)


nicm:
Thanks. Don't you also need to remove `cursorgc` and bit that draws the space as well?

lexurco:
Indeed.  Should be fixed now. Thanks!

nicm:
OK, I have applied to OpenBSD now with some trivial changes, will be in GitHub later.

Unfortunately a default cursor of black doesn't really work because the text is black and it doesn't reverse so the actual character under the cursor becomes invisible, so I made it cyan instead.

Thanks!

PR #4176, opened 2024-10-09
Better tab handling near the end of the line

lexurco:
Hello!

Currently horizontal tabs input when cursor is at the end of the line are ignored, and when cursor is on the penultimate cell, it just gets moved right once. The result is that in these cases tab does not create any spacing between characters.

This changes the behaviour to be more like it is in xterm:

- If the cursor it too close to the end of the line (last, or penultimate cell), tab instead moves the cursor starting from the beginning of the next line
- In case of a 1 column-wide screen, tab is represented by moving further down an additional line

I am not exactly sure if that might cause problems in terminal raw mode, though as far as I know, most applications requiring raw mode do their own tab handling.

Thanks.

### DIFF
--- a/input.c
+++ b/input.c
@@ -1230,9 +1230,16 @@ input_c0_dispatch(struct input_ctx *ictx)
 		screen_write_backspace(sctx);
 		break;
 	case '\011':	/* HT */
-		/* Don't tab beyond the end of the line. */
-		if (s->cx >= screen_size_x(s) - 1)
-			break;
+		/* Tab on the next line if too close to the end. */
+		if (screen_size_x(s) == 1 || s->cx >= screen_size_x(s) - 2) {
+			screen_write_linefeed(sctx, 1, ictx->cell.cell.bg);
+			screen_write_carriagereturn(sctx);
+			if (screen_size_x(s) == 1) {
+				screen_write_linefeed(sctx, 1, ictx->cell.cell.bg);
+				screen_write_carriagereturn(sctx);
+				break;
+			}
+		}
 
 		/* Find the next tab point, or use the last column if none. */
 		do {


nicm:
It does not seem to be how xterm works for me:
Screenshot 2024-10-09 at 15 22 32

nicm:
I don't know, anything in your .Xresources or .Xdefaults for xterm? What xterm version?

lexurco:
Nothing I can put my finger on.

`XTerm/OpenBSD(393)`

```
XTerm*termName: xterm-256color
XTerm*cursorUnderLine: true
XTerm*cursorBlink: never
XTerm*loginShell: true
XTerm*saveLines: 0
XTerm*scrollBar: false
XTerm*showBlinkAsBold: true
XTerm*pointerMode: 2
XTerm*VT100*translations: #override \n\
    Ctrl ~Shift <Key>H: string(0x7f) \n\
    Ctrl Shift <Key>H: string(0x08) \n\
    Ctrl Shift <Key>C: copy-selection(CLIPBOARD) \n\
    Ctrl Shift <Key>V: insert-selection(CLIPBOARD) \n\
    Ctrl Shift <Key>Y: insert-selection(PRIMARY)
XTerm*VT100*metaSendsEscape: true
XTerm*VT100*altSendsEscape: true
```

nicm:
I don't see any OpenBSD xterm changes could cause this either, but in any case this is how most terminals behave so I don't think we should change unless you can pinpoint an option or something in xterm that explains it.

ThomasDickey:
Perhaps this FAQ: [That description of wrapping is odd, say more?](https://invisible-island.net/xterm/xterm.faq.html#vt100_wrapping).

But as noted in the FAQ, some programs do less well than others.  See [this](https://github.com/mattiase/wraptest), which is representative but not exhaustive.

lexurco:
I investigated the issue a little further. On my system, when I run xterm "normally" (i.e. from the window manager), it behaves as I describe. However, if I run it from xterm, it starts behaving like tmux, other terminal emulators, and it seems from above, actual VTs, i.e. by just ignoring tabs on the last column. Upon a quick glance, the output of `stty -a` is different, though I have not investigated it in detail:

- xtrem-from-wm:
```
speed 38400 baud; 24 rows; 80 columns;
lflags: icanon isig iexten echo echoe echok echoke -echonl echoctl
	-echoprt -altwerase -noflsh -tostop -flusho pendin -nokerninfo
	-extproc -xcase
iflags: -istrip icrnl -inlcr -igncr -iuclc ixon -ixoff -ixany -imaxbel
	-ignbrk -brkint -inpck -ignpar -parmrk
oflags: opost onlcr -ocrnl -onocr -onlret -olcuc oxtabs -onoeot
cflags: cread cs8 parenb -parodd hupcl -clocal -cstopb -crtscts -mdmbuf
cchars: discard = ^O; dsusp = ^Y; eof = ^D; eol = <undef>; eol2 = ^@;
	erase = ^?; intr = ^C; kill = ^U; lnext = ^V; min = 1; quit = ^\;
	reprint = ^R; start = ^Q; status = <undef>; stop = ^S; susp = ^Z;
	time = 0; werase = ^W;
```

- xterm-from-tmux:
```
speed 38400 baud; 24 rows; 80 columns;
lflags: icanon isig iexten echo echoe echok echoke -echonl echoctl
	-echoprt -altwerase -noflsh -tostop -flusho pendin -nokerninfo
	-extproc -xcase
iflags: -istrip icrnl -inlcr -igncr -iuclc ixon -ixoff ixany imaxbel
	-ignbrk brkint -inpck -ignpar -parmrk
oflags: opost onlcr -ocrnl -onocr -onlret -olcuc -oxtabs -onoeot
cflags: cread cs8 -parenb -parodd hupcl -clocal -cstopb -crtscts -mdmbuf
cchars: discard = ^O; dsusp = ^Y; eof = ^D; eol = <undef>; eol2 = ^@;
	erase = ^?; intr = ^C; kill = ^U; lnext = ^V; min = 1; quit = ^\;
	reprint = ^R; start = ^Q; status = <undef>; stop = ^S; susp = ^Z;
	time = 0; werase = ^W;
```

From the above links, it appears that the current behaviour is correct, and I should instead investigate what's going on with my xterm. Nevertheless, it is noted to be problematic, inc. in the wraptest repo linked above:

> The fact that TAB is an LCF-resetting operation can have negative consequences. Consider a text file containing a line longer than the width of the terminal. If the file is sent unmodified to the terminal, the long line will wrap, but if it contains an ASCII TAB character at an unfortunate position, the result can be the visual disappearance of a printing character.

So, perhaps there should be an option to choose the desired behaviour?

nicm:
> oxtabs

The tty layer is converting the tabs into spaces, see termios(5).

nicm:
You will see the same behaviour in tmux if you do `stty oxtabs`, regardless of the terminal tmux is running in.

xterm will inherit the termios attributes from the tty where it is run, but if it can't it looks like the default might include `oxtabs`. It has a `ttyModes` resource to set the defaults.

lexurco:
Yes, it appears to be that. So, there's no need for that change. Thanks.

PR #4178, opened 2024-10-10
Added dockerfile and script for building/running in container.

ubinator:
Below is the commit message of the single commit that this pull request consists of. Maybe I overdid it and wrote too much. Sorry. This pull request may be a little bit rough around the edges, but it seems to work nicely. I would like to know whether this feature is useful or not. If it is useful, what should I improve upon before it can be merged? I also find it unsatisfying that it's so detached from the build system, but I'm no autotools expert and I'm not exactly sure what the best way to integrate it more into the build system would be, so advice about that would be appreciated as well.

If you already have tmux installed on your system, testing your changes can be a little annoying. This fixes that. It also is a fixed environment which is the same for everyone, which can be useful as well.

IMPORTANT NOTE: Make sure to clean repo between switching from building/using tmux in container and on host or vice versa. Build products stay around and aren't recognized as belonging to different target. SO YOU NEED TO CLEAN BUILD WHEN SWITCHING!!!!

You can do:

sudo docker build -t some-individual-name .
sudo docker run -it --rm -v .:/tmux some-individual-name bash -c './autogen.sh && ./configure && make'

OR you can use the small helper script like so:

sudo ./run-container.sh interactive --> build for container, start bash
sudo ./run-container.sh build --> build for container
sudo ./run-container.sh cmd 'command' --> execute arbitrary cmd in container

You can use the cmd subcommand like so for example:

sudo ./run-container.sh cmd 'make -C regress' --> run regression tests

NOTE: I know you can test tmux on host systems with tmux already installed if you just change the unix socket that the client uses, but maybe this docker container commit will be useful to someone. If it isn't, just scrap it.

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -1,0 +1,26 @@
+FROM debian:stable
+
+RUN apt update && apt install -y \
+gcc \
+bison \
+autoconf \
+pkgconf \
+make \
+libevent-dev \
+libncurses-dev
+
+RUN useradd app
+
+RUN mkdir /tmux-bin
+RUN chown app:app /tmux-bin
+
+RUN mkdir /tmux
+RUN chown app:app /tmux
+
+WORKDIR /tmux
+
+USER app
+
+ENV PATH="$PATH:/tmux-bin"
+
+CMD [ "bash" ]
--- a/run-container.sh
+++ b/run-container.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -o pipefail
+
+function output_help() {
+	echo &&
+	echo "'build': build repo for container, build products persistently located in host repo" 1>&2 &&
+	echo &&
+	echo "'interactive' (default): automatically calls 'build', afterwards open interactive bash to the container. From there, you can use the built tmux executable from within the container. It's automatically put in the PATH, so you can simply write 'tmux ...' to use the built executable." 1>&2
+	echo &&
+	echo "'cmd': run the argument that follows as a command in bash from within the container. Automatically builds tmux for the container, just like 'interactive'."
+}
+
+function assert_arg_count() {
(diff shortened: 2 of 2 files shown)


nicm:
Thanks but we do not maintain scripts for different package managers in the tmux repository. Normally I would suggest you make your own GitHub repo and maintain it there, but I see on the [installing page](https://github.com/tmux/tmux/wiki/Installing) someone has already made a Docker package [here](https://github.com/ferryman0608/Dockerfile-tmux), so unless yours is substantially different I don't think we need to link to another.

ubinator:
I would argue that mine is a little bit different. The one you linked is just to quickly get your hands on a tmux binary. Mine mounts the repo as a volume in the container for development. I'm afraid I don't know exactly how package managers relate to my commit. My original thought was that the container could give you a nice dev environment if your environment is annoying to work with, like if you have tmux already or something.

With my commit, you can change some source files, and then simply type 'sudo ./run-container.sh', and then it builds into a container and you can test your changes. It's very simple and I figured maybe it added some value to the project because reproducibility and simplicity.

But if it's not worth it's weight then fair enough, that makes sense.

nicm:
OK I thought it was a package management system rather than another way to build tmux, but all the same I don't think there is much call for scripts like this in tmux itself. If you want to maintain it in your own repo we can link to it from the installing page as well, or maybe you could submit your improvements as a PR on the existing scripts.

ubinator:
I guess I can keep it in my fork and you can link to it if you want to. With other scripts do you mean your autotools stuff? I was thinking the same thing, but I have to get more familiar with how autotools does things first and then I'll think about it.

nicm:
No I mean the existing Docker build scripts I linked to.

PR #4184, opened 2024-10-12
Optimize sixel_print

bptato:
Previously, the algorithm scanned each pixel several times; once to find out which colors are active, and then once for every single active color to actually construct the output string.

Now it constructs the compressed sixel patterns in the first pass (now x * 12 iters), so we can reduce the second pass (the really expensive part, at active colors * x * 6 iters) to just appending these to the output buffer.

On my test images, printing time is reduced roughly tenfold (e.g. 80ms -> 8ms, though this also depends on the average number of colors active per line). That's significant enough that scrolling around images in chawan over tmux becomes tolerable again. (Actually, with this patch, tmux beats my own sixel printing routine :p)

### DIFF
--- a/image-sixel.c
+++ b/image-sixel.c
@@ -52,6 +52,19 @@ struct sixel_image {
 	struct sixel_line	*lines;
 };
 
+struct sixel_chunk {
+	u_int			 next_x;
+	u_int			 next_y;
+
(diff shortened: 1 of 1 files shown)


nicm:
Looks good and works well for me, applied now, thanks!

PR #4190, opened 2024-10-17
tty: Pass unrequested OSC 52 sequence to client

holio0:
Problem:

An OSC 52 escape sequence received outside a designated TTY_OSC52QUERY is not passed on to the client application. For example, a program running inside the tmux sends the OSC 52 escape sequence to the host terminal. The host terminal writes data content to tmux to response the request, but tmux ignores the data in tty_keys_clipboard because the TTY_OSC52QUERY wasn't set, so the client program never receives the data.

Solution:

Add a tty_keys_clipboard return code `-2` as a "leave it to the foreground program" indicator.

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -747,6 +747,7 @@ tty_keys_next(struct tty *tty)
 	switch (tty_keys_clipboard(tty, buf, len, &size)) {
 	case 0:		/* yes */
 		key = KEYC_UNKNOWN;
+	case -2:	/* yes, but leave it to the foreground program. */
 		goto complete_key;
 	case -1:	/* no, or not valid */
 		break;
@@ -1340,7 +1341,7 @@ tty_keys_clipboard(struct tty *tty, const char *buf, size_t len, size_t *size)
 
 	/* If we did not request this, ignore it. */
 	if (~tty->flags & TTY_OSC52QUERY)
-		return (0);
+		return (-2);
 	tty->flags &= ~TTY_OSC52QUERY;
 	evtimer_del(&tty->clipboard_timer);
 


nicm:
No, this is not correct, the program is querying tmux's clipboard, not the outside terminal's.

holio0:
Or is there a chance to add a new option, or maybe use an existing option like `allow-passthrough` or `set-clipboard`, for the client program to specify the desired target?

nicm:
Possibly... I think I would like to see an implementation where this escape sequence could trigger tmux to request from the terminal itself... but it might be tricky... for a start which client does it send it to?

Maybe we should allow it to pass through but the general principle is that tmux tries not to send input to panes that it doesn't know will be understood. An option is maybe not a bad idea, but what about other escape sequences? What if tmux later does want to process this itself, the option would no longer work?

holio0:
> which client does it send it to?

My use case is that I am running neovim inside the tmux. The neovim supports accessing system clipboard by sending the OSC 52 sequence all the way to the terminal emulator, so the idea of this change is allowing tmux to route the OSC 52 sequence it doesn't want (== TTY_OSC52QUERY is not set) to pane. 

And for these sequences like OSC 52 that designed for terminal emulator to respond client requests, I think it makes sense for tmux to just pass the "unwanted" escape sequence inputs through, without interpreting them. This is based on the assumption that as long as tmux doesn't want an escape sequence, these are a reply from terminal emulator to the program sends the request earlier, so it's reasonable for tmux to assume the pane knows what it's doing.

> What if tmux later does want to process this itself, the option would no longer work?

It works for me that in this case the option would no longer work as long as the sequence is "wanted" to tmux, which means it's a reply to request tmux sends earlier, and routing these to panes probably is not useful.

holio0:
I added an option in the latest revision. The new option allows users to choose whether to forward or discard unrequested escape sequences. Does this meet your expectations? Please let me know.

nicm:
Please try the change in this comment https://github.com/tmux/tmux/issues/4275#issuecomment-3467124509 instead. Thanks!

PR #4191, opened 2024-10-17
Fix options ordering in tmux.1

tgyurci:
Switch `pane-colors[]` and `cursor-style` to be options in alphabetical order.


nicm:

PR #4192, opened 2024-10-17
Use singular phrase for list-buffers and list-panes

tgyurci:


nicm:
This came up before and I don't think it is necessary, it is just making it more complicated for no real reason.

tgyurci:
The real reason is for me to be grammatically correct, but I accept your opinion.

nicm:
Yes I know, but I don't think it matters here enough to make the code more complicated, thanks anyway.

PR #4194, opened 2024-10-17
Respect the attribute for menu style

japinli:
See: https://groups.google.com/g/tmux-users/c/7o7vVOqg078


nicm:
Sorry, I hadn't forgotten about this... I was going to look and see if it made sense to set up a format so that the dim for items with a leading `-` could be expressed in the style rather than the code (so the default would become something like `#{?#{m|#{menu_item}|-*,dim,default}`). This would mean applying the style separately for each item though.

Anyway in the meantime I have applied this, it should be in GitHub later when it syncs up. Thanks!

PR #4195, opened 2024-10-18
Instructions to bypass alias when testing tmux's default config

Arnaudv6:


nicm:
Thanks but I think it is self evident you should run the tmux you want to test and not something we need to document.

PR #4201, opened 2024-10-20
Preserve tabs for use in copy mode

lexurco:
This changes the handling of tab characters to save tabs as extended cells instead of just moving the cursor. As a result, tabs can now be copied in copy mode. I didn't test it extensively, however it seems to work correctly in most cases.

There are still a few issues that I'd like to address separately, however:

- The `search-*` and `jump-*` commands don't accept tabs. For this `command-prompt` will have to be modified. Perhaps `command-prompt -1` could simply treat control characters as normal visible characters, and without `-1`, `command-prompt` could accept ^V^X sequences for inputting control characters.
- It would be nice if the cursor would automatically snap to the beginning/end of a tab when moving up/down. That, however, requires extensive modification of `window-copy.c`.
- A tab may be currently lost when the cursor is already at the end of the line. This is how terminal emulators behave normally, but perhaps there should be a `tab-wrap` option to wrap tabs like in a text editor.
- It'd be nice if "short tabs", that reached the edge of the screen before the next tab stop, were to be redrawn on screen resize.


lexurco:
That should fix `capture-pane`. As far as I can tell, `grid_string_cells` is only used for `capture-pane` and `comand -F '#{C:string}'`, which is also fixed now.

I am not exactly sure how to test `capturep -P`, or if that should be altered.

Not sure what you mean about debugging. I don't see anything I forgot to remove in `git diff master @`.

nicm:
Thanks. I must have been looking at the .patch earlier which includes every commit so it had your debugging.

Can you move `w` and `maxsz` to the head of the function with the other declarations in `grid_get_cell1` and change `if (!(gc.data.size == 1 && *gc.data.data == ' '))` to `if (gc.data.size != 1 || *gc.data.data != ' ')` in `input_c0_dispatch`?

Otherwise it looks OK but I will try it again later.

lexurco:
OK, done.

lexurco:
Sorry, forgot to add input.c to the last commit. Both changes are in now.

lexurco:
I discovered that this branch actually introduces a bug to the `jump-*` commands (i.e. if you do manage to jump to a tab, it doesn't actually do anything). I'll let you know when it's fixed.

lexurco:
Fixed.

nicm:
Even if it did have to be extended, you shouldn't need to set `GRID_FLAG_EXTENDED` in `input.c` - `grid_set_cell` will work that out with `grid_need_extended_cell`.

lexurco:
Thank you. Good noticing about setting the flag in `input.c` -- must be something I forgot to remove after testing. Allowing tabs to be non-extended cells makes the code a bit trickier, as we now have to account for 2 cases. Both done now.

nicm:
Sorry I think maybe it was better always as an extended cell :-/. I think it is better to reduce the special cases by just treating it as a cell of width N with N spaces.

But doesn't `input.c` also need to fill with spaces? The cell you are setting the bit on is only guaranteed to be one space.

I think you also need a check in input.c to ignore tabs which are > sizeof gc.data.data and they should be ignored.

lexurco:
OK, I reverted to treating all tabs as extended cells. :^)

I also introduced a check in `input.c` to see if the width exceeds the size of data. In this case, the fallback behaviour of just moving the cursor N cells to the right is used.

I don't think `input.c` needs to fill with spaces. There are basically two scenarios:

1. All cells up to the next tab stop are already empty, i.e. spaces. In this case a cell with GRID_FLAG_TAB flag is written.
2. There is at least one non-empty cell, in which case we fall back to the old behaviour of just moving the cursor.

We check for it in the `do { ... } while(...);` loop:

```
  1240			/* Find the next tab point, or use the last column if none. */
  1241			do {
  1242				grid_get_cell(s->grid, cx, line, &gc);
  1243				if (gc.data.size != 1 || *gc.data.data != ' ')
  1244					has_content = 1;
  1245				cx++;
  1246				if (bit_test(s->tabs, cx))
  1247					break;
  1248			} while (cx < screen_size_x(s) - 1);
```

So, writing spaces in case 1. is unnecessary, and in case 2. would diverge from the normal terminal behaviour. For example, when you write `printf '12345678\r\t9\n'` you would get `        9` instead of `123456789`, etc.

lexurco:
Actually now that I think of it, there are also weird cases of background, and other attributes being different between these cells. E.g., `printf '    '; tput rev; printf '    '; tput sgr0; printf '\r\tX\n'` should give you 4 normal spaces, followed by 4 reverse-video spaces, followed by a normal X.

Probably in cases of mismatching attributes we should also just move the cursor. Otherwise, we might as well write all the spaces explicitly, but it shouldn't make any real difference.

nicm:
If you don't put spaces in the cell you need to special case tab in the grid code but if it already had the spaces it will be treated like any other wide character. You could put the spaces in and then have screen-write.c skip drawing it, although that is swapping one special case for another I suppose.

> Actually now that I think of it, there are also weird cases of background

If you want to do it this way with the tab as a single wide cell (which is a perfectly reasonable way to do it) then you will have to compromise and say that basically anything weird will ignore the tab. The other way would be to make tab purely a flag, but the disadvantage is that it would be hard to do tab overwriting because you wouldn't know when you need to look for a tab.

nicm:
I'm not wild about tab cells sometimes being `\t` and sometimes spaces, I think it would be clearer to write the tab cell into the grid as it is meant to come out - so width = size = N, data = N * space, flags = TAB - then let the `screen-write.c`/`grid.c` code worry about how to store it in the same way as it does for any wide characters (although you will still need to check the tab flag in `grid_need_extended_cell` for the N = 1 case). This has the advantage that any code that doesn't check the TAB flag will still handle it just as N spaces like it (should) handle any other wide character.

I'd make the `screen_write_cell` do any optimization to omit writing it to the tty (it will still need to write the cell and any padding cells into the grid) but it can probably just skip writing out any tab cells entirely.

lexurco:
I took a look at how urxvt handles the "weird" case, and it does exactly that: just moves the cursor without storing a tab.

The above two commits deal with that case, and also set the cell to correct size and data in `input.c`.

nicm:
OK this looks great. A few of minor things:

- I have applied your `grid-reader.c` helper function change now without the tab bits since it is an improvement so may as well get it out of the way.

- I think the comparison in `input.c` could use `grid_cells_look_equal`?

- How about adding a helper to set up the tab like this [x.diff.txt](https://github.com/user-attachments/files/17518283/x.diff.txt) instead of doing it inline in two places? A bit tidier? 

- I think something like this will stop it writing out the tab? [y.diff.txt](https://github.com/user-attachments/files/17518334/y.diff.txt) Does that seem right?

nicm:
> About '\t' vs N * ' '. Do you mean some of the window-copy.c functions checking for GRID_FLAG_TAB to decide what string to return?

I thought it was `\t` in `input.c` but actually it wasn't, never mind. I was actually hoping the tab could be stored as a UTF-8 character but I forgot that the representation only allows two bits of width, so that won't work. But I think a check on decoding it from/to extended is fine.

It seems about three times as slow to store tabs (try `tmux detach; tmux clear-history; clear; time seq -f "$(printf '\t%%g\n')" 200000` then wait for a few seconds before attaching again). Part of this probably it being an extended cell and part having to fill in padding. I don't really know if that is a problem... how many tabs can there be?

lexurco:
For me there is only about 2 times difference, and it stays almost the same if I set `gc.data.size`, and all bytes in `gc.data.data` to 0 (which prevents it from being printed). So, the slow-down is probably due to using extended cells.

lexurco:
One thing we could do if this is a problem, is to use a non-extended cell with a single space for tab, but with `GRID_FLAG_TAB` set, and then set `GRID_FLAG_PADDING` for the cells to the right until the next tabstop (also containing single space). That would still require rewriting some of the logic, but I think it might be manageable.

nicm:
I think it would be worth trying to make it faster, but shall we get this version in first (if you make the changes in my last but one comment) and then do you want to make a new MR for the performance?

lexurco:
Sure, though before I start working on a faster version, I'd like to also make sure that adjacent things like search and formats also work properly.

Regarding your proposed changes: I'll look at the first 3, but I don't think we should use y.diff.txt because then tabs are not highlighted when selected in copy mode.

nicm:
OK  I can't look again now but what if you change it to:
~~~~
+	if (!selected && gc->flags & GRID_FLAG_TAB)
+		skip = 1;
~~~~
Although it might be better a bit higher up before the `skip = 0;`.

lexurco:
It still doesn't work with marks and search, however. And that information isn't even stored in `struct screen`. And search is one of the things I'd like to work on straight after this. I think it's OK to just write tabs out for now, but later we might be able not to.

nicm:
OK that's fine just leave it.

lexurco:
I merged master, and implemented the other two points, except that I made `grid_set_tab()` call `memset(&gc->data, 0, sizeof gc->data)` first in pattern with `utf8_to_data`.

nicm:
OK I have applied this now to OpenBSD with some trivial style changes, and I also made a separate change to fix an existing bug (`break` should be `continue`) in `format_get_line` which was stopping wide characters being copied and added tabs to the same function while I was there. Let me know if you see any problems with this, or have any other fixes (I would like to see if it can be made faster if you have time to look at that). Thanks!

lexurco:
This should fix searching for tabs.

lexurco:
And this should hopefully be the last formats fix.

lexurco:
I investigated the slowdown further, and it seems like almost all of the delay comes from padding cells being extended, as per condition in `grid_need_extended_cell()`: `if (gc->data.size != 1 || gc->data.width != 1)`. I don't entirely understand, however, why padding cells need to be stored as extended. Coludn't this statement be replaced with `if (gc->data.size > 1 || gc->data.width > 1)`?

Incidentally, it seems to confirm my suspicion that much better performance could be achieved by storing tabs as a series of single-space cells with either `GRID_FLAG_TAB` or `GRID_FLAG_PADDING` set.

nicm:
I have applied both these changes. I don't see any reason that padding has to be an extended cell if you want to change it, the only thing that is useful in it is the padding flag.

lexurco:
This should hopefully fix the remaining issues with search.

I am still seeing some edge-cases where tabs get converted to cursor movement, so I'll work on that next.

Thanks!

lexurco:
And that should fix some tabs getting converted to spaces.

nicm:
This bit doesn't look right: in `window_copy_search_marks` the line `px += tmp_width;` is outside the loop, but if `window_copy_search_mark_at` returns nonzero it will either be initialized or left over from the previous loop.

lexurco:
Yes. I also noticed that `padding` was being initialised twice (at the beginning of the outer loop, and of the function). Both fixed now.

Unfortunately, after some more testing I still seem to be getting some weird glitches that I don't know how to reproduce yet exactly. Will work on that.

lexurco:
Fixed that problem, and also fixed a bug in grid_set_tab(), which I should have noticed earlier.

nicm:
I have applied this now, I changed `tmp_width` to `tw` to fix some long lines and some other similar changes, although I'm not wild about that... maybe a helper function for the loop body would be a better move but I can't quickly see how to obviously break it out. Anyway, let me know if I broke anything. Thanks!

lexurco:
> maybe a helper function for the loop body would be a better move

Something like this?
[x.diff.txt](https://github.com/user-attachments/files/17686451/x.diff.txt)

nicm:
Perfect, applied now, thanks!

PR #4203, opened 2024-10-21
Add new function window_copy_get_current_offset

mgrant0:
Add new function to get the current offset of the position (in lines) and size (also in lines) of the pane in copy mode.


nicm:
I've applied this with some trivial changes, thanks.

PR #4206, opened 2024-10-23
Accept control characters in command-prompt

lexurco:
Hello!

As I mentioned in PR #4201, limitations of the command prompt make the new tab feature somewhat hard to use. This solves the issue:

- `command-prompt -1` now accepts input without processing it first.
- Otherwise, control characters can now be input with Ctrl-V Ctrl-<Key> (and Ctrl-V Tab, Ctrl-V Esc, etc.).

Thanks.

### DIFF
--- a/status.c
+++ b/status.c
@@ -1262,6 +1262,12 @@ status_prompt_key(struct client *c, key_code key)
 	}
 	key &= ~KEYC_MASK_FLAGS;
 
+	if (c->prompt_flags & PROMPT_SINGLE) {
+		if (key & KEYC_CTRL)
+			key = (key & ~KEYC_CTRL) & 0x1f; 
+		goto append_key;
+	}
+
 	keys = options_get_number(c->session->options, "status-keys");
 	if (keys == MODEKEY_VI) {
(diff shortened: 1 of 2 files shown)


nicm:
Looks good except a couple of things:

- `quotenext` should not be a `static` variable - a bit in `c->prompt_flags` would probably work instead.

- `key &= (key & KEYC_CTRL) ? 0x1f : 0xff;`- I think this will make a mess of Unicode keys.

lexurco:
Thanks. I changed it to use `c->prompt_flags` instead of a static variable, and improved handling of keys >0x7e.

nicm:
This looks good but the cursor position is wrong if I type control characters at the command prompt, it seems like it doesn't realise they should be width 2.

I'm also a bit concerned `^@` might cause trouble but I guess maybe we can just see how that goes.

nicm:
OK I changed it to use `utf8_data` years ago but I guess the cursor position was broken either then or at some point since.

I've applied now with some minor changes (use some temporaries to tidy up long lines), thanks. It works for me in that I can do for example `setb` with some tabs, but searching doesn't seem to work.

PR #4209, opened 2024-10-24
Fix infinite loop in copy mode search-* commands

lexurco:
Hello! It is currently possible to enter an infinite loop by attempting to search for a control character in copy mode. Can be tested with `tmux bind -T copy-mode / "send -X search-forward -- '	'"` (that's a tab inside single quotes). This fixes the issue by immediately terminating the search for "0 bytes long" strings. Thanks.


nicm:
Applied now, thanks!

PR #4212, opened 2024-10-26
Command prompt fixes

lexurco:
This fixes the following issues in command prompt:

- The `PROMPT_QUOTENEXT` flag is unset after quoting, not before, meaning that accidentally scrolling the mouse wheel with `set mouse on` and similar don't break quoting.
- The mid-quoted state is indicated by writing a '^' character.
- Cursor now correctly moves over wide characters.


nicm:

PR #4221, opened 2024-10-30
Scrollbars bugfix large window size

mgrant0:
Fixes:

> There is a bug - oversize windows are not handled correctly. Try this (increase the 200 if your terminal is more than 200 columns) and you will see the right pane scrollbar is always visible even when it should be hidden, and if you split and move around the panes you will see it is not redrawn in the right position.
> 
> tmux new \; set -g pane-scrollbars on \; splitw -dh
> tmux resizew -x200


nicm:
Why do `i` and `j` need to be `int`, they can't be negative?

I think you could maybe get rid of a lot of the casts by making `pad_col` and `sb_w` `int` and adding `int` temporaries for some of the others like:
~~~~
int xoff = wp->xoff, yoff = wp->yoff, ox = ctx->ox, oy = ctx->oy;
~~~~
The compiler will not warn about implicit casts like this.

mgrant0:
I did try this but it ended up requiring even more casting.  Let me have a go at cleaning that up.  It's not that `i` or `j` can be negative, it's when they are used with `px` and `py`.  Yes, I agree, let me clean that up, this is the easy bit now.  At least the logic seems to work.  Still testing this myself, let me know if you find some edge case.

mgrant0:
This is cleaned up now.  I think you'll like it (I like it much better now).  I'm off to work on the other issues.  I gather you would prefer a separate PR for each of those against this branch.

nicm:
It's up to you, separate PRs is good if it is easy but if you want to roll them together that's fine.

mgrant0:
> You redraw the scrollbars in cmd-select-pane.c. I'm not saying this is wrong, but why is is necessary? What could happen when the active pane changes that needs the scrollbars to be redrawn?

It seems like it can be deleted.  

Previously the scrollbars were being forced to be redrawn in `screen_redraw_draw_borders` or `screen_redraw_draw_border_cell` and I saw a call to `server_redraw_window_borders` here so it was an attempt to figure out where I needed to add that redraw flag and I clearly missed removing this here after the dust settled.

It's possible there are some other places this happend as well.  I will probably need to remove each of those and see what happens.

mgrant0:
I had a look at `sl.diff.txt`.  I see that this allows one to configure the fg and bg separately for the slider vs the "shaft".  But is that useful if the slider is a space character?  Are you thinking at some point another character other than a space might be used?  Adding that variable check there means it'll go through that code each time the scrollbar is redrawn.

By the way, I originally thought I would be using some unicode box characters for the scrollbar slider but it dawned on me a simple inverse space was sufficient and way simpler.  When the scrollbar width changes, this would have meant more complex box drawing. 

I think adding a variable to the format is overly complicated for someone who might want to change the setting.  First they need to figure out how conditional variables work, then, they might go searching for is_slider in the docs.  When I first saw it, I wasn't sure if 'is_slider' meant that the scrollbar was visible or not.  I'm already pretty familiar with this and I had to go search for it and look at the code.

If you still think a separate fg and bg for the slider is useful (which it might be), then, what about a second optional `pane-scrollbars-slider-style` option which, if present, affected the slider, otherwise the slider took on the reverse fg/bg of `pane-scrollbars-style`?

It just seems like that would be more straight forward from a user perspective than using `is_sider` as a variable.  Getting it wrong, deleting a space, forgetting or not understanding the comma separating the true from the false, messing up the brackets... it leaves a lot of room for the user who might want to change that and guessing how to fix it.  I know, I've been there!

mgrant0:
> In server-client.c, c->redraw_panes is reused for both PANE_REDRAW and PANE_REDRAWSCROLLBAR. This makes it impossible to redraw the scrollbars without redrawing the entire pane, which means that for example window_copy_scroll_up will end up redrawing the entire pane as well as the lines which have scrolled. I think this probably needs to be split into two bitmasks.

Yeah I wanted to mention this to you too, glad you brought it up.  I had this split in an earlier version but didn't understand the client back then.  I thought this fixed that but you are right, it looks like I missed the reverse case that the scrollbars end up forcing a screen redraw.

If I create a client flag `CLIENT_REDRAWSCROLLBARS`, then redraw all of the scrollbars in all of the clients, would this solve it?  I thought of this but it seemed like it was going to redraw other scrollbars that didn't need redrawing.  

Once I set `CLIENT_REDRAWSCROLLBARS`, then `PANE_REDRAWSCROLLBAR` gets unset and I no longer know which scrollbar needs attention.  I clear the `PANE_REDRAWSCROLLBAR` flag in `server_client_loop` since we expect the scrollbars to be updated as part of the client.  Perhaps I shouldn't do that.

I have an idea...  I have a feeling I may have already tried this but worth trying again.  Instead of clearing the `PANE_REDRAWSCROLLBAR` flag, leave it set.  Then, in `screen_redraw_screen`, instead of checking `CLIENT_REDRAWWINDOW`, I would loop over each wp looking to see if the scrollbar needs updating.   I remember thinking it seems wasteful to loop over every wp inside `screen_redraw_screen` when I'm inside a client, it seems inside out.  Like once I know a specific wp's scrollbar needs to be redrawn, every client should update that wp's scrollbar rather than having to loop over each wp to find it.  Am I overthinking it?

It felt wrong to call `screen_redraw_draw_pane_scrollbars` from `server_client_check_redraw`.  I could check `PANE_REDRAWSCROLLBAR` in `screen_redraw_draw_pane_scrollbars` and unset it after.  But this function really is meant to redraw the scrollbars regardless, it would then need a force flag and be called if the whole pane is redrawn.  

Do you see my conundrum?  What's the right way to fix this?  Or should I just loop over all the wp's in `server_client_loop` even though it's not the most efficient way?

nicm:
> I had a look at sl.diff.txt. I see that this allows one to configure the fg and bg separately for the slider vs the "shaft". But is that useful if the slider is a space character?

The foreground doesn't matter, the point is to allow the background colour of the slider, at the moment it is always just the reverse of the scrollbar itself.

nicm:
> If you still think a separate fg and bg for the slider is useful (which it might be), then, what about a second optional pane-scrollbars-slider-style option which, if present, affected the slider, otherwise the slider took on the reverse fg/bg of pane-scrollbars-style?

No, no new options if they can be trivially configured as part of an existing option.

nicm:
> If I create a client flag CLIENT_REDRAWSCROLLBARS, then redraw all of the scrollbars in all of the clients, would this solve it? I thought of this but it seemed like it was going to redraw other scrollbars that didn't need redrawing.

It should not redraw all the scrollbars, it should work like `CLIENT_REDRAWPANES`: if it is unset, redraw no panes, if it is set, redraw only panes with `PANE_REDRAW` set.

> I have an idea... I have a feeling I may have already tried this but worth trying again. Instead of clearing the PANE_REDRAWSCROLLBAR flag, leave it set. Then, in screen_redraw_screen, instead of checking CLIENT_REDRAWWINDOW, I would loop over each wp looking to see if the scrollbar needs updating. 

Yes. It already loops over all the panes to check if they need to be redrawn, you need to also check if they need scrollbar redrawn. I don't mind if `PANE_REDRAW` implies `PANE_REDRAWSCROLLBAR` as well (that is, redrawing content always redraws scrollbar), but there needs to be a way to only redraw the scrollbar and not the pane content.

> It felt wrong to call screen_redraw_draw_pane_scrollbars from server_client_check_redraw.

That is fine, it calls `screen_redraw_pane` already.

> I could check PANE_REDRAWSCROLLBAR in screen_redraw_draw_pane_scrollbars and unset it after. But this function really is meant to redraw the scrollbars regardless, it would then need a force flag and be called if the whole pane is redrawn.

No, don't do that, `PANE_REDRAWSCROLLBAR` should be handled in `server-client.c` like the other flags. `screen_redraw_pane` should not really be checking it either.

nicm:
>The foreground doesn't matter, the point is to allow the background colour of the slider, at the moment it is always just the reverse of the scrollbar itself.

A better option may be to use `fg` for the slider and `bg` for the rest of the scrollbar, since at the moment neither use `fg`. That closes off the possibility of drawing scrollbars with anything except spaces (without also changing the meaning of the option at least) but is that realistically ever going to be needed?

nicm:
(Don't worry about the commit messages for this branch, that will all be rewritten because I will apply the total diff to OpenBSD in one commit.)

nicm:
This does the fg=slider/bg=rest split: [x.diff.txt](https://github.com/user-attachments/files/17588325/x.diff.txt) I don't really mind which way we go here, it seems sensible to reuse `fg` for the slider if we are not going to need it for something else later.

mgrant0:
It's funny the way you reformat the declarations.  Doesn't bother me, however you like it.  I kind of try to do it closer to how things are used to make it easier to see what's going on, but I know this isn't the BSD style.  Personally I also like to declare variables close to where they are used or if possible inside {...} causes.  But I'm trying to match your style, just don't always suceed!

On the x.diff, yeah let's do the second try using fg for the slider and bg for the shaft or whever it's called.  Do I need to first apply the first x.diff.txt and then the second one on top of it?

nicm:
I'll just apply it here.

nicm:
Applied the diff to the scrollbars branch now.

mgrant0:
I did those other 2 changes now.  Going to test more over weekend.  Not yet sure if I'm only redrawing the scrollbars when specifically required but will be testing that.  Let me know if this is more what you were looking for.

nicm:
This is close but not quite right - the bits with `redraw_scrollbars` look right, but when it comes to it, it still calls `screen_redraw_pane`:
~~~~
			else if (c->flags & CLIENT_REDRAWSCROLLBARS)
				redraw = !!(c->redraw_scrollbars & (1 << bit));
			bit++;
			if (!redraw)
				continue;
			log_debug("%s: redrawing pane %%%u", __func__, wp->id);
			screen_redraw_pane(c, wp);
~~~~
So it will still redraw the entire pane, instead of just the scrollbar.

You need to add `screen_redraw_scrollbar` and call that instead if the scrollbar is being redrawn (and remove the scrollbar bits from `screen_redraw_pane` - if you want `PANE_REDRAW` to imply `PANE_REDRAWSCROLLBAR` then do it in `screen-redraw.c`).

mgrant0:
I did this by adding a `redraw_scrollbar_only` flag to `screen_redraw_pane`.  Would you prefer I add a separate `screen_redraw_scrollbar` function only to redraw the scrollbar?  Up to you, if you feel that's cleaner, I willl change that.  It just started to feel like there were already a lot of redraw scrollbar functions in `screen-redraw.c` already.

You go to quite a lot of trouble to create this bitmask then loop over the bits to redraw the panes versus the whole screen.  I'm trying to understand why you did that with a bitmask.  Why not just call `screen_redraw_pane` when looping through the `TAILQ_FOREACH(wp, &w->panes, entry)` again?  Is it that much more efficient to create a bitmask?  

All that aside, it seems to me the scrollbars are part of or perhaps "child" of the pane, not like a "peer" of the pane if you know what I mean.  In my earlier thinking, I was trying to get the pane to deal with its scrollbar like it deals with its pane status line.  But by creating this extra bitmask and treating the scrollbar nearly separately from the pane, it feels like we've elevated it to a separate thing.  Is that bad?

mgrant0:
I have prepared a commit which manages the alternate screen. I can push it here or separetly, just let me know.  It's pretty straight forward.

nicm:
This looks good to me now.

> You go to quite a lot of trouble to create this bitmask then loop over the bits to redraw the panes versus the whole screen. I'm trying to understand why you did that with a bitmask. Why not just call screen_redraw_pane when looping through the TAILQ_FOREACH(wp, &w->panes, entry) again? Is it that much more efficient to create a bitmask?

The point is to keep track of any panes that need to be redrawn from a previous visit to this function, when redraw was deferred because the terminal has not consumed all the previous data. There is no point in redrawing the same pane multiple times if the terminal has not finished with the previous one.

nicm:
I've put this in the branch now, I'll have another look at it later and then it can probably go in to OpenBSD.

mgrant0:
Ok I merged it.  I basically took everything you changed which looked like cosmetic changes.  It still seems to work.

mgrant0:
I'm now working on the pad and width style.  Couple questions.

I stored them in the style struct.  I see I can get these numbers by calling `options_string_to_style` which returns a style struct.  But I don't think it's efficient to continually call `options_string_to_style` each time I need them.  Where should I store them so they can be gotten at efficiently? 

I added a `width` style keyword.  Should I also add a `height` keyword for consistency even though nothing uses it at the moment?

nicm:
I've applied the `scrollbars` branch to OpenBSD now, it will be in GitHub `master` later on.

> I stored them in the style struct. I see I can get these numbers by calling options_string_to_style which returns a style struct. But I don't think it's efficient to continually call options_string_to_style each time I need them. Where should I store them so they can be gotten at efficiently?

You will need to cache it in the pane since it is a pane option so it can have different values for each pane.

> I added a width style keyword. Should I also add a height keyword for consistency even though nothing uses it at the moment?

No, only add what is needed.

mgrant0:
Ok, I've prepared 2 branches:

https://github.com/mgrant0/tmux/tree/scrollbars-alt-screen

and

https://github.com/mgrant0/tmux/tree/scrollbars-pad-width

The alt-screen fix is going to need a mod to the mouse code to do the same check.

What should I do, should I do these as PRs to this branch so you can look at them more easily?  Or should I wait?  At any rate, let me know if you have comments on either of those patches.

nicm:
I'll look at them after GitHub syncs up and the previous changes are there.

PR #4222, opened 2024-11-02
don't store 0-width cells as extended

lexurco:
Continuing the discussion in PR #4201. I am becoming less convinced that storing tabs as non-extended cells is at all a good idea. However, just storing padding cells as non-extended seems to be a large improvement already. This small patch does that.

Thanks.


nicm:

Issue #4215, opened 2024-10-28
tmux source-file can enter an infinite loop


gpakosz:
### Issue description

It's possible to make make the `source-file `command enter an infinite loop with the following `.conf` files:

```
source -F '#{current_file}'
```

or

```
source -F '#{d:current_file}/*.conf'
```

Maybe tmux could inspect `config_files` before including another file again? As a safety net.

### Required information

* tmux version (`tmux -V`): `tmux 3.5a`
* Platform (`uname -sp`): `Darwin arm`
* $TERM inside and outside of tmux (`echo $TERM`): `tmux-256color` and `xterm-256color`
* Logs from tmux (`tmux kill-server; tmux -vv new`): I don't think it's needed

nicm:
There are two ways to fix it, either keep a list of all the previous files and do not allow one to be included again, or just have a hard limit on the amount of nesting (say 20 or 100 or something). The former will prevent a file conditionally including itself or a file that includes itself so the latter is probably better.

gpakosz:
Having a configuration that includes the same file multiple times, and relies on variables having specific values, and using `%if` statements is somewhat convoluted but possible 🙃

Having a hard limit sounds reasonable. And in that case `source-file` should return an error.

FadiAfani:
> There are two ways to fix it, either keep a list of all the previous files and do not allow one to be included again, or just have a hard limit on the amount of nesting (say 20 or 100 or something). The former will prevent a file conditionally including itself or a file that includes itself so the latter is probably better.

can nesting be counted by the number of references to a given file or a client ? by looking at the source code it seems to me a new client_file instance is created with each source-file call for a given client, am I getting this wrong ?

nicm:
I don't really know where it would go because it will need to be somewhere that subsequent `source-file` commands can access. I guess it would need to go into the `cmdq_item`.

nicm:
This should be fixed now, thanks.

gpakosz:
Thank you @FadiAfani and @nicm 🙏

PR #4223, opened 2024-11-02
fixed source-file command causing infinite loops

FadiAfani:
this should address #4215, although using a hash table to track duplicate files might be better


nicm:
I don't think this need to be that complicated and track every file separately. It just needs a count of how many files deep it is, it doesn't matter if they are different files or the same file. So it can just be a single counter instead of a list of files.

FadiAfani:
> I don't think this need to be that complicated and track every file separately. It just needs a count of how many files deep it is, it doesn't matter if they are different files or the same file. So it can just be a single counter instead of a list of files.

that makes a lot more sense actually, I thought that nesting was to be checked per file, I will try fixing this thanks !

nicm:
Are you still working on this? Thanks.

FadiAfani:
> Are you still working on this? Thanks.

Sorry, I haven't done much with it since last time I got caught up in a few things, I will get back to it if it hasn't been fixed

nicm:
I have applied this now with some changes to tidy it up a bit and to put the depth count inside `struct client` if a client is available. Will be in GitHub later when it syncs up. Thanks!

PR #4231, opened 2024-11-08
Dismiss scrollbar in alternate screen

mgrant0:
Check if the window is in the alternate screen and don't draw the scrollbar.  Call layout_fix_panes on alternate screen on and off.

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -30,8 +30,8 @@ const struct cmd_entry cmd_copy_mode_entry = {
 	.name = "copy-mode",
 	.alias = NULL,
 
-	.args = { "deHMs:t:uq", 0, 0, NULL },
-	.usage = "[-deHMuq] [-s src-pane] " CMD_TARGET_PANE_USAGE,
+	.args = { "deHMqSs:t:u", 0, 0, NULL },
+	.usage = "[-deHMqSu] [-s src-pane] " CMD_TARGET_PANE_USAGE,
 
 	.source =  { 's', CMD_FIND_PANE, 0 },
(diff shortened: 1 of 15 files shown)


nicm:
I like this, but there is a bug, note the extra half border:
Screenshot 2024-11-12 at 09 45 37
Also I don't like adding `alt_screen` everywhere when it is only used once. How about adding a macro, for example:
~~~~
#define PANE_ALTERNATE_SCREEN(wp) ((wp)->screen->saved_grid == NULL)
~~~~
Or make it work on the screen instead so `SCREEN_IN_ALTERNATE(s)` or something?

mgrant0:
Needing to check the scrollbars option is negative occurs in only one place.  You know what, I'm going to simplify this and just remove that check and always pass it in.

nicm:
Looks good but yes either remove `alt_screen` and check `s->saved_grid` directly or make it a macro, I don't mind which, but if you make it a macro it might be best to do it separately and change the other places `s->saved_grid` is checked for that purpose as well.

mgrant0:
I made those changes now, thanks.  Looks even cleaner.

Ok I redid the MR against main.  Please let me know if I didn't do that correctly.

mgrant0:
The macro is:

```
#define PANE_ALTERNATE_SCREEN(s) ((s)->saved_grid != NULL)
```
Is that the correct name?  Or `SCREEN_ALTERNET_SCREEN` since it checks the `screen` struct?

nicm:
I would go with `SCREEN_IS_ALTERNATE` or something like that, definitely not PANE since it operates on a screen.

nicm:

PR #4233, opened 2024-11-09
add pad and width style tags

mgrant0:
Added `width` and `pad` tags to the style.  They are not specific to scrollbars.  Added support to the scrollbar code to pulll the width and pad from the style.  I believe I also fixed some bugs with respect to unexpected large and small values of the width and pad.  

I made the pad a number as in the number of columns to pad.  I get it, padding is probably never more than a single column, it could be a flag.  But then the `pad` tag wouldn't be as useful for other things like adding padding in the status line like someone else wanted.  Treating both `pad` and `width` as numbers just seems to be more self evident.

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -30,8 +30,8 @@ const struct cmd_entry cmd_copy_mode_entry = {
 	.name = "copy-mode",
 	.alias = NULL,
 
-	.args = { "deHMs:t:uq", 0, 0, NULL },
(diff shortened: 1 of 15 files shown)


nicm:
This looks good on quick look, thanks. A few comments:
- You should be able to rebase this on to master now, I am going to delete the `scrollbars` branch soon.
- I do not think you should change the style inside the option. Instead, copy the style into `wp->scrollbar_style` (so make it a `struct style` instead of `struct style *`) and then change it.
- If I set `width=123123123`, tmux hangs.

PR #4241, opened 2024-11-14
Scrollbars pad width

mgrant0:
Add support to set the width and padding of the pane scrollbars.


nicm:
Is this instead of #4233 or as well as it?

mgrant0:
I looked at all the changes in both this branch and #4233 and they are identical with one exception.  #4233 has some extra changes to window-copy.c which are not mine.  It contains the commit [Tidy up loop, from Alexander Arch.](https://github.com/tmux/tmux/pull/4233/commits/f527412d9b15ab9c5b75959a83b96d7a689990bc).

I'm far from an expert at git but it looks like it shouldn't matter if you pull in one or both.

I am going to resolve the conflict now in this branch now that you pulled in the alternate screen MR.

nicm:
OK I've closed the other one if this is the same.

mgrant0:
looks good.  Conflicts merged.  Looks good, compiles without errors!

nicm:
Great, I have applied this to OpenBSD now with a few trivial changes, it will be in GitHub later. Thanks! I really like the modal scrollbar, bit tempted to turn it on by default but people might complain, we'll see...

Is there anything else you were going to work on or is that this now complete?

mgrant0:
I really like the modal scrollbar too and I may set that as my default.

I think this work is complete!  

There is some possible tidying in server-client.c to pull some of the scrollbar if statements out into a function for clarity.  They get difficult to read getting wrapped by being sqeezed to the right.  On the other hand, calling an extra function through that loop causes yet another context switch.  I'm happy to do this if you want it to be more clear, so just let me know.

nicm:
If you think it would be tidier or easier to follow then helper functions are fine, the compiler will inline them itself if it makes sense, there won't be any appreciable extra runtime cost.

mgrant0:
Ok I'll do that in another branch with another MR and you can decide if it's cleaner or not.

PR #4242, opened 2024-11-15
Add no-detach-on-destroy client flag

laur89:
As brought up in https://github.com/orgs/tmux/discussions/4237

Note I decided to implement it so session option `detach-on-destroy` would still be in effect given it's capable of producing a replacement session. If not, we employ same logic as if `detach-on-destroy` were `off`.


server-fn.c
```diff
@@ -438,15 +443,31 @@ server_destroy_session(struct session *s)
 		s_new = session_previous_session(s);
 	else if (detach_on_destroy == 4)
 		s_new = session_next_session(s);
+
 	if (s_new == s)
 		s_new = NULL;
```
laur89, line 448: Is this check a legacy remnant or is it kept as an extra security against potential future issues?
As far as I can tell none of the replacement-session-seeking functions could return the original `s`ession.

nicm:
This looks mostly OK but:

- Don't use the same flag for `CLIENT_NO_DETACH_ON_DESTROY` as `CLIENT_ACTIVEPANE`.

- Please don't use `unsigned` on its own, use `u_int`, but...

- ...I don't think using a bitmask to check for two values adds anything here, just do `(detach_on_destroy == 1 || detach_on_destroy == 2)`.

- `s_new_effective` seems a bit long, what about `s_change` or `use_s` or something?

server-fn.c
nicm: It is probably left over from old code.

laur89:
Thanks for feedback, pushed the changes.
Frankly it's been ages since I did any work on C codebase, so I need help with the `CLIENT_NO_DETACH_ON_DESTROY` bit selection.

Note the defensive but no-longer-needed `if (s_new == s)` check was also removed.

nicm:
I think the bit was OK before, it was just in the wrong place in the list so I thought it was the same as the one next to it.

I have applied this to OpenBSD now with some minor changes (some style, and I rewrote your comment to make it clearer and the man page bit, also `s_new` needs to stay initialized to `NULL`). It'll be in GitHub later on when it syncs up. Thanks!

PR #4246, opened 2024-11-17
show stderr from `run-shell` commands, not just stdout

jyn514:
when commands fail, they often print error messages to stderr. previously, failures were quite hard to debug because tmux would hide stderr.

before:
```
; tmux run-shell 'echo oops >&2; exit 1'
'echo oops >&2; exit 1' returned 1
```

after:
```
; tmux run-shell 'echo oops >&2; exit 1'
oops
'echo oops >&2; exit 1' returned 1
```

### DIFF
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -44,8 +44,8 @@ const struct cmd_entry cmd_run_shell_entry = {
 	.name = "run-shell",
 	.alias = "run",
 
-	.args = { "bd:Ct:c:", 0, 1, cmd_run_shell_args_parse },
-	.usage = "[-bC] [-c start-directory] [-d delay] " CMD_TARGET_PANE_USAGE
+	.args = { "bd:Ct:Es:c:", 0, 1, cmd_run_shell_args_parse },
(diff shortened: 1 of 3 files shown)


nicm:
I wonder if this should be a flag rather than the default...

jyn514:
hmm, when would someone want to *never* see this error output? they still have the option manually hide it with `2>/dev/null`.

nicm:
I don't know but anyone who is currently using `run-shell` w/o `2>/dev/null` will suddenly get a bunch of output they don't expect.

nicm:
I think would be better as be a `-E` or something flag to `run-shell`, not an option.

jyn514:
ok, done.

i didn't quite understand how the `template` parsing works. i saw that a letter after the option means it does not take a flag, but i'm not sure what the letter means. i just picked on at random.

nicm:
`E` on its own means `-E`, `E:` means `-E arg`.

nicm:
https://man.openbsd.org/getopt.3#DESCRIPTION

nicm:
I've applied this with one minor change - I renamed `success` to `do_close` and inverted the sense of it - and added `-E` to the man page. Should be in GitHub later. Thanks!

PR #4247, opened 2024-11-17
fix(doc): vertically <-> horizontally

vdegenne:
Unless I am completely wrong...


nicm:
No, these are correct as they are.

PR #4253, opened 2024-11-19
fix grid_reader_cursor_* functions moving over tabs

lexurco:
Hello!

This fixes next-word, next-word-end and previous-word not working correctly on lines with tabs.

Thanks.

### DIFF
--- a/grid-reader.c
+++ b/grid-reader.c
@@ -242,7 +242,7 @@ grid_reader_cursor_next_word(struct grid_reader *gr, const char *separators)
 	}
 	while (grid_reader_handle_wrap(gr, &xx, &yy) &&
 	    ((width = grid_reader_in_set(gr, WHITESPACE))))
-		gr->cx += n;
+		gr->cx += width;
 }
 
 /* Move cursor to the end of the next word. */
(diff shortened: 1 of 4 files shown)


lexurco:
Somehow didn't include these into the first commit: a correct variable name, and a fix for back-to-indentation.

lexurco:
On some further thought, I think something like this might work better. Though I am definitely not a fan of duplicating so much code in the two *_in_set functions. I'll take another look at it later.

[x.diff.txt](https://github.com/user-attachments/files/17818695/x.diff.txt)

lexurco:
It looks better now, I think, but it should probably be tested with UTF-8 characters, particularly in word-separators.

lexurco:
OK, that seems to work with UTF-8 characters too.

lexurco:
This should do it for that bug. It's a bit weird how grid_in_set() can now return >1 for tabs, but 1 for other wide characters. However, I think I'd rather address that when I make it possible for the cursor to snap to the first/last cell of tab (and perhaps any wide character).

nicm:
Good stuff, applied now, thanks!

botovq:
Many thanks for fixing this.

PR #4254, opened 2024-11-19
Add compatibility for older systemd

FractalBoy:
In the version of systemd on my system (Rocky Linux 8), there is no SD_ID128_UUID_FORMAT_STR macro.

This PR shims it in so that tmux can be built with --enable-systemd on versions without that macro.


nicm:
Applied, although I moved the define to the top of the file. Thanks!

PR #4257, opened 2024-11-21
Fix issue mouse resize not working issue

mgrant0:
When the mouse clicked on the border between two horizontal panes, `where` was not being set properly.
Pulled the checking of where the mouse click in the pane was into separate function for clarity.


mgrant0:
I fixed the issue.  I feel like I should be able to clean up the nested if statements in the new function more.

nicm:
Looks good, applied to OpenBSD now with some trivial changes. Thanks for looking at this!

PR #4260, opened 2024-11-21
man page: fix typo for `save-buffer`

ilyagr:
Feel free to treat this as a (minor) docs bug report or as a PR.

As an aside, I think a command like `read-buffer` might be nice to have. It would be easier to discover than `load-buffer -`.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -7068,7 +7068,8 @@ If
 .Ar path
 is
 .Ql - ,
-the contents are read from stdin.
+the contents are written to stdout, like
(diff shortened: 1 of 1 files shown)


nicm:
Applied to OpenBSD now, although I left out the reference to `show-buffer`. Will be in GitHub later. Thanks!

I don't think we need a new command for a small feature which is documented and is relatively common Unix command behaviour in any case. If you want `read-buffer` you could add it with the `command-alias` option.

PR #4266, opened 2024-11-23
Don't revert to black block cursor in popup

lexurco:
Hi!

This is a small patch to use cursor-colour and cursor-style from options in popup.

Thanks.


nimbuuus (approved):

nicm:

PR #4274, opened 2024-11-26
Allow gifting sessions to tmux (tmux -g + add-window -s)

nabijaczleweli:
`tmux -g 4 -- whatever-command ...` runs `whatever-command...` but also sends fd 4 to the server with it.

This allows for `tmux -g 4 -- new-window -s` to, instead of opening a new pty, reuse the master sent alongside with the command.

I usually have the fore-thought to start a very long job in tmux, but not always, and when it stings it stings (what sparked this particular instance is a like-ten-minute VM settle/poweroff/boot/restore/continue cycle).

st+iso14755, adding this to iso14755():
```c
	if (us && !strcmp(us, "totmux\n") && !fork()) {
		snprintf(codepoint, sizeof(codepoint), "%d", cmdfd);
		_exit(execlp("tmux", "tmux", "-g", codepoint, "--", "new-window", "-s", (char *)NULL));
		return;
	}
```
cooperatively gifts the session to tmux. [Demo](https://101010.pl/@nabijaczleweli/113551201086443574):

https://github.com/user-attachments/assets/e106494c-d95e-4044-b5ba-19d4389d6294

On systems that allow dup-from-remote-process, this can be used to steal any session from any terminal emulator. This example can be used on Linux:
```c
#define _GNU_SOURCE
#include <sys/syscall.h>
#include <unistd.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <err.h>
#include <dirent.h>

int main(int argc, char ** argv) {
	if (argc != 2 && argc != 3 && argc != 4)
		fprintf(stderr, "usage: %1$s xterm-pid xterm-/dev/ptmx-fd [./tmux]\n"
		                "       %1$s xterm-pid\n", argv[0]);

	pid_t pid = atoi(argv[1]);
	int pidfd = syscall(SYS_pidfd_open, pid, 0);
	if (pidfd == -1)
		err(1, "pidfd_open(%d)", pid);

	if (!argv[2]) {
		char path[128], *name;
		sprintf(path, "/proc/%d/fd", pid);
		DIR *fds = opendir(path);
		if (!fds)
			err(1, "opendir(%s)", path);

		int ret = 1;
		for(struct dirent *ent; (ent = readdir(fds));) {
			int mfd = syscall(SYS_pidfd_getfd, pidfd, atoi(ent->d_name), 0);
			if((name = ptsname(mfd))) {
				printf("%s\t%s\n", ent->d_name, name);
				ret = 0;
			}
			close(mfd);
		}
		return ret;
	}

	int rfd = atoi(argv[2]);
	int mfd = syscall(SYS_pidfd_getfd, pidfd, rfd, 0);
	if (mfd == -1)
		err(1, "pidfd_getfd(%d)", rfd);

	fprintf(stderr, "Stole %d's fd %d %s\n", pid, rfd, ptsname(mfd));

	const char * tmux = argv[3] ?: "tmux";
	char mfds[32];
	sprintf(mfds, "%d", mfd);
	fprintf(stderr, "execl(%s -g %s -- new-window -s)\n", tmux, mfds);

	fcntl(mfd, F_SETFD, 0);
	execlp(tmux, tmux, "-g", mfds, "--", "new-window", "-s", (char*)0);
	return -1;
}
```

To steal a session from st or gnome-terminal (which hosts all masters in a single gnome-terminal-server process, so the listing mode is useful there). [Demo](https://101010.pl/@nabijaczleweli/113551092840876376):

https://github.com/user-attachments/assets/6ba85b23-9c9d-44fb-9b7d-4c6fd19b387b

### DIFF
--- a/client.c
+++ b/client.c
@@ -230,7 +230,7 @@ client_exit(void)
 /* Client main loop. */
 int
 client_main(struct event_base *base, int argc, char **argv, uint64_t flags,
-    int feat)
+    int feat, int gift_fd)
 {
 	struct cmd_parse_result	*pr;
 	struct msg_command	*data;
@@ -388,7 +388,7 @@ client_main(struct event_base *base, int argc, char **argv, uint64_t flags,
 		size += sizeof *data;
 
 		/* Send the command. */
-		if (proc_send(client_peer, msg, -1, data, size) != 0) {
+		if (proc_send(client_peer, msg, gift_fd, data, size) != 0) {
 			fprintf(stderr, "failed to send command\n");
 			free(data);
 			return (1);
--- a/cmd-new-window.c
+++ b/cmd-new-window.c
@@ -38,8 +38,8 @@ const struct cmd_entry cmd_new_window_entry = {
 	.name = "new-window",
 	.alias = "neww",
 
-	.args = { "abc:de:F:kn:PSt:", 0, -1, NULL },
-	.usage = "[-abdkPS] [-c start-directory] [-e environment] [-F format] "
+	.args = { "abc:de:F:kn:PSt:s", 0, -1, NULL },
+	.usage = "[-abdksPS] [-c start-directory] [-e environment] [-F format] "
 		 "[-n window-name] " CMD_TARGET_WINDOW_USAGE " [shell-command]",
 
 	.target = { 't', CMD_FIND_WINDOW, CMD_FIND_WINDOW_INDEX },
@@ -56,6 +56,7 @@ cmd_new_window_exec(struct cmd *self, struct cmdq_item *item)
 	struct cmd_find_state	*current = cmdq_get_current(item);
 	struct cmd_find_state	*target = cmdq_get_target(item);
 	struct spawn_context	 sc = { 0 };
+	int	steal_master_fd;
 	struct client		*tc = cmdq_get_target_client(item);
 	struct session		*s = target->s;
 	struct winlink		*wl = target->wl, *new_wl = NULL;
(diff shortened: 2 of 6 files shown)


nicm:
Don't use `proc_get_last_fd` because it will be going away as soon as imsg is fixed on Linux.

This seems like a very narrow use case and requires an additional custom program as well, so I'm not sure it belongs in tmux. Is this more generally useful?

nabijaczleweli:
What's the preferred interface, then? The internal API has largely defeated me tbh

It can be integrated trivially into any terminal emulator (see the first hunk and demo 1), you only need the program to steal the ptmx non-cooperatively (demo 2).

The use-case is "moving your xterm session to tmux so it can keep running when your workstation isn't". I'd call it a USP, since it facilitates tmux saving your ass when you need it most, but.

nicm:
The fd will come with the message so you will call `imsg_get_fd` at the point it arrives. It should be fixed soon so you can probably just change it when it is done.

The problem with this is that I think hardly anyone is going to write or find and compile their own extra little program to actually do this, so it seems like we will be adding a complex extra feature to tmux that very few people will actually use. So, if this is really the only use case and `-g` is not generally useful for other purposes, I think it would only be appropriate if it can be fully integrated into tmux, so instead of `tmux -g 123` you do something like `tmux new-window -g pid:fd` and tmux does all the work for you. Obviously since this is entirely nonportable it would need to be mostly wrapped up in `osdep-linux.c` or a file in `compat`.

What happens to the fd from xterm's point of view? If I keep typing in the xterm window, does it mess with tmux? If so - we can't really have that, can it be closed?

nabijaczleweli:
Can't really see a way to get back to the imsg from the command callback, if I'm reading 0f308bd18f8051ac7a725e02e18ad234b4f467fa right, this would need to get saved somewhere in `server_client_dispatch_command()` and passed there some other way.

I'm pretty sure that on openbsd `new-window -s pid:fd` is unimplementable but the `totmux` patch (or another equivalent implementation) would still work; the ptrace restrictions are also suboptimal in principle but realistically I don't think they matter.

If you don't consider this interface upstreamable then rather than rework it I'd prefer to drop this altogether tbh; I wasn't expecting much either way and I can keep this locally. Maybe it is niche.

This is entirely transparent to the xterm, and the xterm and tmux windows are indistinguishable. You can type into either perfectly fine, but both xterm and tmux race on poll to read and display the output. Realistically this means that either one or the other wins consistently, until one dies. This is hardly avoidable.

nicm:
OK. I wasn't suggesting making it work on OpenBSD, it would be fine if it only worked on Linux and was a noop on other platforms if it could not be supported.

PR #4301, opened 2024-12-15
correct repeated preset layout name.

charlottia:
`main-vertical` is written twice, and `main-vertical-mirrored` is not mentioned here.


nicm:

charlottia:
Thanks!

PR #4303, opened 2024-12-16
Spread cells more evenly

tobbez:
Previously when spreading out cells, the space that could be evenly
distributed was, with any remaining space assigned to the last cell.

Update the code to instead spread the remaining N lines/columns evenly
across the first N cells (one each).

This is a significant improvement when layouting large numbers of cells,
since the remaining space grows linearly with the number of cells. For
example, with 20 windows, up to 19 lines/columns would be assigned to
the last cell (which is quite noticable).


nicm:

PR #4312, opened 2024-12-26
Bugfix job tranfer popup with null job to pane

mgrant0:
This seems to fix the bug when transferring a popup to a pane when the job has exited.


nicm:

PR #4322, opened 2025-01-07
Fallback dark gray to gray instead of black in 8-color terminal

dimich-dmb:
tmux running in 8-color terminal (for example, Linux virtual console with TERM==linux) passes only 3 bits of color code, so dark gray (color 8) turns to black (color 0) and becomes invisible on black background.

This change maps dark gray (8) to gray (7) if background is black.

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -2880,6 +2880,8 @@ tty_check_fg(struct tty *tty, struct colour_palette *palette,
 				gc->fg &= 7;
 				if (colours >= 16)
 					gc->fg += 90;
+				else if (gc->fg == 0 && gc->bg == 0)
+					gc->fg = 7;
 			}
 		}
 		return;


nicm:
7 is not necessarily grey, it is meant to be white, but I think this is a good idea anyway - it is probably better to use white than black-on-black.

nicm:
No, I mean mapping bright-black (8) to white (7), but I think it is OK. I'm away but I will make your change next week.

dimich-dmb:
> No, I mean mapping bright-black (8) to white (7)

Ah, indeed. There was no grey color, it was white (7) on 8-color terminal and white (7) / bright white (15) on 16-color.
Historically 16-color terminals were actually 15-color. There are 3 bits for RGB and 1 bit for brightness controlling common RGB levels attenuation. When background was black and program set foreground to "bright black" (color 8), it was black-on-black indeed. But here we are getting color 8 from our 256->16 mapping table, not from a program.

> I will make your change next week.

Thank you!

nicm:
I have applied this to OpenBSD now, it will be in GitHub later. Thanks!

PR #4351, opened 2025-01-27
tmux lscm: add some forgotten spaces after `CMD_TARGET_PANE_USAGE`

ilyagr:
Previously, `tmux lscm splitw` did not have a space between `[-t target-pane]` and `[shell-command]`.


nicm:

Issue #4341, opened 2025-01-20
`list-commands` does not use the same abbreviation logic as the rest of tmux


ilyagr:
I found it confusing that `tmux split` works but `tmux lscm split` outputs nothing.

I think that `tmux lscm split` should output the same thing as `tmux lscm splitw`, following the same command resolution logic as `tmux split` does. This also applies to `tmux lscm at`, etc.

I'm running tmux 3.5a.

**Update:** For an even more minor bug, note how there is no space between `[-t target-pane]` and `[shell-command]` at the end of the output:

```console
% tmux lscm splitw
split-window (splitw) [-bdefhIPvZ] [-c start-directory] [-e environment] [-F format] [-l size] [-t target-pane][shell-command]
```

**Update 2:** The missing space is solved by #4351.

ilyagr:
This is now implemented.

PR #4352, opened 2025-01-27
tmux list-commands: expand command prefixes 

ilyagr:
Make `tmux lscm split` equivalent to `tmux lscm splitw`, since
`tmux split` is equivalent to `tmux splitw`.

Fixes https://github.com/tmux/tmux/issues/4341

This required making a previously-static function public.

-------

I am a little uncomfortable that the `cause` variable may be allocated but never freed, but it seem that it's not freed anywhere else either (that I could find), so I hope that's OK.

On second thought, whether it's OK depends on whether it's allocated by the client (I'm guessing it's short-lived in the `tmux lscm` case) or by the server. In any case, let me know if I should change it. I could add a special-case to `cmd_find` when `**cause` is NULL, and pass NULL to it.

-------

I am not a C programmer, let me know if there are any novice issues or if there are some conventions I'm violating. I'm not sure how to format the code either, and whether there is a command to automatically reformat my code correctly.

I also didn't see any place to add tests for this.

### DIFF
--- a/cmd-list-keys.c
+++ b/cmd-list-keys.c
@@ -321,6 +321,29 @@ out:
 	return (CMD_RETURN_NORMAL);
 }
 
+static void
+cmd_list_single_command(const struct cmd_entry *entry, struct format_tree *ft, const char *template, struct cmdq_item *item) {
+	const char *s;
+	char			 *line;
+
+	format_add(ft, "command_list_name", "%s", entry->name);
+	if (entry->alias != NULL)
+		s = entry->alias;
+	else
+		s = "";
+	format_add(ft, "command_list_alias", "%s", s);
+	if (entry->usage != NULL)
+		s = entry->usage;
+	else
+		s = "";
+	format_add(ft, "command_list_usage", "%s", s);
+
+	line = format_expand(ft, template);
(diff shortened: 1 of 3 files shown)


nicm:
This is almost exactly right - I made a few minor style changes and also made it so that if the command is invalid (such as `tmux lscm l`) it shows the error string in `cause` instead of just showing nothing and returning success (this matches how `show-option` works as well). `cause` does need to be freed and it is freed in the other callers of `cmd_find` but it is not obvious where because it is passed upwards through the callers `cmd_find` -> `cmd_parse` -> `cmd_parse_build_command`.

I have applied to OpenBSD now, it will be in GitHub later. 

Thanks!

ilyagr:
Thank you very much for the review and the fixes!

Very minor: It's probably too late to fix, but the commit description of 4c12ac9 is a bit confusing. `tmux lscm` already showed either all commands or exactly one command before this change, since it was checking only for exact matches of names in the latter case.

ilyagr:
Oh, BTW, the reason it was difficult for me to find the freeing of `cause` is that I didn't know about `autogen.sh` when I went looking.

PR #4361, opened 2025-02-07
Add support for popups to control mode

gnachman:
This was prompted by [a request](https://gitlab.com/gnachman/iterm2/-/issues/12149) from an iTerm2 user.

A few thoughts:
* The concept of popup size is not well defined in control mode. The popup could be shown in a modal dialog, for example. I tried to find a compromise since a control mide client might want to display the popup in a TUI.
* I need to pass an optional wp. It's normally a u_int but I used int so I could use -1 for "invalid". Maybe there's a better way to do this that I missed.
* popup_append belongs with other string handling code, but I couldn't find a good home for it.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -432,7 +432,16 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 		w = tty->sx;
 	if (h > tty->sy)
 		h = tty->sy;
-	if (!cmd_display_menu_get_position(tc, item, args, &px, &py, w, h))
+	if (tc->flags & CLIENT_CONTROL) {
+		/* Control clients may not have a window size, so provide a
+		 * reasonable default so popups can still work. */
+		if (w == 0)
+			w = 80;
+		if (h == 0)
+			h = 25;
+                px = 0;
+                py = 0;
+	} else if (!cmd_display_menu_get_position(tc, item, args, &px, &py, w, h))
 		return (CMD_RETURN_NORMAL);
 
 	value = args_get(args, 'b');
@@ -498,5 +507,7 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 	free(cwd);
 	free(title);
 	cmd_free_argv(argc, argv);
+	if (tc->flags & CLIENT_CONTROL)
+		return (CMD_RETURN_NORMAL);
 	return (CMD_RETURN_WAIT);
 }
--- a/cmd-find.c
+++ b/cmd-find.c
@@ -36,6 +36,7 @@ static const char *cmd_find_map_table(const char *[][2], const char *);
 
 static void	cmd_find_log_state(const char *, struct cmd_find_state *);
 static int	cmd_find_get_session(struct cmd_find_state *, const char *);
+static int	cmd_find_get_popup(struct cmd_find_state *, const char *);
 static int	cmd_find_get_window(struct cmd_find_state *, const char *, int);
 static int	cmd_find_get_window_with_session(struct cmd_find_state *,
 		    const char *);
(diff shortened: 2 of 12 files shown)


nicm:
Thanks for this. A few minor comments:

- `int` for `wp` is fine if you want to use `-1` as a special value.

- I think `popup_append` is fine as just as a copy, its only a few lines. Or you could put it as `append_string` or something in `tmux.c` with the other utility functions?

- `free(NULL)` is always OK so you don't need the `if (cb->line != NULL)` in `control_free_block` (you do still need the conditional for `cb->buffer`).

- In `control_write_buffer`, `ap` is unused apart from the `va_end`, so it can be removed?

- Can you wrap lines at 80 columns? Wrap at the nearest comma for functions.

gnachman:
PTAL

nicm:
Sorry for the delay on this. The code looks OK to me but it doesn't seem to work. I run `tmux -Lx kill-server; tmux -vvCLx new` and then from another terminal `tmux -Lx attach` and `tmux popup`, but I do not see any `%popup`s in the first client.

gnachman:
Hm, I hadn't considered the use of popups interactively. I was under the impression they were mostly used to modally show info, but I now see that may not be not the case—they are more like window panes. I think that a more general approach would be better. Let me know if you agree.

gnachman:
I've updated my PR to add support for job I/O to control mode. Popup is the only feature that takes advantage of it, but there are various things I've wanted job support for anyway (such as to enable autocomplete in iTerm2's composer UI) so we can kill two birds with one (really big) stone.

nicm:
Sorry for the delay on this. 

A general question (sorry for not asking this earlier, I think I kind of misunderstood what it was intended to do) - what is the use case here? I ask because:
   1) popups are transient, so as soon as the control client disconnects, they will close - they are not making use of tmux for persistence (like panes etc);
   2) popups belong to clients, so a control client cannot access popups from other clients.

So maybe the question is  - why does iTerm2 need to have tmux manage the popup? Couldn't it just draw a popup window itself and the behaviour would be exactly the same as storing it in tmux?

As far as the code itself goes, it reads OK to me but:
- When I do `tmux -Lx new` and `tmux -Lx -C a -freport-popups`, then type `popup` in the control client:
   - If I then press `Enter` to exit the control client, tmux crashes.
   - If I close the popup with `send -t^2 exit Enter`, it closes OK but then when I press `Enter` to exit the control client, tmux still crashes.
   - If I do `popup -C` to kill the popup, tmux crashes immediately.
- Since each client can only have one popup at a time, is there any need for job IDs?
- The new popup notifications etc need to be added to `tmux.1`.

I have partly been holding off on it because I was talking to @mgrant0 about getting rid of popups in favour of sort of "floating panes"... but that would be quite a big job and I'm not sure it will happen soon, so if we want this we are better just to do this now and sort out any incompatibilities later.

Here is an updated diff with some style changes, mostly spacing and line length, but also renaming the enum members in `popup_data` to be less generic: [popup.diff.txt](https://github.com/user-attachments/files/21330636/popup.diff.txt). And here is the same but as an incremental diff your changes [popup_changes.diff.txt](https://github.com/user-attachments/files/21330639/popup_changes.diff.txt).

gnachman:
The impetus was this issue: https://gitlab.com/gnachman/iterm2/-/issues/12149

> why does iTerm2 need to have tmux manage the popup

I often see people whose workflow is to run a tmux command at the shell rather than going through the native UI (e.g., `tmux new-window` rather than cmd-N). Furthermore, I can't use a local shell for a popup because you could be ssh'ed to another host and the popup should execute on that machine. I suppose I could create a tmux window for the popup and try to hide it from the rest of the UI, but it would still be visible in other contexts, and that just seems really messy.

This change would be a big one for me as well so if popups are teetering on the brink maybe it would be best to wait. This isn't a very high priority item fo rme.

nicm:
Well, it's not so much on the brink, it's just we would like them to work differently, so they are more like floating panes than a separate thing... but it could be months or could be years or could be never. There are some changes to popups as they are now that will happen first anyway, although they wouldn't affect your changes.

nicm:
I'm going to close this because I don't want to add anything further to popups now that floating panes are there, they are likely to change.

PR #4363, opened 2025-02-09
Add display-message -C flag to update pane while message is displayed

### LINKED ISSUE #4362: Feature request: add an option for display-message to disable tty freezing
When invoking `display-message`, it always freezes tty until message times out and disappear. This is caused because of this line, where `TTY_FREEZE` flag is always being set.

https://github.com/tmux/tmux/blob/ef68debc8d9e0e5567d328766f705bb1f42b7c51/status.c#L517

I'm currently working on a companion tool for NetHack that uses `capture-pane` command to analyse screen contents and give dynamic background feedback on status bar via `display-message`. Unfortunately, even without `-N` option it freezes output until key is pressed or delay times out (which makes it unacceptable to use long delays, such as 5 seconds).

However, recompiling tmux without `TTY_FREEZE` flag in `status_message_set` seems to work just fine, with dynamic output being updated in pane while message in status line is still being displayed. Which makes me think if there are really hidden technical limitations and pitfalls that require always setting `TTY_FREEZE` flag during message display.

It seems possible to add an extra option to `display-message` command that would disable setting `TTY_FREEZE` flag for current display-message invocation.

ostrosablin:
Add a `-C` flag to prevent display-message from freezing tty and continue updating pane while message is still being displayed. This should benefit scripts that give real-time background feedback in status line without distrupting output of interactive applications that dynamically update screen.

Fixes #4362.

### DIFF
--- a/alerts.c
+++ b/alerts.c
@@ -315,11 +315,11 @@ alerts_set_message(struct winlink *wl, const char *type, const char *option)
 		if (visual == VISUAL_OFF)
 			continue;
 		if (c->session->curw == wl) {
-			status_message_set(c, -1, 1, 0, "%s in current window",
-			    type);
+			status_message_set(c, -1, 1, 0, 0,
+			    "%s in current window", type);
 		} else {
-			status_message_set(c, -1, 1, 0, "%s in window %d", type,
-			    wl->idx);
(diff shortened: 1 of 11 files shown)


nicm:

PR #4364, opened 2025-02-10
Fix a case of failure to overwrite sixel image cells

hpjansson:
When the cell to be overwritten contained part of an image, no update would be issued if the backing cell character appeared unchanged.

We address this by issuing cell updates when the cell intersects with an image, allowing images to e.g. be reliably erased with spaces.

I chose to add a new function `image_intersect_area()` instead of adding a bool argument to `image_check_area()` to make image removal optional, since we can break out of the loop early in the pure intersection case. I've no idea if this affects performance at all, though - conceivably it could have an impact if the maximum number of concurrent images is increased in the future.

This was prompted by https://github.com/hpjansson/chafa/issues/241, which contains some examples and analysis (big thanks to @niksingh710!).

### DIFF
--- a/image.c
+++ b/image.c
@@ -150,6 +150,21 @@ image_check_area(struct screen *s, u_int px, u_int py, u_int nx, u_int ny)
 }
 
 int
+image_intersect_area(struct screen *s, u_int px, u_int py, u_int nx, u_int ny)
+{
+	struct image	*im;
+
+	TAILQ_FOREACH(im, &s->images, entry) {
+		if (py + ny <= im->py || py >= im->py + im->sy)
+			continue;
+		if (px + nx <= im->px || px >= im->px + im->sx)
+			continue;
+		return (1);
+	}
+	return (0);
+}
+
+int
 image_scroll_up(struct screen *s, u_int lines)
 {
 	struct image		*im, *im1;
--- a/screen-write.c
+++ b/screen-write.c
@@ -2022,6 +2022,10 @@ screen_write_cell(struct screen_write_ctx *ctx, const struct grid_cell *gc)
 	if (skip) {
 		if (s->cx >= gl->cellsize)
 			skip = grid_cells_equal(gc, &grid_default_cell);
+#ifdef ENABLE_SIXEL
+		else if (image_intersect_area(s, s->cx, s->cy, width, 1))
+			skip = 0;
+#endif
 		else {
 			gce = &gl->celldata[s->cx];
 			if (gce->flags & GRID_FLAG_EXTENDED)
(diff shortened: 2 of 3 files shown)


nicm:
Thanks - but is this enough? What stops tmux drawing the image again over the top of the characters that were just added?

A few comments on `image_intersect_area`:
   - please remove the debug logging;
   - `isect` could be removed and just `return (1)` inside the loop and `return (0)` outside;
   - you can just use `TAILQ_FOREACH` not `TAILQ_FOREACH_SAFE` since this does not modify the `TAILQ`.

hpjansson:
Thanks for the review! I think I've addressed the issues now.

You're right - this doesn't fix every inconsistency, only one particular case that's common with previewers like fzf. If you switch to a different window and then back, the erased image comes back.

One way to cover more cases would be something like this:

- Store a special cell state (either a flag or a Unicode character, perhaps U+fffc or U+fffd) in cells underneath the image when it's first drawn to the window.
- Allow text (including spaces) to overwrite this state.
- On redraw, draw images first, and then text on top. Allow spaces to overdraw images. Don't overdraw in cells that still have the image flag/character.

Also, I noticed that on redraw, it flickers a bit as the stack of images is redrawn. It would be nice to dispose of images that become completely covered by another image (common case, not too hard), the union of higher-priority images (slightly harder) or text (hard to do efficiently?).

Granted this is just off the top of my head, and I'm not too familiar with the tmux code base.

There are a few more cases where I think it's not being 100% consistent, but I think it would take a bigger patch and introduce more risk. Maybe it's better done as a separate PR?

nicm:
I don't think we should make changes to the terminal that are then undone as soon as tmux redraws.

The way you describe to support this sounds like it would work, but if you don't want to implement that in the meantime it would be better to remove the image entirely when it is overwritten by any text (or ignore the text instead).

Removing images that are wholly hidden by another image is a good idea as well, that should be a separate PR yes.

hpjansson:
I actually tried the remove-on-overwrite approach first. Unfortunately, that makes the fzf use case worse, as the previews will then be printed and then immediately removed. The reason is that fzf prints guide text on top of the image - as seen in foot, top right:

fzf-preview-foot

So if you'd rather not have this workaround, it's probably best to reject the PR and wait for a bigger overhaul of the tmux sixel code. I can look into that if you'd like to, but I don't know when I'll have time. Do you want me to open an issue as a reminder? There are more details we could discuss there too.

nicm:
OK - I don't mind if it removes the image or doesn't print the text, but when you switch panes or do `C-b r` it should not undo the result. If you want to look at this or any other improvements that would be great... I don't need a reminder issue but if you have questions when you get to it you can either create a new issue or ask here and I can reopen this one. Thanks!

hpjansson:
Ok - the reminder would be for me :-) I'll open it as a separate issue because sixel discussions can sometimes get lengthy.

PR #4371, opened 2025-02-18
Added to .gitignore to ignore vim swapfiles.

ubinator:
My vim swapfiles come up in git status, which is annoying. I figure other people probably have the same problem. Added a line to the .gitignore. Maybe it's worth putting in the upstream.


nicm:
Applied, thanks.

PR #4372, opened 2025-02-18
Stopped ignoring error on client command parse.

ubinator:
It's always been kind of strange how tmux doesn't tell you the issues in the arguments unless it can connect to a server or start a server or whatever.

Like if you do:

    tmux thisdoesntexist

and there's no server, it'll just say can't connect to server instead of telling you that you messed up the cmdline.

I understand it from a software architecture perspective. The server parses the command, and if a connection can't be established then that is the first error that happens and as such it's reasonable that it's the error that is reported.

But it strays from almost every other command-line tool I've ever used. I just figure maybe it's worth doing in a more expected way, principle of least astonishment and such.

I can imagine 1 or 2 edge cases where it might be of miniscule importance to have this. Like if you write a script with a tmux command but don't have a server running. Then you're going to get "can't connect to server" when you test it. If you only test it once and then save it for a rainy day when you do happen to have a server running, you expect it to work. But it might not because once the server is up you find out you misspelled the command in your script. Then you have to go back and change it, which is extra work.

None of this matters that much. I mainly just want to ask what the policy is about this. Personally I think I would find it slightly better if the error were to be reported early instead of late, but I could see why you would make the opposite case.

I think optimally, we would parse it on the client, report errors early, and then forward the parsed cmd structure to the server, that way the server doesn't have to parse again.

This PR doesn't contain that. I wanted to ask what the policy is on this. If you agree with me, I would be interested in seeing what I can do about removing the client/server redundancy in the way that I mentioned above.

### DIFF
--- a/client.c
+++ b/client.c
@@ -266,8 +266,11 @@ client_main(struct event_base *base, int argc, char **argv, uint64_t flags,
 			if (cmd_list_any_have(pr->cmdlist, CMD_STARTSERVER))
 				flags |= CLIENT_STARTSERVER;
 			cmd_list_free(pr->cmdlist);
-		} else
+		} else {
+			fprintf(stderr, "%s\n", pr->error);
 			free(pr->error);
+			return 1;
+		}
 		args_free_values(values, argc);
 		free(values);
 	}


nicm:
Yes, it seems to make sense... although I have a funny feeling there was a reason not to do this but I can't remember what it was...

ubinator:
So you'd be open to the idea? If so I can see what I can do about getting a branch with non-redundant behavior. If I end up finding whatever reason there was not to do this in the first place, I will let you know. If it works out though, you'd be up for merging it in?

nicm:
I don't know if there is much more to do beyond this change. Serializing the command and sending it over the socket is going to be more complicated and probably just as expensive as simply parsing it again, and sticking it somewhere for the server to use after fork will only work for the very first client.

ubinator:
@nicm So nothing to do then? But we are on the same page about the argument errors being reported on the client right? Is there something wrong with this pull request, because you haven't merged it yet? If you don't want to eliminate the redundancy then this PR seems to be exactly what you want.

I must admit I didn't take the time to fully understand your policy on memory management on error. I freed the error text but didn't free the other things that are freed just below the if statement. There were some other parts where you didn't free on error, so I assumed maybe you'd like this the most.

As far as I can tell, the error reporting seems to match what it would look like if the server did it, down to the character. At least from what I can tell when using it on the command line.

If I may, I'd like to make a selfish request. You took the change in my other PR, improved it and committed it with my name in the title (thank you). I really really like tmux. I use it everyday and I'm a big fan of your work. And your code is very clean and I like how you've done this whole project. If it's not too much trouble for you, it'd be an honor to see myself in the git blame, or at least whatever's left after you maybe improve the few lines that I wrote. I haven't necessarily deserved it given the simplicity of my PRs, but I figure I might ask anyway.

nicm:
I think this is OK but I'll take another look at it at some point relatively soon.

All changes go into OpenBSD CVS so I will take the diff from the PR and commit it, then they are copied to GitHub automatically. You will be credited in the commit message.

ubinator:
Alright. Thought I might ask, but at the end of the day either way is fine for me, no problem :) I really do like tmux though. Great program. Like I said, use it everyday :) You're workflow is interesting, autotools and CVS. I'm going to guess you've been in the programming scene for quite a while then right? Why not just use git?

For the record, I agree with your notion of not adding unnecessary complexity. And since the redundancy was already there in order to know whether or not to start the server. This error stuff seems like a no-brainer. I do wonder though how hard it would really be to forward the parsed cmd structure through the protocol. I don't know much about the protocol side of tmux yet, but if it's implemented like I expect it to be implemented, it'd probably be a doable task. But this definitely isn't a priority and potentially adding bugs because of it isn't worth it I guess. So I understand where you're coming from. I think your solution is a good middle-ground.

nicm:

nicm:
I have reverted this change (d938ab5dd7698e9f4dd92986d5b9746b2cd12834) for now, because it breaks command aliases - they are not set in the server but not the client. See https://github.com/tmux/tmux/issues/4424 for an example. I think the only way to do it would be to have `cmd_parse_from_arguments` differentiate somehow between "definitely an error" from "maybe retry in the server".

Issue #4370, opened 2025-02-17
Cannot select-pane between two horizontal panes with pane-scrollbar on


davidsierradz:
### Issue description

Hi

tmux.conf:

```sh
set-option -g pane-scrollbars on
```

Steps to reproduce:
1. Open a horizontal panel (`C-b %`)
2. Try to move between panels (`C-b Left` and/or `C-b Right`)

With `set-option -g pane-scrollbars modal` if the two panels are in copy-mode, the same happens

Friendly ping @mgrant0

Thanks

### Required information

Please provide the following information:

* tmux version: next-3.6 (9a377485becdd34dda695f38cb73ee5082d9088b)
* Platform: NixOS
* $TERM inside: `tmux-256color` and outside of tmux: `foot`
* Logs from tmux (`tmux kill-server; tmux -vv new`)

[tmux-client-816019.log](https://github.com/user-attachments/files/18826752/tmux-client-816019.log)
[tmux-out-816022.log](https://github.com/user-attachments/files/18826750/tmux-out-816022.log)
[tmux-server-816022.log](https://github.com/user-attachments/files/18826751/tmux-server-816022.log)

mgrant0:
Ok thanks, i will get to this.  Busy traveling atm.

mgrant0:
I've fixed this in this PR: https://github.com/tmux/tmux/pull/4374

nicm:
This PR is now applied to OpenBSD so should be in GitHub later. Thanks!

PR #4374, opened 2025-02-21
Bugfix for issue 4370: Cannot select-pane between two horizontal panes with pane-scrollbar on

mgrant0:
Bugfix for https://github.com/tmux/tmux/issues/4370

Added the width of the scrollbars to the calculation of the width of the window panes when finding the adjacent panes.


nicm:
Applied with some trivial changes, thanks!

PR #4378, opened 2025-02-22
account for scrollbar width when drawing pane-border-status

mgrant0:
I noticed this the other day, when scrollbars and pane-border-status both are enabled, the pane-border isn't properly coloured all the way across when the pane is not the active pane.  Some example images:

Screenshot 2025-02-22 152752

Screenshot 2025-02-22 152709

1. I added taking into account the width of the scrollbar if enabled, but this wasn't enough!
2. Instead of subtracting 4 characters, I had to reduce that to 2

n.b. sb_w is 0 when scrollbars are disabled.

I was a bit surprised I had to change the 4 to a 2 but it didn't seem to break things in the cases I tested.  Leaving it at 4 for me leaves a single character at the end of the border unchanged when you switch panes.  If this breaks something please let me know!

### DIFF
--- a/screen-redraw.c
+++ b/screen-redraw.c
@@ -434,9 +434,14 @@ screen_redraw_make_pane_status(struct client *c, struct window_pane *wp,
 	struct format_tree	*ft;
 	char			*expanded;
 	int			 pane_status = rctx->pane_status;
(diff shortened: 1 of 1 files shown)


nicm:
Applied now, thanks!

Issue #4330, opened 2025-01-11
Popup windows should not be draggable while the mouse is still pressed during the selection process


drowning-cat:
### Issue description

https://github.com/user-attachments/assets/c5d0ccb8-ca09-43da-ae58-4049151bcf40

### Expected behavior

1. Prevent dragging the floating window while the mouse button is held down during text selection.
>
2. **Or**, allow text selection inside a floating window, even if the mouse is dragged outside the popup while the mouse button is still pressed, like in a browser.

Demo for p2: https://codepen.io/drowning-cat/pen/emOMMjo

https://github.com/user-attachments/assets/2c7bbf40-4798-4b4d-b922-a0bab6858a1f

### Required information

Please provide the following information:

* tmux version (`tmux -V`).
  - tmux next-3.6
* Platform (`uname -sp`).
  - Linux unknown
* $TERM inside and outside of tmux (`echo $TERM`).
  - alacritty
  - tmux-256color

drowning-cat:
Thanks to @mgrant0 the issue can be closed.

PR #4379, opened 2025-02-22
Non blocking popup windows

mgrant0:
This is a work in progress, not ready to pull quite yet but I wanted to discuss it.

Note how I did this, I created a new POPUP_FOCUSED flag.  But I also re-used the PANE_FOCUSED flag for a slightly different purpose.  Is this a good idea?  

After doing this, I started thinking maybe I should use w->active instead setting PANE_FOCUSED?  

However, this does let you:
```
tmux display-popup -E bash & 
```
and get a floating shell.  But it's a little strange, you can only have a single popup like this. 

I'd like some comments on this method of solving this please.  Thanks


mgrant0:
This does not quite work.  The problem with this approach is that this popup is part of the client.  This means if you reattach from a different client, you don't see the popup.  If you start a popup in the second client and try moving the focus around strange things start happening.  

I can't reuse PANE_FOCUS the way I did and to know if the popup in the client is focused, I need to store this elsewhere or somehow check if POPUP_FOCUS is set instead.

mgrant0:
This seems to work well now.  I added a new flag `CLIENT_OVERLAYPOPUP_FOCUSED` to the client struct.  Each client therefore has its own overlay popup and corresponding flag which indicates if its popup is focused or if it's the window behind.

mgrant0:
There is a bug with the cursor.  When the cursor is behind the overlay, it shows the block cursor.

mgrant0:
I'd love some feedback on this now!  How I have tested this:

1. Start an overlay shell: `tmux display-popup -E bash & `
2. Split the window: `tmux split-window`
3. Enable scrollbars: `tmux set-option -g pane-scrollbars on`
4. Drag the overlay around

Remember, you only get one overlay-popup per client.

nicm:
I like this and seems to work well.

A few comments:

- I think we should add `-b` to `display-popup` to not block the client (like `-b` to `run-shell` and `if-shell`). Just passing in `NULL` for `item` to `popup_display` should be enough I think, that will stop it calling `cmdq_continue` in `popup_free_cb`.

- `CLIENT_OVERLAYPOPUP_FOCUSED` is a bit of a mouthful, why not just `CLIENT_OVERLAYFOCUSED`?

- The mouse doesn't work. Mouse clicks outside the pane should be passed to the pane and ideally it should be possible to open the pane menu as normal (I'm not sure how best to open the popup menu, maybe only on the border?).

- Could you skip the loop in `server_client_handle_key` if `done` is set? It's going to delete the overlay anyway?

mgrant0:
Sure I can add a -b to not to block.  I guess this is to not break 
backwards compatibility if someone depends on it blocking?  Does anyone 
actually use this for something where they depend on it blocking?

CLIENT_OVERLAYFOCUSED sounds great.

Not only doesn't the mouse work, nor does the prefix key, nor 
scrollbars, nor other things.  I can work on the mouse and perhaps even 
the prefix.  I still feel a more general floating pane might be better 
value for effort here.  I spent quite a few days staring at the code to 
try to understand how to do that.  If you have some thoughts on this at 
some point, please let me know.

I will look at skipping the loop in server_client_handle_key.

Thanks

------ Original Message ------
From "Nicholas Marriott" ***@***.***>
To "tmux/tmux" ***@***.***>
Cc "Michael Grant" ***@***.***>; "Author" 
***@***.***>
Date 24/03/2025 19:30:47
Subject Re: [tmux/tmux] Non blocking popup windows (PR #4379)

nicm:
Yeah -b will not break existing uses, also it matches run-shell where the default is blocking.

For floating panes there will be quite a bit to do. You'll need to handle them basically everywhere a pane is handled.

For the first step I would probably look at the redrawing, so I'd add a list of per-window floating panes, with a struct like:
~~~~
struct floating_pane {
    u_int px;
    u_int py;
    u_int sx;
    u_int sy;
    u_int z;
    RB_ENTRY(floating_pane) entry;
};
~~~~
Then add a tree ordered by `z` to `struct window `, some test code to add a few floating panels to every window and start looking at how to draw them. You'll need to look at both the `screen-redraw.c` and `input.c` redraw code. The overlay code will be a start but floating panes will be relative to the window view rather than the client, so they would need to be offset, clipped etc as needed.

Once drawing is working the next step would be resizing, moving, changing z order and maybe hiding panes entirely. Then once that is done you could link up the struct above to an actual `struct window_pane`, so allow a pane to have either a `struct layout_cell` or `struct window_pane` (an intermediate "pane view" struct with a type could be useful there). And finally everywhere that panes are used or usable would need to be looked at to decide how it works with floating panes.

nicm:
The other option would be to make floating panes part of the layout so there is a floating `layout_cell`. That would allow the layout code to include them when it applied a layout set and would also potentially allow gaps (dead space) in the layout. A lot the redraw code would still need to become "floating aware" but it might be neater because for redraw purposes you could treat the existing layout as just cells with a z order that means they are always on the bottom.

mgrant0:
I'll work on finishing off the things for the non blocking popups you 
asked about and also look at issue #4330 which I can likely fix while 
making the mouse work.

I am guessing you are not thinking of replacing these session based 
overlay popups with any sort of new floating pane.  I'll work on the 
floating panes after this perhaps.

I should be able to get the mouse working again easily.  prefix key too 
i hope.

------ Original Message ------
From "Nicholas Marriott" ***@***.***>
To "tmux/tmux" ***@***.***>
Cc "Michael Grant" ***@***.***>; "Author" 
***@***.***>
Date 25/03/2025 09:50:53
Subject Re: [tmux/tmux] Non blocking popup windows (PR #4379)

mgrant0:
I also thought about making floating panes part of the layout.

Maybe a bit radical but I suppose there could be something like a `pane-policy` which could be either `tiled` or `floating`.  If set to `floating`, any pane could just be moved around.  If a pane was split, it's split in place and you have 2 panes and each could be moved away from the other separetly.  You would have to have borders around all the panes edges to make this work well I think, so you'd lose a bit of screen realestate.

mgrant0:
> I like this and seems to work well.
> 
> A few comments:
> 
> * I think we should add `-b` to `display-popup` to not block the client (like `-b` to `run-shell` and `if-shell`). Just passing in `NULL` for `item` to `popup_display` should be enough I think, that will stop it calling `cmdq_continue` in `popup_free_cb`.
> * `CLIENT_OVERLAYPOPUP_FOCUSED` is a bit of a mouthful, why not just `CLIENT_OVERLAYFOCUSED`?
> * The mouse doesn't work. Mouse clicks outside the pane should be passed to the pane and ideally it should be possible to open the pane menu as normal (I'm not sure how best to open the popup menu, maybe only on the border?).
> * Could you skip the loop in `server_client_handle_key` if `done` is set? It's going to delete the overlay anyway?

`-b` was already used, how about `-D`? (as in Don't block)

I think I have the mouse working.  But you need to apply my fix to #4450 after because it fixes a segv.  With that fix, you can do 
```
tmux/tmux display-popup -D -E 'tmux/tmux new'
```
and the mouse works, you can select stuff in the window.  Getting into an editor, the scrollwheel and left mouse button work for me.

`tmux.1` needs updating.  Any other problems with this?

mgrant0:
There is still an issue with opening the pane's menu.  It's opening the overlay's menu.  However, left clicking is going to the pane.

mgrant0:
after pulling the current master, I'm getting a segv when the overlay closes, working on that...

mgrant0:
The segv is happening in `popup_free_cb`.  It seems that `cmdq_get_client(item)` is returning garbage as if something got freed but still returning a pointer to it.

mgrant0:
Left-clicking the mouse outside the overlay is working.  However, right-clicking is not (yet).  I am currently working on this.

mgrant0:
This is getting closer but still does not fully work as you would expect.  You can now start a display popup overlay (blobked or not), and you can right-click the mouse on the border of the overlay or even outside the overlay and get the overlay's window _as long as it's focused_.  If you click in the pane which brings the focus to the pane, then right click, it brings up the pane's menu.  But it clears the overlay popup!

Here's what's going on:

When the overlay popup is visible and focused, right-clicking builds the menu which floats over the overlay like this in popup_key_cb:
```
  menu:
          pd->menu = menu_create("");
          if (pd->flags & POPUP_INTERNAL) {
                  menu_add_items(pd->menu, popup_internal_menu_items, NULL, c,
                      NULL);
          } else
                  menu_add_items(pd->menu, popup_menu_items, NULL, c, NULL);
          if (m->x >= (pd->menu->width + 4) / 2)
                  x = m->x - (pd->menu->width + 4) / 2;
          else
                  x = 0;
          pd->md = menu_prepare(pd->menu, 0, 0, NULL, x, m->y, c,
              BOX_LINES_DEFAULT, NULL, NULL, NULL, NULL, popup_menu_done, pd);
          c->flags |= CLIENT_REDRAWOVERLAY; 
```

When you right-click in the pane, the menu is built when `cmd_display_menu_exec` calls `menu_display`.  `menu_display` looks like this:
```
  int
  menu_display(struct menu *menu, int flags, int starting_choice,
      struct cmdq_item *item, u_int px, u_int py, struct client *c,
      enum box_lines lines, const char *style, const char *selected_style,
      const char *border_style, struct cmd_find_state *fs, menu_choice_cb cb,
      void *data)
  {
          struct menu_data        *md;

          md = menu_prepare(menu, flags, starting_choice, item, px, py, c, lines,
              style, selected_style, border_style, fs, cb, data);
          if (md == NULL)
                  return (-1);
         server_client_set_overlay(c, 0, NULL, menu_mode_cb, menu_draw_cb,
              menu_key_cb, menu_free_cb, NULL, md);
          return (0);
  }
```

The call to `server_client_set_overlay` clears the overlay popup and uses the overlay for the menu.  

How to fix this?  Could `menu_display` be easily fixed so it doesn't use the overlay?

nicm:
There are a couple of things you can do here.

To do it really nicely I think you would need an overlay stack, so the pane menu can appear at the same time as a popup. Keys and mouse could always go to the top item in the stack, but redraw would redraw all (taking care of skipping cells in an older overlay that will be overwritten by a newer).

A quicker solution may be to take advantage of the fact there can only be one popup and one menu and either just splitting them into two separate overlays (pretty much the same as a stack but just with two), or adding code when a menu is opened to save the popup state if a popup exists and to the menu redraw code to draw the popup (similar to how the popup code draws it's own menu).

mgrant0:
Finally some progress.  I seem to have found the SEGV.

mgrant0:
@nicm please try this, it seems to work.

1. Please decide if you are ok with the way I added the callbacks to the window struct or if you prefer those functions being called directly.
2. The `menu_check_cb` callback is conspicuously missing from this group of callbacks now in the window struct, maybe I should add it too?  It's not set in `server_client_set_overlay`.

More testing now.  Floating windows still on the table, this just fixed outstanding things with the current overlay.

mgrant0:
Note that you get the menu for whatever is focused.  If the overlay popup is focused, right-clicking outside it gives you the overlay's menu.  If the window is focused, you get that menu unless you unless you are right-clicking inside the overlay.  This seems the most user friendly way to get at the menu in both worlds.

nicm:
Sorry for the delay here, I have not had much time to look at "big" changes recently.

I think this is a sensible way for it to behave (use the menu of the one that is focused).

It works well but there is one bug. If I focus the pane and then open the pane menu so that it covers the popup, moving down the menu seems to affect the popup - sometimes it resizes it, sometimes it opens the popup menu, occasionally it makes tmux crash. Here is a video:

https://github.com/user-attachments/assets/dfbd631e-19f7-41e4-9c7b-0a5187c5b8ce

mgrant0:
No worries about the delay!  I know you get to things eventually!

Oh nice catch, that's a bug not a feature.  Let me fix that.

mgrant0:
Ok please let me know if you see any other bugs.

mgrant0:
There is a bug around binding a menu to other buttons.  Working on that now.

mgrant0:
I have been working on this.  There are currently 3 issues which I have been trying to track down:

1. If you bind `MouseDown1Pane` to the same thing that `MouseDown3Pane` is bound to by default, a strange thing happens.  When you get the menu, it also starts selecting text in the pane.  Maybe this is expected because it's the first mouse button?
2. If you bind the same menu to `MouseDown2Pane`, you can't click on anything in the menu.  This may be because I can't seem to produce a drag event on this mouse button on my laptop.  Maybe I need to bind something else to capture a MouseDown1 in the menu?
3. When the popup is open, the menu now correctly displays and when you move the mouse over the menu, each item in the menu is highlighted when you roll over it.  However, if the popup is not displayed, the rollover highlighting does not work and I don't know why.

(3) clearly needs to get fixed.  (1) and (2), I'm not sure if this is expected.  I literally just copied the `MouseDown1Pane` binding to the other keys.

I'm a bit stuck at the moment.  I will figure out out eventually but if you have any clues that could move me along faster that would be much appreciated.

mgrant0:
I am back looking at this today.  I can't reproduce (3) above unless I bind `MouseDown1Pane`, so it appears it's not a general issue.  The other 2 issues above may be expected or non-issues.  If there are other menus that are bound to things that I need to test please let me know.

In this PR, I did not attempt to add listening for the prefix key or anything to do with the mouse, as in selecting text or putting the popup in a mode such as scrolling back.  Perhaps a future feature.

This PR could be ready to go pending any other things you want checked or done to it.

hovsater:
@nicm @mgrant0 I would love to see this one get over the finish line. Is there something I could do to help out?

nicm:
This has been superseded by floating panes (`floating_panes` branch).

mgrant0:
Actually there are things in this branch that could be pulled into the floating panes branch.  This branch specifically lets popups be unblocked.  But it's true that you can do much more with floating panes.  

@nic Is it worth pulling the relevant changes into the floating panes branch to give client side popups this -D option to not block?  Let me know, I can do this.

mgrant0:
@hovsater floating panes gives you server side "popups", they are full fledged panes, they have a copy mode you can scroll back.  You can have more than one on the screen at once, and in multiple windows at once.  

Client side overlay popups, you get one of these, no scrolling back, no prefix character, escape quits them.  They are basically only useful for a modal popup message.  Leaving them to block the way they do are definitely consistent with a popup modal message.  But like i said I could merge in the -D changes if there was a need to do non-blocking modal popups.  Just not sure its worth it!

nicm:
Maybe eventually but I would resist adding features to the branch until the core functionality (opening, redrawing, moving, closing floating panes) is well tested, crash free and ready for master because the more changes there are the harder it will be to merge into master and then back into the branch.

hovsater:
I wasn't aware that work was going on in the `floating_panes` branch. I'm not sure I fully understand the difference.

My specific use case is that I'd like to spawn programs in "popups" that does not disrupt my current layout and then still be able to interact with say, my editor. The current popup mechanics is nice but like you said, it blocks everything under it.

Perhaps floating panes solve that?

mgrant0:
Floating panes 100% solves this.  `tmux newp` opens a shell in a floating pane.  Run your command, do whatever you would do in a tiled pane.  When done, ctrl-d or exit the shell and it goes away.

`tmux newp cmd arg arg arg` pops up a floating pane with cmd running in the pane and you still have the shell prompt in the tiled pane.  However, at the moment the pane closes when the process exits.  Meanwhile `tmux 'newp cmd args...; read'` acomplishes waiting until enter is pressed to exit the new pane if you need it to block.

There is a todo item to add an option to block like the current popups do.

The floating panes branch is fairly stable.  I run it as my every day tmux at the moment.  Feedback welcome here: https://github.com/tmux/tmux/issues/4800

nicm:
I'm closing this now in favour of floating panes.

Issue #4383, opened 2025-02-23
";=" syntax not documented in tmux.1


mml:
I read the latest tmux.1 from github and couldn't find it there.

In the default status format, we have

```
#{T;=/#{status-left-length}:status-left}"
```

I tried to find an explanation for the `;` and the `=` in the manpage, but couldn't.  Reading the source, I think the `;` just is an alternate way to indicate the end of the `T` modifier?  And the `=`, I _guess_ (maybe?) is saying to expand/interpret `status-left-length`.  But I didn't find that in the code.

Maybe this is in the manpage and I just didn't read the right section.  If it's not, I'm happy to send a patch explaining it.  And if it's in there and I could not find it, I'll also be happy to author a change making the manpage a bit more clear.

nicm:
`;` just separates multiple modifiers.

PR #4384, opened 2025-02-24
Document the use of ';' as a modifier separator.

mml:
Fixes github issue #4383.


nicm:

PR #4399, opened 2025-03-07
Don't copy padding cells into the selection

jyn514:
Tmux represents tab padding using '!' padding markers. Most commands check `GRID_FLAG_PADDING` and know not to render these. However, regex searching did not respect the padding flag.

Steps to reproduce the issue this fixes:
1. `tmux set -g mode-flags vi`
2. `printf '\tfoo`
3. `C-b [ ?\sfoo<Enter>` (enter copy mode, search for "whitespace foo", copy the selection)
4. `C-b ]` (or any other method of pasting the current selection)

before this PR, that will show `	!!!!!!!foo` instead of `	foo`.


nicm:

PR #4401, opened 2025-03-07
Cleanup window_get_active_at function.

mgrant0:
Cleanup patch.  Simplified the `window_get_active_at` function to use the new `window_pane_full_size_offset` function which was created to fix issue #4370.


nicm:

PR #4410, opened 2025-03-14
tmux unlimited history-limit

yash-fn:
the command line arg is apparently converted into long long which can fit signed int. So why not allow negative values for history-limit like with how bash allows negative values for history as well. Like with bash, negative values disable history trimming i.e. all history is kept in buffer. Consumer level ram machines are becoming more readily available (see apple m3 ultra with 512gb ram). Maximum addressable # of lines with 32-bit int max is like 1.6tb with utf-8, but about 400gb with ascii (this was a table napkin calculation with assumptions).


nicm:
What is the need for this? The maximum is already INT_MAX and nobody will ever reach that without running out of memory. Not to mention that exceeding UINT_MAX is likely to cause problems.

yash-fn:
Well assuming ascii and 212 characters per line you get to 400gb < 512 gb ram machine. But main motivation is just easily requesting unlimited history like with bash. Whether truly unlimited or set to whatever the stable max is I guess fine too. But isn't uint 32bit? Couldn't 64bit be used instead?

nicm:
Do you have such a machine? Did you try it? What happens when the number of lines wraps INT_MAX and UINT_MAX?

It sounds like your forgot about attributes, you'll find that 400 GB with ASCII is more like 3 million lines, to wrap INT_MAX you will need about 2 terabytes of memory.

If you just want to make it easier, it would be best to make it so that history-limit -1 means INT_MAX.

nicm:
Change the option to have min -1 and anywhere it is used check for -1 and use INT_MAX instead.

PR #4411, opened 2025-03-16
Increase escape-time delay to a minimum of 500 ms when requesting device attributes

jyn514:
This avoids treating the terminal response as user input when it comes over a network connection.

I confirmed that this fixed the issue for me when ssh-ing from a Windows host (Windows Terminal, OpenSSH_for_Windows_9.5) to a Linux host.

This is the same as the suggested fix in https://github.com/PowerShell/Win32-OpenSSH/issues/2275#issuecomment-2462437774.

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -937,6 +937,16 @@ partial_key:
 	delay = options_get_number(global_options, "escape-time");
 	if (delay == 0)
 		delay = 1;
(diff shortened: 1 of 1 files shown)


nicm:
Applied to OpenBSD now with some minor changes (I removed the comment and tweaked the log text), will be in GitHub later. Thanks!

jyn514:
you're very welcome!

PR #4414, opened 2025-03-17
Expose additional formatting variables

shiro:
## Description

As mentioned in https://github.com/orgs/tmux/discussions/4407#discussioncomment-12487995, I would like to access more cursor related information. Hooking up some variables to the format function seems to work well, thanks for the pointers!

## Changes

- exposes new `${cursor_colour}` variable
- exposes new `${cursor_shape}` variable
- exposes new `${cursor_blinking}` variable
- exposes new `${cursor_very_visible}` variable

## Notes
It's my first PR here, let me know if something's not right :bowing_man:


nicm:
Good stuff, applied now to OpenBSD with some minor style changes (mostly to `format_cb_cursor_colour` to keep lines under 80 columns). Will be in GitHub later when it syncs. Thanks!

PR #4415, opened 2025-03-18
Make it easy to swap windows from choose-tree

### LINKED ISSUE #4388: Suggestion: add keys in choose-tree to swap the selected window with the one above or below
### Issue description

`move-window` and `swap-window` work well to reorder windows, but it would be nice to be able to do that graphically too. I was thinking that if `choose-tree` had additional keybindings to swap the currently selected/highlighted window with the one directly above or below (which might not be `-1` and `+1`), it would make it a lot easier to rearrange windows while seeing what I'm doing.

Personally, I'd like to use this while sorting by name, but only to rearrange windows that share the same name. I can see how that could be confusing though if a user tries to swap windows with different names, and the order shown by `choose-tree` doesn't change. Maybe it would be worth showing an error message instead, if swapping indexes with the window above or below would not change the order with the current sorting?

### Required information

* tmux version (`tmux -V`). tmux 3.5a
* $TERM inside and outside of tmux (`echo $TERM`). outside: xterm-256color, inside: tmux-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`). N/A, I think

dseomn:
I picked the new keys for this because of the similarity to switching vs swapping tabs in a browser or other tabbed UI. There, Ctrl+PgUp switches to one tab previous and Ctrl+Shift+PgUp swaps with the previous tab. So I figured adding Shift to swap instead of move might be somewhat natural here.

Some of the swapping code is copied from cmd-swap-window.c

Fixes #4388

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -58,6 +58,7 @@ struct mode_tree_data {
 	mode_tree_menu_cb         menucb;
 	mode_tree_height_cb       heightcb;
 	mode_tree_key_cb	  keycb;
+	mode_tree_swap_cb	  swapcb;
 
 	struct mode_tree_list	  children;
 	struct mode_tree_list	  saved;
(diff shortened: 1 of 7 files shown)


nicm:
Can you pick normal keys as well (`U` and `D`? or `M-u` and `M-d`?)? Not all terminals will have Shift function keys

dseomn:
Sure. Do you think `K`/`J`, `C-P`/`C-N`, and `KEYC_WHEELUP_PANE|KEYC_SHIFT`/`KEYC_WHEELDOWN_PANE|KEYC_SHIFT` would make more sense, to match this code?

https://github.com/tmux/tmux/blob/3eb93383a3302bda0ec0ba509ce47d92e31b354a/mode-tree.c#L1138-L1149

Should I add the additional keys to `tmux.1`? It looks like it only has one key per function at the moment.

nicm:
Don't do `C-P`, most terminals don't distinguish `C-p` and `C-P`. Maybe `M-p` and `M-n` if they aren't already used?

I wouldn't do the mouse with modifiers.

We can always add more keys later but the important thing is there should be some key that every terminal will support.

I think it's simpler if man page only has one key maybe even if there are aliases?

dseomn:
> Don't do `C-P`, most terminals don't distinguish `C-p` and `C-P`.

Oh, good point.

> Maybe `M-p` and `M-n` if they aren't already used?

I think those would conflict with the default keys to jump to a line:

https://github.com/tmux/tmux/blob/3eb93383a3302bda0ec0ba509ce47d92e31b354a/window-tree.c#L59-L68

> I wouldn't do the mouse with modifiers.
> 
> We can always add more keys later but the important thing is there should be some key that every terminal will support.
> 
> I think it's simpler if man page only has one key maybe even if there are aliases?

Sounds good. I added `K` and `J` since those didn't seem to be used already, and I left the man page as it was.

nicm:
Code looks perfect to me and it works great, applied to OpenBSD now, will be in GitHub later. Thanks!

PR #4419, opened 2025-03-20
Documentation fixes and slight imrovements

druckdev:
This includes:

  - Syncing between the usage string in code and in the manpage
  - Adding optional arguments that were simply not mentioned (e.g.
    shell-command arguments)
  - Adding square brackets around arguments that are actually optional
  - Fix run-shell's maximum argument count
  - Document default-command as display-popup's fallback
  - Improve the documentation of bright colours slightly
  - Improve documentation on the scope and type of options

### DIFF
--- a/cmd-bind-key.c
+++ b/cmd-bind-key.c
@@ -38,7 +38,7 @@ const struct cmd_entry cmd_bind_key_entry = {
 
 	.args = { "nrN:T:", 1, -1, cmd_bind_key_args_parse },
 	.usage = "[-nr] [-T key-table] [-N note] key "
-	         "[command [arguments]]",
+	         "[command [argument ...]]",
 
 	.flags = CMD_AFTERHOOK,
 	.exec = cmd_bind_key_exec
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -42,7 +42,7 @@ const struct cmd_entry cmd_display_menu_entry = {
 	.usage = "[-MO] [-b border-lines] [-c target-client] "
 		 "[-C starting-choice] [-H selected-style] [-s style] "
 		 "[-S border-style] " CMD_TARGET_PANE_USAGE " [-T title] "
-		 "[-x position] [-y position] name key command ...",
+		 "[-x position] [-y position] name [key] [command] ...",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
@@ -59,7 +59,7 @@ const struct cmd_entry cmd_display_popup_entry = {
 		 "[-d start-directory] [-e environment] [-h height] "
 		 "[-s style] [-S border-style] " CMD_TARGET_PANE_USAGE
 		 " [-T title] [-w width] [-x position] [-y position] "
-		 "[shell-command]",
+		 "[shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -43,7 +43,7 @@ const struct cmd_entry cmd_new_session_entry = {
 	.usage = "[-AdDEPX] [-c start-directory] [-e environment] [-F format] "
 		 "[-f flags] [-n window-name] [-s session-name] "
 		 CMD_TARGET_SESSION_USAGE " [-x width] [-y height] "
-		 "[shell-command]",
+		 "[shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_SESSION, CMD_FIND_CANFAIL },
 
--- a/cmd-new-window.c
+++ b/cmd-new-window.c
@@ -40,7 +40,8 @@ const struct cmd_entry cmd_new_window_entry = {
 
 	.args = { "abc:de:F:kn:PSt:", 0, -1, NULL },
 	.usage = "[-abdkPS] [-c start-directory] [-e environment] [-F format] "
-		 "[-n window-name] " CMD_TARGET_WINDOW_USAGE " [shell-command]",
+		 "[-n window-name] " CMD_TARGET_WINDOW_USAGE
+		 " [shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_WINDOW, CMD_FIND_WINDOW_INDEX },
 
--- a/cmd-respawn-pane.c
+++ b/cmd-respawn-pane.c
@@ -36,7 +36,7 @@ const struct cmd_entry cmd_respawn_pane_entry = {
 
 	.args = { "c:e:kt:", 0, -1, NULL },
 	.usage = "[-k] [-c start-directory] [-e environment] "
-		 CMD_TARGET_PANE_USAGE " [shell-command]",
+		 CMD_TARGET_PANE_USAGE " [shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
--- a/cmd-respawn-window.c
+++ b/cmd-respawn-window.c
@@ -36,7 +36,7 @@ const struct cmd_entry cmd_respawn_window_entry = {
 
 	.args = { "c:e:kt:", 0, -1, NULL },
 	.usage = "[-k] [-c start-directory] [-e environment] "
-		 CMD_TARGET_WINDOW_USAGE " [shell-command]",
+		 CMD_TARGET_WINDOW_USAGE " [shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_WINDOW, 0 },
 
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -44,7 +44,7 @@ const struct cmd_entry cmd_run_shell_entry = {
 	.name = "run-shell",
 	.alias = "run",
 
-	.args = { "bd:Ct:c:", 0, 2, cmd_run_shell_args_parse },
+	.args = { "bd:Ct:c:", 0, 1, cmd_run_shell_args_parse },
 	.usage = "[-bC] [-c start-directory] [-d delay] " CMD_TARGET_PANE_USAGE
 	         " [shell-command]",
 
--- a/cmd-send-keys.c
+++ b/cmd-send-keys.c
@@ -35,7 +35,7 @@ const struct cmd_entry cmd_send_keys_entry = {
 
 	.args = { "c:FHKlMN:Rt:X", 0, -1, NULL },
 	.usage = "[-FHKlMRX] [-c target-client] [-N repeat-count] "
-	         CMD_TARGET_PANE_USAGE " key ...",
+	         CMD_TARGET_PANE_USAGE " [key ...]",
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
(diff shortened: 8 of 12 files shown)


druckdev:
I would like to ask again, if it is possible to keep some kind of git authorship intact, in case you merge (some of) these. I understand that the actual git author does not work, but could you please keep/add the `Co-authored-by` trailer and the end of the commit message? This way Github will link the commits even if you are the committer and author.
Thanks a lot!

nicm:
Thanks for this but no, sorry, I can mention your name or your email or both in the form you prefer but I'm not including Git syntax in CVS commit messages.

I don't know if there would be a way for the sync script to insert them if it saw an email address in the commit message - I normally prefer real names and obfuscate email addresses but I could easily include email addresses as is - we would need to ask @ThomasAdam. But why is this so important? I'll make sure you are credited, what difference does the form make?

Note that if you make significant changes to a file you should add your name and email to the copyright at the top (by significant I mean dozens of lines at least, a few lines of changes are not copyrightable).

druckdev:
Thanks for your reply, but I'm sorry, I don't understand 100% what you're suggesting:

> [...] a way for the sync script to insert them if it saw an email address in the commit message - I normally prefer real names and obfuscate email addresses but I could easily include email addresses as is [...]

You mean that the commit message would include the name and email address, just not prepended by `Co-authored-by`? I don't see how the trailer would make a big difference here, but maybe I'm misunderstanding you.

Is the sync script public somewhere? I saw the `SYNCING` file mention `git-cvsimport(1)`, but can't judge how up-to-date that is, since the mentioned [ThomasAdam/tmux-openbsd](https://github.com/ThomasAdam/tmux-openbsd) repository returns a 404.
In a test I did with `git-cvsimport`, the trailer worked without problems.

Another approach I see that from my perspective is more work though, Is somehow managing to create the cvs commit with a user named like the Github username and using `git-cvsimport`'s `author-conv-file`, so that the commits have the proper author identity. But again, this seems like a lot more work, which is really not my intention for you.

> But why is this so important? I'll make sure you are credited, what difference does the form make?

I would argue that Github is (today) the main way for the average person to interact with the tmux repository (Which I see validated through the amount of commits mentioning Github issues). And seeing it then [list](https://github.com/tmux/tmux/graphs/contributors) only 6 people as contributors feels a bit unjust to the >300 people that have contributed to this project. Especially since in my opinion proper credit is import and makes sense in the space of FOSS. Obviously all made commits are this way now, but I feel like it would never be too late to start with a workflow that credits a bit better.

I see the trailer as a simple solution that does not really change your workflow, since you're mentioning the names already. Admittedly it's a solution to a propblem that I have, not necessarily you; but I'm also not the first person to bring this up.

nicm:
OK I see what you are saying, but the fact is that tmux is part of OpenBSD and GitHub is the portable version of tmux only, this is made clear in the notes for contributors. Until that changes, anyone who contributes to OpenBSD (through tmux or directly) must be aware that the source tree is CVS and that credit is given in the commit message text and not using any decorations Git may offer - which are not necessarily the right or even the best form of "proper credit" in any case. tmux is not the only project to be hosted elsewhere and with only a source code mirror and issue tracking on GitHub.

I don't control GitHub or what they choose to show and I'm not under any obligation to cater to what they want. Don't forget GitHub is a commercial product and it is entirely possible that eventually it will (like SourceForge) stop being a good place to host open source projects. I already mention GitHub issue numbers in commit messages, and I can't say I'm wild about that, given that they will become meaningless if we move, although the benefits do outweigh the potential downsides.

I don't mind if I have to make sure commits end in some *readable English* text matching "From User Name username@email [in GitHub issue 1234]" and the script then translates that to some Git command. @ThomasAdam would need to weigh in on if that is possible/desirable since I don't maintain the sync scripts (I have a feeling they are in the repository? There is a `SYNCING` file anyway).

Some projects maintain a file listing contributors, I never did that partly because I never started and partly because it is redundant with the commit logs. But I don't mind having one and I'll keep it up to date as far as I remember, but someone would need to go through the commit logs and generate it to start with.

nicm:
Two small comments on your diff:

* You can't use `display-menu` without arguments, so `name [key] [command] ...` should probably be `name key command [name key command ...]`?

* `set-buffer` requires an argument as well so `[data]` should just be `data`.

ThomasAdam:
Hi,

@druckdev -- I can see your point, however, there isn't anything I can do to change this in -portable.

Because tmux is developed in-tree of OpenBSD's src repository, the authorship is going to be what CVS records.  As you'll note, nicm is mostly the sole author, along with a few others with commit-bit access for updating documentation, etc.

When it comes to external contributors, they won't have commit rights to CVS in OpenBSD -- and the convention there has always been attribution happens in the commit message; tmux is no different in this regard.

One thing I tried doing very early on with `git-cvsimport` was maintaining a list of CVS names to user name.  I can't remember why I dropped this, but I recall it being a bit more work than I was willing to maintain.  But we could -- for CVS committers -- maintain a .mailmap entry.

So the best we're going to come up with is maintaining a list of contributors through their attribution for non-CVS committers.  We could automate this partially through something like:

```
git log --no-merges --pretty=format:%s | perl -wlne '/from\s+(.*)\.??$/i and print $1;' | sort | uniq -c | sort -nr
```

But note that this is something which needs to be manually cleaned up a lot, and then perhaps an agreement with @nicm about a format on commit messages in the future such that it could be  updated automatically?

Either way, there's some overhead to this -- and I'm not sure I personally want to maintain this.

druckdev:
Thanks for your detailed reply!

At first, just so we're on the same page in case we haven't been; this is how a commit like this would look like:
https://github.com/druckdev/tmux/commit/c17e370bebbf6c559a4115d853ef6ca8c11749d8

> I don't control GitHub or what they choose to show and I'm not under any obligation to cater to what they want. Don't forget GitHub is a commercial product and it is entirely possible that eventually it will (like SourceForge) stop being a good place to host open source projects.

I agree full-heartedly. I'm not a fan of Microsoft at all and would prefer a different host any time, but I can't deny the fact that this is were (most of) the industry is happening right now. But the trailer is also nothing that is Github specific. It is widely used in the linux kernel and git/git and is a feature that is also supported by other hoster, at the least I know of Gitlab. So I don't necessarily see this as solution that is not future-proof.

> I don't mind if I have to make sure commits end in some readable English text matching "From User Name username@email [in GitHub issue 1234]" and the script then translates that to some Git command

If the *readable* is the key breaking point of the trailer, then I understand and can't really say anything against that. I'd happily help in extending support in any script, if wanted.

druckdev:
Thanks for the reply @ThomasAdam!

> So the best we're going to come up with is maintaining a list of contributors through their attribution for non-CVS committers. We could automate this partially through something like:
>
> ```
> git log --no-merges --pretty=format:%s | perl -wlne '/from\s+(.*)\.??$/i and print $1;' | sort | uniq -c | sort -nr
> ```

To gather the >300 number I used:
```
git log --oneline | grep -o 'from [A-Z][^ ]* [A-Z][A-Za-z]*' | sort | uniq -c
```

which seems to not suffer from false-positives, but probably some false-negatives.

> But note that this is something which needs to be manually cleaned up a lot, and then perhaps an agreement with @nicm about a format on commit messages in the future such that it could be updated automatically?
>
> Either way, there's some overhead to this -- and I'm not sure I personally want to maintain this.

Indeed that sounds like a bit of work, but might be doable and would probably get easier with a set format. If you define a fixed format for the credit in commit messages though, I don't really see why one could not just use the trailer then, since one could get the list with one git command without the need for any post-processing (with the added benefit, that any hoster that supports this would properly attribute the commit to the right person):

```
git log --format="%(trailers:key=Co-authored-by,valueonly=true)" | sed '/^$/d'
```

nicm:
I think it's better to not try to include those two cases, the intent of usage is not to be strictly accurate for every case but to (hopefully) make the most common cases more obvious.

druckdev:
Understood. I'll revert it for `set-buffer` and switch to your suggested format for `display-menu`

nicm:
Greatv,the problemis if try to reflect everything for complex commands it just gets totally confusing. Another option is to list multiple usages for a command but that only works in the man page... I don't think it is worth it for these.

druckdev:
I agree that multiple usages would probably not be worth it.

As a last suggestion, to reflect that `key` and `command` should be omitted if `name` is empty:
```
name key command [name [key command] ...]
```

(my previous version also did this wrong, since it suggested that they can be omitted independently)

druckdev:
Also because having a separator line at the beginning probably doesn't really make sense

ThomasAdam:
> I'd happily help in extending support in any script, if wanted.

The issue isn't in writing the script -- I am more than capable of doing that, but whether all of this work is worth it in the end.

I don't believe it is.

Software doesn't exist in a vacuum filled with "git"; and to try and coerce that notion just because that happens to be the $TOOL_DE_JOUR seems somewehat short-sighted to me.  Attribution existed in the form OpenBSD via CVS uses, long before git existed, and no one had an issue with it then.

What happens when git is replaced with $SOMETHING_ELSE?  You're going to have a smimilar issue.

So I think I'd like to reserve judgement on this for now.

Apologes -- this thread seems to have drifted between mdoc updates to commands (whatever), and the philosopical point of attribution.

ThomasAdam:
Hi all,

Apologies to hijack this issue, but one thing which I remembered...

I do maintain a `.mailmap` -- the "issue" is that git at the time did not respect the `.mailmap` file by default when running commands such as `git log`.  I have this in my global `.gitconfig` file:

```
git config --global log.mailmap true
```

Which makes most commits appear as the authors from CVS (where those authors have a commit bit).

Whether git's changed such that it can detect a `.mailmap` file, I'm not sure, but enabling the above might make some of this easier.  Of course, you've still got the peripheral issue that not all online "forges" will honour this setting by default.

I know it won't address the other issue of authorship, but as I've said previously, I'm less inclined to do anything about that.

druckdev:
> The issue isn't in writing the script -- I am more than capable of doing that, but whether all of this work is worth it in the end.
>
> I don't believe it is.

Yes, I understand. And very sorry, I didn't mean to question your abilities!

> Software doesn't exist in a vacuum filled with "git"; and to try and coerce that notion just because that happens to be the $TOOL_DE_JOUR seems somewehat short-sighted to me. Attribution existed in the form OpenBSD via CVS uses, long before git existed, and no one had an issue with it then.
>
> What happens when git is replaced with $SOMETHING_ELSE? You're going to have a smimilar issue.
>
> So I think I'd like to reserve judgement on this for now.

Fully understandable and your right to do. I'd probably have more to say to this, but I don't see value in dragging this out.

> Apologes -- this thread seems to have drifted between mdoc updates to commands (whatever), and the philosopical point of attribution.

Thank you both (@nicm & @ThomasAdam) for being open to this discussion and explaining your points of view!

druckdev:
Also small ping @nicm:

> As a last suggestion, to reflect that key and command should be omitted if name is empty:
>
>     name key command [name [key command] ...]

What do you think of this?

druckdev:
> I do maintain a `.mailmap` (it's in the -portable repo) -- the "issue" is that git at the time did not respect the `.mailmap` file by default when running commands such as `git log`. I have this in my global `.gitconfig` file:
> 
> ```
> git config --global log.mailmap true
> ```
> 
> Which makes most commits appear as the authors from CVS (where those authors have a commit bit).
> 
> Whether git's changed such that it can detect a `.mailmap` file, I'm not sure, but enabling the above might make some of this easier. Of course, you've still got the peripheral issue that not all online "forges" will honour this setting by default.

Thanks for mentioning this and FYI: yes, git turned this on by default in v2.23.0.

nicm:
> What do you think of this?

This looks good to me now with a couple of minor fixes in the man page (one "new sentence, new line" and one "it's" -> "its"). Applied to OpenBSD now, will be in GitHub later. Thanks!

PR #4420, opened 2025-03-21
Issue 4412 choose tree format prints wrong values

mgrant0:
Fixes bug that caused the panes not to show in the choose-tree output.  Also fixes bug which caused the `pane_pid` to be incorrect in the formatted data.


mgrant0:
This is still in testing.

nicm:
Applied now, thanks!

PR #4422, opened 2025-03-21
Add new default-client-command option

### LINKED ISSUE #4416: Suggestion: different default command on client startup
### Issue description

From the man page on what tmux does if no commands are specified:

https://github.com/tmux/tmux/blob/3eb93383a3302bda0ec0ba509ce47d92e31b354a/tmux.1#L257-L259

Personally, I'd rather have it run `new-session -A` so that I don't have to think about whether there's already a session or not.

Do you think changing the default to `new-session -A` makes sense?

Relatedly, what do you think about adding a new option so that users can choose their own default?

If you think either or both is a good idea, I can work on PR(s). Is this the right place in the code to start?

https://github.com/tmux/tmux/blob/3eb93383a3302bda0ec0ba509ce47d92e31b354a/server-client.c#L3470-L3474

### Required information

* tmux version (`tmux -V`). tmux 3.5a, but I was also looking at git master's code
* $TERM inside and outside of tmux (`echo $TERM`). N/A, I think
* Logs from tmux (`tmux kill-server; tmux -vv new`). N/A, I think

dseomn:
Fixes #4416


nicm:
This looks perfect, I'm not wild about the name of the option but I can't think of a better choice so we'll just go with it. Applied to OpenBSD now, will be in GitHub later. Thanks!

dseomn:
Thanks! And yeah, I couldn't think of a great name for the option either. Feel free to change it if you think of something better later.

PR #4435, opened 2025-03-26
Move cgroup dbus requests to the child

danieldg:
This avoids a race where a spawned child that quickly forks will have only the parent process moved to the newly created cgroup, leaving the early children in tmux's own cgroup.  It also avoids problems if the spawned process inspects or changes its own cgroup.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -450,7 +450,7 @@ void		*recallocarray(void *, size_t, size_t, size_t);
 /* systemd.c */
 int		 systemd_activated(void);
 int		 systemd_create_socket(int, char **);
-int		 systemd_move_pid_to_new_cgroup(pid_t, char **);
+int		 systemd_move_to_new_cgroup(char **);
 #endif
 
 #ifdef HAVE_UTF8PROC
--- a/compat/systemd.c
+++ b/compat/systemd.c
@@ -76,7 +76,7 @@ fail:
 }
 
 int
-systemd_move_pid_to_new_cgroup(pid_t pid, char **cause)
+systemd_move_to_new_cgroup(char **cause)
 {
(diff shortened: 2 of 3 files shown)


danieldg:
After doing further testing, there is still a race, but it's due to the fact that systemd doesn't do the cgroup movement before replying to the dbus method.  This will need to use dbus to wait for the unit to be started to fully fix the issue.

nicm:
What does all this JobRemoved stuff do?

danieldg:
The full documentation for the systemd dbus api is at https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.systemd1.html. Requests such as `StartUnit` and `StartTransientUnit` will return after enqueuing a job, issuing a job ID.  The actual work of the job (in this case, actually assigning the process to a new cgroup) is done asynchronously. The `JobRemoved` signal is sent on job completion, and is the recommended way to wait for systemd to actually do something.

If systemd was used to actually launch the process instead of just to assign it a cgroup, tmux would not need to wait for the JobRemoved signal - however, this would mean that tmux is not actually the parent process of the window's main pid, which would mess up the ability to `wait` for windows and more - so I wouldn't suggest doing that.

nicm:
Looks OK to me but can you make a few minor changes:

- Can you rename `job_watch` to `systemd_job_watch`?

- Please make every line less than 80 columns.

- Please remove `{}`s around single line ifs.

- Use `const char *path;` not `const char* path;`.

Normally I'd do this since it is quick but I have no Linux box to build the systemd code so it is better if you do it :-). Thanks!

nicm:
Applied now, thanks!

PR #4444, opened 2025-03-29
mode-tree: Only align window and pane indexes

### LINKED ISSUE #4442: choose-tree: misaligned session names
This was introduced in #4360 to align window numbers but it doesn't play nice with custom named sessions.

example (sorry it's a screenshot, idk how to copy the output of choose-tree):

Image

Notice how session-names are right-aligned (it was left-aligned before #4360). This makes it hard to know which session the window belong to.

dseomn:
Fixes #4442

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -102,6 +102,7 @@ struct mode_tree_item {
 
 	int				 draw_as_parent;
 	int				 no_tag;
+	int				 align;
 
 	struct mode_tree_list		 children;
 	TAILQ_ENTRY(mode_tree_item)	 entry;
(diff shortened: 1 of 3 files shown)


mode-tree.c
```diff
@@ -776,8 +784,8 @@ mode_tree_draw(struct mode_tree_data *mtd)
 		else
 			tag = "";
 		xasprintf(&text, "%-*s%s%*s%s%s", keylen, key, start,
-		    namelen[line->depth], mti->name, tag,
-		    (mti->text != NULL) ? ": " : "" );
+		    names_all_digits[line->depth] ? namelen[line->depth] : 0,
```
dseomn, line 787: I think if I changed this to `(names_all_digits[line->depth] ? 1 : -1) * namelen[line->depth]`, it would align left, but still pad to the full width. Would that be better? I never really use more than one session, so it's not something I've though about much.

nicm:
I think it would be better to add a `no_pad` or `no_align` flag or something, like `no_tag` and `draw_as_parent`, and have `window-tree.c` call it on sessions but not windows and panes?

dseomn:
Or maybe `int align` that can be set to -1 for left, 0 for none (default), or 1 for right? I think that would be very slightly easier to implement. I do like the idea of adding something to `struct mode_tree_item` instead of guessing.

nicm:
Yes that sounds good.

nicm:

PR #4446, opened 2025-03-30
Explain what it means when an issue is closed

dseomn:
When looking through various issues, I think I've seen a few times where I think people misunderstood the intent of closing an issue. Maybe this (or something similar) will help that happen less often? I'm not sure though, and I don't have any experience running an issue tracker for a large open source community, so feel free to ignore this if this won't help.


nicm:
Applied now, thanks!

PR #4450, opened 2025-04-01
Issue 4330 popup window should not be draggable while mouse still pressed

mgrant0:
Fix for issue #4330


nicm:

PR #4451, opened 2025-04-01
Extend format conditionals

### LINKED ISSUE #4438: suggestion: extended conditionals in formats
### Issue description

The current conditional format takes 3 arguments and behaves like an if-else statement:

https://github.com/tmux/tmux/blob/9e1f110db0776ff0ca3f1b41ee3a6c8ccaf6452b/tmux.1#L5595-L5614

This leads to code like this that nests multiple conditionals:

https://github.com/tmux/tmux/blob/9e1f110db0776ff0ca3f1b41ee3a6c8ccaf6452b/window-tree.c#L39-L57

What do you think about extending conditionals to take any number of arguments >= 2, such that:

* The first two arguments are a condition and an if-true value like the current behavior.
* For each additional pair of arguments, the first of the pair is an `else if`-style condition and the second is the value if that condition is true.
* If the number of arguments is even and none of the conditions matches, the result is the empty string.
* If the number of arguments is odd and none of the conditions matches, the result is the last argument.

So instead of `#{?pane_format,...,#{?window_format,...,...}}`, it would be possible to do `#{?pane_format,...,window_format,...,...}`. And it would be backwards compatible.

### Required information

* tmux version (`tmux -V`). I looked at tmux.1 on master to see if it was still the same
* Platform (`uname -sp`). N/A
* $TERM inside and outside of tmux (`echo $TERM`). N/A
* Logs from tmux (`tmux kill-server; tmux -vv new`). N/A

dseomn:
1. Add support for `else if`, so `#{?cond1,value1,#{?cond2,value2,else-value}}` can be changed to `#{?cond1,value1,cond2,value2,else-value}`
2. Add default empty string if there's no `else` value, so `#{?cond1,value1,}` can be changed to `#{?cond1,value1}`

This has some degenerate cases that I don't think are particularly useful, but I also don't think it's worth adding code to make them errors:

```
$ tmux display-message -p '#{?}'

$ tmux display-message -p '#{?foo}'
foo
```

And the more useful cases:

```
$ tmux set @true 1
$ tmux set @false 0
$ tmux display-message -p '#{?@true,foo}'
foo
$ tmux display-message -p '#{?@false,foo}'

$ tmux display-message -p '#{?@true,foo,bar}'
foo
$ tmux display-message -p '#{?@false,foo,bar}'
bar
$ tmux display-message -p '#{?@false,foo,@true,bar}'
bar
$ tmux display-message -p '#{?@true,foo,@true,bar}'
foo
$ tmux display-message -p '#{?@false,foo,@false,bar}'

$ tmux display-message -p '#{?@false,foo,@false,bar,default}'
default
```

And expansion still works:

```
$ tmux set @foo bar
$ tmux display-message -p '#{?#{@true},#{@foo}}'
bar
$ tmux display-message -p '#{?#{@false},foo,#{@foo}}'
bar
```

Fixes #4438

### DIFF
--- a/format.c
+++ b/format.c
@@ -4518,7 +4518,8 @@ format_replace(struct format_expand_state *es, const char *key, size_t keylen,
 {
 	struct format_tree		 *ft = es->ft;
 	struct window_pane		 *wp = ft->wp;
-	const char			 *errstr, *copy, *cp, *marker = NULL;
+	const char			 *errstr, *copy, *cp, *cp2;
+	const char			 *marker = NULL;
 	const char			 *time_format = NULL;
 	char				 *copy0, *condition, *found, *new;
 	char				 *value, *left, *right;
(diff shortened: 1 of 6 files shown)


nicm:
This looks good to me, thanks.

Do you want to change any of the default option values to use this?

Does this need changes (or additions?) to `regress/format-strings.sh`?

dseomn:
> Do you want to change any of the default option values to use this?

Sure. I changed the ones that looked like they'd benefit the most from it. I didn't change all the `#{foo,bar,}` to `#{foo,bar}` because the benefit of those changes seemed dubious and I didn't feel like testing all of them to make sure I didn't mess them up by accident.

> Does this need changes (or additions?) to `regress/format-strings.sh`?

Yes, yes it did. Done.

nicm:

PR #4455, opened 2025-04-03
Assorted changes to make bash completion easier

dseomn:
I'm working on a bash completion script to include in the bash-completion repo itself. To (hopefully) reduce the maintenance burden, I'm trying to parse tmux's usage output as much as possible rather than hard-code any commands or options. These are some changes that I think will make that easier.

* Change `-T type` to `-T prompt-type` in usage output to match the man page.

* Change `-c confirm_key` to `-c confirm-key` to match the man page.

* Make environment variable names consistent with `variable` instead of a mix of `variable` and `name`.

* Add a documented way to get the top-level usage message, instead of needing to rely on an error of some sort.

I can split this up if you want some of those but not others. Or feel free to split it if that would be easier.


nicm:
These all look good to me, will apply them soon.

nicm:

PR #4463, opened 2025-04-11
Allow making key bindings repeatable without a command

druckdev:
This makes it possible to make the default bindings repeatable without knowing their command.

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -201,6 +201,8 @@ key_bindings_add(const char *name, key_code key, const char *note, int repeat,
 				free((void *)bd->note);
 				bd->note = xstrdup(note);
 			}
+			if (repeat)
+				bd->flags |= KEY_BINDING_REPEAT;
 		}
 		return;
 	}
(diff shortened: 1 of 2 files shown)


key-bindings.c
```diff
@@ -197,11 +197,12 @@ key_bindings_add(const char *name, key_code key, const char *note, int repeat,
 	bd = key_bindings_get(table, key & ~KEYC_MASK_FLAGS);
 	if (cmdlist == NULL) {
 		if (bd != NULL) {
-			free((void *)bd->note);
-			if (note != NULL)
+			if (note != NULL) {
+				free((void *)bd->note);
 				bd->note = xstrdup(note);
```
druckdev, line 202: One could theoretically check here if `note` is empty and in case it is, save the `strdup` and just set `NULL`, as `list-keys` skips empty notes either way.

Since this wasn't done previously here and below, I left it as is for now. Is there any reason to keep the empty string, that I'm missing?

nicm:
I think the extra free/strdup is fine, this is not something that needs optimization. Applied to OpenBSD now, will be in GitHub later. Thanks!

druckdev:
Ok. Thanks as well!

PR #4469, opened 2025-04-16
Fix description of `pane_in_mode`

druckdev:
Since 3f6bfbaf2bab ("Allow multiple modes to be open in a pane...") `pane_in_mode` is not just a boolean, but reflects the number of modes a pane is currently in.


nicm:

PR #4470, opened 2025-04-16
Command `trigger-bind`, to trigger an existing `bind-key`

kkovacs:
Hi @nicm, Kristof here -- we actually worked on one tmux thing before together, _"send-keys, but from a process"_ (issue number 1186), and now after 6 years I have _another_ new idea :)

I'm trying to build a nested `display-menu`.

I have a quite complex menu setup already, which [you can see here]( https://github.com/kkovacs/kkrc/blob/e150d37a35d223a2feff5a093c86e8c6c4f00181/.tmux.conf#L414) . I'm already over-abusing `command-alias` to make it a little more livable. :)

I know in theory I can nest `display-menu` items, but with the quoting, that gets ugly VERY fast, especially if I want to make more than two levels.

One solution I can think of is if I had a way to trigger already defined binds. Then I could build something like this:

```
# Submenu 1
bind -T mymenu "1" display-menu -T "Menu One" \
	"Foo" "f" "display-message foo" \
	"Bar" "b" "display-message bar"

# Submenu 2
bind -T mymenu "2" display-menu -T "Menu Two" \
	"Baz" "b" "display-message baz" \
	"Qux" "q" "display-message qux"

# MAIN menu
bind "X" display-menu -T "MAIN MENU" \
	"One" "1" "trigger-bind -T mymenu 1" \
	"Two" "2" "trigger-bind -T mymenu 2"
```

But I found no way to trigger a bind (`send-keys` doesn't work for that), so I created a `trigger-bind` command that does exactly that. It seems to work (the above code actually works with it).

```man
       trigger-bind [-n] [-T key-table] key
               Trigger  the  command  bound  to key as if the key had been pressed.
               By default, the key is looked up in the prefix table.  If -n is given, the root
               table is used instead.  If -T is given, the key is looked up in key-table.
               If the key is found and has a command list bound, the command list is executed.
```

If you would be so kind to accept it, please look at it with a _very_ sharp eye -- I haven't written C in many years, I'm not familiar with tmux's code! I just hope there are no huge mistakes in there.

Thanks in advance, and keep up the GREAT work you are doing with tmux -- we all owe you big time!

Kristof

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -140,6 +140,7 @@ dist_tmux_SOURCES = \
 	cmd-swap-pane.c \
 	cmd-swap-window.c \
 	cmd-switch-client.c \
+	cmd-trigger-bind.c \
 	cmd-unbind-key.c \
 	cmd-wait-for.c \
 	cmd.c \
--- a/cmd-trigger-bind.c
+++ b/cmd-trigger-bind.c
@@ -1,0 +1,101 @@
+/* $OpenBSD$ */
+
+/*
+ * Copyright (c) 2024 Nicholas Marriott <nicholas.marriott@gmail.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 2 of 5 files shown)


nicm:
Isn't this already `send -K`?

In any case I think this should be a flag to `send-keys` not a new command.

kkovacs:
I'm closing the PR then, since you're right, `send-keys -K` is perfect for this, and is already in. Thanks for the tip, and keep up the good work!

nicm:
If you want to specify a key table instead of using the client's defa

PR #4472, opened 2025-04-17
Strip trailing slash from «new -c ""» $PWD

### LINKED ISSUE #4471: Trailing slashes in `$PWD` when running `new-window -c ""`, which happens for config in wiki
Since commit 95baa32383bf095daf1a1e707de6b540ddb05d37 and until recent master, there is the following discrepancy (after starting tmux):

- If you type `prefix : new-window dash<Enter>` and then type `echo $PWD`, you get a path like `/home/ilyagr`, without a trailing slash. This is the correct behavior.
- If you type `prefix : new-window -c "" dash<Enter>` and then type `echo $PWD`, you get `/home/ilyagr/`, with a trailing slash. This is wrong and messes up shell prompts (see below).

For a more mechanical reproduction, you can run:

```console
$ ./tmux -L dev new-session -c "" env HOME=$(mktemp -d) dash -c '{ echo PWD "$PWD" ; pwd; sleep 0.1; } > /tmp/buglog'; cat /tmp/buglog
PWD /home/ilyagr/dev/tmux/
/home/ilyagr/dev/tmux/
[exited]
```

and compare this with

```console
$ ./tmux -L dev new-session env HOME=$(mktemp -d) dash -c '{ echo PWD "$PWD" ; pwd; sleep 0.1; } > /tmp/buglog'; cat /tmp/buglog
PWD /home/ilyagr/dev/tmux
/home/ilyagr/dev/tmux
[exited]
```

Note that, since this uses `new-session`, it does not necessarily take you to the home dir.

As I said, I bisected this to commit 95baa32383bf095daf1a1e707de6b540ddb05d37 from a couple of years ago. Before that commit, you'd get:

```console
$ # At commit ead75c2d
$ ./tmux -L dev new-session -c ""  env HOME=$(mktemp -d) dash -c '{ pwd; echo PWD "$PWD" ; sleep 0.1; } > /tmp/buglog'; cat /tmp/buglog
/home/ilyagr
PWD /home/ilyagr
[exited]
$ ./tmux -L dev new-session  env HOME=$(mktemp -d) dash -c '{ pwd; echo PWD "$PWD" ; sleep 0.1; } > /tmp/buglog'; cat /tmp/buglog
/home/ilyagr/dev/tmux
PWD /home/ilyagr/dev/tmux
[exited]
```

I believe (but haven't tested) that `prefix : new-window dash<Enter>` and `prefix : new-window -c "" dash<Enter>` would work the same at that commit.

### Why this matters/how I came up onto this

I set up my config as recommended in [the wiki](https://github.com/tmux/tmux/wiki/Recipes#create-new-panes-in-the-same-working-directory):

```
bind '"' split-window -c "#{pane_current_path}"
bind % split-window -hc "#{pane_current_path}"
bind c new-window -c "#{pane_current_path}"
```

It turns out that, on my Mac, if a pane is running `sudo sshd -De`, and I press `prefix c`, `#{pane_current_path}` expands to the empty string (not sure why), and `new-window -c ""` gets called.

I noticed that, in this case, my shell prompt would show `~/` as my dir instead of the usual `~`. This happens in bash 5.2.37(1) and in fish up to commit https://github.com/fish-shell/fish-shell/commit/5076cfbd7193e8dd61cec49a0587600386cc939d from yesterday (thanks @krobelus !).

I'm not sure whether this has other implications; I couldn't find anything about trailing slashes in <https://pubs.opengroup.org/onlinepubs/9799919799/utilities/pwd.html>, but I haven't looked very hard in other parts of POSIX or other standards. The inconsistency in tmux behavior is annoying anyway.

### Workarounds

1. In the wiki, replacing the commands with the likes of

```
bind c new-window -c  "#{?pane_current_path,#{pane_current_path},#{HOME}}"
```

should work. 

2. In at least `bash` and `fish`, putting `cd "$PWD"` in the shell init script should also fix the problem. And after https://github.com/fish-shell/fish-shell/commit/5076cfbd7193e8dd61cec49a0587600386cc939d, `fish` works around this automatically.

### Additional weirdness
This is more of a curiosity, but I find it a bit confusing and it does not seem to be documented.

Running `cd /tmp; tmux new-window` or `cd /tmp; tmux new-window -c ""` creates a new window starting in `/tmp`, at least on the latest master (and one of those demos the bug the rest of the issue described). On the other hand, typing `cd /tmp<Enter> prefix : new-window<Enter>` or binding `new-window` to a key takes you to HOME.

Ideally, the docs would explain what happens for `new-window` without `-c` or with empty `-c`

krobelus:
Setting an empty (relative) working directory will
export $PWD with a trailing slash:

	tmux new-session -c "" bash -c 'echo $PWD; read'

Fix this.

Do not touch slashes added by the user.
For example this command still outputs three slashes:

	tmux new-session -c "///" bash -c 'echo $PWD; read'

I checked other occurrences of %s/%s and server_client_get_cwd
and didn't find other affected cases.

There may be one if client_file.path can have a trailing slash.
I'm guessing that even if it did, it probably doesn't leak into the UI.

Fixes #4471

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -229,8 +229,9 @@ spawn_pane(struct spawn_context *sc, char **cause)
 	if (sc->cwd != NULL) {
 		cwd = format_single(item, sc->cwd, c, target->s, NULL, NULL);
 		if (*cwd != '/') {
-			xasprintf(&new_cwd, "%s/%s", server_client_get_cwd(c,
-			    target->s), cwd);
+			xasprintf(&new_cwd, "%s%s%s",
+			    server_client_get_cwd(c, target->s),
+			    *cwd != '\0' ? "/" : "", cwd);
 			free(cwd);
 			cwd = new_cwd;
 		}


nicm:
How about this instead of the switch? Bit shorter:
~~~~
--- spawn.c     10 Jul 2023 09:24:53 -0000      1.33
+++ spawn.c     22 Apr 2025 13:00:55 -0000
@@ -231,8 +231,9 @@ spawn_pane(struct spawn_context *sc, cha
        if (sc->cwd != NULL) {
                cwd = format_single(item, sc->cwd, c, target->s, NULL, NULL);
                if (*cwd != '/') {
-                       xasprintf(&new_cwd, "%s/%s", server_client_get_cwd(c,
-                           target->s), cwd);
+                       xasprintf(&new_cwd, "%s%s%s",
+                           server_client_get_cwd(c, target->s),
+                           *cwd != '\0' ? "/" : "", cwd);
                        free(cwd);
                        cwd = new_cwd;
                }

krobelus:
yes that's much simpler

nicm:

PR #4478, opened 2025-04-20
add note for window and pane menus

tomis007:
I noticed that the default key bindings for tmux include an option to show a window and pane menu with `prefix >` and `prefix <`. I could not find them though in the key mapping list in `prefix ?`. I added a note to the key bindings so that now they show up as expected in `prefix ?` with a default config. Let me know if you would like any changes or other documentation elsewhere! Thanks


nicm:

PR #4481, opened 2025-04-21
Fix border/scrollbar gllitch with swapping panes when scrollbars enabled

mgrant0:
Add calls to `layout_fix_panes` after a pane swap to fix case if one of the panes is in alternate screen mode and had a scrollbar.


nicm:

PR #4490, opened 2025-04-26
Status line2 sessions

mgrant0:


nicm:
I don't think we should add `session_index`, we should not give the impression this is the order of sessions or that they can be reordered. If you must have the ID without `$`, use the `s` modifier to remove it, although it would be better just to show the name.

You could add a flag to `S` to select the sort order like `S/a` for alpha sort by name, `S/i` for ID sort, maybe others? Like in choose mode.

nicm:
I don't think we should add any more style options that are redundant with other options.

nicm:
You should be able to use line drawing characters by using the `acs` and `noacs` style attributes.

nicm:
See `tty-acs.c` for what is what, for example `q` is "horizontal line".

nicm:
Sort order is part of the RB tree that holds the sessions, if you want to sort a different way you will need to add a parallel tree with a different sort order, or copy the list of sessions into a temporary buffer and `qsort` it. But the default sort order should probably be the same as `choose-tree` (is it ID or name by default?).

mgrant0:
> I don't think we should add `session_index`, we should not give the impression this is the order of sessions or that they can be reordered. If you must have the ID without `$`, use the `s` modifier to remove it, although it would be better just to show the name.

ok agreed.  `session_index` is not really necessary.  `#S` prints the ID without the `$` (or the name if named).  

> You could add a flag to `S` to select the sort order like `S/a` for alpha sort by name, `S/i` for ID sort, maybe others? Like in choose mode.

I see how to do this.  It adds some complexity, especially if I were to try and match all the sort orders of `window-tree`.  

How about if I stick only to index order for now.  If you think adding `S/a` and `S/i` and other flags, I can do that, and do that for all of `#W`, `#S`, and `#P` as well later.

I see why the order of the sessions move around now.  In `cmd-rename-session.c` there is:
```
        RB_REMOVE(sessions, &sessions, s);
        free(s->name);
        s->name = newname;
        RB_INSERT(sessions, &sessions, s);
```
Why is it necessary to remove and re-insert the session if it's just being renamed?  I tried to remove the `RB_REMOVE` and `RB_INSERT` here and it seems to work just fine and the order is now always index order which is what would be expected in my opinion.

mgrant0:
Todo:
1. add `#{session_flags} flag and callback function
2. look into adding activity, bell, last, and silence flags and styling like for window-status
3. look at window panes

nicm:
Isn't the session tree by name not index? That's why it adds and removes.

Can't you just use the window formats and it will give eg the window flags for the current window?

mgrant0:
I am having a hard time following the macro.  I see a `session_cmp` function defined and it does indeed compare `s->name`.  I compiled with `-E` and looked at the preprocessor output.  I see it actually calling ` sessions_RB_REMOVE` with `sessions_` pluriel so I don't exactly see how it works but I do see `RB_GENERATE(sessions, session, entry, session_cmp);` and I see how that creates the other functions.  So yes, I believe you must be correct, it should be sorted by name.  

I propose something like this:
```
        l = NULL;
        n = 0;
        RB_FOREACH(s, sessions, &sessions) {
                                l = xreallocarray(l, n + 1, sizeof *l);
                l[n++] = s;
        }
        window_tree_sort = sort_crit;
        qsort(l, n, sizeof *l, format_loop_cmp_session);

        for (i = 0; i < n; i++) {
                /* all that is in the FOREACH goes here*/
        }
        free(l);
```
In `window_tree_cmp_session`, I see it can sort by index, time, and name.  Unfortunately without some more additional work, I can't use that directly as it depends on stuff inside window-tree.c, so I'll need to essentially copy it.  

I guess I can add a /i, /n, and /t options to those 3 loops.  Mise as well fix it while I'm here.  I think I will default it to the index though so the session doesn't jump around when it's renamed.

> Can't you just use the window formats and it will give eg the window flags for the current window?

But for a session, instead of a single window, shouldn't it be for a whole session or have I misunderstood something?  I see some functions in `session.c` that look like they should be helpful to do that, I think.

nicm:
You might be best to look at https://man.openbsd.org/man3/tree.3 for the tree macros. That looks fine but I would not realloc each time, allocate 100 up front and only add more when that limit is hit. You could add a helper function (`session_build_list` or something) to do this.

Sessions don't have flags or activity or anything like that, so it will just be the flags of the current window?

mgrant0:
I left the `/n` arg in there so you could see it in action.  If you remove it, it sorts by id.

nicm:
Yes, the `S` bits look fine to me.

nicm:
Is this ready to go? What are the `format-draw.c` changes for?

Maybe we should bump the existing `status-format[2]` down to `status-format[3]` so if anyone is using it they can put it back easily?

mgrant0:
Let me do a tiny bit of clean up.

Line 3, yes, my thoughts too.  Especially if we eventually used line 2 for a panes list.  Something like that will be needed for floating panes that are buried. But not right now.

mgrant0:
The changes in `format-draw.c` are really a bug fix.  

What's going on there is there's an extra character that gets added to those 2 format variables.  The `%` and `$`.  They are not displayed when you use `:S` or `:P`, and so the mouse click ends up being off by a single character.  Clicking on the first character of the session actually ends up selecting the session immediatly to the left of it.  Took me several days to track this down.

Should I pull this change out into a separate bugfix PR?

mgrant0:
I've moved it to line 3 now and also I added whatever flags I could easily find that were associated with a session.

In doing this, I wanted to change the bg color for the different status lines individually.  It turns out to be quite messy.  You have to very carefully surgically rewrite each of the status lines.

I think it would be useful to add a `status-line-style` array option similar to the `status-format` array.  I see how to do this.  I think I'd like to do that in a separate PR if you are ok with that.  I know how you hate adding options but I did not find an easier way.

nicm:
Can't you use for example `fill=blue bg=blue` in `status-format`?

mgrant0:
Hmm, closer but not entirely.  

I pre-pended that and also pushed it to default in `status-format[0]` in `options-table.c` so I could experiment with it:

```
#define OPTIONS_TABLE_STATUS_FORMAT1 \
        "#[fill=blue bg=blue]" \
        "#[push-default]" \
        "#[align=left range=left #{E:status-left-style}]" \
        "#[push-default]" \
 ...
```

Before:
image

After:
image

1. It didn't change the colour of the window names
2. It didn't change the colour of the space between the window names
3. It didn't change the colour of the range-right

so I tried to change `window-status-style` and `window-status-current-style` like this:
```
tmux set-window-option -g window-status-style "fill=blue fg=white bg=blue"
tmux set-window-option -g window-status-current-style "fill=blue fg=white bg=black"
```
and here you can see it won't change the colour between the names:
image

I feel like I don't understand what `push-default` is doing or something else in the long `status-format[0]` is re-overridding it back to the old default.  

I did manage to make it change all the parts of the status line by modifying `status_redraw` and adding a call to `style_add` but it's not a simple fix and opens more questions.  

Have any other ideas?

mgrant0:
I went down this rat-hole this morning with the idea of adding a `status-line-style` array just to see if that worked.

Seeing as `status-format` is an array, it felt right to have something like `status-style` an array as well, but that name is already used so I called it `status-line-style`.

I tried inserting a call to `style_add` with a single-value format variable and put the style in that variable and it worked.  I then thought maybe I could modify `style_add` to deal with a name with `[...]` in it and have `options_string_to_style` call `option_array_get` to get the index if so.  But here's where it starts to fall apart.  `options_string_to_style` caches the style.  If it's an array, it means caching an array of styles somehow.  I see arrays of format variables were never quite intended to be used with styles.

Instead of an array, I could just append a number to the name, for example `status-line1-style` and then usse `i` in the loop to `xasprintf` to build the style name and then call `style_add` with that.  This works but felt wrong seeing as `status-format` is an array and these individual formats aren't.  

Before trying to modify `status.c`, I did try just modifying the format in `options-table.c` but I just can't seem to get it to change the underlying default which pokes through.

If I could figure out how to get it to do the right thing inside the `status-format` line, I could even just add it from an array element right there with something like:
```
"#[#{E:status-line-style[0]}]" \
 "#[push-default]" \
...
```

and now we're back to where I am with your fill idea which isn't working here either.  

How do you feel about me adding caching of styles to array format variables?  Or have I missed something here and this already exists or isn't necessary?

I have not traced through the `push-default` code and I don't know how the `fill` attribute works exactly.  Might there be some bug here that this has uncovered?  My guess is it's more likely my lack of understand here.

nicm:
We may need a way to set the default rather than just pushing/popping it.

mgrant0:
I had a quick look at this.  I don't immediatly see what's going on.

If you have further thoughts on this maybe i can dig into it more.  I will have to trace through this in more detail.  Thanks

nicm:
This adds `set-default` to replace the default (current and "pushed") completely with the current style which makes it work: [setdefault.diff.txt](https://github.com/user-attachments/files/20158505/setdefault.diff.txt)

mgrant0:
I tested your diff, it seems to work fine.  I gather you did not want to modify `push-default` because that might break something.

Do you want me to include this diff in this branch or will you do that outside it?

I want to commit my current changes to show you where this stands and discuss a couple other small points.

mgrant0:
In the end I included your diff since this branch depends on it.  I can pull it out if you want.

The bugfix that I did (https://github.com/tmux/tmux/pull/4490/commits/537bc3ce1c2ef43f12fd9a96eb52055e10a3864e), was necessary to properly handle the mouse click on the pane to select the pane in the status line.  I hope that doesn't break anything!  I tested by running things like `tmux switch-client -t '%2'` with `$`, `%`, and `@` prefixes and in all cases it switches to the expected session, pane, or window.

mgrant0:
Try this and you will see what I was aiming for:
```
tmux set -g status 3
tmux set-window-option -g status-style "fg=black,bg=darkgreen"
tmux set-window-option -g window-status-current-style "fg=white,bg=black"
tmux set-window-option -g pane-status-style "fg=black,bg=green,fill=green"
tmux set-window-option -g pane-status-current-style "fg=black,bg=darkgreen"
tmux set-window-option -g session-status-style "fg=black,bg=lightgreen,fill=lightgreen"
tmux set-window-option -g session-status-current-style "fg=black,bg=green"
```
They can look like tabs this way.  Here's the little thing that might be objectional.  I used `pane-status-style` to style the entire line 2, and `session-status-style` to style the entire status line 3, _as well as_ the individual panes and sessions ids/names.  I can remove that and if the user wants to colour each of those status lines they can just pre-pend something like `fill=darkgreen,bg=darkgreen` to the option.  I am ok with that but I wanted your opinion here first.  The thing is, I found that if you want to style the whole line's colours, I ended up setting it to the same thing as the individual panes or sessions.  I don't know if that's generally true or if there would be a reason to want the default for those lines not to be this.  Anyway, I'm ok with removing these from the first line of status lines 2 and 3 if you prefer.

mgrant0:
Sorry for the noise here.  I thought about this overnight and I have reconsidered.  I am going to remove those 2 lines and if the user wants to change the colour of those status lines they can simply prepend something like this to one of the lines:

```
tmux set-window-option -g status-format[1] "#[fill=green,bg=green]#[set-default]`tmux show-option -gv status-format[1]`"
```

mgrant0:
I added 4 option variables which I know you don't want.  I could re-use `window-status-current-style` and `window-status-style` but the items in those other lines are not "windows".  And furthermore, if we do that, there's then no way to style the individual lines without the user putting the contents of the line in their `.tmux.conf` and surgically editing them which from experience is a rather tricky.

If you are very much dead set against adding these 4 new option variables then I will reuse the window-status styles in those lines.  Or let me know if you have a better idea.

And please look at my bugfix several comments above and let me know if that might break something.  Thanks!

mgrant0:
The issue this fixes is an "off-by-one" error because of the prefix character: `$`, `%`, or `@` which is contained in the argument.  When the mouse clicks in the status line, it ends up being off by one character because of this extra undisplayed prefix character.  If you click the mouse on the first character of a session index, it was selecting the session immedietly to the left because it sees that target as one character too wide.

To reproduce this, comment out those changes, then start 2 sessions and carefully click on the first character of the second session and you should see it selects the first instead.

I will propose this change in `format-draw.c` as a separate PR.

mgrant0:
> Can you remove the old OPTIONS_TABLE_STATUS_FORMAT2 instead of commenting it out?

Absolutely, again, this was left for comparision to be cleaned up before final commit.

mgrant0:
> I don't like the acs bits in the new default...

hah!  I put acs characters in there at your bequest!  I totally agree things should work on plain ascii too by the way.  I wondered about this on a larger scale.  I was wondering if there was some sort of fallback if acs wasn't supported.  As in maybe using the vertical bar and minus for horizontal and veritical bars and plus for the elbow.  I believe the window borders seem to do that unless I'm mistaken.

Anyway it was really just decoration to make it clear that the `[n]` referred to the session index.  It can be removed completely or I can use something like:
```
|
`->
```
if you prefer.

mgrant0:
> I think we should try to do without the new options for the moment. If you want them for yourself in your config you can use user options for the same purpose.

This is proving difficult for me without adding the style in some sort of variable.  The issue is these styles are reused multiple times in each line.

Secondly, line 1 already has `window-status-current-style` and `window-status-style`.  I was trying to be consistent with existing practice.

What about deprecating `window-status-current-style` and `window-status-style` and moving them to `@window-status-current-style` and `@window-status-style`?  Perhaps some of the other variables in status line 1 as well?

Towards deprecating these, we could it to first check for `@window-status-current-style`, then `window-status-current-style` then `@window-current-style`, then `window-current-style`, else use `"default"`.  

Here's the thing about the options in the status line formats, these formats are user settable.  Once the user changes the format of those lines, those built-in options can become useless or broken.  So to me you shouldn't be putting built-in options in settable formats like that.  If you take this stance, there are definitely a few option variables that we can deprecate.

nicm:
OK just leave them for now and I'll have another look, maybe it isn't too much.

I don't really think we should use `@` options by default, it seems better to just add normal options in that case.

mgrant0:
> What is the change in server-client.c?

Removing this line:
```
		m->wp = -1;
```

This changes goes with the change in `cmd-switch-client.c`.  If you wipe out the value of `m->wp`, then it's impossible to find the pane in `cmd-switch-client.c`.  Secondly, `m->wp` is already initialised to `-1` in the declarations above so this does not seem to be necessary.

I couldn't see how removing this broke something and it allowed `switch-client` to select panes.  Did I miss something?

mgrant0:
> I don't really think we should use @ options by default, it seems better to just add normal options in that case.

So to be clear, I wasn't proposing prefilling any @ option with something like `"default"`, only just referencing it in the status format.  If undeclared, it seems not to hurt, it just returns empty like an empty shell environment variable.  If the user wanted to style that thing in the format they could create that user variable.  That was my thinking.

tmux generates an error when parsing the format with a non-existent option variable, but it seems fine when parsing the format with a non-existent `@` user option.

I agree it's not perfect solution to add more options.

mgrant0:
I removed the acs arrow now.  I tried to find something in plain ascii that did the trick but nothing worked well so I removed the arrow all together.

Now that sessions are in line 3, it doesn't really make any sense to show `[session]` in line 1 when status > 2.  So I added a condition to `status-left` not to show the session if line 3 is shown.  If you hate this, I will remove this condition.

mgrant0:
I pulled out the `format-draw.c` changes https://github.com/tmux/tmux/pull/4515.

I pulled out the `format.c` changes into https://github.com/tmux/tmux/pull/4516/.  I hope I didn't miss something doing that!

mgrant0:
Here is what it looks like with a single status line (the default).  Nothing should have changed if using a single status line:

image

Here is what it looks like setting `tmux set -g status 2`:

image

Here is what it looks like setting `tmux set -g status 3`:

image

mgrant0:
How does this look?  Anything else needed on this?

nicm:
The code looks good, but:

- Can you remove the format.c stuff for now and we'll do it in the other MR?

- This changes `switch-client` behaviour. It is intended to default to sessions. Try this:
~~~~
tmux new \; new -d \; splitw
tmux switch -t1
~~~~
This should switch to the session named "1" but with your change it switches to pane 1 instead. I should need to do `.1` or `:.1` for pane 1.

I don't think the default status line 2 and 3 are very clear: 

- The leading space in front of the pane and session list looks odd. Either there should be no space or it should be lined up with the first window? Or maybe it would be useful to put `panes:` and `sessions:` at the left of the defaults to make clear what they are showing?

- When there are 3 status lines, the session name mysteriously disappears from the first status line to be replaced by a single space. The status line content should not depend on how many there are.

- Maybe the pane title would be more useful than the size?

mgrant0:
On the format of the status lines, all suggestions appreciated here.  

I will get rid of the staggered indent.

What about `P:` and `S:` at the left of lines 2 and 3? 

Regarding the `[WxH]` after the pane id, I didn't add that, that was there already in status line 2.

I had also thought of using the title.  The thing about the pane title is some people put long things in there like the current directory.  It will end up being truncated in the status line.  Probably this is a bad idea.  Maybe add a `pane_name` attribute to `window_pane` like sessions and windows?  I must admit even though i am not crazy about the `[WxH]` it does make it obvious it's a pane.  I don't know what else might be better to be honest.

The mysteriously disappearing session name in line 1, I will put it back.  My thinking was when status line 3 with the sessions displayed that they wouldn't need the `[session]` in line 1.  I had an arrow in there at one point to make it obvious this was the session name.  Ascii arrows just do not look good in all fonts.  It will be a little redundent but I guess that's fine.

Hopefully we can find something reasonable.  I am working on the other issues you brought up.  Thanks.

nicm:
OK I think WxH is fine then, people can always change it.

P: or (P) or something would be fine. I guess it'd be complicated to make it line up always since the session name could be any length, but spaces just looks a bit odd.

I don't think the status lines should be interdependent by default, so if you add status line 2 or 3 it shouldn't change status line 1. If people want that the can configure it easily enough.

mgrant0:
I have backed out changes to `format.c` (they are in the other PR) and `switch-client`.  The change to `switch-client` seems like it wasn't necessary and again, i may have made that change chasing a bug that I caused with a messed up format.  So this is good news.  One less change.

I made the changes to the default status lines.  Currently there is a single space before `P:` and `S:` , really don't mind, let me know if you prefer one or the other:

image

or

image

nicm:
Now the mouse works for me to change session, but not to change pane. Maybe that was why you changed `switch-client`?

How about this change to the default status lines? I don't think we will get anything perfect, most people will want to customize it.
~~~~
--- tmux3/options-table.c       Wed Jul  2 09:27:45 2025
+++ tmux1/options-table.c       Wed Jul  2 09:20:58 2025
@@ -174,7 +174,7 @@
        "#[pop-default]" \
        "#[norange default]"
 #define OPTIONS_TABLE_STATUS_FORMAT2 \
-       "#[align=left] P: "          \
+       "#[align=left][#{n:#{P:x}} panes] " \
        "#[norange default]" \
        "#[list=on align=#{status-justify}]" \
        "#[list=left-marker]<#[list=right-marker]>#[list=on]" \
@@ -199,7 +199,7 @@
                "#[norange list=on default] " \
        "}"
 #define OPTIONS_TABLE_STATUS_FORMAT3 \
-       "#[align=left] S: " \
+       "#[align=left][#{n:#{S:x}} sessions] " \
        "#[norange default]" \
        "#[list=on align=#{status-justify}]" \
        "#[list=left-marker]<#[list=right-marker]>#[list=on]" \

mgrant0:
> Now the mouse works for me to change session, but not to change pane. Maybe that was why you changed switch-client?

Yes, exactly why I made that change.  Selecting the pane selects the pane in the window in the session.  I think I need to put my change back in and try to fix your issue with selecting -t1.

Meanwhile, in all this editing somehow the sessions are broken now, this is what I am seeing now:

image

It's like it's not properly dealing with the `,` where `S:` outputs the stuff after the `,` for the active session and before the `,` for all other sessions.  It's just looping over and printing the whole thing including the `,`.  Huh, I don't know what broke this.  Ideas?

I added your diff above with a little mod to correct the singular/plurial cases.  Up to you here.  Personally I like the more terse `P:` and `S:` but if you prefer this that's fine.  

In line 1, you have `[0]` which is the session number/name.  It took me ages to figure out what that was before I understood the formats.  I do wonder if it would be better if the first line started with `session: ` and remove the brackets.  For example 
```Session 0:  0:bash*```
I thought about `[$0]` to make it clear it was a session but it unfortunately looks like some broken format where `$0` is some failed substitution and it would likely cause people to just ask about it over and over on the group.  If you like the idea of adding the word `Session` in front I'll change it in this PR but I don't have a strong opinion on this.

nicm:
We don't do singulars anywhere else because it is too much hassle, so I would just leave it as `1 panes`.

I don't really mind going back to`P:` and `S:`. They look a bit odd sometimes, but I don't think we will get anything perfect because the session name could be any length, so `P:` looks OK with `[0]` but it looks odd with `[mysession]`.

The other option would be to make the whole lot line up with the session name with something like `#{R: ,#{n:[#{session_name}]}}`.

I don't think we want to change the default status line, everyone is used to it and will all complain if it is changed.

`tmux display -p '#{S:a,b}'` still works for me in master

mgrant0:
@nicm I have tried to make the changes you recommended.  I went back to simply `P:` and `S:` because I can't think of anything else better right now.  Agreed, let's not change the first status line (the default one).  

I fixed the `tmux switch -t1` issue you brought up.  What I did was create a special case for `-t.=` which forces the `=` arg to refer to a pane.  This way I was able to bind the mouse click in the status line to `switch -t.=` which does the right thing.  Is this solution acceptable?

Though it's not strictly necessary, for consistency, would you like me to also add `-t:.=` and maybe `-t%=` to mean the same thing?

nicm:
I have applied this without the `switch-client` bit, with one change to pad the `P:` and `S:` out to match the length of the `[session]` in the first status line, which I think looks better.

For `switch-client` - how about just extending the existing special-case to `=` as well, this seems to work for me and I can't think of any downside:
~~~~
--- cmd-switch-client.c 21 Aug 2021 10:22:39 -0000      1.70
+++ cmd-switch-client.c 14 Aug 2025 07:13:29 -0000
@@ -61,7 +61,8 @@ cmd_switch_client_exec(struct cmd *self,
        const char              *tablename;
        struct key_table        *table;

-       if (tflag != NULL && tflag[strcspn(tflag, ":.%")] != '\0') {
+       if (tflag != NULL &&
+           (tflag[strcspn(tflag, ":.%")] != '\0' || strcmp(tflag, "=") == 0)) {
                type = CMD_FIND_PANE;
                flags = 0;
        } else {

mgrant0:
I think this looks good too.  

I don't yet see these changes yet in master.  I added your change to this PR.  Is there anything more I need to do with this PR?

nicm:
GitHub should be up to date now and I have applied my change as well. Sorry this one took so long. Thanks!

Issue #4491, opened 2025-04-26
no error message for invalid/unknown long options


dseomn:
### Issue description

From discussion at https://github.com/scop/bash-completion/pull/1364#discussion_r2061537610, it looks like tmux doesn't show any error for unknown long options like it does for unknown short options at the top level:

```
$ tmux -a
tmux: unknown option -- a
usage: tmux [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
```

```
$ tmux --a
usage: tmux [-2CDlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
```

In subcommands it does show an error though:

```
$ tmux new-window -A
command new-window: unknown flag -A
```

```
$ tmux new-window --A
command new-window: invalid flag --
```

@akinomyoga thought https://github.com/tmux/tmux/blob/master/compat/getopt.c might be out of date.

### Required information

Please provide the following information:

* tmux version (`tmux -V`). I tested with 3.5a, @akinomyoga tested with 36fba4c.
* Platform (`uname -sp`). `Linux unknown` for me.
* $TERM inside and outside of tmux (`echo $TERM`). N/A, I think
* Logs from tmux (`tmux kill-server; tmux -vv new`). N/A

nicm:
There aren't any long options because they aren't supported?

akinomyoga:
Yes, so something looking like "the long option" would be interpreted as a set of short options, and the error should be printed as if it tried to interpret it as as a set of short options. However, the current behavior is different between the short options and "something looking like the long options". While `tmux` fails to parse "something looking like the long option" as is clear from the behavior printing the usage to stderr, it doesn't show anything about what it failed to recognize.

It's fine to parse it as a set of short options, but it should output the error message in the same way as the usual short option. For example, `ssh` produces the following error (see the first line of the output "`unknown option -- -`"):

```console
$ ssh --this-is-definitely-not-a-valid-option
unknown option -- -
usage: ssh [-46AaCfGgKkMNnqsTtVvXxYy] [-B bind_interface] [-b bind_address]
           [-c cipher_spec] [-D [bind_address:]port] [-E log_file]
           [-e escape_char] [-F configfile] [-I pkcs11] [-i identity_file]
           [-J destination] [-L address] [-l login_name] [-m mac_spec]
[...]

PR #4492, opened 2025-04-27
Use the latest `getopt_long.c` from the upstream OpenSSH

akinomyoga:
This tries to address #4491.

```console
$ ./tmux -A
tmux: unknown option -- A
usage: tmux [-2CDhlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
$ ./tmux --A
tmux: unknown option -- -
usage: tmux [-2CDhlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
$ ./tmux -2a new
tmux: unknown option -- a
usage: tmux [-2CDhlNuVv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [-T features] [command [flags]]
$ ./tmux -2 new -a
no server running on /tmp/tmux-1000/default
```

It seems to work fine as far as I test it manually, but manual testing would be limited. Is there a way to run unit tests? I tried `make check`, but it doesn't seem to do anything.


nicm:
You can do `make` in the `regress` dir but some might fail for unrelated reasons (I don't run them that often).

akinomyoga:
Thank you. I ran the tests. The following table shows the results.

Some are failing as you anticipated, but these fail even with the current `master` branch. There are essentially no difference between the results with the `master` and the HEAD of this PR (Exactly speaking, the output of `input-keys.sh` was different between the `master` and the HEAD, but it seems to change every time I run the test even with the same binary, so I haven't counted it as an essential difference).

| Name                         | master 36fba4cc | PR HEAD b5bc9f47 |
|------------------------------|-----------------|------------------|
| am-terminal.sh               | pass            | pass             |
| capture-pane-hyperlink.sh    | fail            | fail             |
| capture-pane-sgr0.sh         | pass            | pass             |
| combine-test.sh              | pass            | pass             |
| command-order.sh             | pass            | pass             |
| conf-syntax.sh               | pass            | pass             |
| control-client-sanity.sh     | pass            | pass             |
| control-client-size.sh       | pass            | pass             |
| copy-mode-test-emacs.sh      | fail            | fail             |
| copy-mode-test-vi.sh         | fail            | fail             |
| cursor-test1.sh              | pass            | pass             |
| cursor-test2.sh              | pass            | pass             |
| cursor-test3.sh              | pass            | pass             |
| cursor-test4.sh              | pass            | pass             |
| format-strings.sh            | pass            | pass             |
| has-session-return.sh        | pass            | pass             |
| if-shell-TERM.sh             | pass            | pass             |
| if-shell-error.sh            | pass            | pass             |
| if-shell-nested.sh           | pass            | pass             |
| input-keys.sh                | fail            | fail             |
| kill-session-process-exit.sh | pass            | pass             |
| new-session-base-index.sh    | pass            | pass             |
| new-session-command.sh       | pass            | pass             |
| new-session-environment.sh   | fail            | fail             |
| new-session-no-client.sh     | pass            | pass             |
| new-session-size.sh          | pass            | pass             |
| new-window-command.sh        | pass            | pass             |
| osc-11colours.sh             | pass            | pass             |
| run-shell-output.sh          | fail            | fail             |
| style-trim.sh                | fail            | fail             |
| tty-keys.sh                  | fail            | fail             |

akinomyoga:
I adjusted the commit message.

nicm:
Funny, only these ones fail for me:

copy-mode-test-emacs.sh -> tab is now output as tab not spaces
copy-mode-test-vi.sh -> same
new-session-environment.sh -> default TERM is tmux-256color for me not screen
run-shell-output.sh -> running the wrong tmux, duh
tty-keys.sh -> something to do with `M-BSpace` which it seems may still be broken

I've fixed all but the last.

Your change looks fine for me, so I have applied it now, thanks!

akinomyoga:
> Funny, only these ones fail for me:

OK, I had 8 errors, so there are three extra errors in my environment.

### `input-keys.sh` and `style-trim.sh`

They turned out to depend on the user's `~/.bashrc`. I confirm these tests pass when I replace `~/.bashrc` with an empty file.

### `capture-pane-hyperlink.sh`

Here is a patch: [`0001-Use-printf-instead-of-echo.patch`](https://github.com/user-attachments/files/19939860/0001-Use-printf-instead-of-echo.patch)

<details>

This test still failed with a plain `~/.bashrc`:

```console
$ sh capture-pane-hyperlink.sh
/tmp/tmp.vIyfZ0l1OT /tmp/tmp.GHorgEm5kz differ: char 1, line 1
```
With both the previous master branch (before applying this PR) and the current master, the output of `$TMUX capturep -peS0 -E1` seems to be
<pre>
<b>ESC</b>]8;id=1;https://github.com<b>ESC</b>\test1^[]8;;<b>ESC</b>\

</pre>

(where **ESC** is a raw ESC), while the expected value is

```
\033]8;id=1;https://github.com\033\\test1\033]8;;\033\\\n
```

This is due to the use of [`echo $2`](https://github.com/tmux/tmux/blob/f367e3a6c8761328b2233067b301909bb096d6fc/regress/capture-pane-hyperlink.sh#L19). The behavior of the `echo` builtin with an escape sequence is specified to be implementation-defined by POSIX, so the result depends on the shell. In my environment, `sh` is Bash.

> [POSIX XCU 3.37 - echo OPERANDS](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/echo.html#tag_20_37_05)
>
> *string*
>
> A string to be written to standard output. If the first operand consists of a '`-`' followed by one or more characters from the set {'`e`, '`E`', '`n`'}, or **if any of the operands contain a \<backslash> character, the results are implementation-defined.**

One should instead use `printf "$2\n"`.

</details>

nicm:
Yes, IIRC we changed a few `echo` -> `printf` already but this one must have been missed. Applied now, thanks!

I don't think there is much we can do about .bashrc, I guess if we could figure out it was bash we could rerun the shell with `--noprofile --norc` but I don't know if there is a foolproof way to know it is bash...

dseomn:
From https://www.gnu.org/software/bash/manual/bash.html#Invoked-with-name-sh it looks like a non-interactive non-login shell invoked as `sh` shouldn't read any startup files, I think? Did you run the scripts explicitly with bash instead of using the `#!/bin/sh` shebang line?

akinomyoga:
Tmux is a terminal, and these are tests using actual terminal sessions. A terminal session creates an interactive shell session.

dseomn:
Oh gotcha, I hadn't looked at those tests too closely. It looks like input-keys.sh is mostly testing the output of `cat -tv` and the relevant part of style-trim.sh is capturing tmux's own status line, I think. So maybe it's possible to change those tests to avoid interactive or login shells? Do you know what part of your .bashrc is causing the problem so I can try to fix the tests?

akinomyoga:
I agree with @nicm that it wouldn't be practical to support an arbitrary `~/.bashrc` (which can be really anything), so I believe we should instead think about a way to run an interactive session with `--noprofile` and `--norc`, but I took a look into the current situation.

### `input-keys.sh`

First, it seems to have been my misunderstanding that the test `input-keys.sh` depends on `~/.bashrc`. It seems to fail randomly at some probability even with an empty `~/.bashrc`. The probability does not seem to depend on the contents of `~/.bashrc`.

I noticed that two specific `assert_key` lines produce errors with a relatively higher probability (though some other tests also seem to fail). I looped these two `assert_key`, and measured the probability of failure.

```bash
assert_key 'C-i' '^I'  -- 'M-C-i' '^[^I'

# The above failed with the following values 32 times out of 600 calls:
declare -- key="C-i"
declare -- actual_code=$'\t'
declare -- expected_code="^I"
# The above failed with the following values 33 times out of 600 calls:
declare -- key="M-C-i"
declare -- actual_code=$'^[\t'
declare -- expected_code="^[^I"
```

```bash
assert_key 'C-u' '^U'  -- 'M-C-u' '^[^U'

# The above failed with the following values 28 times out of 600 calls:
declare -- key="C-u"
declare -- actual_code=""
declare -- expected_code="^U"
# The above failed with the following values 36 times out of 600 calls:
declare -- key="M-C-u"
declare -- actual_code=""
declare -- expected_code="^[^U"
```

The failure probability seems mostly consistent with each other, and assuming the probability to be the same for those four failure types, the failure probability is estimated to be 5.42(46)%. The statistical error was estimated by assuming the binomial distribution.

### `style-trim.sh`

The test `style-trim.sh` fails 100% with my `~/.bashrc` but succeeds 100% with an empty `~/.bashrc`. The minimal reproducer is this:

```bash
# ~/.bashrc

source /path/to/ble.sh --norc
bleopt prompt_status_line='\u@\h\r\w'
```

Obviously, the status line shown by the shell framework is interfering.

nicm:
You could try something like (untested):
~~~~
tmux set -sF default -command '#{?#{m/r:#{default-shell},bash},bash --noprofile --norc,}'
~~~~
Or some equivalent shell `if`.

Or can anything be done with `BASHOPTS`?

akinomyoga:
Thanks for the suggestion. I think setting `default-command` after creating the new session wouldn't have an effect, (though I might misunderstand your intent). We need to create a new window after setting `default-command`. Then, I ended up simply using `tmux new-window 'bash --noprofile --norc'` (without changing `default-command`).

This is a patch for `style-trim.sh`: [`0001-Run-the-regress-style-trim.sh-tests-in-a-plain-Bash-.patch`](https://github.com/user-attachments/files/19960546/0001-Run-the-regress-style-trim.sh-tests-in-a-plain-Bash-.patch)

nicm:
~~~~
 $TMUX2 -f/dev/null new -d "$TMUX -f/dev/null new"
 sleep 2
+if [ -n "$use_plain_bash" ]; then
+	$TMUX new-window 'bash --noprofile --norc'
+	sleep 1
+fi
~~~~
Maybe it would be nicer to just run the shell in the first window instead of making a second one? Something like:
~~~~
shell=
if command -v bash >/dev/null 2>&1; then
    shell="bash --noprofile --norc"
fi
...
$TMUX2 -f/dev/null new -d "$TMUX -f/dev/null new -- $shell"
~~~~
Also can't it set `HISTFILE` to empty or add `+Ohistory` instead of using a temporary file?

akinomyoga:
Thank you. That is better. I updated the patch to use `new -- $shell`.

> Also can't it set `HISTFILE` to empty or add `+Ohistory` instead of using a temporary file?

OK, I tested these to confirm that the command history can be disabled by specifying empty HISTFILE or by setting `+o history` (though the uppercase `-O`/`+O` are used for `shopt` settings, while `history` is a `set -o/+o` setting).

Updated patch: [`0001-Run-the-regress-style-trim.sh-tests-in-a-plain-Bash-.patch`](https://github.com/user-attachments/files/19961998/0001-Run-the-regress-style-trim.sh-tests-in-a-plain-Bash-.patch)

nicm:
I've applied this now.

I also removed `logit` from `getopt_long.c` because it has a warning, and tmux has its own `warnx` anyway.

Thanks!

akinomyoga:
Thank you! Sorry, I haven't noticed `warnx` in another file.

akinomyoga:
@nicm This is a patch for the mysterious CI error in Solaris: 
[`0001-Include-a-Solaris-header-declaring-opt-before-defini.patch`](https://github.com/user-attachments/files/19979341/0001-Include-a-Solaris-header-declaring-opt-before-defini.patch).

The problem is explained in the code comment added in the patch. A minimal reproducer of the error is this:

```console
[murase@vm-solaris]$ cat macro-const.c
#define optarg BSDoptarg
#include <stdlib.h>
extern const char* optarg;
int main() {}
[murase@vm-solaris]$ gcc macro-const.c
macro-const.c:1:16: error: conflicting types for ‘BSDoptarg’
 #define optarg BSDoptarg
                ^
macro-const.c:3:20: note: in expansion of macro ‘optarg’
 extern const char* optarg;
                    ^
macro-const.c:1:16: note: previous declaration of ‘BSDoptarg’ was here
 #define optarg BSDoptarg
                ^
[murase@vm-solaris]$
```

The above error message is strange, but the description of the real problem is revealed when one removes the macro:

```console
[murase@vm-solaris]$ cat macro-const.c
/*#define optarg BSDoptarg*/
#include <stdlib.h>
extern const char* optarg;
int main() {}
[murase@vm-solaris]$ gcc macro-const.c
macro-const.c:3:20: error: conflicting types for ‘optarg’
 extern const char* optarg;
                    ^
In file included from macro-const.c:2:0:
/usr/include/stdlib.h:202:14: note: previous declaration of ‘optarg’ was here
 extern char *optarg;
              ^
[murase@vm-solaris]$
```

nicm:
Why is optarg `const char *` - isn't it normally `char *`? Wouldn't changing it to `char *` fix this?

akinomyoga:
I've added `const` to suppress errors arising while compiling `getopt_long.c`, which tried to assign literal strings to `char *` variables. If you don't like it, you can remove `const` (see cbb729e8d900616d8ffb868325d152a7607d91b4 for the places where `const` is added) and instead copy the literal strings to new static-storage `char[]` variables (or forcibly cast the literal strings to `(char *)`? I'm not sure if this is safe when someone tries to rewrite `optarg` in the future).

However, even if `const char*` is changed to conventional `char *`, I believe the attached patch should still be applied because rewriting the standard headers through macros shouldn't be intended.

nicm:
Declaring things again of the same type is fine, I don't think we need to do anything special just because they're in system headers.

I'd probably make it `char *` like it's meant to be and then add casts to make `-Wwrite-strings` shut up.

akinomyoga:
> Declaring things again of the same type is fine,

That part is fine, but I'm not sure if rewriting all the identifiers `optarg`, etc. to `BSDoptarg`, etc. in a system header in a translation unit wouldn't cause inconsistency. Technically, headers may have `static` or `inline` function definitions that use those variables.

nicm:
They could but they won't because this is getopt and the declarations are standardized.

nicm:
I changed `optarg` back to `extern char *` in 6cc1a607fcbe1a9a8401642ce96d5648c777556c, could you try this and see if it fixes Solaris? Thanks!

akinomyoga:
Thanks. I tried it in a Solaris 11 virtual machine (Since the toolset of Solaris 11 is too old, `./autogen.sh` produces a broken `./configure` in Solaris, so I instead created a tarball in another host and copied it to Solaris 11). I confirm that the current master successfully compiles on Solaris 11.

nicm:
Great, thanks!

PR #4496, opened 2025-05-03
feat: Use posix_spawn for JOB_NOWAIT jobs

### LINKED ISSUE #3352: tmux slows in proportion to scrollback size, but only when status has subshells
After setting a large `history-limit`, as scrollback is filled with 78 column lines, tmux becomes increasingly sluggish to complete `refresh-client -S` (and everywhere else), but only if there are subshells in status format (`ie #(command ...)`).  Eventually tmux becomes unusable and 100% cpu, by about 40M lines of scrollback, even with no output or interaction and just one tmux window open.

The performance is fair enough (for that many lines) when there is no status-left defined.  I do not understand why scrollback size and presence of subshells in status format are linked and cause this incredible slowdown.

1. build latest master on linux with --enable-utempter --enable-static --enable-utf8proc --disable-systemd --sysconfdir=/etc
2. create rcfile:
```
set-option -g history-limit 2147483647
set-option -g status-left-length 99

set-option -qg \
status-left "\
#(git -C #{pane_current_path} \
  symbolic-ref --short HEAD) \
#(git -C #{pane_current_path} \
  status -uno --porcelain \
  | awk '{if \(!seen[$1]\) {printf\($1\); seen[$1]=1}}')\
#(git -C #{pane_current_path} \
  rev-list --walk-reflogs --count refs/stash 2>/dev/null) \
"
```
3. start a new tmux
4. run
```
perl -le 'for ($i = 0; $i < 10000000; $i++) {print "a" x 78}'
```
5. try `time refresh-client -S`
6. repeat steps 3 and 4, yielding following run times:
10M lines: 0.13s
20M lines: 0.22s
30M lines: 0.31s
40M lines: 0.42s

by that time cpu fans are going, takes 66 seconds to enter copy-paste mode, with lines 0/29999999 shown, scrolling painfully slow, then stops interaction altogether, cpu 100%.  you can wait a long time and it seems to never stop.  You can interact with tmux, just everything is very slow, on timescale of minutes or tens of seconds even for tmux prompt input to echo

*without* **status-left** as above,`refresh-client -S` completes instantly, no matter how many lines of scrollback!!  Note that `pane_current_path` does not need to be inside git directory for this to happen, or for tmux to have ever been inside one from start.

Not sure if this is related to #2551 or not.  After this test, it takes several minutes to exit the test window and full cpu is being used the whole time.  But it will exit properly and do orderly shutdown.
If not exiting, Tmux never recovers interactive performance after this, even if all windows that had scrollback are closed, and new empty ones opened instead, it will always be very slow even several hours later and continue to use high cpu even with one window, no scrollback, running a shell doing nothing.

eMPee584:
Ah wow, have been testing this for two weeks and making the status calls (JOB_NOWAIT) go via posix_spawn() seems to fix #3352, no accumulating sluggishness like before. Awesome : D
If AI co-created code is not acceptable, at least this provides a proof of concept..
Oh and the environment handling, probably only useful for a very small subset of status line commands, I didn't ask for it but neither did I try ripping it out yet..

--------------------------

Implement posix_spawn() in job_run() for jobs flagged with JOB_NOWAIT and not requiring a PTY. This avoids a full fork() for simple background tasks.

Includes:
- Added posix_spawn path in job_run().
- Added environ_get_envp() and environ_free_envp() helpers.
- Set up pipe and bufferevent to capture stdout for spawned jobs.
- Addressed various compilation and runtime issues during implementation.

Note: The cwd parameter is currently ignored for posix_spawn jobs.

[**AI crafted using roo code and gemini 2.5 pro**]

### DIFF
--- a/environ.c
+++ b/environ.c
@@ -279,3 +279,52 @@ environ_for_session(struct session *s, int no_TERM)
 
 	return (env);
 }
+
+/* Create an envp array from an environment. */
+char **
+environ_get_envp(struct environ *env)
+{
+	struct environ_entry	*envent;
+	char			**envp;
+	u_int			  i, count;
+
+	/* Count entries. */
+	count = 0;
+	RB_FOREACH(envent, environ, env) {
+		if (envent->value != NULL &&
+		    *envent->name != '\0' &&
+		    (~envent->flags & ENVIRON_HIDDEN))
+			count++;
+	}
+
+	/* Allocate envp array. */
+	envp = xcalloc(count + 1, sizeof *envp);
+
+	/* Populate envp array. */
+	i = 0;
+	RB_FOREACH(envent, environ, env) {
+		if (envent->value != NULL &&
+		    *envent->name != '\0' &&
+		    (~envent->flags & ENVIRON_HIDDEN)) {
+			xasprintf(&envp[i], "%s=%s", envent->name, envent->value);
+			log_debug("%s: adding to envp: %s", __func__, envp[i]); /* Log added env var */
+			i++;
+		}
+	}
+	envp[i] = NULL;
+
+	return (envp);
+}
+
+/* Free an envp array created by environ_get_envp. */
+void
+environ_free_envp(char **envp)
+{
+	u_int	i;
+
+	if (envp == NULL)
+		return;
+	for (i = 0; envp[i] != NULL; i++)
+		free(envp[i]);
+	free(envp);
+}
--- a/job.c
+++ b/job.c
@@ -21,8 +21,10 @@
 #include <sys/socket.h>
 #include <sys/wait.h>
 
+#include <errno.h> /* Added for errno */
 #include <fcntl.h>
 #include <signal.h>
+#include <spawn.h> /* Added for posix_spawn */
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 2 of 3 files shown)


nicm:
Well, I don't mind doing this, although I'd like to know the benefit. fork has been COW for years (decades?) so what is the benefit of posix_spawn?

Also I'm not sure I can take this since you have told me you didn't write it. What is the license of this code? It is a large enough change to be potentially copyrightable, although it is quite generic. You didn't write it so you don't own the copyright. Can you find out if this code is public domain or can be BSD licensed or what the story is? You may need to find a statement on Google's website.

eMPee584:
Sorry my browser session crashed and ate the reply, then I got swamped in other random stuff to deal with.. I'll need a few more days to get back to this. Any reason to rush conclusions?

eMPee584:
Also just figured out my mail account started to feed all github notification emails into the spam folder, so that didn't help my responsiveness on various issues..

eMPee584:
> what is the benefit of posix_spawn?

Well yeah supposedly fork() does COW, but for whatever reason, without this patch there was heavy session _aging_ (any tmux interaction get's more and more sluggish due to multiple shell commands launched from status-line on every refresh).. now, even with many windows and large histories: **zero latency** 🥲 ... like, finally can feel the power of my ancient hexacore rig, fluidly zapping through tens and tens of tmux windows AT LAST 
… what I am unsure about though is the whole environment variable support, but I guess it's necessary if you don't want to restrict the versatility of this call method.. 🤔

> [**Use of Generated Content**](https://ai.google.dev/gemini-api/terms#use-generated)
> 
> Some of our Services allow you to generate original content. **Google won't claim ownership over that content**. You acknowledge that Google may generate the same or similar content for others and that we reserve all rights to do so.
> 
> As required by the API Terms, you'll comply with applicable law in using generated content, which may require the provision of [attribution](https://ai.google.dev/api/rest/v1beta/CitationMetadata) to your users when returned as part of an API call. Use discretion before relying on generated content, including [code](https://g.co/legal/generative-code). You're responsible for your use of generated content, and for the use of that content by anyone you share it with.

Licensing-wise this seems to suggest all copyright ownership or infringement responsibility is on the user.

smemsh:
> fork has been COW for years (decades?) so what is the benefit of posix_spawn

Data pages copy on write, but as per #3352, that does not include the PTEs representing those pages, which can accumulate to large quantities over time in some tmux configurations.

Also a side note related to that issue: `tcsetpgrp()` semantics for setting controlling tty are available race-free in glibc `posix_spawn()` now -- see bminor/glibc@6289d28d -- by setting it up with `posix_spawn_file_actions_addtcsetpgrp_np()` and passing the controlling terminal fd.  So `posix_spawn()` could be used for panes as well, in theory.

I haven't looked at other platforms though... I was just getting around to looking at this again, as it remains an issue that tmux gradually slows down, but then I noticed this PR... hrmmm AI :-/

smemsh:
After using this PR, *wow*, it's much faster now with even small amounts of scrollback (couple million 80-col lines globally was already over 100ms, now 0).  This is a considerable improvement.  I had backgrounded+disowned my `refresh-client -S` in $PROMPT_COMMAND shell function already before, but it still caused an accumulating lag from all that PTE copying.

It may be better to have long-running procs for the `#()` expansions -- then tmux is just doing IO and not forking -- but if I understand how that works, it would need to keep writing a new line every second (the minimum quantum) to have the latest info (tmux uses the latest line whenever it needs to render).   Whereas this patch allows the expansions to work in the obvious way, by running single commands on demand only, which terminate, and don't unexpectedly take a long time and cause growing CPU drag as scrollback increases.

Using `posix_spawn()` for new panes also would be great, but a lot harder for reasons mentioned in the other issue.  However this patch seems like a good start and is immediately useful.  Thanks for posting it @eMPee584.

nicm:
I think this is OK but it needs to be cleaned up a bit, at minimum I would use a separate function for the `posix_spawn` path (and probably the `forkpty` path too). I also don't see why it is limited to `JOB_NOWAIT`, it just tells tmux not to care if the process has exited?

PR #4505, opened 2025-05-12
Point to new bash completion file

dseomn:
I can't read Japanese, so I just guessed about the change to README.ja.

From https://github.com/imomaliev/tmux-bash-completion:
> **Archived**: Bash completion for tmux is now part of the [bash-completion](https://github.com/scop/bash-completion/), please read more [here](https://github.com/imomaliev/tmux-bash-completion/issues/13#issuecomment-2873735797)


nicm:
Applied, thanks!

PR #4513, opened 2025-05-20
Add border case for middle window.

mgrant0:
When there are more than two horizontal windows and the active window is in not on an edge this fix correctly highlights both its left and right borders.


nicm:
Applied, thanks!

PR #4515, opened 2025-05-22
Fix lengths of panes and windows in status bar.

mgrant0:
The issue in format-draw.c fixes is an "off-by-one" error because of the prefix character: $, %, or @ which is contained in the argument. When the mouse clicks in the status line, it ends up being off by one character because of this extra undisplayed prefix character. If you click the mouse on the first character of a session index, it was selecting the session immedietly to the left because it sees that target as one character too wide.


mgrant0:
This appears not to be a bug.  Closing this.

PR #4516, opened 2025-05-22
Feature: session loop sorting and session activity format variable

mgrant0:
Adds sorting flags to the S: format.  

Adds `session_alert` format which returns `#` for activity, `!` for bell, and `~` for silence if one of the windows in has those alerts.

Adds `session_bell_flag` format which is true if one of the windows had rung the bell in one of the windows in a session.

### DIFF
--- a/format.c
+++ b/format.c
@@ -131,6 +131,18 @@ enum format_type {
 	FORMAT_TYPE_PANE
 };
 
+/* Format loop sort type. */
+enum format_loop_sort_type {
+	FORMAT_LOOP_BY_INDEX,
+	FORMAT_LOOP_BY_NAME,
+	FORMAT_LOOP_BY_TIME,
+};
+
+static struct format_loop_sort_criteria {
+	u_int	field;
+	int	reversed;
+} format_loop_sort_criteria;
+
 struct format_tree {
 	enum format_type	 type;
 
@@ -539,6 +551,31 @@ format_cb_session_attached_list(struct format_tree *ft)
 		xasprintf(&value, "%.*s", size, EVBUFFER_DATA(buffer));
 	evbuffer_free(buffer);
 	return (value);
+}
+
+/* Callback for session_alert. */
+static void *
+format_cb_session_alert(struct format_tree *ft)
+{
+	struct session	*s = ft->s;
+	struct winlink	*wl;
+	char		 alerts[1024];
+
+	if (s == NULL)
+		return (NULL);
+
+	*alerts = '\0';
+	RB_FOREACH(wl, winlinks, &s->windows) {
+		if ((wl->flags & WINLINK_ALERTFLAGS) == 0)
+			continue;
+		if (wl->flags & WINLINK_ACTIVITY)
+			strlcat(alerts, "#", sizeof alerts);
+		if (wl->flags & WINLINK_BELL)
+			strlcat(alerts, "!", sizeof alerts);
+		if (wl->flags & WINLINK_SILENCE)
+			strlcat(alerts, "~", sizeof alerts);
+	}
+	return (xstrdup(alerts));
 }
 
 /* Callback for session_alerts. */
@@ -2471,6 +2508,22 @@ format_cb_window_bell_flag(struct format_tree *ft)
 	return (NULL);
 }
 
+/* Callback for session_bell_flag. */
+static void *
+format_cb_session_bell_flag(struct format_tree *ft)
+{
+	struct winlink		*wl;
+
+	if (ft->s != NULL) {
+		RB_FOREACH(wl, winlinks, &ft->s->windows) {
+			if (wl->flags & WINLINK_BELL)
+				return (xstrdup("1"));
+			return (xstrdup("0"));
+		}
+	}
+	return (NULL);
+}
+
 /* Callback for window_bigger. */
 static void *
 format_cb_window_bigger(struct format_tree *ft)
@@ -3224,6 +3277,9 @@ static const struct format_table_entry format_table[] = {
 	{ "session_activity", FORMAT_TABLE_TIME,
 	  format_cb_session_activity
 	},
+	{ "session_alert", FORMAT_TABLE_STRING,
+	  format_cb_session_alert
+	},
 	{ "session_alerts", FORMAT_TABLE_STRING,
 	  format_cb_session_alerts
 	},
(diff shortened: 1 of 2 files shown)


mgrant0:
Should I add the sorting options to `S:` and perhaps also `P:`?

nicm:
I thought I had replied to this already but apparently not.

Anyway, I have applied the `S:` bits with some minor changes for line length.

I left out `format_cb_session_alert` for now because if multiple windows have the same alert it seems like it will include them multiple times? Is that what you want?

Adding sorting for `P` and `W` would be nice too.

ThomasAdam:
> I left out `format_cb_session_alert` for now because if multiple windows have the same alert it seems like it will include them multiple times? Is that what you want?

Indeed -- I hope not.  This would really irritate me unless it can be predicated by checking if the window is linked...

mgrant0:
> I left out format_cb_session_alert for now because if multiple windows have the same alert it seems like it will include them multiple times? Is that what you want?

I see what you mean.  Let me fix that.  There should be at most one type of alert for each.  

@ThomasAdam I don't quite understand what you mean.  Unless I misunderstand I think Nic sees a bug in my code where if multiple windows in a session had a bell for example, then, there would be multiple `!` characters in the session_alert.  The intent with the session alert was to be able to show that at least one window within the session had one of those alerts and maybe the user would then click that session and see what's going on.

Linked windows, hmm, I haven't tried that but my expectation would be if a window was linked to 2 sessions and that window caused an alert that both sessions the alert added to their session_alert in the 3rd status line.

nicm:
Looks good, can you add them to tmux.1 as well? There is a list of format variables.

nicm:
Oh `format_table` needs to be sorted or `bsearch` will not work.

Don't you want `session_activity_flag` etc as well?

mgrant0:
I added `/i`, `/n`, `/t`, `/r` sorting to `W:` and `/r` sorting to `P:` because there's no pane name or pane last activity time.

I had a small issue with `W:` which caused me a bit of drama.  When you sort it other that by index ascending (the default), the callback `format_cb_window_end_flag` returns the last window by index which would be wrong if you sort it differently.  `OPTIONS_TABLE_STATUS_FORMAT1` uses `window_end_flag` to put a separator in between windows.

I fixed this by adding a new flag `loop_last_flag` which returns `1` on the last item in the loop.  (or feel free to rename it).  This, then, can be used in all the new and improved status lines.  To get this to work, I add a new FORMAT_LAST flag into the ft->flags.  I modified the first status line to use this new `loop_last_flag` instead of `window_end_flag` which I did not want to modify in case someone already uses that.

I also added the sorting flags to L: for clients, but since a client doesn't have a numeric id, I just used the order clients came out of the TAILQ.

mgrant0:
> Oh `format_table` needs to be sorted or `bsearch` will not work.

Did I fix this with the most recent commit by sorting the list before and using a for loop to add the `nft`?  If not, where are you referring to?

> Don't you want `session_activity_flag` etc as well?

I think so yes for consistency.  I see the activity time.  How and where do you recommend this being turned into a flag?

nicm:
I mean that `session_bell_flag` does not go after `window_bell_flag` in `format_table`, it needs to go between `session_attached_list` and `session_created`, or `bsearch` will get confused.

There is `window_activity_flag` and `window_silence_flag`, the session versions of those can work the same way as `session_bell_flag` does for `window_bell_flag`?

mgrant0:
Got it.  I think that was the only one out of order.  I did this to check it:
```
grep  '{ "' format.c > 1
sort 1 > 2
diff 1 2
```
I get no differences now.

mgrant0:
I've added these so far:
- format_cb_session_silence_flag
- format_cb_session_activity_flag
- format_cb_session_active
- format_cb_last_session_index
- format_cb_session_index

What others might be useful?  Zoomed (as in, at least one window zoomed within a session) does not seem terribly useful.  Similar with a marked pane, does not seem useful to me in context of a session.  I was not going to implement a session stack to track last session.

I just need to clean things up, fix the ordering in the file and try to make sure they work before pushing these.

If there's something more you want here, let me know.

mgrant0:
The session bell and silence flags, I am not so sure how to test those but the others seem to work.

mgrant0:
Let me know what else you want here.

nicm:
Looks good but:

- There is no such thing as session index, only session ID, so please get rid of `session_index` and `last_session_index`.

- `qsort_r` is not portable, so if you want to use it you will have to add a `compat` version.

nicm:
Since tmux is single threaded, a `compat` version of `qsort_r` should be trivial - it can call `qsort` with a global to store the context pointer.

nicm:
Actually, no, OpenBSD has no `qsort_r` yet so we can't use it.

mgrant0:
I went back to the global variable variant way.  Do you prefer a qsort_r in compat?  I can find one if you do.  (Maybe I was too quick to revert this back to using the global variable!)

Personally I would say using something like qsort_r without needing to pass info in a global variable would be better.  But you have globals like this in several places so it's not so awful.  I should make it static though if i leave it like this.

nicm:
No, if OpenBSD had it then compat would be better but since it doesn't it's better to just stick with qsort so we don't have a difference between OpenBSD and portable.

mgrant0:
Ok hopefully those commits address your comments.  Thanks.

nicm:
This looks good, I have applied it with a few changes:

- I don't like having a variable called something as small as `sc` at global scope so I changed it back to `format_loop_sort_criteria`.

- `format_cmp_pane` had a redundant assignment of `result` which I removed.

- If no arguments were being provided, the sort criteria was not being cleared, so if you did `W/ir` it would stick for the next `W` (the `if (fm->argc < 1)` check).

I didn't test all the new formats, I assume you did that :-).

It'd be nice if some tests for the sorting went into `regress/format-strings.sh`.

format.c
```diff
@@ -4356,6 +4477,39 @@ format_window_name(struct format_expand_state *es, const char *fmt)
...
+{
+	const struct winlink *const	*a = a0;
+	const struct winlink *const	*b = b0;
+	const struct window		*wa = (*a)->window;
+	const struct window		*wb = (*b)->window;
+	int				 result = 0;
+
+	switch (sc.field) {
+	case FORMAT_LOOP_BY_INDEX:
+		result = wa->id - wb->id;
```
naruaway, line 4491: Due to this, windows are sorted by id by default, not by index, on the status bar

naruaway: Fixing this in https://github.com/tmux/tmux/pull/4551

PR #4353, opened 2025-01-27
Add mode 2031 support (automatic dark/light mode)

jonathanslenders:
See: https://github.com/tmux/tmux/issues/4269

This PR implements:
- Enable mode 2031 when the terminal is Vt100-like to subscribe to theme updates from the terminal.
- Handle mode 2031 for applications running within tmux.
- Adds a `client_theme` variable for usage in the formatting.
- This adds two hooks: `client-dark-theme` and `client-light-theme` that allows a tmux config to respond to theme changes and update colors.

Usage instructions
-------------------

For anyone coming across this PR:

1. Use a terminal emulator that also supports mode 2031 like Ghostty or Kitty. Ghostty can automatically follow the system theme, when configured like that. E.g., in `~/.config/ghostty/config`, do:

```
theme = "light:rose-pine-dawn,dark:rose-pine"
```

2. In the tmux config, bind shell scripts to the `client-{dark,light}-theme` events. In these shell scripts, you can then perform tmux commands that toggle the style of the status bar and tmux panes.

```
set-hook -g client-dark-theme 'run-shell ~/bin/tmux-dark-theme.sh'
set-hook -g client-light-theme 'run-shell ~/bin/tmux-light-theme.sh'
```

For debugging, you can display the theme in the status bar:
```
set -g status-right " %H:%M %d-%b-%y [#{client_theme}]"
```

3. For neovim, use a recent neovim version that supports mode 2031. This will automatically toggle the `bg` setting between light and dark following tmux and ghostty. Then use a color scheme like rose-pine that follows the bg setting.

Testing the protocol
--------------------

Query whether mode 2031 is supported (should return `<csi>?2031;2$y`):
```sh
echo -e '\e[?2031$p'
```

Within a tmux pane, the following should report the current theme (`<CSI>?997;1n` for light a `<CSI>?997;2n` for dark):
```sh
echo -e "\e[?996n"
```

To subscribe for theme updates:
```sh
echo -e "\e[?2031h"
```

To unsubscribe:
```sh
echo -e "\e[?2031l"
```

When tmux is styled with a background color of its own, this should automatically be taken into account and reported whenever this changes. Do `echo -e "\e[?2031h"` and then type the following in the tmux prompt:
```
set window-style 'fg=#000000,bg=#ffffff'
```

PR #4519, opened 2025-05-25
Document DEC mode 2031 hooks in manpage

ccjmne:
See Git commit eaf70c95, GitHub issue #4353

--

Hello,

Here's an immensely simple commit, but I suppose it's a good way to get
one's feet wet: I would like to contribute as well as possible.

It's my first time using CVS and I figured I'd be more comfortable
asking here :)

Can I simply post to the `tech` mailing list with a Git patch?  It
seems odd, but the few patches I can find on that mailing list are
Git-formatted.  I can't seem to find a patch there that corresponds to a
commit on GitHub either... I must be missing something.

By the way, the last line of the `man` page displays a literal
`$Mdocdate$` for me. Is that an issue or my end, or could I attempt
addressing that nitpick myself, as a bit of an exercise?

Thanks for the excellent tool!
Cheers,

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -5471,6 +5471,10 @@ Run when focus exits a client
 Run when a client is resized.
 .It client-session-changed
 Run when a client's attached session is changed.
+.It client-light-theme
+Run when a client switches to a light theme.
+.It client-dark-theme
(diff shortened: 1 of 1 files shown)


nicm:

PR #4534, opened 2025-06-16
Add command rename-client.

michael-stevenson:
I've recently found myself wanting to name clients some human-friendly name so that I can easily target them with other commands.   I was a bit surprised to find that this command didn't already exist.   So, I'm hoping that it will be found worthy of being upstreamed..   Contribution guidelines focused on issue reporting, not so much on PRs, so I do apologize for any anti-pattern.   Anyway, I would appreciate consideration for inclusion.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -114,6 +114,7 @@ dist_tmux_SOURCES = \
 	cmd-pipe-pane.c \
 	cmd-queue.c \
 	cmd-refresh-client.c \
+	cmd-rename-client.c \
 	cmd-rename-session.c \
 	cmd-rename-window.c \
 	cmd-resize-pane.c \
--- a/cmd-rename-client.c
+++ b/cmd-rename-client.c
@@ -1,0 +1,79 @@
+/* $OpenBSD$ */
+
+/*
+ * Copyright (c) 2007 Nicholas Marriott <nicholas.marriott@gmail.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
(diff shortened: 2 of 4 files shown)


nicm:
I don't know that this would do any harm, but what is your the use for this? Clients are usually transitory.

michael-stevenson:
> Clients are usually transitory.

Not for me.   I like to integrate with other tools that target a client to, for example, switch which session its currently pointed at.  A concrete example: I use taskwarrior and when I switch tasks, I want to send my "task client" a command to switch which task session its attached to.   Doing this with a proper name allows me to not have to jump through hoops of "remembering" (via a file or such) which is the "task client".

> I don't know that this would do any harm

Obviously not definitive, but I haven't noticed any breakages while using this for a few weeks time.   Its not as if I introduced anything new (the client name was already there and was always synonymous with the tty - i.e. redundant).   Hopefully, anything that was actually *trying* to get the tty wasn't using the name as a proxy.

nicm:
Sorry for the delay, I'm just a bit reluctant to add new commands, especially for rarely-used features, as we already have way too many. I'm wondering about making it part of `refresh-client`, but that is already a bit overloaded... maybe `rename-client` just makes sense.

nicm:
OK, let's go with this. Can you make two changes:

- Add your copyright line to the top of the new `cmd-rename-client.c` file (after mine) with copyright year 2025. If you don't want to include your email address that is fine, just leave it out.

- Add `client-renamed` to the list in the `HOOKS` section of `tmux.1`.

Thanks

PR #4539, opened 2025-06-22
Detect support for OSC 52 using the device attributes report

### LINKED ISSUE #4532: Consider using the device attributes report to detect OSC 52 support
### Issue description

As far as I understand it, tmux will automatically make use of the `OSC 52` clipboard functionality on terminals that have the _Ms_ capability. But that's not particularly reliable, because many terminals will set their `TERM` to some `xterm` variant without necessarily matching all of xterm's features. And those that do support `OSC 52` are not guaranteed to have the functionality enabled anyway.

So instead I want to you to consider using the primary device attributes report, and looking for the feature number `52` in the list of supported extensions. This is a recent extension that a number of terminals have agreed to support, and indicates that the terminal is capable of copying to the clipboard with `OSC 52` using a _Pc_ value of `c` or an empty _Pc_ value.

I'm not suggesting you make this change immediately, because I don't think any terminals have actually shipped with this extension in their `DA` report yet, but I just wanted to raise the issue now to see if you would be interested in making use of the functionality once it is actually available.

I should also be clear that this extension only indicates support for _writing_ to the clipboard. It does not guarantee that you'll be able to read from the clipboard with an `OSC 52` query.

### Required information

I don't think any of this is relevant, because this suggestion is applicable to multiple platforms and terminals, but fwiw...

* tmux version (`tmux -V`): next-3.6
* Platform (`uname -sp`): Linux x86 64
* $TERM inside tmux: tmux-256color
* $TERM outside of tmux: xterm-256color

j4james:
Detect support for OSC 52 using the device attributes report

This is an extension agreed upon by modern terminals to indicate that
they support copying to the clipboard with XTerm's OSC 52 sequence.
On terminals that can disable clipboard access, this extension is only
reported when writing to the clipboard is actually allowed, so this is
a more reliable indicator than the Ms terminfo capability.

Fixes #4532

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -1459,6 +1459,8 @@ tty_keys_device_attributes(struct tty *tty, const char *buf, size_t len,
 				tty_add_features(features, "margins", ",");
 			if (p[i] == 28)
 				tty_add_features(features, "rectfill", ",");
+			if (p[i] == 52)
+				tty_add_features(features, "clipboard", ",");
 		}
 		break;
 	}


j4james:
@nicm I'm not sure if this is what you had in mind, but I've tested on Windows Terminal by setting my `TERM` to vt525, and then checking whether I could copy to the clipboard from the tmux buffer with <kbd>Ctrl</kbd>+<kbd>W</kbd>. On the release build of Windows Terminal it doesn't work (because the _Ms_ cap isn't in the vt525 terminfo), but on the canary build it does (because that now includes the `52` extension parameter in its device attributes report).

nicm:

PR #4542, opened 2025-06-26
Support more OSC 133 start of prompt commands

SuperStormer:
Add N and P OSC 133 commands based on the [semantic prompts spec](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md). For example, they're used in [wezterm's shell integration](https://github.com/wezterm/wezterm/blob/main/assets/shell-integration/wezterm.sh).


nicm:
Which terminals support these?

PR #4546, opened 2025-07-04
Fix Hangul Jamo width problem (#4543)

### LINKED ISSUE #4543: Incorrect spacing in ls output for Korean filenames in tmux on macOS
### Issue description

In tmux 3.5a and next-3.6 (built with `--enable-utf8proc`) on macOS, filenames containing Hangul (Korean) characters are misaligned in the output of commands like `ls`, `eza`, and `lsd`. The columns become uneven due to incorrect character width calculations inside tmux.

### Steps to reproduce

1. Open a tmux session in any terminal emulator (iTerm2, WezTerm, or macOS Terminal).
2. Create files with Hangul in their names. (e.g., via `touch` or Finder).
3. Run `ls` several times.
4. Observe that alignment breaks over repeated runs - extra spaces appear after Hangul characters.

Image

### Expected behavior

- Columns should align properly, treating Hangul characters as width 2.
- Output inside tmux should match the alignment seen outside of tmux.

### Additional observations

- `printf '터미널에서 프린터(print directly in terminal)'` outputs correctly inside tmux.
- Outside of tmux, `ls` output is consistently aligned.
- Files created purely via the terminal (`touch`) do **not** show misalignment.
- However, once these files are modified via Finder (macOS GUI), the issue appears inside tmux.
- The issue does **not** appear immediately: the output is initially fine, but becomes misaligned after a few `ls` runs.
- This suggests repeated screen redraws trigger width miscalculations specifically for filenames altered by Finder.
- Possibly related to macOS's use of decomposed Unicode (NFD) normalization vs NFC.

### Additional context

- Verified that `utf8proc` is correctly linked to tmux (`otool -L $(which tmux)`).
- The issue persists across different terminal emulators (iTerm2, WezTerm, and macOS Terminal).
- Output is consistently correct outside of tmux in the same terminals.

### Environment

- macOS 14.7.2 (Sonoma)
- `uname -sp`: Darwin arm
- tmux versions: 3.5a (Homebrew), next-3.6 (built from source with utf8proc)
- utf8proc version: 2.10.0 (Homebrew)
- Terminal emulators tested: iTerm2, WezTerm, macOS Terminal
- `$TERM`: xterm-256color

phi4grv:
This pull request fixes the incorrect width handling of individual Hangul Jamo characters (U+1100–U11FF), as described in issue #4543.

### Context

While it's uncommon for users to type standalone Hangul Jamo characters directly, they often appear in filenames on macOS. This is because macOS uses Unicode Normalization Form D (NFD), which decomposes Hangul syllables into individual Jamo characters.

For example, the syllable "한" (U+D55C) may be stored as:
- U+1112 HANGUL CHOSEONG HIEUH
- U+1161 HANGUL JUNGSEONG A
- U+11AB HANGUL JONGSEONG NIEUN

In the current behavior, tmux assigns widths of 2 to Choseong, 1 to Jungseong, and 1 to Jongseong — resulting in a total width of 4 for a single syllable. This leads to visual misalignment in terminal applications such as `ls` or Vim when filenames contain decomposed Hangul characters.

### Fix

This patch adds a workaround for decomposed Hangul sequences by adjusting character widths as follows:

- When a sequence of Hangul Jamo characters appears (i.e., **Choseong (initial) + Jungseong (medial) + Jongseong (final)**):
  - The **first character (Choseong)** is given width **2**
  - The **subsequent characters (Jungseong and Jongseong)** are given width **0**

This approximates the width of a fully composed syllable and resolves alignment issues when rendering filenames.

While this is not a complete or linguistically precise solution, it provides a practical fix for the macOS NFD normalization case and avoids rendering problems in real-world usage.

### Result

With this patch:
- Decomposed Hangul sequences in macOS filenames display correctly
- Terminal layout remains consistent when listing or editing files

### Related

Fixes #4543

### DIFF
--- a/input.c
+++ b/input.c
@@ -2583,14 +2583,8 @@ input_exit_rename(struct input_ctx *ictx)
 	server_status_window(w);
 }
 
-#define HANGULJAMO_TO_CODEPOINT(p) \
-    (((((u_char*)(p))[0] & 0x0F) << 12) | \
-     ((((u_char*)(p))[1] & 0x3F) << 6)  | \
-     (((u_char*)(p))[2] & 0x3F))
-
 static void
 set_width_chosung(struct input_ctx *ictx, struct utf8_data *ud) {
-	log_debug("HangulJamo: U+%04X[%.3s] width:%d: Chosung set width 2", HANGULJAMO_TO_CODEPOINT(ud->data), ud->data, ud->width);
 	ud->width = 2;
 	ictx->flags &= ~INPUT_UTF8_HANGULJAMO;
 	ictx->flags |= INPUT_UTF8_HANGULJAMO_CHOSEONG;
@@ -2599,10 +2593,8 @@ set_width_chosung(struct input_ctx *ictx, struct utf8_data *ud) {
 static void
 set_width_jungseong(struct input_ctx *ictx, struct utf8_data *ud) {
 	if (ictx->flags & INPUT_UTF8_HANGULJAMO_CHOSEONG) {
-		log_debug("HangulJamo: U+%04X[%.3s] width:%d: Jungseong set width 0", HANGULJAMO_TO_CODEPOINT(ud->data),ud->data, ud->width);
 		ud->width = 0;
-	} else {
-		log_debug("HangulJamo: U+%04X[%.3s] width:%d: Jungseong without Choseong set width 2", HANGULJAMO_TO_CODEPOINT(ud->data), ud->data, ud->width);
+	} else { // Jungseong without Choseong
 		ud->width = 2;
 	}
 	ictx->flags |= INPUT_UTF8_HANGULJAMO_JUNGSEONG;
@@ -2611,10 +2603,8 @@ set_width_jungseong(struct input_ctx *ictx, struct utf8_data *ud) {
 static void
 set_width_jongseong(struct input_ctx *ictx, struct utf8_data *ud) {
 	if (ictx->flags & INPUT_UTF8_HANGULJAMO_CHOSEONG && ictx->flags & INPUT_UTF8_HANGULJAMO_JUNGSEONG) {
-		log_debug("HangulJamo: U+%04X[%.3s] width:%d: Jongseong set width 0", HANGULJAMO_TO_CODEPOINT(ud->data), ud->data, ud->width);
 		ud->width = 0;
-	} else {
-		log_debug("HangulJamo: U+%04X[%.3s] width:%d: Jongseong without Choseong and Jungseong set width 2", HANGULJAMO_TO_CODEPOINT(ud->data), ud->data, ud->width);
+	} else { // Jongseong without Choseong and Jungseong
 		ud->width = 2;
 	}
 	ictx->flags &= ~INPUT_UTF8_HANGULJAMO;
@@ -2623,7 +2613,6 @@ set_width_jongseong(struct input_ctx *ictx, struct utf8_data *ud) {
 static void
 handle_hanguljamo(struct input_ctx *ictx, struct utf8_data *ud) {
     if (ud->size != 3 || ud->data[0] != 0xE1) { // not Hangul Jamo
-		log_debug("HangulJamo: U+%04X[%.3s] width:%d: Not HangulJamo", HANGULJAMO_TO_CODEPOINT(ud->data), ud->data, ud->width);
 		ictx->flags &= ~INPUT_UTF8_HANGULJAMO;
 		return ;
 	}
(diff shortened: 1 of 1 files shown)


nicm:
Thanks, but this isn't the right place to do this, it won't work for UTF-8 in other places like formats. Also what happens if an application prints Choseong and then nothing else, does tmux just sit there forever waiting for the next character and never display the one it has?

It would seem best if this could be done in `screen_write_combine` in `screen-write.c` like ZWJ. But it might be tricky there too, because you have three characters to check for rather than two, so if you get for example Choseong-Jungseong-X, it may be difficult to correctly "undo" the Jungseong to make it width 2 once you receive the X.

Another option might be to allow `utf8_open`/`utf8_append` to return multiple characters (up to 3 would be enough) - there aren't that many places that use it and on the face of it, I think they could all be changed to handle this in some acceptable way. And it could do ZWJ too. But that would give us the same problem as your diff - what happens if we get Choseong and nothing else? `input.c` at least would need some sort of timer.

So I think either:

1) see if you can integrate this into `screen_write_combine`;

2) or change `utf8_open`/`utf8_append` to return up to 3 characters and change the callers to handle them.

phi4grv:
Makes sense — I agree this isn’t the right approach for a general solution.  
This patch was more of a quick fix for a specific issue I was facing.

That said, in this patch, all input is passed through immediately.  
If a character is Jungseong or Jongseong and can be combined, it's simply given width 0.  
No buffering or waiting is involved.

Doing this in `screen_write_combine` seems like it could work — I’ll give it a try when I get the chance.

As for extending `utf8_open`/`utf8_append`, they’re mainly focused on decoding a single codepoint from a UTF-8 sequence, so they might not be the right place to handle composition logic.
Plus, handling composition there isn’t straightforward — for example, if a Choseong comes in and we expect a Jungseong, but the next character isn't, it makes things tricky to recover.

phi4grv:
### Update

As discussed earlier, this PR has been updated to handle Hangul Jamo processing inside `screen_write_combine`.

The Hangul Jamo composition logic is simplified by classifying Jamo characters into three types:  
choseong (initial consonant), jungseong (vowel), and jongseong (final consonant).

This implementation combines only two composable cases:  
- choseong + jungseong  
- choseong + jungseong + jongseong  

For choseong characters, `screen_write_combine` returns 0 (not zero width),  
while for non-composable combinations (e.g., jungseong + jongseong without choseong), it returns 1 (zero width).

nicm:
This looks good apart from code style issues.

My main question is: is it the case that Korean is the only language with this kind of behaviour? Or would making this more generic be useful, so that it can be extended to other languages should anyone wish to do that? I'm thinking mainly of simple things like changing the names of the functions and members not be tied so much to Korean (so for example maybe things like `HANGULJAMO_STATE_CHOSEONG` -> `UTF8_COMBINE_STATE_KOR_CHOSEONG` or even something like `UTF8_COMBINE_STATE_INTERMEDIATE`).

I also wonder if `utf8_is_modifier`/`utf8_has_zwj` could be built into the same API, so there is one function that returns a state telling `screen_write_combine` what to do.

phi4grv:
As far as I know, other languages also use combining characters. However, I am only familiar with the specific structure used in Korean Hangul, so I focused this implementation on that case.

I used the terms Choseong, Jungseong, and Jongseong because they appear in the Unicode specification. However, I’m open to renaming them if the terms might be difficult to understand.

Refactoring to generalize the three cases is possible. Since there are only three cases and each follows slightly different rules, it might make the code a bit harder to read.

nicm:
OK I think this is fine as is, if anyone wants to add similar support for other languages we can make it more general at that point. If you can make the code style closer to the tmux code style it would be helpful.

phi4grv:
Okay, is there a style guide or example in tmux I should follow? I’ll update the code.

nicm:
There is https://man.openbsd.org/style but generally if you just follow the existing code style. If you get it most of the way I will do anything else I notice before I commit it. Notably:

- Space between `if` or `switch` and `(`.
- No `{}` around single-line blocks.
- Do not run `if` and the block together, always a newline. So `if (x)\ny;` not `if (x) y;`.
- Newline before `{` for functions (but not `if`/`switch`/etc).
- All declarations at top of function not in inner blocks.
- `hanguljamo_check_state` and `hanguljamo_state` in `tmux.h` need to go under the comment for `utf8-combined.c` not `tty-acs.c`.

Thanks!

phi4grv:
I've updated the PR with the style fixes you mentioned.

nicm:
Thanks, I have applied this to OpenBSD now with some minor style changes, it should be GitHub when it syncs in a few hours.

If you have time, it might be nice to add a test to the `regress` directory to reduce the risk of this getting broken again by accident.

Thanks!

PR #4550, opened 2025-07-10
Use MAC_OS_X_VERSION_MIN_REQUIRED macro for OS version guards

sevan:
tmux now builds on OS X 10.4 & up. Tested every release up to 10.11.
Terminal.app on OS versions prior to 10.11 print `1;2c` when tmux is first invoked (fresh install, default config).
`__unused` is defined in `sys/cdefs.h`.


nicm:
Applied now, thanks!

PR #4551, opened 2025-07-12
Fix format_cmp_window: windows should be sorted based on index, not id

naruaway:
I am using the current master branch (7e439539377e272f37d18bb10dbff374b87acee6) and realized that the ordering of the windows in status bar is broken.

It turned out that `format_cmp_window` is sorting windows based on id, not by index by default now due to [a recent change](https://github.com/tmux/tmux/pull/4516/files#r2202331906).

`format_cmp_client` has the exact same logic [here](https://github.com/tmux/tmux/blob/7e439539377e272f37d18bb10dbff374b87acee6/format.c#L4685-L4686) so I think this was just a mistake.

## ✅ Verification

I created a session with 3 windows on macOS 15.5 and checked the status bar. I did not use any custom tmux config so the status bar format is the default one.
After creating windows, I used swap-window to "move" windows around.
Note that I also confirmed that before [this recent change](https://github.com/tmux/tmux/pull/4516/files#r2202331906), the issue did not exist so this issue is a regression introduced on `master` and not yet released.

### ❌  Before this PR (7e439539377e272f37d18bb10dbff374b87acee6)

We see that the windows are NOT ordered by index

ok

### ✅ After this PR (9ac982a73812fea80c6d5b6195ad13ca461b6be5)

We see that the windows are ordered by index even after `swap-window`

ng

### DIFF
--- a/format.c
+++ b/format.c
@@ -4490,7 +4490,6 @@ format_cmp_window(const void *a0, const void *b0)
 
 	switch (sc->field) {
 	case FORMAT_LOOP_BY_INDEX:
-		result = wa->id - wb->id;
 		break;
 	case FORMAT_LOOP_BY_TIME:
 		if (timercmp(&wa->activity_time, &wb->activity_time, >)) {


nicm:
This fix is not quite right because it will return every window as comparing equal (the compare function will always return zero) and `qsort` does not guarantee that the result will be stable in this case. Also it will not correctly reverse the list if `/r` is given. Instead, I have changed it to work like `format_loop_clients`. Will be in GitHub later. Thanks!

naruaway:
Sorry for the wrong fix and thanks for the correct fix 🙏 
I just confirmed that your fix (9831175a2d9228efd75cb08849b8954b61feda39) works as expected on my machine as well 🎉

PR #4568, opened 2025-07-30
Fix rpath issue on macOS 15

ikolesnikes:
Running tmux fails with the error

dyld[85540]: Library not loaded: @rpath/libevent_core-2.2.1.dylib
  Referenced from: <04573DE8-0B9F-3941-AF91-5205FEE73089> tmux
  Reason: no LC_RPATH's found

This change adds /usr/local/lib to rpath.


nicm:
I don't think it is tmux's responsibility to set this. I imagine it should probably be in the libevent .pc file.

ikolesnikes:
> I don't think it is tmux's responsibility to set this. I imagine it should probably be in the libevent .pc file.

That makes sense I think.

PR #4570, opened 2025-07-30
man page: fix reverse documentation of display-popup -s and -S

rjbs:
The summary of options has things backward.  The rest of the text had it correct.  This makes the short form correct.  That's all!


nicm:

PR #4577, opened 2025-08-04
Fix resource leak in file_read() on early exit

Eomtaeyong820:
### Describe the bug

In `file_read()`, when `evbuffer_add()` or `ferror()` fails, the function exits early via `goto done` **without calling `fclose(f)`**, resulting in a **file descriptor leak**.

This may not cause an immediate crash, but over time — especially during long-running sessions or under memory pressure — it can trigger **"Too many open files"** errors.

---

### Affected Code

```c
f = fopen(cf->path, "rb");
if (f == NULL) {
    cf->error = errno;
    goto done;
}
for (;;) {
    size = fread(buffer, 1, sizeof buffer, f);
    if (evbuffer_add(cf->buffer, buffer, size) != 0) {
        cf->error = ENOMEM;
        goto done;  // 🔴 f not closed
    }
    if (size != sizeof buffer)
        break;
}
if (ferror(f)) {
    cf->error = EIO;
    goto done;  // 🔴 f not closed
}
fclose(f);
goto done;
Vulnerability
fopen() succeeds → file is open

an error occurs → goto done; skips fclose(f)

→ file descriptor leak
```
---

### Fix
This patch introduces a cleanup: label before done: and moves fclose(f) there.
All early exit paths (which previously used goto done) now use goto cleanup, ensuring that the file is properly closed on failure.

---

### Patched Code (Excerpt)
```c
if (evbuffer_add(...) != 0) {
    cf->error = ENOMEM;
    goto cleanup;
}
if (ferror(f)) {
    cf->error = EIO;
    goto cleanup;
}
fclose(f);
goto done;

cleanup:
if (f != NULL)
    fclose(f);
done:
file_fire_done(cf);
```

---

### Why This Should Be Merged
Prevents file descriptor leaks on error paths

Follows tmux’s existing goto cleanup convention

Minimally invasive change scoped to one function

Improves program stability under edge-case and resource-constrained conditions

---

### Thanks!

### DIFF
--- a/file.c
+++ b/file.c
@@ -392,17 +392,20 @@ file_read(struct client *c, const char *path, client_file_cb cb, void *cbdata)
 			size = fread(buffer, 1, sizeof buffer, f);
 			if (evbuffer_add(cf->buffer, buffer, size) != 0) {
 				cf->error = ENOMEM;
-				goto done;
+				goto cleanup;
 			}
 			if (size != sizeof buffer)
 				break;
 		}
 		if (ferror(f)) {
 			cf->error = EIO;
-			goto done;
+			goto cleanup;
 		}
-		fclose(f);
-		goto done;
+		goto cleanup;
+
+                cleanup:
+                if (f != NULL)
+                    fclose(f);
 	}
 
 skip:


nicm:
Thanks, but I think it would be tidier just to do the `fclose` at the existing `done` label:
~~~~
--- file.c      17 Apr 2023 17:58:35 -0000      1.15
+++ file.c      12 Aug 2025 12:38:29 -0000
@@ -364,7 +364,7 @@ file_read(struct client *c, const char *
        size_t                   msglen;
        int                      fd = -1;
        u_int                    stream = file_next_stream++;
-       FILE                    *f;
+       FILE                    *f = NULL;
        size_t                   size;
        char                     buffer[BUFSIZ];

@@ -404,7 +404,6 @@ file_read(struct client *c, const char *
                        cf->error = EIO;
                        goto done;
                }
-               fclose(f);
                goto done;
        }

@@ -427,6 +426,8 @@ skip:
        return cf;

 done:
+       if (f != NULL)
+               fclose(f);
        file_fire_done(cf);
        return NULL;
 }

nicm:

PR #4580, opened 2025-08-08
Go to previous entry in completion menu on Shift-Tab

proski:


nicm:

benknoble:
Apologies in advance for derailing the thread:

FWIW, I still[^1] find it frustrating that tracing the lineage of code that lands in tmux sources is made difficult by the lack of attribution (_e.g._, the [CVS commit for this PR](https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/tmux/input.c?rev=1.234&content-type=text/x-cvsweb-markup&sortby=date) lists nicm as the author, which is not true; further, it doesn't contain a URL that leads back here (though an issue number helps somewhat)). This "disconnect" between the relevant pieces makes it much harder to dig up relevant information when needed, and probably muddies the waters on licensing (since it's not easy to understand which authors have interest in what parts of code).

The disconnect is of course further reflected in the Git history.

I don't know if CVS (or OpenBSD) supports it, but it would be nice if you could apply sent patches _and_ retain authorship. One common way that's done with Git is a `Co-authored-by:` [trailer](https://git-scm.com/docs/git-interpret-trailers#_description) if you can't `git cherry-pick` or similar to preserve separate author/committer metadata to begin with. You could also add trailers like `Issue: <url>` or `PR: <url>` to create some sort of linkage back to the original.

I suppose the other way to do it might be to sync from Git to CVS instead of the other way 'round, assuming the relevant parties (OBSD base?) support it. Then you could preserve all the information Git gives you, and decide on a conversion to CVS (lossless or lossy, depending).

[^1]: I wrote about it in 2019 after my first contribution here, but the text reflects an immature understanding of both the development process for tmux and for distributed version control writ-large, so I won't bother linking it here.

PR #4581, opened 2025-08-09
Fix PageDown in pop-up menu, it should go 5 positions down

proski:
The current behavior is to get the cursor to the last position. That's not what PageUp down, and it's not what the code was meant to do. `i` starts at 5 but never reaches zero because it's incremented rather than decremented.


nicm:

Issue #4584, opened 2025-08-11
Transparent pane-borders


Paulchen5:
### Issue description

I am using iTerm2 with a custom background image. Is there a way to hide / make the pane borders transparent?

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

* tmux version (`tmux -V`): `tmux 3.5a`
* Platform (`uname -sp`): `Darwin arm`
* Terminal in use (xterm, rxvt, etc): `iTerm2`
* $TERM *inside* tmux (`echo $TERM`):  `tmux-256color`
* $TERM *outside* tmux (`echo $TERM`):  `xterm-256color`

nicm:
No there is no easy way to do this, although it would be relatively easy to add a new choice to `pane-border-lines` to make it use spaces.

paulchen5:
Thank you for your response. I guess I am as many others not that much experienced in C.

If you are not willing to implement this by your own I would be happy if you could provide some code sections for me which would help me to implement this by my own👍

nicm:
Add a new type like `spaces` to the end of `options_table_pane_border_lines_list` in `options-table.c` then in `tmux.h` add a define for the same index to `enum pane_lines` and in `screen_redraw_border_set` in `screen-redraw.c` add a case that does `utf8_set(&gc->data, ' ');`.

nicm:
IIRC for popups we use `padded` so that might be better than `spaces`.

paulchen5:
Thanks for your help👍

I will have a look into the source code later and will let you know if I need furhter help.

PR #4587, opened 2025-08-13
Add `spaces` option for pane borders

Paulchen5:
Hi, I followed the instructions of @nicm from #4584 .

How can I install the modified source code and test it on my own?


nicm:
If you made the changes you can do `sh autogen.sh` then `./configure` then `make` then test your new tmux by doing for example `./tmux -Ltest new`.

paulchen5:
Ok seems to work. Because I am new to tmux I don't know what to test else to cover some possible edge cases. Feel free to review the changes 👍

nicm:

PR #4596, opened 2025-08-21
Fix OSC 4

ur4t:
Some TUI apps may want to get color values via OSC 4.

Reference: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html


nicm:

PR #4605, opened 2025-08-26
Just some minor things

BertramScharpf:
... that I detected while reading the code. Mostly space inconsistencies.


nicm:
General policy is not to do this kind of thing unless changing the file for some other reason, so could you just keep the warning fixes and visible indentation fixes and leave out the spaces to tabs (unless you are changing the file anyway)? Thanks

PR #4608, opened 2025-08-28
fix enable-debug issue in configure

zhangjiakouzf:
when appending enable-debug=yes in configure, the Optimization options are "-std=gnu99 -O2 -g". It is better to remove "-O2" and add "-O0"


nicm:
No, it is intended to be `-O2`. `--enable-debug` is the default for builds from master and `-O0` is too slow.

PR #4612, opened 2025-08-28
feat(popup): add -k flag

### LINKED ISSUE #4611: display-popup: dismiss with any keypress, not just escape
I would like for `display-popup` to have an option (such as `-k`) that causes any key, not just Escape, to dismiss the popup if the command inside has exited.

Currently I emulate this by using `-E` and letting the shell command wait for a key press before exiting. This has some downsides:

1. `display-popup` uses `$SHELL`, unlike `run-shell` which uses `/bin/sh`, which makes the 'wait for keypress' command depend on your shell. For example, it's `read -rsk` for zsh, and `read -rsN1` for bash.
2. Making this work with `-EE` requires constructs like `script && read`, `script || read`, or `. script; read`
3. It doesn't work at all with the multi-argument command form `display-popup script foo bar`.

My proposal is for the following behaviour:

||||
|-|-|:-|
|-k|| Dismiss via any key|
|-k|-E|Dismiss automatically on exit (cancels out -k)
|-k|-EE|Dismiss automatically on success, dismiss via any key otherwise

I'm not familiar with tmux's input code yet, but I could try to make a PR for this feature if desired.

LunarLambda:
Allows any key to dismiss the popup once the command has exited. Closes #4611

That aside, I'm marking this as a draft because I couldn't quite figure out how to check that a key event is a keyboard key and not a mouse or a paste, which probably shouldn't dismiss the popup.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -54,8 +54,8 @@ const struct cmd_entry cmd_display_popup_entry = {
 	.name = "display-popup",
 	.alias = "popup",
 
-	.args = { "Bb:Cc:d:e:Eh:s:S:t:T:w:x:y:", 0, -1, NULL },
-	.usage = "[-BCE] [-b border-lines] [-c target-client] "
+	.args = { "Bb:Cc:d:e:Eh:ks:S:t:T:w:x:y:", 0, -1, NULL },
+	.usage = "[-BCEk] [-b border-lines] [-c target-client] "
 		 "[-d start-directory] [-e environment] [-h height] "
 		 "[-s style] [-S border-style] " CMD_TARGET_PANE_USAGE
 		 " [-T title] [-w width] [-x position] [-y position] "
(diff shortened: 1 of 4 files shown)


nicm:
Are you sure you killed your existing tmux before testing? Or use `-Ltest` to use a different socket.

LunarLambda:
Yeah the -Ltest was necessary. Silly mistake. Any advice on the keyboard check?

nicm:
You can do `!KEYC_IS_MOUSE(key) && !KEYC_IS_PASTE(key)`.

LunarLambda:
Got it! Thank you. Tested it and it works.

nicm:

LunarLambda:
Nice! Curious, since 3.5a is almost a year old, is there the possibility of cutting a 3.5b release? Not sure what the release convention is. If not, no biggie, I can compile from master no issue.

nicm:
Not sure, some time in the next few months maybe.

PR #4613, opened 2025-08-30
Add .idea to .gitignore

AsadunnobieIshaan:
This update adds the JetBrains .idea directory to .gitignore to prevent IDE-specific configuration files from being committed. These files are machine-dependent, add unnecessary noise to the repo, and can cause merge conflicts. Ignoring them ensures the repository remains clean, portable, and focused only on source code and project-relevant files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ fuzz/*-fuzzer
 tags
 tmux
 tmux.1.*
+.idea


TahsinAyman:
seens like a valid pull request.

nicm:
Thanks but I don't really see a need to clutter `.gitignore` with files for all sorts of IDEs, unless it is one that someone contributing to tmux regularly is using.

PR #4614, opened 2025-08-30
input.c: add replies for DECRQSS cursor, DECRQM ?12, and others

alberti42:
# Summary

This patch extends `input.c` to support proper replies to terminal state/status queries, so that applications running *inside tmux* can query cursor style, cursor blink status, and CSI u (keyboard enhancement protocol) just as they would against a real terminal emulator.

Specifically:

* **DECRQSS “ q”** (`DCS $ q q ST`)  
    → Reply with the current cursor shape (`Ps=0–6`).  
    If the application has explicitly set a style via DECSCUSR, that is reported; otherwise tmux falls back to the configured `cursor-style` option (so queries now reflect the effective cursor users actually see on screen).
* **DECRQM ?12$p**  
    → Reply with the current blink state.  
    If blink has been explicitly controlled by the application (`MODE_CURSOR_BLINKING_SET`), we report that; otherwise we infer from the configured `cursor-style` option (`blink-*` vs steady).
* **CSI ? u**  
    → Reply with the current keyboard enhancement protocol state:
    * `ESC[?0u` = off
    * `ESC[?1u` = mode 1 active (disambiguate escape codes)
    * `ESC[?1;2u` = mode 2 active (alternate keys)
* **DECRQM ?2004, ?1004, ?1006**  
    → Reply with bracketed paste, focus event, and SGR mouse reporting status, based on the screen mode bits.

These replies are emitted from `input_csi_dispatch` and `input_dcs_dispatch` alongside the existing query handlers.

---

# Rationale

* Many TUIs (e.g. [Yazi](https://github.com/sxyazi/yazi?utm_source=chatgpt.com), editors, Ratatui apps) **probe terminal capabilities and status** on startup.
* Before this patch, tmux would not reply to queries like `\eP$q q\e\` (cursor shape) or `\e[?u` (CSI u), leading applications to assume features were unsupported.
* This forced developers to resort to passthrough hacks or disabled features inside tmux.
* With this patch, tmux responds consistently with the **effective state the user sees**:
    * If an app sets DECSCUSR, that takes precedence.
    * Otherwise, tmux reports the configured defaults (`cursor-style`).
    * Status queries for modes (bracketed paste, focus, mouse) reflect `s->mode`.
* Applications can now initialize correctly inside tmux without passthrough, because queries and replies match the semantics of real terminals like xterm or alacritty.

---

# Notes

* Replies use standard DECSCUSR / DECRQM semantics (`Ps=0–6` for shape; `;1`/`;2` for on/off status).
* We intentionally **keep application state and option defaults separate**: tmux still allows apps to override cursor style at runtime, but query replies fall back to the configured option if no app override exists.
* This matches user expectations (“queries should show what’s on screen”) without taking control away from applications.

---

# Testing

Verified using a shell script that queries:

* `\eP$q q\e\` → cursor shape
* `\e[?12$p` → blink status
* `\e[?u` → CSI u
* `\e[?2004$p`, `\e[?1004$p`, `\e[?1006$p`

All return the correct values in tmux, consistent with the cursor appearance and active modes. Yazi and Crossterm apps now initialize correctly inside tmux.

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -326,6 +326,7 @@ tty_start_start_timer(struct tty *tty)
 	struct timeval	 tv = { .tv_sec = TTY_QUERY_TIMEOUT };
 
 	log_debug("%s: start timer started", c->name);
+	evtimer_del(&tty->start_timer);
 	evtimer_set(&tty->start_timer, tty_start_timer_callback, tty);
 	evtimer_add(&tty->start_timer, &tv);
 }


alberti42:
[term_query](https://github.com/alberti42/term_query) is a bash script I wrote to test the new features in `tmux`. You can simply run:

```bash
term_query.sh
```

This is what I get on my setup:

```text
Preparing terminal...
Sending DIRECT queries to the current terminal...
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 50 31 24 72 20 71 36 ┊ 20 71 1b 5c 1b 5b 3f 31 │•P1$r q6┊ q•\•[?1│
│00000010│ 32 3b 32 24 79 1b 5b 3f ┊ 30 75 1b 5b 3f 31 3b 32 │2;2$y•[?┊0u•[?1;2│
│00000020│ 63 1b 5b 3e 38 34 3b 30 ┊ 3b 30 63 1b 5b 3f 32 30 │c•[>84;0┊;0c•[?20│
│00000030│ 30 34 3b 32 24 79 1b 5b ┊ 3f 31 30 30 34 3b 32 24 │04;2$y•[┊?1004;2$│
│00000040│ 79 1b 5b 3f 31 30 30 36 ┊ 3b 32 24 79             │y•[?1006┊;2$y    │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
→ DECRQSS cursor shape: Ps=6  [steady bar]
→ DECRQM ?12 (blink): steady (off)  [state=2]
→ DA1 (primary): ?1;2c  [VT100 w/ Advanced Video Option]
→ DA2 (secondary): >84;0;0 c  [type=tmux  version=0  cartridge=0]
→ DA3 (tertiary): no reply
→ CSI u: supported, but currently OFF  [flags=0]
→ DECRQM ?2004: bracketed-paste: disabled
→ DECRQM ?1004: focus-events: disabled
→ DECRQM ?1006: mouse-sgr: disabled
Test finished.
```

nicm:
Thanks.

I do not see `CSI ? u` in ctlseqs.txt and xterm does not respond to it, where is this from?

It looks like it should respond `DCS 0 $ r  ST` for invalid `DECRQSS` requests, not ignore them.

Can you please at least fix the indentation and if possible also the rest of the code style?

alberti42:
You’re right — `CSI ? u` is not in ctlseqs.txt and current xterm does not implement it.

This sequence originates from Paul Evans’ *fixterms* proposal (2020), which introduced the `CSI u` encoding to fix longstanding problems with keyboard input handling:

- [Fix Keyboard Input on Terminals – Please (Paul Evans, 2020)](https://www.leonerd.org.uk/hacks/fixterms/)

Kovid Goyal extended this into the [kitty Keyboard Protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), which adds flags and a query mechanism. Modern emulators including kitty, alacritty, foot, iTerm2, rio, wezterm, and ghostty implement it.

Applications and libraries such as Crossterm, notcurses, and Yazi now routinely probe with `CSI ? u`. Without a reply, they assume no support, which disables extended key handling inside tmux panes.

This patch lets tmux reply to that query, so these modern TUIs can work correctly.

Regarding DECRQSS: agreed, unsupported queries should reply with `DCS 0 $ r Pt ST` rather than being silently ignored. In the new commit, I adjusted the handler accordingly and also fixed the indentation/style.

Let me know if you need further adjustments/tests.

nicm:
OK I could have guessed that it was added by kitty, their choices of escape sequence are usually deranged... but I guess it is done now.

Can you leave out CSI ? u for the moment and we will do the others first?

ThomasDickey:
The comment above seems to imply that `\e[?u` is a DECRQSS, which would get a response.  But it is not.  It is a separate control, which is treated as an error by xterm, and simply discarded.

alberti42:
I have removed the `CSI ? u` query from this PR so we can proceed with the
other (standard) queries as you suggested. Let's discuss `CSI ? u` separately
in a future PR.

For context: `CSI ? u` is implemented in several modern terminal emulators.
I tested it personally with Ghostty, Kitty, WezTerm, Alacritty, and iTerm2;
it is also supported by foot and rio. Terminals that do not recognize the
query (for example, macOS Terminal) simply return no reply — you can verify
this with my test script [term_query](https://github.com/alberti42/term_query).

I agree that bringing this discussion to a separate PR makes it easier to track changes and avoids blocking the
current PR on a newer, non-standard extension.

nicm:
You've changed a lot of the existing indentation to be wrong, can you fix that and just correct the indentation in your changes?

alberti42:
Apologies — I was too quick and dropped my commits on my fork while trying to restore the latest version of `input.c` without my changes. This automatically closed the PR, and I was unable to reopen it despite restoring my previous commit chain. I’ve therefore opened a new PR #4618 that squashes everything we discussed into a single commit. I’ve added a link there back to this page so the previous discussion remains accessible.

PR #4615, opened 2025-08-31
Ensure break-pane -P prints when only one pane is left.

chaoyi:
When a window has only one pane, break-pane -P fails to print any output.  This was happening because the function would exit early in the single-pane case, skipping the final printing logic. The fix adjusts the code path to ensure the printing block is always reached.


nicm:

PR #4616, opened 2025-08-31
Fix HLS image display

bptato:
Currently, tmux limits Sixel color components to one byte in size. However, this is wrong, as the HLS color space as supported by Sixel uses 360 degrees in its first component (hue).

The easiest way to fix this is to just give another bit to the first component; this patch does that.

Additionally, I've extended the sanity check to the components so that invalid colors are now rejected too (previously, only the type was checked).


nicm:
Applied, thanks!

PR #4618, opened 2025-09-02
input.c: add replies for DECRQSS cursor, DECRQM ?12, and others

alberti42:
# Summary

Note: this is a new PR that reopens previous PR #4614 where the conversation initially started; PR #4614 was inadvertently closed; this is why it is reopened here.

This patch extends `input.c` to support proper replies to terminal state/status queries so that applications running *inside tmux* can query cursor style and cursor blink status just as they would against a real terminal emulator.

Specifically:

* **DECRQSS “ q”** (`DCS $ q q ST`)  
    → Reply with the current cursor shape (`Ps=0–6`), falling back to the cursor-style option if no explicit DECSCUSR; handle invalid requests.
    If the application has explicitly set a style via DECSCUSR, that is reported; otherwise tmux falls back to the configured `cursor-style` option (so queries now reflect the effective cursor users actually see on screen).
* **DECRQM ?12$p**  
    → Reply with the current blink state.  
    If blink has been explicitly controlled by the application (`MODE_CURSOR_BLINKING_SET`), we report that; otherwise, we infer from the configured `cursor-style` option (`blink-*` vs steady).
* **DECRQM ?2004, ?1004, ?1006**  
    → Reply with bracketed paste, focus event, and SGR mouse reporting status, based on the screen mode bits.

These replies are emitted from `input_csi_dispatch` and `input_dcs_dispatch` alongside the existing query handlers.

---

# Rationale

* Many TUIs (e.g. [Yazi](https://github.com/sxyazi/yazi?utm_source=chatgpt.com), editors, Ratatui apps) **probe terminal capabilities and status** on startup.
* Before this patch, tmux would not reply to queries like `\eP$q q\e\` (cursor shape) or `\e[?u` (CSI u), leading applications to assume features were unsupported.
* This forced developers to resort to passthrough hacks or disabled features inside tmux.
* With this patch, tmux responds consistently with the **effective state the user sees**:
    * If an app sets DECSCUSR, that takes precedence.
    * Otherwise, tmux reports the configured defaults (`cursor-style`).
    * Status queries for modes (bracketed paste, focus, mouse) reflect `s->mode`.
* Applications can now initialize correctly inside tmux without passthrough because queries and replies match the semantics of real terminals like xterm or alacritty.

---

# Notes

* Replies use standard DECSCUSR / DECRQM semantics (`Ps=0–6` for shape; `;1`/`;2` for on/off status).
* We intentionally **keep application state and option defaults separate**: tmux still allows apps to override cursor style at runtime, but query replies fall back to the configured option if no app override exists.
* This matches user expectations (“queries should show what’s on screen”) without taking control away from applications.

---

# Testing

Verified using a shell script [term_query.sh](https://github.com/alberti42/term_query) that queries:

* `\eP$q q\e\` → cursor shape
* `\e[?12$p` → blink status
* `\e[?2004$p`, `\e[?1004$p`, `\e[?1006$p`

All return the correct values in tmux, consistent with the cursor appearance and active modes. Yazi and Crossterm apps now initialize correctly inside tmux.

### DIFF
--- a/input.c
+++ b/input.c
@@ -2489,15 +2489,14 @@ input_dcs_dispatch(struct input_ctx *ictx)
 				return (0);
 			}
 			/* 
-			 * Unrecognized DECRQSS ($ q Pt): DCS 0 $ r Pt ST
-			 * kitty, wezterm, and possibly other terminals omit Pt and
-			 * simply reply DCS 0 $ r ST. We closely follow ctlseqs,
-			 * thus returning Pt as well.
-			 */
-			if (len > 1)
-				input_reply(ictx, "\033P0$r%.*s\033\\", (int)(len - 1), buf + 1);
-			else
-				input_reply(ictx, "\033P0$r\033\\");
+			 * Unrecognized DECRQSS ($ q Pt): DCS 0 $ r ST
+			 * Note that the VT510 manual states that the terminal should echo
+			 *`Pt` in the failure reply (`DCS 0 $ r Pt ST`). In practice, xterm
+			 * (and kitty, wezterm, etc) *do not* echo `Pt`; they reply just
+			 * `DCS 0 $ r ST`. We decided to omit `Pt` in order to match xterm
+			 * (and what most apps expect).
+			 */	
+			input_reply(ictx, "\033P0$r\033\\");
 			return (0);
 		}
 
--- a/input.c
+++ b/input.c
@@ -2488,14 +2488,7 @@ input_dcs_dispatch(struct input_ctx *ictx)
 				input_reply(ictx, "\033P1$r q%d q\033\\", ps);
 				return (0);
 			}
-			/* 
-			 * Unrecognized DECRQSS ($ q Pt): DCS 0 $ r ST
-			 * Note that the VT510 manual states that the terminal should echo
-			 *`Pt` in the failure reply (`DCS 0 $ r Pt ST`). In practice, xterm
-			 * (and kitty, wezterm, etc) *do not* echo `Pt`; they reply just
-			 * `DCS 0 $ r ST`. We decided to omit `Pt` in order to match xterm
-			 * (and what most apps expect).
-			 */	
+			/* Unrecognized DECRQSS ($ q Pt): DCS 0 $ r ST */	
 			input_reply(ictx, "\033P0$r\033\\");
 			return (0);
 		}
--- a/input.c
+++ b/input.c
@@ -1560,6 +1560,33 @@ input_csi_dispatch(struct input_ctx *ictx)
 		break;
 	case INPUT_CSI_QUERY_PRIVATE:
 		switch (input_get(ictx, 0, 0, 0)) {
+		case 12:  /* cursor blink */
+			/* 1=set(blink), 2=reset(steady) */
+			if (s->mode & MODE_CURSOR_BLINKING_SET) {
+				n = (s->mode & MODE_CURSOR_BLINKING) ? 1 : 2;
+			} else {
+				int opt_ps;
+				if (ictx->wp != NULL)
+					opt_ps = (int)options_get_number(ictx->wp->options, "cursor-style");
+				else
+					opt_ps = (int)options_get_number(global_options, "cursor-style");
+				/* blink for 1,3,5; steady for 0,2,4,6 */
+				n = (opt_ps == 1 || opt_ps == 3 || opt_ps == 5) ? 1 : 2;
+			}
+			input_reply(ictx, "\033[?12;%d$y", n);
+			break;
+		case 2004:  /* bracketed paste */
+			n = (s->mode & MODE_BRACKETPASTE) ? 1 : 2;
+			input_reply(ictx, "\033[?2004;%d$y", n);
+			break;
+		case 1004:  /* focus reporting */
+			n = (s->mode & MODE_FOCUSON) ? 1 : 2;
+			input_reply(ictx, "\033[?1004;%d$y", n);
+			break;
+		case 1006:  /* SGR mouse */
+			n = (s->mode & MODE_MOUSE_SGR) ? 1 : 2;
+			input_reply(ictx, "\033[?1006;%d$y", n);
+			break;
 		case 2031:
 			input_reply(ictx, "\033[?2031;2$y");
 			break;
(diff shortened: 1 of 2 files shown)


nicm:
How do you get xterm to echo back `D .. D` for the unrecognized case? I can't get it to send anything other than `DCS 0 $ r ST`.

https://vt100.net/docs/vt510-rm/DECRPSS.html does say:

> The terminal does not send a data string (D...D) to the host when the terminal receives an invalid request.

But I guess this is not considered an invalid request because it is sending back Ps = 0 not 1...

nicm:
Other than that the mechanics of this looks good and it works for me, I have made some minor changes mostly to keep lines under 80 columns, for example I broke the DECRQSS code into a helper function. Here is the new diff - it won't apply to portable tmux because OpenBSD doesn't have SIXEL and there is some code for that right under your changes, still you might want to look over it: [reply.diff.txt](https://github.com/user-attachments/files/22091636/reply.diff.txt)

alberti42:
Thanks for checking.

**On the data string for invalid DECRQSS:** the VT510 manual states that the
terminal should echo `Pt` in the failure reply (`DCS 0 $ r Pt ST`). In practice,
xterm (and kitty/wezterm) *do not* echo `Pt`; they reply just `DCS 0 $ r ST`.
For example, sending `DCS $ q invalid ST` to xterm returns only `DCS 0 $ r ST`.

If we want tmux to match xterm (and what most apps expect), we should omit `Pt`
in the invalid case as well. I’m happy to adjust the patch accordingly. I’d also
add a brief comment noting that this intentionally follows xterm behavior rather
than the VT510 text, to aid future readers.

**On `Ps` meaning:** the VT510 manual says `Ps=0` = valid, `Ps=1` = invalid,
but xterm (and ctlseqs) use the opposite convention: `1$r …` for valid and
`0$r …` for invalid. This patch follows the xterm/ctlseqs convention, since that
is what current software expects. I can add a code comment here too, for clarity.

Let me know if you prefer we keep echoing `Pt` for invalid requests; otherwise
I’ll update the patch to omit it and match xterm.

nicm:
I think we should just follow what xterm does.

alberti42:
- Regarding your reply:
 
> I think we should just follow what xterm does.

Yes, I also think it is the best way. I committed the changes 69bb1f2dc1cfb505689c60847190b8745486f174.

- Moreover, I corrected a tiny bug (and inconsistency with xterm behavior), which I fixed in commit 7f2d5e7596064f2d79ef8433cd1e1899bdbef23f. The explanation is provided in the commit comment.

- I also checked the changes you made in the reply.diff.txt file you sent me. Everything worked nicely: the code refactoring is an improvement and maintained all changes.

nicm:
Where does the manual say it should echo `Pt`, it reads to me as if it shouldn't?

I don't know why xterm has 0 and 1 the other way round from the VT420 but it doesn't really matter, we can just do what xterm does.

alberti42:
Sorry, the manual https://invisible-island.net/xterm/ctlseqs/ctlseqs.html does not say what I thought. Here is what it says:

```text
xterm responds with DCS 1 $ r Pt ST for valid requests, replacing the Pt with the corresponding CSI string, or DCS 0 $ r  ST for invalid requests.
``` 

My mistake originates from https://gist.github.com/nmichlo/5196879a6637ca42d5d0af22ee6848bf. I assumed that all comments in the code were a copy&paste from the manual. Obviously, in this case, it was not correct:

```python
#           xterm responds with DCS 1 $ r Pt ST for valid requests,
#           replacing the Pt with the corresponding CSI string, or DCS 0 $
#           r Pt ST for invalid requests.
```

I will make a new commit and eliminate the misleading and possibly wrong comment.

nicm:
Oh right... if you say "the manual" everyone will probably think "the DEC manual" not the "xterm manual", and if you say "xterm manual" they will think xterm(1) not the ctlseqs page. I'd probably say "the xterm documentation" in this case.

I don't think you need to set the `BLINKING_SET` flag like this. `BLINKING_SET` only applies to `SCREEN_CURSOR_DEFAULT`, that is when tmux does not know the cursor style, it just means "we don't know the cursor style but we know it is/isn't blinking". So you can change this code:
~~~~
               if (s->mode & MODE_CURSOR_BLINKING_SET)
                        blinking = (s->mode & MODE_CURSOR_BLINKING) != 0;
                else
                        blinking = 0; /* treat unspecified as steady */
~~~~
To just `blinking = (s->mode & MODE_CURSOR_BLINKING) != 0;`, because the blinking flag should never be unspecified unless the style is `SCREEN_CURSOR_DEFAULT` - if the user or application has set a cursor style, they have to have also set the blinking flag implicitly.

j4james:
> I don't know why xterm has 0 and 1 the other way round from the VT420 but it doesn't really matter, we can just do what xterm does.

FYI, this has been tested on several of the original DEC terminals, and it was confirmed that the VT420/VT510/VT520 manuals are incorrect, and got those numbers backwards. The DEC STD 070 manual documents the correct interpretation.

ThomasDickey:
Per [documentation](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Technical-manuals):
```
o   Sometimes the manuals are simply incorrect.  For example, testing a
    DEC VT420 in 1996 showed that the documented code for a valid or
    invalid response to DECRQSS was reversed.
```

nicm:
I have applied this to OpenBSD now based on my diff with the later changes. It should be in GitHub later on. Let me know if you see any problems. Thanks!

alberti42:
Happy to see the changes upstream! There is still a small issue: when tmux
applies DECSCUSR (`ESC [ Ps q`) to change the cursor style (and blink bit),
it does not mark the blink state as explicitly set. As a result, a subsequent
`DECRQM ?12$p` (cursor blink status) falls back to the option-derived default
and can report the wrong value.

Example with the default of a steady cursor:

```bash
tmux set-option -g cursor-style bar

printf '\e[?12$p'  # reports ^[[?12;2$y → steady
printf '\e[5 q'    # sets a blinking bar cursor (on screen it blinks)
printf '\e[?12$p'  # tmux still reports ^[[?12;2$y → steady (wrong)
```

In xterm, the same sequence produces `^[[?12;1$y`, correctly reporting a  
blinking cursor as the user sees it.

I believe tmux should follow xterm here, so that queries about the cursor  
style always return the *effective on-screen state*. The fix is in commit  
7f2d5e7596064f2d79ef8433cd1e1899bdbef23f: it makes DECSCUSR update  
`MODE_CURSOR_BLINKING_SET` just like the `CSI ? 12 h/l` handlers do. This  
ensures `DECRQM ?12$p` accurately reflects the runtime state rather than  
falling back to defaults.

nicm:
It should only fall back to the options if `MODE_CURSOR_BLINKING_SET` is clear and the current style is default, if the current style is not default, the blinking flag must have been set.

nicm:
Does this do it?
~~~~
--- input.c     2 Sep 2025 15:15:12 -0000       1.237
+++ input.c     3 Sep 2025 19:14:49 -0000
@@ -1558,7 +1558,8 @@ input_csi_dispatch(struct input_ctx *ict
        case INPUT_CSI_QUERY_PRIVATE:
                switch (input_get(ictx, 0, 0, 0)) {
                case 12: /* cursor blink: 1 = blink, 2 = steady */
-                       if (s->mode & MODE_CURSOR_BLINKING_SET)
+                       if (s->cstyle != SCREEN_CURSOR_DEFAULT ||
+                           s->mode & MODE_CURSOR_BLINKING_SET)
                                n = (s->mode & MODE_CURSOR_BLINKING) ? 1 : 2;
                        else {
                                if (ictx->wp != NULL)
~~~~
I'm not sure whether we should clear `MODE_CURSOR_BLINKING_SET` and `MODE_CURSOR_BLINKING` when we go back to default since it is really indeterminate again at that point...

alberti42:
Thanks, Nicholas!

Your suggested change:

```c
if (s->cstyle != SCREEN_CURSOR_DEFAULT || (s->mode & MODE_CURSOR_BLINKING_SET))
    n = (s->mode & MODE_CURSOR_BLINKING) ? 1 : 2;
else
    /* derive from option */
```

does make the response correct in most cases. However, it still leaves a couple of edge cases where two “sources of truth” can diverge:

### 1) Reverting to default via `DECSCUSR Ps=0`
- `screen_set_cursor_style(0, …)` sets `cstyle = DEFAULT` but does **not** touch `MODE_CURSOR_BLINKING_SET`.
- If `MODE_CURSOR_BLINKING_SET` was previously set (from `?12 h/l` or `Ps=1..6`), the condition remains true due to the `|| (…BLINKING_SET)`, so queries keep reporting the **stale** `MODE_CURSOR_BLINKING` bit even after returning to default.
- Result: `?12$p` may report “blink” while the effective default (option) is steady (or vice versa).

### 2) Mixing `DECSCUSR Ps=0` after `CSI ? 12 h/l`
- If you first set blink with `?12h`, then send `Ps=0`, we’re back to “default,” but because `BLINKING_SET` remains set, queries still report the explicit blink bit instead of falling back to the option-derived default.

---

## Proposed fix

When processing **`DECSCUSR Ps=0`**, explicitly clear the “explicit” marker so that subsequent queries fall back to the option:

```c
case INPUT_CSI_DECSCUSR: {
    int ps = input_get(ictx, 0, 0, 0);
    if (ps == -1) break;

    screen_set_cursor_style(ps, &s->cstyle, &s->mode);

    if (ps == 0) {
        /* Back to default → make it indeterminate again */
        screen_write_mode_clear(sctx, MODE_CURSOR_BLINKING_SET);
        /* Leave MODE_CURSOR_BLINKING as-is; it’s ignored when DEFAULT */
    }
    break;
}
```

With `cstyle == DEFAULT` and `BLINKING_SET` cleared, the `?12$p` handler will derive from the tmux option, matching what’s actually on screen.

This approach:
- Preserves the two-stage model (explicit runtime choices vs. defaults).
- Prevents queries from reporting stale explicit state after an app requests “back to default.”
- Mirrors `CSI ? 12 h/l` semantics (those always assert “explicit”), so moving to default should always de-assert it.

An alternative would be to clear `MODE_CURSOR_BLINKING_SET` inside `screen_set_cursor_style()` for `style == 0`, but that function is a lower-level mapper. Keeping this logic in the **DECSCUSR input handler** avoids side effects in other contexts in the future.

For reference, my earlier commit 7f2d5e7596064f2d79ef8433cd1e1899bdbef23f implemented the same idea inside `screen_set_cursor_style()`, but I agree that doing it in the handler is the cleaner layering.

nicm:
I think you're right, can give me a diff or make an MR (or update this one)?

alberti42:
Sure, this diff works for me. Take a look if it also works for you:

```diff
diff --git a/input.c b/input.c
index 4393fd7d..462c0f73 100644
--- a/input.c
+++ b/input.c
@@ -1562,7 +1562,7 @@ input_csi_dispatch(struct input_ctx *ictx)
 	case INPUT_CSI_QUERY_PRIVATE:
 		switch (input_get(ictx, 0, 0, 0)) {
 		case 12: /* cursor blink: 1 = blink, 2 = steady */
-			if (s->mode & MODE_CURSOR_BLINKING_SET)
+			if (s->cstyle != SCREEN_CURSOR_DEFAULT || s->mode & MODE_CURSOR_BLINKING_SET)
 				n = (s->mode & MODE_CURSOR_BLINKING) ? 1 : 2;
 			else {
 				if (ictx->wp != NULL)
@@ -1746,8 +1746,15 @@ input_csi_dispatch(struct input_ctx *ictx)
 		break;
 	case INPUT_CSI_DECSCUSR:
 		n = input_get(ictx, 0, 0, 0);
-		if (n != -1)
+		if (n != -1) {
 			screen_set_cursor_style(n, &s->cstyle, &s->mode);
+
+			if (n == 0) {
+				/* Back to default blinking state → clear MODE_CURSOR_BLINKING_SET */
+				screen_write_mode_clear(sctx, MODE_CURSOR_BLINKING_SET);
+			}
+		}
+
 		break;
 	case INPUT_CSI_XDA:
 		n = input_get(ictx, 0, 0, 0);
```

I think this is the last remaining bit. I don't see any other issue at the moment.

nicm:
Great, applied to OpenBSD now with some trivial tweaks, will be in GitHub later. Thanks!

PR #4621, opened 2025-09-05
Fix OOB access in sixel printing

bptato:
Fixes OOB access in sixel_print by ensuring that the number of chunks actually matches the highest color used on the image, even if not defined as a color register.

We still pass those color numbers through, as such images are perfectly valid - indeed, on the VT340, all images shared the same color registers, so not defining any at all is not unheard of.

(I first thought this was my mistake, but at a second glance it looks like the precursor to the chunks/active variables had the same issue.)

### DIFF
--- a/image-sixel.c
+++ b/image-sixel.c
@@ -43,6 +43,7 @@ struct sixel_image {
 
 	u_int			*colours;
 	u_int			 ncolours;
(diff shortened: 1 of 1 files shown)


nicm:
Applied, thanks.

Issue #4529, opened 2025-06-05
Artifacts are introduced for SIXEL graphics


bytesnake:
### Issue description

Apparently tmux skips sometimes lines in SIXEL rendering and introduces with this some artifacts. The
annoyance is minor, but still noticable when reading rendered documents.

Direct rendering in Alacritty

![Image](https://github.com/user-attachments/assets/ef157d84-e12a-4111-ad9d-015111d82165)

Pass-through tmux

![Image](https://github.com/user-attachments/assets/326a14bc-7569-4946-8ed3-b7688e439df9)

I also attach a problematic SIXEL sample, where I see the skipping occur at several lines.

[sample.sixel.txt](https://github.com/user-attachments/files/20609908/sample.sixel.txt)

### Required information

Please provide the following information:

* tmux version tmux 3.5a
* Platform Linux x86_64
* $TERM inside (tmux-256color) and outside (alacritty-direct) of tmux

nicm:
What does this show inside tmux?
~~~~
tmux lsc -F '#{client_name} #{client_cell_width} #{client_cell_height} #{window_cell_width} #{window_cell_height}'
~~~~

bytesnake:
results in `/dev/pts/2 11 28 11 28`

nicm:
I do not see this in iTerm2:

<img width="1262" height="418" alt="Image" src="https://github.com/user-attachments/assets/0e8e0443-a0e3-4a6c-a2a0-f708a8ead317" />
<img width="1262" height="590" alt="Image" src="https://github.com/user-attachments/assets/7570baa8-89ec-46ac-9de5-4751858fd1c2" />

Do you see this in other terminals? xterm? Do you have to do anything to reproduce?

j4james:
Isn't this just a scaling artefact? If I remember correctly, tmux stores sixel images with a virtual cell size of something like 16x32. If the target terminal has a lower resolution that than, you'll inevitably end up dropping scan line and columns when the image is scaled down. It's just a basic point sampling algorithm as far as I can see.

https://github.com/tmux/tmux/blob/b13005e802df23652e87c98f136f9eb13f096374/image-sixel.c#L462-L468

nicm:
They already showed that the cell size is the same for both window and client. Although I thought we did not rewrite the image in that case, but it looks like we now do, so you are probably right and it is caused by that.

j4james:
[...]

PR #4622, opened 2025-09-05
Do not squeeze sixel image height to multiples of 6

bptato:
This removes the problematic truncation discussed in #4529.  Besides the linked ticket, this also fixes things like

```
printf '\033Pq"1;1;100;28#0;2;99;0;0#1;2;0;99;0#1!100~-!100~-!100~-!100~-!100N\033\\\n'
```

which on current tmux bleeds some arbitrary color into the last line - depending on the terminal, you get register 0/red (XTerm, foot), register 255 (Windows terminal - I *think*), or the terminal background.

---

The comment at the top of the function provides some rationale:

> We want to get the section of the image at ox,oy in image cells and
> map it onto the same size in terminal cells, remembering that we
> can only draw vertical sections of six pixels.

...but this is a misconception.  Sixel draws sections of six pixels at once, but you can simply omit painting the bottom rows to draw images of arbitrary heights, provided you send image dimensions ("set raster attributes").

I suspect the actual reason for the truncation was that previously, tmux did not handle raster attributes at all.  So an image with an uneven number of lines (before or after scaling) would have resulted in filling the last lines with some terminal-dependent color, and scaling to multiples of six is a simple workaround for that.

But since we now support both raster attributes and P2=1 (transparency), this workaround is just causing problems without any real benefit.

### DIFF
--- a/image-sixel.c
+++ b/image-sixel.c
@@ -420,8 +420,7 @@ sixel_scale(struct sixel_image *si, u_int xpixel, u_int ypixel, u_int ox,
 
 	/*
 	 * We want to get the section of the image at ox,oy in image cells and
-	 * map it onto the same size in terminal cells, remembering that we
-	 * can only draw vertical sections of six pixels.
+	 * map it onto the same size in terminal cells.
 	 */
 
 	sixel_size_in_cells(si, &cx, &cy);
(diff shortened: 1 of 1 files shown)


nicm:
Applied, thanks.

PR #4623, opened 2025-09-06
Quote glob characters using dedicated function

proski:
`utf8_stravis` is a wrong function to escape path to use with `glob`. It escapes `*` and `?` using octal encoding when `VIS_GLOB` is set, regardless of other flags. However, `glob` only accepts backslash as the escape character.

The new implementation uses backslash to escape all ASCII characters other than letters, digits and the forward slash. I tested it with directories containing various ASCII symbols and UTF-8 characters, and I was able to load files from those locations, using both full names and wildcards. The current implementation cannot load configuration files from directories containing e.g. `#` in the name.

### DIFF
--- a/cmd-source-file.c
+++ b/cmd-source-file.c
@@ -149,6 +149,22 @@ cmd_source_file_add(struct cmd_source_file_data *cdata, const char *path)
 	cdata->files[cdata->nfiles++] = xstrdup(path);
 }
 
+static char *quote_for_glob(const char *path)
+{
+	char *const   quoted = xmalloc(2 * strlen(path) + 1);
+	const char   *p = path;
+	char         *q = quoted;
+
+	while (*p != '\0') {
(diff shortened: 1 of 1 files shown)


proski:
Successfully tested in a directory with a name containing all ASCII characters except `'\0'` and `/` (they are not allowed in filenames).

Considering my choices what to escape:

I don't want to have an exhaustive list of symbols to be escaped. `~` and `{` can be used with GNU extensions. It's better to escape everything unless it definitely doesn't need escaping (letters digits, non-ASCII characters). Forward slash can be escaped too, by I felt it would be safer not to escape it, as it has a special meaning (path separator) and we don't to cancel that meaning.

nicm:
Makes sense, applied to OpenBSD with a couple of minor changes, will be in GitHub later. Thanks!

Issue #4624, opened 2025-09-06
RFE: Please make an optional limit for the number of columns in the tiled layout


vkadlcik:
### Issue description

I often use a higher number of panes in the tiled layout. I often need the panes rather wide regardless their height, therefore in just two columns. For years I have been using a patched build of tmux, enforcing two columns.

I dislike maintaining the patch and I think that optional limiting of the number of columns can be useful to more users than just me. Therefore this RFE.

vkadlcik:
I've tried to implement it myself, see the PR above

PR #4625, opened 2025-09-06
Optionally limit number of columns in tiled layout

vkadlcik:
Introduce a new window option: tiled-layout-max-columns. It configures the maximum number of columns in the tiled layout. The default value is 0 which means no limit and makes the change backward-compatible.

Implements RFE #4624


nicm:

PR #4635, opened 2025-09-14
cmdq_error: don't uppercase first letter, it can be a filename

proski:
For instance, run `source-file abc.conf`. The error message would be `Abc.conf: No such file or directory`, which is confusing.


nicm:
Hmm. The reason for this is that the convention for errors on the command line is lowercase, for example:
~~~~
$ tmux abc
unknown command: abc
~~~~
But this looks odd when it is shown without an uppercase letter as a message inside tmux.

Maybe it doesn't matter too much.

proski:
I can modify the error message so that the filename doesn't appear first: `No such file or directory: abc.conf`

Issue #3159, opened 2022-04-14
Feature: add ability to center cursor vertically in copy mode.


tlietz:
In vi copy mode, I think the keybinding for this should be `zz` to match vi. 

I am not sure what the keybinding should be for standard copy mode. Perhaps `<prefix>z`?

nicm:
Added to todo list. Even if someone does it, it won't be `zz` however because we don't support double keys.

PR #4662, opened 2025-10-26
Cursor centering

peripherium:
Hello tmux developers,

While reviewing the feature suggestions, I came across [issue #3159](https://github.com/tmux/tmux/issues/3159), which proposed adding a way to center the cursor vertically in copy mode.  
I found this feature useful and decided to implement it.

This PR adds two related features:

- **Vertical centering** — moves the cursor to the vertical center of the visible area.  
- **Horizontal centering** — moves the cursor to the horizontal center of the current line.

Although the original issue mentioned a `zz` key binding, tmux already provides a similar *scroll-middle* action.  
Therefore, I assigned the new bindings to **`C-l`** (vertical centering) and **`M-l`** (horizontal centering), since `l` was unused and the modifier keys felt intuitive.

If the maintainers prefer to only include vertical centering, I can remove the horizontal feature easily.

Feedback and suggestions are very welcome!

---

Testing
Tested on **Linux** (`tmux next-3.6`, `xterm-256color`).  
Both bindings work as expected in copy mode.

---

Thank you for reviewing!


nicm:
Thanks. This looks good but can you make the code use "centre" rather than "center" to match the rest of tmux, and I think the commands would be better as `cursor-centre-horizontal` and `cursor-centre-vertical`.

peripherium:
Thanks, I wasn’t aware tmux uses British spelling. I’m from Germany, so I used "center" by habit. Updated to "centre" now.

nicm:
OK I've applied this with a couple of minor changes:

- The keys you listed in `tmux.1` didn't match the actual bindings, so I fixed `tmux.1`.
- The functions need to redraw when they move the cursor in case there is a selection.

It's in OpenBSD now and should appear in GitHub next time it syncs up. Thanks!

PR #4665, opened 2025-10-27
query terminal for palette colors 0-7 at startup

rockorager:
Query the terminal for OSC 4 palette colors 0-7 at startup, similar
to how OSC 10 and 11 are queried for foreground and background colors.
Store the responses in tty->palette[8] alongside tty->fg and tty->bg.

When applications query palette colors via OSC 4, respond with the
pane's palette if set (via OSC 4 or pane-colours option), otherwise
fall back to the colors queried from the terminal at startup. This
allows applications inside tmux to discover the terminal's default
color scheme.

Add window_pane_get_palette_tty() following the same pattern as
window_pane_get_fg(), which loops through attached clients to find
the palette color from the terminal. Add colour_palette_set_default()
and colour_palette_from_tty() to support storing queried colors in
the default palette.

This matches the existing behavior for OSC 10/11 foreground and
background colors.

Amp-Thread-ID: https://ampcode.com/threads/T-7f1c58ac-6049-42b6-8022-78a12a062e50
Co-authored-by: Amp <amp@ampcode.com>

### DIFF
--- a/colour.c
+++ b/colour.c
@@ -1123,6 +1123,48 @@ colour_palette_set(struct colour_palette *p, int n, int c)
 	return (1);
 }
 
+/* Set a default colour in a palette. */
+int
+colour_palette_set_default(struct colour_palette *p, int n, int c)
+{
+	u_int	i;
+
+	if (p == NULL || n > 255)
+		return (0);
+
+	if (c == -1 && p->default_palette == NULL)
+		return (0);
+
+	if (c != -1 && p->default_palette == NULL) {
+		p->default_palette = xcalloc(256, sizeof *p->default_palette);
+		for (i = 0; i < 256; i++)
+			p->default_palette[i] = -1;
+	}
+	p->default_palette[n] = c;
+	return (1);
+}
+
+/* Populate default palette from client tty palette. */
+void
+colour_palette_from_client(struct colour_palette *p, struct client *c)
+{
+	u_int	i;
+
+	if (p == NULL || c == NULL)
+		return;
+
+	if (p->default_palette == NULL) {
+		p->default_palette = xcalloc(256, sizeof *p->default_palette);
+		for (i = 0; i < 256; i++)
+			p->default_palette[i] = -1;
+	}
+
+	for (i = 0; i < 8; i++) {
+		if (c->tty.palette[i] != -1)
+			p->default_palette[i] = c->tty.palette[i];
+	}
+}
+
 /* Build palette defaults from an option. */
 void
 colour_palette_from_option(struct colour_palette *p, struct options *oo)
--- a/input.c
+++ b/input.c
@@ -2803,6 +2803,8 @@ input_osc_4(struct input_ctx *ictx, const char *p)
 		s = strsep(&next, ";");
 		if (strcmp(s, "?") == 0) {
 			c = colour_palette_get(palette, idx|COLOUR_FLAG_256);
+			if (c == -1 && idx >= 0 && idx < 8)
+				c = window_pane_get_palette_client(ictx->wp, idx);
 			if (c != -1)
 				input_osc_colour_reply(ictx, 4, idx, c);
 			s = next;
--- a/tmux.h
+++ b/tmux.h
@@ -1545,6 +1545,7 @@ struct tty {
 	int		 mode;
 	int              fg;
 	int              bg;
+	int		 palette[8];
 
 	u_int		 rlower;
 	u_int		 rupper;
@@ -1579,6 +1580,7 @@ struct tty {
 #define TTY_WINSIZEQUERY 0x1000
 #define TTY_WAITFG 0x2000
 #define TTY_WAITBG 0x4000
+#define TTY_WAITPALETTE 0x8000
 #define TTY_ALL_REQUEST_FLAGS \
 	(TTY_HAVEDA|TTY_HAVEDA2|TTY_HAVEXDA)
 	int		 flags;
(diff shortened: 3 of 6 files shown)


nicm:
The code looks OK but what is the practical use case? Do any applications use this? It is going to add a lot of extra queries on startup.

Also, since you wrote this with AI and it is a large enough change to be potentially copyrightable, can you please show me a statement on their website saying they do not assert any ownership over the code changes, or that they are happy for you to ISC license it?

rockorager:
> The code looks OK but what is the practical use case? Do any applications use this? It is going to add a lot of extra queries on startup.

Generic use case is to get the user's configured palette as RGB to allow applications to blend / darken / brighten colors and stay on theme. You can also do contrast checking to be sure that a color combination will work when you have the RGB values.

Specific use case is that I work on the ampcode.com TUI, we query for colors to be able to show a nicer diff view (IE background highlighted in green or red, but dimmed closer to your background color rather than using reverse). 

CleanShot 2025-10-28 at 06 37 39@2x


As some more examples, both of my TUI libraries (libvaxis / vaxis) allow on-demand querying for colors using OSC 4, and notcurses queries the entire 256 palette at startup to expose it to library users.

I'm happy to provide some benchmarks of startup times with and without this patch if that would help. Probably something over SSH just to simulate worst case.

> Also, since you wrote this with AI and it is a large enough change to be potentially copyrightable, can you please show me a statement on their website saying they do not assert any ownership over the code changes, or that they are happy for you to ISC license it?

Sure thing! We don't have one right now actually but our legal team will get one up shorty. I've got no expectations of you accepting this until you see that :).

nicm:
I'm just thinking that realistically any terminal where a user is going to be tweaking colours will support at least 256 if not full RGB colour, so wouldn't your application (or any application) use those instead of the ANSI colours? If you want to dim to the background colour you only need OSC 10/11 which we already have.

Querying the ANSI colours means tmux is doing a big extra query on every attach for something that seems of limited use. The whole 256 colour set would be more useful but that seems too many to be practical...

TBH we really need a way to proxy queries on demand (which we could use for this and also OSC 52), so if the application asks and the pane doesn't know it can work out which client to ask and send a request and then have some way of getting the reply back to the pane. I guess that wouldn't work if the session is detached though.

Thanks

rockorager:
> I'm just thinking that realistically any terminal where a user is going to be tweaking colours will support at least 256 if not full RGB colour, so wouldn't your application (or any application) use those instead of the ANSI colours? If you want to dim to the background colour you only need OSC 10/11 which we already have.

The problem with using 256 is staying within the user's color choices. There's a big difference between the green from Gruvbox and the green from TokyoNight. And if I want to put that green as the background and the user's blue as the foreground, I can't possibly know if there will be enough contrast - the only way to be more sure is to use a blue from the 256 palette but then I'm outside of the user's configured colors.

I intentionally limited this to the first 8 of the palette to keep the query as small as possible but still provide theme information. A ton of themes don't actually set 8-15, and beyond that some terminals allow configuring 16-255 but I have never actually seen anyone do that.

I agree that the best approach would be like an "OSC tunnel" to the host. Applications that are querying for these will already need to handle cases where the terminal doesn't respond so I think tmux doesn't need to do much around that, just forward the requests and responses as it gets them.

nicm:
OK a few comments:

- I'm not wild about `palette_count` since it will not work if you get the same response multiple times. I think it'd be better just to loop over them and check that none are -1, there are only eight of them.

- I think `colour_palette_from_tty` should be `from_client` and take a `struct client` and same for `window_pane_get_palette_tty` . A client contains a tty but client is really what other stuff should be talking about.

- I don't think you need to call `session_theme_changed` when you receive palette responses since theme only depends on the background colour.

- Declaration of `wp` needs to go at the top of `tty_keys_next` not in an inner block.

- Wrap the very long lines for tty_puts in tty_send_requests and tty_repeat_requests, or use a loop (maybe in a static helper function).

rockorager:
Thanks for the feedback! I've made those changes, once I have the code ownership statement I'll put the link in here too for you.

nicm:
Thanks. I had a look at making the request on-demand and it was not terribly difficult. Please take a look at this: [input_req.diff.txt](https://github.com/user-attachments/files/23205235/input_req.diff.txt)

The main thing I don't like about this is that if a program sends a request multiple times in quick succession, tmux will send it multiple times as well, which means it could hammer the outside terminal with requests without tmux stopping it. But it is a bit of a hassle to make tmux only send duplicate requests once without getting replies in the wrong order (if you do `\033]4;1;?;27;?;1;?\033\\` we will get 1 first but we can't use that to also reply to the second 1 immediately because we haven't got the 27 yet, we'd need to cache the result). Some sort of rate limit on requests might be easier, but not sure I like that idea either.

nicm:
Here is an updated diff with one fix (counting requests was wrong): [input_req.diff.txt](https://github.com/user-attachments/files/23205408/input_req.diff.txt)

rockorager:
Yeah this approach looks good. I assume the underlying request logic for palettes would first look for configured colors in the existing palette (the case today), if not there query the host terminal.

nicm:
Yes, that's right, it only sends a query if it doesn't have an entry set.

rockorager:
> Yes, that's right, it only sends a query if it doesn't have an entry set.

Nice. Do you want to go that route and I close this PR?

nicm:
Yes, I'm going to go with this, I think it is OK. I'll check it again and probably apply it tomorrow.

I've extended it to do OSC 52 as well and it works, but I'll apply that separately, because I'm not sure if I want to get rid of or modify `refresh-client -l` at the same time.

Thanks!

rockorager:
Cool, thanks!

nicm:
This has been applied to OpenBSD now, it will be in GitHub later when it syncs up. I know we didn't use your code in the end but it was useful and at least there is some solution now :-). Thanks!

rockorager:
Glad it was useful, and thanks for finding a solution! I think this way will work great.

PR #4666, opened 2025-10-28
Avoid starting cmdq_error() with a filename

proski:
This is an alternative to #4635 (although both can be applied if desired). This change ensures that an error message produced by `cmdq_error()` won't start with a filename. Since `cmdq_error()` turns the first letter of the error message to uppercase, the filename in the initial position can be changed, making the message inaccurate.


nicm:
Let's go with that - applied to OpenBSD now, will be in GitHub later. Thanks!

Issue #3655, opened 2023-08-05
Ability to change popup color within the popup


mgpinf:
### Issue description
* It would be nice if there was an ability to change `tmux` popup color from within the popup (especially while running a script that calls it)
* Went through the manual, where it is mentioned that `Panes are not updated while a popup is present.`.
* I think it would be a good idea to edit at least popup options from within the popup window (if not the pane options)
* Asking this because while calling `fzf` within `tmux` popup window, if the `tmux` popup has border beforehand, and the `fzf` window also has borders, it would look cluttered. So, if there is an option to change popup options in a popup window, then the popup border is initially deactivated, and after calling the `fzf` function, the popup border would be activated, (once `fzf` function call ends), thereby obtaining a more homogeneous appearance
* Current tmux popup window in which a script calling `fzf` runs: (as seen, there are 2 borders, one for `fzf`, other one for `tmux`)
  
<img width="1385" alt="image" src="https://github.com/tmux/tmux/assets/56510798/8b98ceaf-952c-496e-a0aa-410735ff3ddc">

* It's a feature request

Steps to reproduce:
* `tmux display-popup 'script'`
* `script` contains command to change tmux popup border color
* Initially the popup window has a border of a certain color
* Unable to change it once the popup window starts

### Required information

* tmux version (`tmux -V`).
`tmux 3.3a`
* Platform (`uname -sp`).
`Darwin arm`
* $TERM inside and outside of tmux (`echo $TERM`).
  * Inside `xterm-256color`
  * Outside `screen-256color`
* Logs from tmux (`tmux kill-server; tmux -vv new`).

nicm:
I have added this to the todo list but I don't think it would be trivial and it doesn't seem to have wide use cases, so I wouldn't hold your breath...

Can't you ask fzf not to show a border instead?

PR #4678, opened 2025-11-09
Add `modify-popup` command to dynamically change popup styles and behavior

peripherium:
This implements a new command, `modify-popup`, following discussion in issue #3655 (https://github.com/tmux/tmux/issues/3655).

`modify-popup` allows updating the popup's color, border style, and post-exit behavior **after it has been displayed**. Parameters `-s`, `-S`, `-b`, `-B`, `-E`, `-EE`, `-k` behave analogously to `display-popup`.

A new option, `-N`, clears any previously set `-E`, `-EE`, or `-k` options, allowing full control over the popup's post-exit behavior.

**Example usage:**
Open a popup with a script
tmux display-popup 'myscript.sh'

After the script ends, modify the popup border to red if it failed or green, if the script was executed successfully.
tmux modify-popup -S fg=red

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -33,8 +33,6 @@ static enum cmd_retval		cmd_display_menu_exec(struct cmd *,
 				    struct cmdq_item *);
 static enum cmd_retval		cmd_display_popup_exec(struct cmd *,
 				    struct cmdq_item *);
-static enum cmd_retval		cmd_modify_popup_exec(struct cmd *,
-				    struct cmdq_item *);
 
 const struct cmd_entry cmd_display_menu_entry = {
 	.name = "display-menu",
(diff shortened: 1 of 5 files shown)


nicm:
Will take a closer look later but I'd prefer this to be a flag on `display-popup` not a new command.

peripherium:
I’d be happy to implement your suggestion of integrating this feature into cmd_display_popup_exec(). My initial idea was actually similar: if popup is called from within another popup, the current popup would be modified. I was just a bit concerned about potential ambiguities, but I think your approach could work well.

peripherium:
Calling display-popup from within a popup now updates that popup's style
and properties instead of creating a new one. This replaces the previously
proposed separate modify-popup command.

nicm:
I have applied this with a couple of minor changes:

- I renamed `isapopup` to `popup_present` to match the naming scheme.

- By reordering some of the code in `cmd_display_popup_exec` its possible to have more of it within one `if (!modify)` which I think makes it a bit more obvious.

Let me know if you see any problems.

It is in OpenBSD now and should be in GitHub later on.

Thanks!

PR #4688, opened 2025-11-13
Avoid losing spaces in `pane-border-format`

ryabkovdv:
Currently trailing spaces in `pane-border-format` are rendered in a strange way.

Current behavior (`pane-border-status` is set to `top`):
1. When `pane-border-format` is `"#[default] 12345678 "` all spaces rendered for both active and inactive panes:
   size8
1. When `pane-border-format` is `"#[default] 123456789 "` trailing space does not show up for inactive pane:
   size9
1. When `pane-border-format` is `"#[default] 12345678901 "` all spaces show up again:
   size11
1. When there are multiple trailing spaces only some of them are rendered: for example with `"#[default] 123456789      "` (6 spaces) only 3 are visible:
   size9-6spaces
1. Adding extra dummy `#[default]` (e.g. `"#[default]#[default] 123456789 "`) makes spaces visible for sizes 9 and 10, but instead they are missing for sizes 5, 6.
1. This issue only happens with the default style - changing either `pane-border-style` or `#[default]` to some other value makes it go away.
1. It can be reproduced regardless of size by setting `pane-border-format` to only spaces:
   spaces

After some debugging I think I found why this happens:
1. `format_draw()` firstly formats pane border text into temporary empty screens (with sizes set to a length of a string before processing style directives).
1. `screen_write_cell()` does not insert a cell if:
   1. storage for it is allocated and the new cell is equal to the old cell (not applicable here because temporary screens are empty);
   1. storage for it is not allocated yet and it is equal to the default cell (space, `fg=8`, `bg=8`).

   This is why this problem does not reproduce for non-default style.
1. Allocated storage for cells is proportional to sizes of screens (via `grid_expand_line()`) - this is why adding dummy `#[default]`-s changes behavior.
1. `screen_write_fast_copy()` does not copy cells that are not allocated.

Here I fixed it by adding flag to `screen_write_ctx` that makes `screen_write_cell()` append cell even if it is equal to the default cell (flag is only set for temporary screens in `format_draw()`). Not sure if it the right place to fix it, but seems to be working.

### DIFF
--- a/format-draw.c
+++ b/format-draw.c
@@ -745,6 +745,7 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 		screen_init(&s[i], size, 1, 0);
 		screen_write_start(&ctx[i], &s[i]);
 		screen_write_clearendofline(&ctx[i], current_default.bg);
+		ctx[i].flags |= SCREEN_WRITE_DEFAULT_CELL;
 		width[i] = 0;
 	}
 
--- a/screen-write.c
+++ b/screen-write.c
@@ -2023,9 +2023,7 @@ screen_write_cell(struct screen_write_ctx *ctx, const struct grid_cell *gc)
 
 	/* If no change, do not draw. */
 	if (skip) {
-		if (s->cx >= gl->cellsize)
-			skip = grid_cells_equal(gc, &grid_default_cell);
-		else {
+		if (s->cx < gl->cellsize) {
 			gce = &gl->celldata[s->cx];
 			if (gce->flags & GRID_FLAG_EXTENDED)
 				skip = 0;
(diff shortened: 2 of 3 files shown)


nicm:
Good analysis, thanks. I wonder if it would be better to make `screen_write_fast_copy` only stop if both the source and target are beyond the end of the allocated line?

This would work because the status line is already filled with border characters.

Like this:
~~~~
--- a/screen-write.c
+++ b/screen-write.c
@@ -583,7 +583,8 @@ screen_write_fast_copy(struct screen_write_ctx *ctx, struct screen *src,
                if (wp != NULL)
                        screen_write_initctx(ctx, &ttyctx, 0);
                for (xx = px; xx < px + nx; xx++) {
-                       if (xx >= grid_get_line(gd, yy)->cellsize)
+                       if (xx >= grid_get_line(gd, yy)->cellsize &&
+                           s->cx >= grid_get_line(ctx->s->grid, s->cy)->cellsize)
                                break;
                        grid_get_cell(gd, xx, yy, &gc);
                        if (xx + gc.data.width > px + nx)

ryabkovdv:
Yes, that's sounds better.

nicm:
Good stuff, I have applied my change to OpenBSD now, it should be in GitHub later. Let me know if you see any problems. Thanks!

PR #4696, opened 2025-11-20
menu: preserve menu on resize and reposition if needed

peripherium:
This changes menu behaviour on resize events. Previously, any SIGWINCH
caused the menu overlay to be cleared. This made menus disappear when
the terminal font size changed.

A new resize callback (menu_resize_cb) is added which keeps the menu
open and repositions it if needed so it remains fully visible.

Tested with tmux next-3.6 on xterm-256color.

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -439,6 +439,37 @@ chosen:
 }
 
 static void
+menu_resize_cb(struct client *c, void *data)
+{
+	int			nx, ny;
+	int			w, h;
+	struct menu_data	*md = data;
+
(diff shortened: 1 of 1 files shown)


nicm:
I've applied this with some minor changes (use `u_int` rather than `int` to avoid cast, and it is OK for menu to cover the status line so there is no need to reduce the size by 1 (and if it wasn't you would need to use `status_line_size` not 1 anyway)). Should be in GitHub later, let me know if you see any problems. Thanks!

PR #4697, opened 2025-11-22
Added clock-mode-seconds option

augustus-7613:
Added a `clock-mode-seconds` option to enable or disable seconds in clock-mode. Works with both 12 and 24 hour clock.

Is toggled by 
```
set-window-option -g clock-mode-seconds on
set-window-option -g clock-mode-seconds off
```

With 12 hour clock
image

With 24 hour clock
image

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -36,7 +36,7 @@ static const char *options_table_mode_keys_list[] = {
 	"emacs", "vi", NULL
 };
 static const char *options_table_clock_mode_style_list[] = {
-	"12", "24", NULL
+	"12", "24", "12-with-seconds", "24-with-seconds", NULL
 };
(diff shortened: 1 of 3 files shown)


nicm:
This looks OK, thanks, but I'm not sure we want another option for a mode that is essentially an example. Maybe `clock-mode-style` could be extended to have `12`, `24`, `12-with-seconds` and `24-with-seconds`. Also you forgot to update `tmux.1`.

augustus-7613:
Thanks for the feedback. I agree that only having  `clock-mode-style` is better than introducing another option. I have added the options `12-with-seconds` and `24-with-seconds` now. I also updated `tmux.1` to reflect this.

nicm:

PR #4703, opened 2025-11-26
Make image limit configurable

s21v1d9p:
## Description

This PR makes the sixel image limit configurable instead of being hard-coded to 10.

## Changes

- Added a new server option `image-limit` in `options-table.c` with a default value of 10
- Modified `image.c` to use the configurable option instead of the hard-coded value
- The hard-coded limit was marked with a `/*XXX*/` comment indicating it should be made configurable

## Usage

Users can now configure the image limit using:
```
set-option -g image-limit <value>
```

## Backward Compatibility

The default value remains 10, maintaining backward compatibility with existing configurations.

## Testing

- Built successfully on macOS (darwin25.1.0)
- Code follows existing patterns for server options in tmux

### DIFF
--- a/image.c
+++ b/image.c
@@ -97,6 +97,7 @@ struct image*
 image_store(struct screen *s, struct sixel_image *si)
 {
 	struct image	*im;
+	u_int		 limit;
 
 	im = xcalloc(1, sizeof *im);
 	im->s = s;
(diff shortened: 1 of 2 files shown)


nicm:
I don't think this should be configurable. If 10 is not enough, we should increase it.

s21v1d9p:
What value would you suggest increasing it to?

nicm:
I don't know. 20? 25? What value would you have set the option to?

s21v1d9p:
I noticed the hard-coded value suggesting it should be made configurable, so I implemented that. I don't have a specific use case.

nicm:
OK, well, let's leave it then until someone has a use case with more than 10 images visible at a time, then we can see how many they need?

veltza:
This would require 25 images:

```
chafa -f sixels --grid 5x5 --label on ~/Pictures/Wallpapers/*
```

And this would require 100 images:

```
chafa -f sixels --grid 10x10 --label on ~/Pictures/Wallpapers/*
```

So the configurable limit doesn't sound like a bad idea to me.

PR #4720, opened 2025-11-30
Bugfix: Add horizontal border case to server_client_check_mouse_in_pane.

mgrant0:
This fix should fix issue 4715.  It returns BORDER when the mouse clicks the horizontal border when pane-border-status is either top or bottom.


nicm:
Applied, thanks for looking at this!

Issue #4722, opened 2025-12-01
'set-option -g pane-scrollbars on/modal' causes vertical line of ASCII 195 character along right split panel when <PREFIX>-E pressed


jmtysonjr:
### Issue description

cat << EOF >  ~/.tmux.conf
set-option -g pane-scrollbars modal
EOF

From the BASH prompt, type 'tmux' to enter TMUX.
Press 'PREFIX %', 'PREFIX-E' to resize the panels.
A vertical line of the extended ASCII box-drawing characters '├' (ASCII 195) appear along the right edge of the panel.
Closing the panel does not clear the artifact.
This happens when '_set-option -g pane-scrollbars_' is set to _modal_ or _on_; it does not occur when set to _off_.
It also happens with this morning's _tmux-master_ branch code.

### Required information

* tmux version: tmux 3.6
* Platform: Ubuntu 24.04.3 LTS
* Terminal in use: PuTTY 0.83
* $TERM *inside* tmux: screen-256color
* $TERM *outside* tmux: xterm
* Logs from tmux (`tmux kill-server; tmux -vv new`): Can provide if necessary

nicm:
@mgrant0 --^

mgrant0:
I should be able to fix this.  Thanks for spotting it.

mgrant0:
I believe I have fixed this in this PR: https://github.com/tmux/tmux/pull/4724

jmtysonjr:
@mgrant0 ,
I cloned master and overlaid your updated _layout.c_ file.
I am no longer able to reproduce the issue.
Your fix appears to have worked!
Thanks!

nicm:
Should be fixed as soon as the change makes it to GitHub later today. Thanks!

jmtysonjr:
Doesn't look like this fix made the 3.6a release.

mgrant0:
Nic didn't want to put too many fixes into 3.6a that haven't been well tested yet.  Hopefully 3.6b but will eventually be in the release.

jmtysonjr:
> Nic didn't want to put too many fixes into 3.6a that haven't been well
tested yet. Hopefully 3.6b but will eventually be in the release.

Roger that. I've just cloned master which includes the fix for #4724
<https://github.com/tmux/tmux/pull/4724> and overlaid your fixes for #4731
<https://github.com/tmux/tmux/pull/4731> and it works beautifully!

Once again, thanks!

PR #4724, opened 2025-12-02
Fix the size caluation for left-right windows used to spread out cells

mgrant0:
Fixes the horizontal size calculation in `layout_spread_cell`.  This function does not need to take into account the width of the scrollbar but rather the width of the layout cell.  Taking out the scrollbar width from this function fixed the spreading out issue in issue #4722.


nicm:
Applied, thanks!

PR #4728, opened 2025-12-02
Add check that the pane is not in alternate screen mode when in copy-mode.

mgrant0:
Add check to window_pane_show_scrollbar to chek bwp->base to see if pane is in alt screen and do return 0 to not enable scrollbars in this case.


nicm:
Applied, thanks!

PR #4730, opened 2025-12-03
Improve code readability in colour_palette_get and colour_palette_set

proski:
In `colour_palette_get`, rename `c` to `n` for consistency with other code where `n` is the index in the palette and `c` is the color.

In `colour_palette_set`, check both limits of `n`. Checking just the upper limit sets a bad example, it pretends to be a complete check, but it's actually incomplete. The callers do their own checks, so I'm OK with not checking any limits at all as an alternative.

Also simplify the logic without changing the behavior. The goal of the code is to allocate the palette if it's not allocated yet; the value of `c` doesn't matter (the special case of `c = -1` without a palette has already been handled).

### DIFF
--- a/colour.c
+++ b/colour.c
@@ -1082,22 +1082,22 @@ colour_palette_free(struct colour_palette *p)
 
 /* Get a colour from a palette. */
 int
-colour_palette_get(struct colour_palette *p, int c)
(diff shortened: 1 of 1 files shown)


nicm:

Issue #4727, opened 2025-12-02
Scrolling pane-scrollbar by click-drag whilst in show-messages causes copy-mode to be invoked


jmtysonjr:
### Issue description
cat << EOF > ~/.tmux.conf
set-option -g mouse on
set-option -g pane-scrollbars modal
EOF

Running release tmux 3.6 (and build from latest master), invoke show-messages by pressing 'PREFIX ~'.
While in '_show-messages_', click-grabing the pane-scrollbar and draging up & down causes '_copy-mode_' to be invoked.
Pressing 'q' returns back to '_show-messages_' contents.
Scrolling the mouse-wheel does not cause the jump to '_copy-mode_'.

Is this expected behavior?

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

* tmux version: tmux 3.6
* Platform: Ubuntu 24.04.3 LTS
* Terminal in use: PuTTY 0.83
* $TERM inside tmux: screen-256color
* $TERM outside tmux: xterm
* Logs from tmux (tmux kill-server; tmux -vv new): Can provide if necessary

nicm:
@mgrant0 I think this is likely to be something you would be best to look at as well :-). Should the scrollbar enter copy mode even if it is already in a mode?

mgrant0:
Definitely _not_ the expected behavior!  Thanks for spotting this.  Should be easy to fix, let me look at it.

mgrant0:
I have created PR https://github.com/tmux/tmux/pull/4731 for a proposed fix to this.  I'd like Nic to look at it just to make sure it doesn't break something else unexpecte.

nicm:
This should now be fixed, thanks!

PR #4731, opened 2025-12-03
Fix for scrolling pane-scrollbar whilst in show-messages causes copy-mode to be invoked

mgrant0:
Add logic to window_pane_set_mode to not go into copy-mode from a different mode.

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -64,14 +64,6 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 	struct session		*s;
 	struct window_pane	*wp = target->wp, *swp;
 
-	if (args_has(args, 'S')) {
-		if (TAILQ_EMPTY(&wp->modes))
-			window_pane_set_mode(wp, wp, &window_copy_mode, NULL,
-			args);
-		window_copy_scroll(wp, c->tty.mouse_slider_mpos, event->m.y,
-		    args_has(args, 'e'));
-		return (CMD_RETURN_NORMAL);
-	}
 	if (args_has(args, 'q')) {
 		window_pane_reset_mode_all(wp);
 		return (CMD_RETURN_NORMAL);
@@ -93,7 +85,6 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 		swp = source->wp;
 	else
 		swp = wp;
-
 	if (!window_pane_set_mode(wp, swp, &window_copy_mode, NULL, args)) {
 		if (args_has(args, 'M'))
 			window_copy_start_drag(c, &event->m);
@@ -102,6 +93,11 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 		window_copy_pageup(wp, 0);
 	if (args_has(args, 'd'))
 		window_copy_pagedown(wp, 0, args_has(args, 'e'));
+	if (args_has(args, 'S')) {
+		window_copy_scroll(wp, c->tty.mouse_slider_mpos, event->m.y,
+		    args_has(args, 'e'));
+		return (CMD_RETURN_NORMAL);
+	}
 
 	return (CMD_RETURN_NORMAL);
 }
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -480,9 +480,9 @@ key_bindings_init(void)
 		"bind -n M-MouseDown3Pane { display-menu -t= -xM -yM -T '#[align=centre]#{pane_index} (#{pane_id})' " DEFAULT_PANE_MENU " }",
 
 		/* Mouse on scrollbar. */
-		"bind -n MouseDown1ScrollbarUp { copy-mode -u }",
-		"bind -n MouseDown1ScrollbarDown { copy-mode -d }",
-		"bind -n MouseDrag1ScrollbarSlider { copy-mode -S }",
+		"bind -n MouseDown1ScrollbarUp { if -Ft= '#{pane_in_mode}' { send -X page-up } {copy-mode -u } }",
+		"bind -n MouseDown1ScrollbarDown { if -Ft= '#{pane_in_mode}' { send -X page-down } {copy-mode -d } }",
+		"bind -n MouseDrag1ScrollbarSlider { if -Ft= '#{pane_in_mode}' { send -X scroll-to-mouse } { copy-mode -S } }",
 
 		/* Copy mode (emacs) keys. */
 		"bind -Tcopy-mode C-Space { send -X begin-selection }",
--- a/tmux.1
+++ b/tmux.1
@@ -2213,6 +2213,33 @@ but also exit copy mode if the cursor reaches the bottom.
 Scroll so that the current line becomes the middle one while keeping the
 cursor on that line.
 .It Xo
+.Ic scroll-to-mouse
+.Xc
+Scroll pane in copy-mode when bound to a mouse drag event.
+.Fl e
+causes copy-mode to exit when at the bottom.  By default
+.Ar MouseDrag1ScrollbarSlider
+is bound to:
+.Pp
+.Bd -literal -offset indent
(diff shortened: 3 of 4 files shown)


mgrant0:
This fixes issue #4727.  Will this break anything else?

mgrant0:
Some background on this...  Clicking or dragging in the scrollbar executes the `copy-mode -S` function which then calls the proper scrolling functions.  The original intent of the scrollbars was to be able to easily manipulate a scrollbar and go up into copy-mode.  But it turned out that the scrollbars just worked everywhere and are especially useful for things like view mode where there's more lines than screen lines.

When not in a mode, `copy-mode -S` intuitively makes sense that it's going to put the pane into copy-mode.  It starts to make less sense (or rather it feels hacky) if the pane is in another mode and you are just calling a copy-mode function to scroll it.

Perhaps a better, cleaner fix for this would be to bind the scrollbar mouse "keys" to new command, say `scrollbar-action` and in that cmd, it could have a '-c' flag which would "go into copy mode if not already in a mode."  Thoughts?

nicm:
I think you should just change the behaviour of `copy-mode -S` - if there is no mode, enter copy mode and scroll, otherwise scroll the active mode.

mgrant0:
Ok, let me rework this PR.  I'll modfy `cmd-copy-mode.c` instead.

nicm:
This is not right, it will break the ability of entering copy mode when in another mode.

I think you need to move `-S` to the start of the function where `-q` is and add some logic to it so it looks at what the active mode is and decides what to do (enter copy mode or scroll the existing mode without changing it).

mgrant0:
I did not move the '-s' (src pane) option up too, does this matter?  Would the `-s` and `-S` options ever make sense together?

nicm:
I don't think you need to do anything with `-s`.

Your change works for `-S` but `-u` and `-d` are still a problem. What about just doing this instead (below)?

This means if you use the mouse in a mode it doesn't switch, but non-mode key bindings for example `C-b PPage` will still do so.
~~~~
--- cmd-copy-mode.c     12 Nov 2024 09:32:56 -0000      1.49
+++ cmd-copy-mode.c     4 Dec 2025 14:01:45 -0000
@@ -85,7 +85,9 @@ cmd_copy_mode_exec(struct cmd *self, str
                swp = source->wp;
        else
                swp = wp;
-       if (!window_pane_set_mode(wp, swp, &window_copy_mode, NULL, args)) {
+
+       if ((!KEYC_IS_MOUSE(event->key) || TAILQ_EMPTY(&wp->modes)) &&
+           !window_pane_set_mode(wp, swp, &window_copy_mode, NULL, args)) {
                if (args_has(args, 'M'))
                        window_copy_start_drag(c, &event->m);
        }
~~~~

nicm:
Or maybe it would be more consistent not to change `copy-mode -S` but instead do it like the other bindings, so we change the binding to:
~~~~
bind -n MouseDrag1ScrollbarSlider if -F "#{pane_in_mode}" { send-keys -M } { copy-mode -S }
~~~~
Then we have copy mode handle the slider itself, same way it handles `WheelUpPane` etc.

That might be a slightly bigger change?

mgrant0:
So that I understand what you're thinking here, when the mouse starts dragging, if the pane isn't in a mode, then `copy-mode -S` puts it into copy-mode.  Then, the next drag event causes `send -M`.  

This causes `send_keys_exec` to call `window_pane_key` which looks for `wme->mode->key` in the `window_copy_mode` struct which isn't there (yet).  If I understand,  you are saying to implement `window_copy_key` that handles the position of the scrollbar?

Will `key` actually be KEYC_MOUSEDRAG1_SCROLLBAR_SLIDER in that function and I call window_copy_scroll?  

It doesn't sound too difficult, but am I on the right path here?

nicm:
Yes that's right. So mouse events in copy mode never fire the `copy-mode` command and the mode itself handles it.

The way the other keys work is you bind them to a command, so for example:
~~~~
bind-key    -T copy-mode    MouseDrag1Pane       select-pane \; send-keys -X begin-selection
bind-key    -T copy-mode    MouseDragEnd1Pane    send-keys -X copy-pipe-and-cancel
~~~~
Then the command callback gets the `struct mouse_event` as its last argument (so for example `begin-selection` uses it to start dragging).

I don't know if we can do the same for the scrollbar (it will be all clicks on the scrollbar not just the slider). I guess at the moment key presses on the slider are bypassing the mode key table somehow?

mgrant0:
The slider itself is not separate from the pane, so a key press on the slider is sent to the pane.  Tracing through it for understanding, the ↑ key is the `Up` key in tmux which is bound to `send -X cursor-up` which is sent to `cmd_send_keys_exec` which then calls `window_copy_command` which looks up `cursor-up` in the `window_copy_cmd_table` and calls `window_copy_cmd_cursor_up`.  

It seems strange to me to implement `window_copy_key` just for the scrolling.  What if I just add a new command `scroll-to-mouse` to the `window_copy_cmd_table` array?  Then the scrolling could be:

```
bind -n MouseDrag1ScrollbarSlider if -F "#{pane_in_mode}" { send -X scroll-to-mouse } { copy-mode -S }
```

Would that also solve this issue in the cleanest possible way?

nicm:
Yes you should add a new -X command if there is no suitable one already. Don't forget about the other Scrollbar key bindings as well.

mgrant0:
Yes, `MouseDown1ScrollbarUp` and `MouseDown1ScrollbarDown`.  And I gather they would be bound similarrly to `MouseDrag1ScrollbarSlider` with the `if` cause?  I'll do this, it makes sense.

nicm:
Yes, that's the idea, that's how we do it for the other mouse key bindings.

mgrant0:
I have this working now locally.  An observation, if someone wants copy-mode to exit automatically at the bottom, the `-e` arg needs to be supplied to _both_ `scroll-to-mouse` and `copy-mode` because if only on `scroll-to-mouse` what happens is first `scroll-to-mouse` gets to the bottom and resets the mode, then, the mouse fires again (on the drag end?) and pane is no longer in a mode so `copy-mode` is called and if you don't have `-e` there, it pops back into copy-mode and this would be very difficult for a user to understand what's going on without actually tracing the code in the debugger.  Similarrly, if you only supply it on `copy-mode`, it does not exit copy-mode at the bottom so even on the final event, it still calls `scroll-to-mouse` without the `-e` option.

```
bind -n MouseDrag1ScrollbarSlider { if -Ft= '#{pane_in_mode}' { send -X scroll-to-mouse -e } { copy-mode -Se } }
```

Instead of having to specify `-e` on both, is there a more intuitive way? 

Or at worst I could note this in the scroll-to-mouse part of the man page (to be written)?

nicm:
You can't distinguish DragEnd and not fire the command for it? I think `-e` for both is fine if that's the way it is.

mgrant0:
ok, a small note in the `scroll-to-mouse` man section.

mgrant0:
Please let me know if anything more to do on this.  Thanks.

mgrant0:
I just reread your comment Nic, hmm, maybe I can distinguisd dragend here.  It shouldn't call copy-mode -S on drag-end, but i need to trace that, i'm not 100% sure that's what it's doing.  I just suspect so.

mgrant0:
The final drag key is `MouseDrag1ScrollbarSlider`, so no, not getting a end drag notification.  Nevermind, this is good enough I think.

nicm:
This is ready to go?

mgrant0:
I did not see any obvious way to improve the key binding but I am willing to try harder if you want, otherwise, I'm ok to stop  here.

@jmtysonjr ?  This completely address the issue you found?

Is there anything else this change could affect that should be tested?  This is my more my concern.  I would not want this to break something else unexpected.

nicm:

jmtysonjr:
@mgrant0, I've been running with your fix for several days and haven't
noticed any issues.
Looks good to me.

PR #4736, opened 2025-12-04
fix: support skin tone for dancing lady emoji

jake-stewart:
the 💃 was missing from `utf8_should_combine` causing rendering problems.


nicm:

PR #4738, opened 2025-12-04
Fix y offset of mouse if status at top.

mgrant0:
Take into account if the status line is at the top of the screen when calculating the position of the scrollbar.


nicm:
I've applied this now, thanks!

PR #4744, opened 2025-12-05
Support synchronized output mode (DECSET 2026) from applications

chrislloyd:
When an application enables synchronized output (`ESC[?2026h`), tmux now defers flushing pane output until the application disables it (`ESC[?2026l`) or a 1 second timeout expires. This reduces tearing for high-frequency updating applications running inside tmux.

The implementation follows the [synchronized output specification](https://github.com/contour-terminal/vt-extensions/blob/master/synchronized-output.md):
- Mode 2026 enables/disables synchronized output
- 1 second timeout prevents hanging if app misbehaves
- Mode is cleared on pane resize
- Redraw cancels sync mode (tmux can always redraw when needed)

This demo is performed in Apple Terminal which _doesn't support_ synchronized output. The demo draws 10 frames with a 0.01s delay between each bar in a frame. In tmux 3.5a, [the output is unbuffered and the frame tears](https://asciinema.org/a/dbZuuSTsO4QyO523Y4Jvvn1rw). With this patch applied, [the output is buffered with no tearing](https://asciinema.org/a/loDOJJicYtr9ZV84EkxrPscb8).

https://github.com/user-attachments/assets/00f4289b-9b36-4d1b-8366-b1901fa2a8cc

## Test plan

`regress/synchronized-output.sh` passes all tests:
- Initial state is 0
- `ESC[?2026h` sets flag to 1
- `ESC[?2026l` clears flag to 0  
- Auto-clear after 1 second timeout
- Clear on window resize
- Nested BSU is idempotent (multiple `ESC[?2026h` don't break things)
- Content buffered during sync mode
- Content visible after sync ends or timeout

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/format.c
+++ b/format.c
@@ -1947,6 +1947,18 @@ format_cb_origin_flag(struct format_tree *ft)
 	return (NULL);
 }
 
+/* Callback for synchronized_output_flag. */
+static void *
+format_cb_synchronized_output_flag(struct format_tree *ft)
+{
+	if (ft->wp != NULL) {
+		if (ft->wp->base.mode & MODE_SYNC)
+			return (xstrdup("1"));
+		return (xstrdup("0"));
+	}
+	return (NULL);
+}
+
 /* Callback for pane_active. */
 static void *
 format_cb_pane_active(struct format_tree *ft)
@@ -3442,6 +3454,9 @@ static const struct format_table_entry format_table[] = {
 	},
 	{ "start_time", FORMAT_TABLE_TIME,
 	  format_cb_start_time
+	},
+	{ "synchronized_output_flag", FORMAT_TABLE_STRING,
+	  format_cb_synchronized_output_flag
 	},
 	{ "tree_mode_format", FORMAT_TABLE_STRING,
 	  format_cb_tree_mode_format
--- a/input.c
+++ b/input.c
@@ -898,6 +898,8 @@ input_free(struct input_ctx *ictx)
 	evbuffer_free(ictx->since_ground);
 	event_del(&ictx->ground_timer);
 
+	screen_write_stop_sync(ictx->wp);
+
 	free(ictx);
 }
 
@@ -1901,6 +1903,11 @@ input_csi_dispatch_rm_private(struct input_ctx *ictx)
 		case 2031:
 			screen_write_mode_clear(sctx, MODE_THEME_UPDATES);
 			break;
+		case 2026:	/* synchronized output */
+			screen_write_stop_sync(ictx->wp);
+			if (ictx->wp != NULL)
+				ictx->wp->flags |= PANE_REDRAW;
+			break;
 		default:
 			log_debug("%s: unknown '%c'", __func__, ictx->ch);
 			break;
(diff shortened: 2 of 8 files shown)


nicm:
Thanks. Can you show me two videos of a program in tmux with and without this change that shows the problem happening and it solving it?

You may want to turn off tmux's own use of the sequence (`set -as terminal-overrides '*:Sync@'`).

chrislloyd:
Thanks for the fast reply @nicm. Added a video to the PR + ASCII casts of the output if you want to inspect further. Also added a few more tests specifically for the output and the timeout.

nicm:
Thanks. TBH I'm not all that convinced by your example because it is contrived demo rather than a real application, but that's not to say we can't do it - it isn't a huge change and we can take it it will help something:

- Because you wrote this with AI, it isn't clear that you have the copyright, so we need a link to a statement from the Claude website showing that they release any copyright over this code to you (and therefore you own it and are able to license it with the ISC license used by tmux).

- You need to delete the timer if it is running when the `input_ctx` is freed, or tmux will crash. So you will need to delete it in `input_free`, but...

- Having the timer in `input.c` but `MODE_SYNC` check in `screen-redraw.c` and and `screen-write.c` seems wrong. Shouldn't the timer be in `screen-write.c` as well?

- I'm not sure the application should be able to prevent tmux redrawing the whole pane if it wants to, so instead of the timer blocking `screen_redraw_draw_pane`, it should probably be the other way round - `screen_redraw_draw_pane` should cancel the timer and sync flag? Or does this stop it working?

- How does this interact with tmux's existing `Sync` support? Can we delete that (it doesn't work very well)?

- `synchronized_output_flag` needs to be added to `tmux.1`.

chrislloyd:
Thanks for the review @nicm - I'm new to the codebase and appreciate your hand-holding.

I work on [Claude Code](https://claude.com/product/claude-code), a terminal AI agent. CC is interactive and sometimes clears/redraws large amounts of scrollback in order to render changes there. It sends `BSU`/`ESU` to mark the start and end of these draw cycles and to prevent tearing for users. While this works when a user is running CC directly in a terminal (like Ghostty or Kitty), when they run it via tmux those commands are swallowed and people see the tearing.

This PR is orthogonal to tmux's own sync support - this handles *incoming* sync from applications, while tmux's existing sync handles *outgoing* sync to the parent terminal. People would see the benefit of this even if tmux had no sync support itself.

(Also, [Anthropic waives all rights of the outputs of Claude.](https://www.anthropic.com/legal/consumer-terms#:~:text=Subject%20to%20your%20compliance%20with%20our%20Terms%2C%20we%20assign%20to%20you%20all%20of%20our%20right%2C%20title%2C%20and%20interest%E2%80%94if%20any%E2%80%94in%20Outputs.))

nicm:
Sorry didn't see your comment until now.

Did you make these changes? I can take another look.

nicm:
OK I think this looks good.

nicm:
I've applied this to OpenBSD now, it should be in GitHub later. Thanks!

nicm:
This is in GitHub now, I was going to apply your regression test as well but it fails (test 7 and 9a). Does it work for you?

chrislloyd:
Sorry for the slow reply @nicm (turned off Github over the break) - looking now.

Issue #4032, opened 2024-07-07
Use popups as toast notifcations?


Garulf:
I have been improving my workflow using tmux's awesome `display-popup` and `display-menu` command. Would it be possible to have an option for a popup menu to not steal focus? This would allow for some cool use cases like toast notifications.

I included some examples below:

![image](https://github.com/tmux/tmux/assets/535299/03f0d353-58ae-4cb7-a180-355d7ca5c119)

![image](https://github.com/tmux/tmux/assets/535299/9caa0019-4978-4c69-9e83-81ce0a6f0ec2)

nicm:
I think it would be possible for popups not to steal key presses from the panes now that they do not block redrawing, but I don't know how difficult it would be. Look at the overlay section in `server_client_handle_key`, I would start by disabling that and see if it works, if it does then you would need to add a flag to say "this is noninteractive overlay".

PR #4655, opened 2025-10-19
Feature floating window panes

mgrant0:
This adds a new type of window pane which is floating (as in not tiled). 

1. It creates a new type of window pane that is not part of the tiled pane layout.  These new panes are functionally identical to the tiled window panes except that they can be dragged around and resized.
a. Dragging the left, lower, or right edges, or dragging one of the corners resizes the pane.  Dragging the top border of the pane moves it around.
b. A new command, new-pane (newp), to create a new floating window pane, bound to `PREFIX *`.
c. A new menu item in the mouse menu to start a floating pane.
d. command options to be able to specify x, y, width and height of new pane.

3. Like the existing overlay popup, a floating pane can be started with a command or not and used to send output to.  The major difference is floating panes run in the session, not in the client, and they do not block the terminal when started.

4. Floating panes can be zoomed and unzoomed like tiled panes.

5. Floating panes can be minimised (made invisible).  
a. To make a pane visible again, you can click on it in the status line 2. 
b. A new command to cycle through the floating panes, unminimising them either one by one or all of them.
c. There is no reason tiled panes should not also be able to minimise and adjacent pane expaneds into the empty space, this is similar to zooming and unzooming.  A minimised tiled pane is simply hidden in the layout tree.  `layout_fix_panes` ignores minimised panes.

6. Floating panes have a z-index so that one floating pane can slide behind another of a lower z-index.  
a. The tiled panes are always below the floating panes.  
b. The "desktop" (the background of middle dots) is always behind everything and is the lowest layer.
c. The dot character used for the desktop can be modified.  A future mod could allow for ascii background art or maybe even sixel images as the desktop.

7. A floating pane has a `layout_cell` but is not part of the tiled layout tree.  It will be possible to convert a floating pane to a tiled pane by inserting it into the layout tree. 
a. A new command that takes a the active pane and a selected tiled pane and inserts it next to it in the layout tree.  
b. Allow the user to drag a floating pane to a specific place, for example, the top of the window or top pane border of a tiled pane and the floating pane inserts itself there in the layout tree.  In this way, a floating pane could be dragged around and inserted in real time where the user desired or they could drag away and the pane continues to be a floating pane.  Need to save the layout tree before modifying it mid-drag in case the user drags away, you want the tiled panes to return to the way they were.  There should be a small time delay before automatically inserting a pane to avoid excessive insert/uninsert actions as a pane is dragged around.
  
[...]

PR #4746, opened 2025-12-05
Add -P flag to display-popup for passthrough/non-blocking mode

patrick-motard:
## Summary

This PR adds a new `-P` flag to `display-popup` that makes the popup non-blocking: keyboard input is passed through to the underlying pane instead of being captured by the popup.

**Use case:** Toast-style notifications that display information without interrupting the user's workflow. Combined with `-E`, popups can auto-dismiss when their command exits while allowing uninterrupted typing.

https://github.com/user-attachments/assets/7a1f3231-3d27-4dfc-ba24-a54777af32ef

**Example:**
```bash
display-popup -P -E -w 30 -h 3 "echo 'PR cache refreshed'; sleep 2"
```

This shows a centered popup for 2 seconds, but the user can continue typing in their pane while it's displayed.

## Changes

- `tmux.h`: Added `POPUP_PASSTHROUGH` flag
- `cmd-display-menu.c`: Added `-P` flag parsing
- `popup.c`: Return `-1` from key callback in passthrough mode to signal keys should pass through (Escape/Ctrl-C still close the popup)
- `server-client.c`: Handle `-1` return from overlay key callback to pass keys to underlying pane
- `tmux.1`: Documentation for the new flag

## Implementation notes

Based on @nicm 's suggestion in #4032 to look at `server_client_handle_key` overlay handling. The overlay key callback now returns:
- `0`: Key consumed by overlay
- `1`: Close overlay  
- `-1`: Pass key through to underlying pane (new)

The `-P` flag can be combined with other flags like `-E` (close on exit) for the toast notification use case.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -54,8 +54,8 @@ const struct cmd_entry cmd_display_popup_entry = {
 	.name = "display-popup",
 	.alias = "popup",
 
-	.args = { "Bb:Cc:d:e:Eh:kNs:S:t:T:w:x:y:", 0, -1, NULL },
-	.usage = "[-BCEkN] [-b border-lines] [-c target-client] "
+	.args = { "Bb:Cc:d:e:Eh:kNPs:S:t:T:w:x:y:", 0, -1, NULL },
+	.usage = "[-BCEkNP] [-b border-lines] [-c target-client] "
 		 "[-d start-directory] [-e environment] [-h height] "
 		 "[-s style] [-S border-style] " CMD_TARGET_PANE_USAGE
 		 " [-T title] [-w width] [-x position] [-y position] "
@@ -503,6 +503,11 @@ cmd_display_popup_exec(struct cmd *self, struct cmdq_item *item)
 		if (flags == -1)
 			flags = 0;
 		flags |= POPUP_CLOSEANYKEY;
+	}
+	if (args_has(args, 'P')) {
+		if (flags == -1)
+			flags = 0;
+		flags |= POPUP_PASSTHROUGH;
 	}
 
 	if (modify) {
--- a/popup.c
+++ b/popup.c
@@ -153,6 +153,10 @@ popup_mode_cb(__unused struct client *c, void *data, u_int *cx, u_int *cy)
 	if (pd->md != NULL)
 		return (menu_mode_cb(c, pd->md, cx, cy));
 
+	/* In passthrough mode, keep cursor in underlying pane. */
+	if (pd->flags & POPUP_PASSTHROUGH)
+		return (NULL);
+
 	if (pd->border_lines == BOX_LINES_NONE) {
 		*cx = pd->px + pd->s.cx;
 		*cy = pd->py + pd->s.cy;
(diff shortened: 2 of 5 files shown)


patrick-motard:
I noticed just now that this feature is already being worked on elsewhere. I can keep going to add mouse support, or abandon this in favor of the other approach. Whoops!

nicm:
Changes to popups are kind of on hiatus for a bit because @mgrant0 is working on floating panes and once they are done we might look at either removing or neutering popups, or at least have a think about what we want them to be able to do...

https://github.com/tmux/tmux/pull/4655 is his floating panes code and https://github.com/tmux/tmux/pull/4379 is the nonblocking popups which I guess you found already.

patrick-motard:
Got it! I'll close this. Gonna keep it alive in my fork though until the alternative comes. Im using pop-ups for notifications when background processes I have tmux spawn complete certain actions.

mgrant0:
Hi @patrick-motard I think the floating panes mod is what you want.  The existing popups have several limitations.  They are in the client, they don't have mouse support and they are basically implemented with the same code as the menu popups.  They are like "second class citizens".

The floating panes work lets window panes be anywhere in the window and are the same "class" as the other window panes.  

I am on the verge of pushing some changes to that, hopefully today if I can sort my "one last thing" out.  I'd love some feedback and QA on that if you are willing.

nicm:
If you want to use a pane for "display only" (that is, don't run a process in it), you can do it by making it an empty pane:
~~~~
P=$(tmux splitw -dP '')
printf '\033[42mmy output\033[0m'|tmux display -It$P
~~~~
I assume this will also work with a floating pane :-).

mgrant0:
image

Yes it does!

mgrant0:
One difference is that 'q' doesn't quit this floating pane.  PREFIX x does.  

Should there be some sort of option to the floating panes to allow pressing any key or 'q' to quit the pane?

nicm:
Couldn't you do that yourself? Something like:
~~~~
bind -n q if '#{@kill_pane_on_q}' killp send

mgrant0:
I was thinking that if a new pane was created (or split like your example) with the sole purpose to output something, maybe there could be a default keybinding like this.  I'm thinking similar functionality to the popup window.

mgrant0:
Like if these things were started with `''` or maybe some option flag which would assign a different key table.

nicm:
I don't think we need shorthands for things that you can do a slightly longer way when there is no evidence people use them enough to warrant it.

mgrant0:
I'm not seeing what you are seeing.  Where would you put this `bind -n q` key binding so it affected the new window?

mgrant0:
I see how to do something like `tmux bind -Tfoo q if '#{@kill_pane_on_q}' killp send` but I don't see how to select this new key table `foo` in the new pane.  Would it make sense to add a `-T` option to these commands to set the default key table?  or am I missing something?

nicm:
You can't set key tables on panes, only on sessions. I'm not sure it is worth worrying about this right now, once floating panes are done we can see how they are being used and what is missing and if they need key tables they can be added then. Initially at least if you want to set keys only for a floating pane, you can do it the same way you would for any pane, by binding them in the root table and checking a pane option or the pane ID or some other criteria.

nicm:
I'm going to close this for the moment. Thanks!

PR #4747, opened 2025-12-06
Add "box" option to pane-border-indicators

patrick-motard:
Add a new "box" option for `pane-border-indicators` that draws a complete box border around the active pane instead of using shared separator borders between panes.

- Add `PANE_BORDER_BOX` constant and `window_pane_box_mode()` helper function
- Reduce internal screen buffer and PTY by 2 in each dimension to reserve space for box borders
- Offset pane content drawing by 1 to account for box border
- Adjust cursor positioning for box mode offset
- Draw styled box border for active pane, blank borders for inactive panes
- Skip normal pane separator borders when box mode is active
- Clear separator cells between panes that would otherwise show old borders

---

Hey @nicm - There's still a bit of polish I want to add before submitting it for a full review, but I wanted to share this draft with you on the off chance that you may be interested in a fully implemented version of this being added to `tmux/tmux`. 

I have a heck of a time determining which pane is in focus when there are multiple panes in a window. The default boarder behavior trips me up. Updating my config to make the unfocused panes to be dimmed helped, but whenever panes have neovim open in them those panes aren't dimmed which leads to me still being unsure which pane is focused. So I decided to take a crack at adding support for adding a frame around all sides of the focused panel. 

Here's a screencast showing the current behavior of this PR.

https://github.com/user-attachments/assets/45e0a432-a0ad-4c5e-80fc-9fb643a4a724

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -73,7 +73,7 @@ static const char *options_table_pane_status_list[] = {
 	"off", "top", "bottom", NULL
 };
 static const char *options_table_pane_border_indicators_list[] = {
-	"off", "colour", "arrows", "both", NULL
+	"off", "colour", "arrows", "both", "box", NULL
 };
 static const char *options_table_pane_border_lines_list[] = {
 	"single", "double", "heavy", "simple", "number", "spaces", NULL
@@ -1282,8 +1282,9 @@ const struct options_table_entry options_table[] = {
 	  .scope = OPTIONS_TABLE_WINDOW,
 	  .choices = options_table_pane_border_indicators_list,
 	  .default_num = PANE_BORDER_COLOUR,
-	  .text = "Whether to indicate the active pane by colouring border or "
-		  "displaying arrow markers."
+	  .text = "Whether to indicate the active pane by colouring border, "
+		  "displaying arrow markers, or drawing a box around the "
+		  "active pane."
 	},
 
 	{ .name = "pane-border-lines",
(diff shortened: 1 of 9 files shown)


nicm:
Thanks. This looks quite nice and the code looks OK, some comments on behaviour:

- The mouse doesn't work on the borders, you have to drag where the split would have been which is invisible, this isn't very intuitive.

- I find the way that the borders of all the inactive pane disappears completely and only the active pane has a border quite attractive, but others may disagree, and not having a way to have a box round inactive panes as well seems like it would be a missed opportunity (maybe `box-all` or something?).

- If the boxes round inactive panes *are* invisible, there is no need to reserve space for them - you could put the bottom line of the top pane's box on the same line as the pane below's box's top line and avoid wasting a line/column for each pane.

- It seems like it gets confused if the window is bigger than the visible area. Try `resize-window -x500 -y500`.

patrick-motard:
Thank you for the feedback @nicm. Iterating and will get back to you.

patrick-motard:
Mouse re-sizing is working. Will push soon.


https://github.com/user-attachments/assets/a008c1c6-0fb3-44f3-baf3-c23ef93ff6c3

eMPee584:
@patrick-motard
By the way, check out `set-window-option -g pane-border-indicators arrows` along with https://github.com/eMPee584/tmux-pane-flash .. for a different scratch to the same itch: easier visual detectability of the active pane 🙃

Oh and another by the way, please, @admins: the auto-closing and locking of issues and PRs is driving digital hygiene a bit over the top imho; preventing any further discussion on a topic is really not the most useful thing to do as a default, is it 🤣

nicm:
I'm going to close this for the moment, but please let me know if you get back to it. Thanks!

PR #4748, opened 2025-12-07
Added startup.rc, startup.sh

roccohimel:
Simply starts custom run commands for tmux


nicm:
Did you open this by mistake?

PR #4750, opened 2025-12-07
Fix list-keys output for commands with different groups

### LINKED ISSUE #3455: [3.3a] Newline in braces is converted to `\;\;`
### Issue description

I have this in tmux conf:

~~~
bind A { run true
         run true
       }
~~~

`tmux lsk` outputs this:

~~~
$ tmux lsk -Tprefix A
bind-key -T prefix A run-shell true \;\; run-shell true
~~~

As it shows the newline in braces is converted to double `\;`. It does not impact functionality. Just does not look nice.

### Required information

* tmux version (`tmux -V`): 3.3a
* Platform (`uname -sp`): Darwin i386

patrick-motard:
### Problem

When using newlines within braces in a tmux configuration:

```
bind A { run-shell true
         run-shell true
       }
```

The output from `tmux list-keys` showed an invalid `\;\;` separator:

```
bind-key -T prefix A run-shell true \;\; run-shell true
```

The `;;` syntax is not understood by the tmux parser, so the output cannot be fed back into a config file.

Fixes #3455

### Solution

Modified `cmd_list_print()` in `cmd.c` to always use `\;` (single semicolon) as the command separator, regardless of whether commands have different groups.

The group information is still preserved internally for error handling during command execution - this change only affects the display/serialization of command lists.

### Before/After

| Input | Before | After |
|-------|--------|-------|
| `bind A { cmd1`<br>`  cmd2 }` | `bind-key -T prefix A cmd1 \;\; cmd2` | `bind-key -T prefix A cmd1 \; cmd2` |

### Testing

1. Create a test config with multi-line brace syntax:
   ```bash
   cat > /tmp/test.conf << 'EOF'
   bind A { run-shell true
     run-shell true
   }
   EOF
   ```

2. Start tmux with the config and check list-keys output:
   ```bash
   tmux -L test -f /tmp/test.conf new-session -d
   tmux -L test list-keys | grep "prefix.*A"
   ```

   Expected output:
   ```
   bind-key -T prefix A run-shell true \; run-shell true
   ```

3. Verify round-trip parsing works (output can be re-used as input):
   ```bash
   tmux -L test list-keys | grep "prefix.*A" > /tmp/roundtrip.conf
   tmux -L test2 -f /tmp/roundtrip.conf new-session -d
   tmux -L test2 list-keys | grep "prefix.*A"
   ```

   Both outputs should be identical.

4. Cleanup:
   ```bash
   tmux -L test kill-server
   tmux -L test2 kill-server
   ```

### DIFF
--- a/cmd-list-keys.c
+++ b/cmd-list-keys.c
@@ -109,7 +109,8 @@ cmd_list_keys_print_notes(struct cmdq_item *item, struct args *args,
 		key = key_string_lookup_key(bd->key, 0);
 
 		if (bd->note == NULL || *bd->note == '\0')
-			note = cmd_list_print(bd->cmdlist, 1);
+			note = cmd_list_print(bd->cmdlist,
+			    CMD_LIST_PRINT_ESCAPED|CMD_LIST_PRINT_NO_GROUPS);
 		else
 			note = xstrdup(bd->note);
 		tmp = utf8_padcstr(key, keywidth + 1);
@@ -288,7 +289,8 @@ cmd_list_keys_exec(struct cmd *self, struct cmdq_item *item)
 			tmpused = strlcat(tmp, " ", tmpsize);
 			free(cp);
 
-			cp = cmd_list_print(bd->cmdlist, 1);
+			cp = cmd_list_print(bd->cmdlist,
+			    CMD_LIST_PRINT_ESCAPED|CMD_LIST_PRINT_NO_GROUPS);
 			cplen = strlen(cp);
 			while (tmpused + cplen + 1 >= tmpsize) {
 				tmpsize *= 2;
--- a/tmux.h
+++ b/tmux.h
@@ -2734,6 +2734,8 @@ void		 cmd_list_append(struct cmd_list *, struct cmd *);
 void		 cmd_list_append_all(struct cmd_list *, struct cmd_list *);
 void		 cmd_list_move(struct cmd_list *, struct cmd_list *);
 void		 cmd_list_free(struct cmd_list *);
+#define CMD_LIST_PRINT_ESCAPED 0x1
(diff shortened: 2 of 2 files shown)


nicm:
Yes... but this means we don't see them in the logs which is why it was changed to show them as `;;`. I think you will need to make it optional for `cmd_list_print` to include or not include them. Nicest way I think would be to change `int escaped` to `int flags` and add two flags, something like `CMD_LIST_PRINT_ESCAPED 0x1` and `CMD_LIST_PRINT_NO_GROUPS 0x2`.

nicm:

patrick-motard:
## Testing Steps

```bash
# 1. Create a test config with multi-line brace syntax (creates different command groups)
cat > /tmp/test.conf << 'CONF'
bind A { run-shell true
  run-shell true
}
CONF

# 2. Start tmux with the test config
./tmux -L test -f /tmp/test.conf new-session -d

# 3. Verify list-keys shows \; (single semicolon) - can be round-tripped
./tmux -L test list-keys | grep "prefix.*A"
# Expected: bind-key -T prefix A run-shell true \; run-shell true

# 4. Test round-trip: save output and reload it
./tmux -L test list-keys | grep "prefix.*A" > /tmp/roundtrip.conf
./tmux -L test2 -f /tmp/roundtrip.conf new-session -d
./tmux -L test2 list-keys | grep "prefix.*A"
# Expected: identical output to step 3

# 5. Cleanup
./tmux -L test kill-server
./tmux -L test2 kill-server
```

The key behavior change:
- `list-keys` output now uses `\;` instead of `\;\;` for commands with different groups
- This allows the output to be parsed back by tmux (round-trip)
- Logs and other internal callers still show `;;` for different groups (via the flags mechanism)

nicm:

PR #4751, opened 2025-12-07
Add F1 help key binding to choose modes

### LINKED ISSUE #4357: FR: make key bindings for `choose-tree`, `choose-buffer`, `customize-mode`, etc self-documenting
I wish it was possible to see the bindings for commands like `choose-tree` without having to search through the man page every time. Ideally, they would be discoverable from the `choose-xxx` interface. This could use a binding, or a special entry in the list that shows help in the preview window.

It's not immediately clear what the best binding is. `F1` is probably good. The F keys are, these days, hard to access on some keyboards, so it'd be nice to have an additional binding. In Vim mode, `?` and `h` are taken, but `c-h` and/or `c-/` would work.

In any case, the first step might be to have an option to `tmux list-keys`, e.g. `tmux lsk -c choose-tree`, that lists the bindings for such commands.

------

An alternative approach would be to make the docs (currently in the man page) for `choose-xxx` commands (or for all commands) easier to reach. E.g., they could be reachable with `tmux lscm -h choose-buffer` or `man tmux-choose-buffer` or `tmux man choose-buffer | man -`. This deserves a whole separate FR, though; the main obstacle here seems to be how to do it portably and without increasing the maintenance burden.

patrick-motard:
### Problem

Users find it difficult to discover key bindings for choose modes (choose-tree, choose-buffer, choose-client, customize-mode) without consulting the manual.

Fixes #4357

### Solution

Add an F1 key binding that displays a help menu showing all available key bindings in choose modes. The menu lists navigation, selection, tagging, searching, and sorting keys.

### Key Choice

F1 was chosen because the more intuitive alternatives are already in use:

- `h` - Used for collapse/left navigation (same as Left arrow)
- `?` - Used for forward search (same as `/`)
- `H` - Used in window-tree.c to jump to the current pane ("Home")

F1 is a conventional help key that doesn't conflict with existing bindings.

### Testing

1. Start tmux: `tmux new-session`
2. Enter choose-tree mode: `Ctrl+b w`
3. Press `F1` to see the help menu
4. Press `q` or `Escape` to close the help menu

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -133,6 +133,32 @@ static const struct menu_item mode_tree_menu_items[] = {
 	{ NULL, KEYC_NONE, NULL }
 };
 
+static const char mode_tree_help_text[] =
+	"Up, k              Move cursor up\n"
+	"Down, j            Move cursor down\n"
+	"g                  Go to top\n"
+	"G                  Go to bottom\n"
+	"PgUp, C-b          Page up\n"
+	"PgDn, C-f          Page down\n"
+	"Left, h            Collapse item\n"
+	"Right, l           Expand item\n"
+	"M--                Collapse all items\n"
+	"M-+                Expand all items\n"
+	"t                  Toggle item tag\n"
+	"T                  Untag all items\n"
+	"C-t                Tag all items\n"
+	"/                  Search forward\n"
+	"?                  Search backward\n"
+	"n                  Repeat search forward\n"
+	"N                  Repeat search backward\n"
+	"f                  Filter items\n"
+	"O                  Change sort order\n"
+	"r                  Reverse sort order\n"
+	"v                  Toggle preview\n"
+	"q, Escape          Exit mode\n"
+	"\n"
+	"Press any key to close";
+
 static int
 mode_tree_is_lowercase(const char *ptr)
 {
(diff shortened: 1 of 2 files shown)


nicm:
Thanks:

- Can you make it `C-h` as well as `F1` like emacs?

- This is not much use if nobody knows about it, so can you add it to `tmux.1`?

- I think if this is going to be a menu then choosing an item should execute the command as if the key was pressed. If must be static then it should be a popup not a menu.

patrick-motard:
I can add those. Thanks for the review!

nicm:

patrick-motard:
https://github.com/user-attachments/assets/40b91af3-4c37-4c42-a2e7-9b599722d61f

patrick-motard:
Sorry for the delay. Let me know if [this](https://github.com/tmux/tmux/pull/4751/commits/c807f698b60a00a33138d88fd56e9ea4ddf34610) is what you had in mind.

patrick-motard:
1. C-h key binding - Added alongside F1
2. Popup instead of menu - Help now displays as a static popup that closes on any key
3. Documentation in tmux.1 - Added F1/C-h help entries to all four choose modes (choose-tree, choose-client, choose-buffer, customize-mode)

nicm:
Thanks.

I actually liked the menu, it just needed to execute the command when you chose it :-).

But a popup is OK instead, but we can't run `cat` just to show some output.

Please take a look at this which adds a way to directly write into a popup: [help.diff.txt](https://github.com/user-attachments/files/24450166/help.diff.txt)

I also added some formatting to the text.

The only problem now is that this is only the common keys, it doesn't include stuff like `D` in client mode, or `f` or `e` in buffer mode. I would probably fix this by just allowing a mode to completely replace the text, so adding a callback like:
~~~~
struct window_mode {
...
    const char* (*help_text)(u_int *width, u_int *height, size_t *size);
...
}
~~~~
I don't see any need to get complicated and try to merge common and individual mode help, I think we just copy the common keys into each one.

nicm:
Did you get a chance to look at my changes? Do you want to work on this further or shall I do it?

patrick-motard:
I can get this over the finish line. My availability comes in small batches. I'll take  look at the code you shared. I can update it to be an interactive menu that executes selected commands.

nicm:
Good stuff, thanks. I don't mind if it is a menu or a popup, I think it looks OK as a popup the way I have it. But it should include the commands that are special to each mode.

nicm:
I extended my diff to include mode-specific text. Is in OpenBSD now, will be in GitHub later. Thanks!

PR #4754, opened 2025-12-09
Updated README file to show libncurses-dev as a required package

roccohimel:
Somehow Debian 13 doesn't ship with libncurses-dev, I added that to README to make notice of it being required


nicm:
`ncurses` is already listed right above this in the "Dependencies" section, I don't think we need to mention specific packages for different platforms.

What might be useful is a link to https://github.com/tmux/tmux/wiki/Installing which does list the packages required for major platforms.

roccohimel:
Oh sorry didn't see

Issue #4712, opened 2025-11-27
The clock isn't accurate to the second


stevenxxiu:
### Issue description

*Tmux* v3.6 has #4697 which adds seconds to the clock. I tried this in 2 panes, but the seconds aren't in sync, hence isn't accurate. Personally I want to use this to sync irl clocks, so accuracy is important.

I use <https://github.com/nwtnni/tock>, which is in sync between terminal windows.

My config:

```tmux
set-option -g clock-mode-style '12-with-seconds'
```

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

- tmux version (`tmux -V`). `tmux 3.6`
- Platform (`uname -sp`). *Arch Linux*
- Terminal in use (xterm, rxvt, etc). *Ghostty*
- \$TERM *inside* tmux (`echo $TERM`). `tmux-256color`
- \$TERM *outside* tmux (`echo $TERM`). `xterm-ghostty`
- Logs from tmux (`tmux kill-server; tmux -vv new`). Can provide if needed, don't think it's too important here.

nicm:
This should be fixed as soon a GitHub syncs up. Thanks!

PR #4760, opened 2025-12-16
Fix clock synchronization in clock-mode across multiple panes.

deechtejoao:
Previously, clock panes were not synchronized as each used its own independent timer.
This caused seconds to drift between panes, affecting precise time synchronization.

All clock panes now update simultaneously at exact second transitions, ensuring synchronization.


ghost:
This problem was discussed previously in https://github.com/tmux/tmux/issues/4712, I think this should fix all the problems described here.

nicm:
Nice one, applied with a minor change to made the `event_add` bit a common function. Will be in GitHub later. Thanks!

PR #4765, opened 2025-12-19
Add prompt-command-cursor-style option

joshua-cooper:
This PR adds prompt-command-cursor-style to let you set the cursor style separately in insert/normal mode in the command prompt when status-keys is vi.

This lets the cursor shape act as a lightweight mode indicator without needing different colours or prompt styles. This is done in neovim by default too.


nicm:

Issue #3503, opened 2023-03-23
Is it possible to reorder tmux panes by dragging them using mouse?


j-rahman:
### Issue description
This is similar to issue #656 but, instead of windows in the status bar, I would like to drag the borders of panes in order to rearrange them within a single window.

Is that possible? If yes, could you please show me how? If not, might I suggest this as a feature request? Thank you.

### Required information

* tmux version: `tmux 3.1c`.
* Platform: `Linux x86_64 RHEL 8.7 (Ootpa)`.
* $TERM inside and outside of tmux: `screen-256color` and `xterm-256color`.

nicm:
Not at the moment, I tried to do this before and it didn't work...

nicm:
Added to todo list.

Issue #656, opened 2016-11-18
renumber/reorder windows by dragging using mouse


dmd:
It would be nice if, in mouse mode, one could drag window titles to reorder/renumber them (like tabs in a browser).

dmd:
It would be nice to leave a comment like 'won't do this' or something instead of just closing with no explanation at all.

KamarajuKusumanchi:
I am curious why this bug is closed? This will be a nice features to have!

sudobash1:
For those who find this issue in a Google search, you need to add a `-d` flag to `swap-window` now (as of version 3.0 I believe). This prevents the window you are swapping from being selected. If you don't do this, you will get tons of stuttering behavior when dragging windows.

```
bind-key -n MouseDrag1Status swap-window -d -t=
```

(Thanks to @nicm for pointing this out)

PR #4766, opened 2025-12-19
{active} token to rearrange windows and panes

peripherium:
Issues #3503 and #656 request the ability to rearrange panes or windows using the mouse.

This change introduces a special target token `{active}` (shorthand `@`) which expands to
the currently active pane or window in the client context. This allows existing commands
such as swap-window and swap-pane to operate on the active object without requiring an
explicit target.

Two example bindings, C-MouseDown1Status and C-MouseDown1Pane, are added to demonstrate
rearranging windows and panes using the mouse. The underlying change is generic and not
limited to mouse usage.

### DIFF
--- a/cmd-find.c
+++ b/cmd-find.c
@@ -991,7 +991,9 @@ cmd_find_target(struct cmd_find_state *fs, struct cmdq_item *item,
 	if (target == NULL || *target == '\0')
 		goto current;
 
-	if (strcmp(target, "@") == 0 || strcmp(target, "{active}") == 0) {
+	if (strcmp(target, "@") == 0 || strcmp(target, "{active}") == 0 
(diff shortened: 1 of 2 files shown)


nicm:
Good idea, thanks, but for windows the terminology we use is the "current window" not the "active window".

peripherium:
I’ve updated the PR accordingly: the changes are now reflected both in the code and in tmux.1.
In addition to {active}, there is now a {current} token; the two are implemented as simple synonyms.
I don’t think it’s necessary to prevent users from using {active} with windows or {current} with panes, since I don’t see this having any harmful effect on program behaviour.

nicm:
That's fine, thanks, will look next week.

nicm:

PR #4767, opened 2025-12-19
Fix cursor position after escape with status-keys vi

joshua-cooper:
When using status-keys vi, pressing Escape to enter prompt command mode now moves the cursor left by one, matching vi.


nicm:

PR #4768, opened 2025-12-19
Fix: In Cygwin, tmux's left-right split-pane display becomes garbled:…

gaheadus:
… when the left pane fills the screen and continues output, it causes the right pane to jump lines


nicm:
I don't think so, this is up to the terminal to advertise.

You should be able to do `set -as terminal-overrides '*:Cmg@:Clmg@'` to force it off.

Issue #782, opened 2017-02-16
[Feature request] Focus follows mouse for panes


choco:
I was just wondering, since https://github.com/tmux/tmux/commit/dd0c8147795c518de443c33895c614e52b42677f introduced support for 1003 mouse mode, would it be possible to add a focus follows pane feature to tmux or an event to hook on?

nicm:
It would be possible but I have no plans to do it.

PR #4771, opened 2025-12-20
Add focus-follows-mouse session option

bwsd0:
Implements the feature requested in #782. Adds a new session option `focus-follows-mouse` that automatically selects a pane when the mouse moves over it.

Requires the mouse option to be enabled. Uses all-motion mouse mode (1003h) to receive movement events without requiring a button press.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -664,6 +664,13 @@ const struct options_table_entry options_table[] = {
 	  .text = "Time for which status line messages should appear."
 	},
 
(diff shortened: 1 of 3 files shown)


nicm:
This looks OK to me.

bwsd0:
Thank you. I've been running this in a local build for 3 days without issues. PR is ready.

nicm:

PR #4778, opened 2025-12-30
Added case modifiers \u, \l, \U, \L, \E to regex.

mmjo:
Allow case modifiers in format substitutions.
e.g.:
tmux display-message -F '#{s/.*/\U\0/:pane_tty}'

(i use this with pane_mode, but that wouldn't make a good example)

Regards, Mike

### DIFF
--- a/regsub.c
+++ b/regsub.c
@@ -38,7 +38,6 @@ static void
 regsub_case_cat(char **buf, ssize_t *len, char cfirst, char crest,
     const char *text, size_t start, size_t end)
 {
-	int (*casemod)(int);
 	u_int i;
 
 	regsub_copy(buf, len, text, start, end);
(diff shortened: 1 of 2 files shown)


nicm:
Where is this syntax from? Did you invent it?

mmjo:
Vim and Perl both have it.
man perlre
vim -c ':help \\u'

nicm:
I am happy enough to add this but:
~~~~
        int (*casemod)(int);
...
        casemod = tolower;
~~~~
`tolower` and `toupper` are permitted to be macros, are we sure that they are functions on all of the platforms tmux cares about?

It would be simple enough and more portable to put the `if` inside the loop instead of using a function pointer, so maybe that would be better?

mmjo:
Good catch! I completely forgot about that.
Added a fix.

nicm:
Are you sure `regsub_case_cat` is right? Why does it change the case of the characters in `*buf` starting from `start`? `start` is an index into `text`, not into `*buf` - shouldn't it be walking from the original value of `*len` to the new value?

Perhaps it would be neater (and more efficient) if instead of copying the text and then walking it to change the case, it just copied and changed case at the same time, so basically make a version of `regsub_copy` where the `memcpy` is replaced with a loop that copies and changes case if needed?

mmjo:
Yeah, sorry, you are completely right. Hopefully it was the Christmas dinner to blame.

Also, the \U and \L must run until \E or until end of string (not only to end of e.g. \1).

I will come up with something else later.

nicm:

mmjo:
On Tue, Jan 20, 2026 at 12:26:59AM -0800, Nicholas Marriott wrote:
> nicm left a comment (tmux/tmux#4778)
> 
> Are you still working on this?

Certainly, I have some code locally, that needs finishing.
Other things distracts me at the moment, sorry.

Regards, Mike Jonkmans

nicm:

PR #4781, opened 2025-12-31
Remove unused decl of buf.

mmjo:
Silence a gcc warning.

Regards, Mike

### DIFF
--- a/compat/imsg-buffer.c
+++ b/compat/imsg-buffer.c
@@ -699,8 +699,6 @@ msgbuf_queuelen(struct msgbuf *msgbuf)
 void
 msgbuf_clear(struct msgbuf *msgbuf)
 {
-	struct ibuf	*buf;
-
 	/* write side */
 	ibufq_flush(&msgbuf->bufs);
 


nicm:
Is this already fixed in the same way in the upstream OpenBSD copy of this file?

mmjo:
That could be. I only see the github repo.
I'll see if i can find the OpenBSD repo...
It's only one line deleted though.

nicm:
Well I'm just wondering if we should sync the whole lot if it is fixed upstream.

mmjo:
Oh dear, OpenBSD is still using CVS; which I used 20 years ago at the latest.
Though funny that since the original author of CVS, was from the same university as I attended, I might have been using it since the mid 80's.

Anyway, if the patch could be a problem, just ignore it. There's no real bug.

nicm:
As it happens, the only upstream change is the same fix:
~~~~
--- a/compat/imsg-buffer.c
+++ b/compat/imsg-buffer.c
@@ -1,4 +1,4 @@
-/*     $OpenBSD: imsg-buffer.c,v 1.35 2025/06/04 09:06:56 claudio Exp $        */
+/*     $OpenBSD: imsg-buffer.c,v 1.36 2025/08/25 08:29:49 claudio Exp $        */

 /*
  * Copyright (c) 2023 Claudio Jeker <claudio@openbsd.org>
@@ -699,8 +699,6 @@ msgbuf_queuelen(struct msgbuf *msgbuf)
 void
 msgbuf_clear(struct msgbuf *msgbuf)
 {
-       struct ibuf     *buf;
-
        /* write side */
        ibufq_flush(&msgbuf->bufs);

nicm:
Fixed now, thanks!

PR #4786, opened 2026-01-03
Move the escape logic outside the loop in cmd_list_print

proski:
Readability improvement and marginal speedup. No visible changes in the behavior.

* Avoid nested `if`
* Give names to string constants for more readability
* Use one `strlcat` instead of four
* Check `escaped` only outside the loop


nicm:

PR #4788, opened 2026-01-04
Redraw borders after alternate screen goes on.

mmjo:
When the alternate screen is used, by e.g. less,
the pane-border is not (immediately) updated with that info.
I use (a.o.):
`set -ag pane-border-format '#{?alternate_on,⮺ }'`

The patch fixes that.
But maybe you have other ideas on this?

Regards, Mike

### DIFF
--- a/screen-write.c
+++ b/screen-write.c
@@ -2461,8 +2461,10 @@ screen_write_alternateoff(struct screen_write_ctx *ctx, struct grid_cell *gc,
 	screen_write_collect_flush(ctx, 0, __func__);
 	screen_alternate_off(ctx->s, gc, cursor);
 
-	if (wp != NULL)
+	if (wp != NULL) {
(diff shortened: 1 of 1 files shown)


nicm:
Thanks but what about `screen_write_alternateoff`?

mmjo:
Somehow screen_write_alternateoff properly updates the pane-border.
I checked it, but i can't tell why the difference is there.

nicm:
Can't see why that would be, but i would make the same change to both functions, it can't do any harm.

mmjo:
I can't trace it in the code.
Logically, when alternate goes on, the screen gets saved. Then the program draws on a new, empty screen.
When alternate goes off, a redraw of the saved screen takes place.

Anyway, i added the border redraw to screen_write_alternate_off also.

BTW, screen_write_alternate_(on,off) are almost equal.

nicm:

Issue #4777, opened 2025-12-30
tmux 3.6a crashes with memory allocation error when entering copy mode on macOS (Apple Silicon)


antonyr:
### Environment
- **tmux version**: 3.6a
- **Platform**: macOS 26.2 (25C56), Darwin 25.2.0
- **Hardware**: Mac16,7 (Apple Silicon M4 Pro)
- **Terminal**: Both Alacritty and Ghostty (terminal-agnostic)
- **Installation**: Homebrew (`/opt/homebrew/bin/tmux`)

### Description
tmux server crashes with a memory allocation error when entering copy mode in long-running sessions. The crash is 100% reproducible and terminal-agnostic.

### Error Message
```
malloc: *** error for object: pointer being freed was not allocated
*** BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED
```

### Stack Trace
```
cmd_copy_mode_exec
→ window_pane_set_mode
→ window_copy_init
→ window_copy_clone_screen
→ screen_reinit
→ grid_clear_lines
→ grid_free_line [CRASH: double-free]
```

### Reproduction Steps
1. Start tmux session
2. Let session run for extended period (8+ hours recommended, 8 days confirmed)
3. Enter copy mode (e.g., `Ctrl-b [`)
4. Server crashes with SIGABRT

### Memory Leak Evidence
Comparing two crashes shows memory accumulation over time:

**Crash 1** (8-day session):
- MALLOC: 277.4M (61 regions)
- VM errors: "VM Fault hit memory shortage"

**Crash 2** (25-hour session):
- MALLOC: 209.4M (44 regions)
- VM errors: "Fault was interrupted" (4×)

### Crash Reports

<details>
<summary>Crash Report 1 (Dec 23, 2025 - Alacritty)</summary>

```json
{
  "procLaunch": "2025-12-15 12:30:41.6461 +0530",
  "captureTime": "2025-12-23 17:04:07.2229 +0530",
  "responsibleProc": "alacritty",
  "exception": {"type":"EXC_CRASH","signal":"SIGABRT"},
  "asi": {"libsystem_c.dylib":["crashed on child side of fork pre-exec"]},
  "ktriageinfo": "VM - (arg = 0x0) VM Fault hit memory shortage\nVM - (arg = 0x0) Fault was interrupted",
  "vmSummary": "MALLOC: 277.4M (61 regions)",
  "frames": [
[...]

PR #4790, opened 2026-01-05
Fix double-free in grid line memory management

antonyr:
## Summary

Fixes a double-free crash that occurs after extended sessions (8+ hours) when entering copy mode.

- Add defensive NULL check in `grid_free_line` to prevent double-free
- Fix pointer aliasing in `grid_move_lines` by zeroing source positions within destination region after memmove
- Fix stale pointers in `grid_trim_history` by zeroing entries at end of array after memmove

## Root Cause

The crash occurs due to shallow copying of `grid_line` structures via `memmove`, which copies `celldata`/`extddata` pointers without clearing the source. This creates aliased pointers that accumulate over time, eventually causing a double-free when copy mode triggers `grid_clear_lines`.

## Test Plan

- [x] Build completes without warnings
- [ ] Run extended session test with heavy scrolling
- [ ] Enter copy mode repeatedly to verify no crash
- [ ] Test with AddressSanitizer: `make CC="cc -fsanitize=address"`

Fixes: https://github.com/tmux/tmux/issues/4777

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -361,9 +361,13 @@ grid_compare(struct grid *ga, struct grid *gb)
 static void
 grid_trim_history(struct grid *gd, u_int ny)
 {
+	u_int	remaining;
+
 	grid_free_lines(gd, 0, ny);
+	remaining = gd->hsize + gd->sy - ny;
 	memmove(&gd->linedata[0], &gd->linedata[ny],
-	    (gd->hsize + gd->sy - ny) * (sizeof *gd->linedata));
+	    remaining * (sizeof *gd->linedata));
+	memset(&gd->linedata[remaining], 0, ny * (sizeof *gd->linedata));
 }
 
 /*


nicm:
Thanks...

Are you sure all these changes are needed?

The `grid_trim_history` change makes sense.

The `grid_free_line` changes don't actually have any effect because `free(NULL)` is no-op?

I don't think the change to `grid_move_lines` can be right. It now does this:
~~~~
        memmove(&gd->linedata[dy], &gd->linedata[py], ny * (sizeof *gd->linedata));
...
	for (yy = py; yy < py + ny; yy++) {
		if (yy >= dy && yy < dy + ny)
			memset(&gd->linedata[yy], 0, sizeof gd->linedata[yy]);
....
~~~~
But if there are overlapping regions this will clear lines we have just moved.

For example, say we have py = 10, dy  = 8, ny = 5, the `memmove` will move 10-14 to 8-12, and then the loop will now walk from 10 to 15 and clear lines 10, 11 and 12, all of which we just moved, and we don't want to clear.

The intent of the loop is to clear those lines which are now "uncovered" by the move, so in this example lines 13 and 14 only.

antonyr:
Thanks for the review!

1. grid_free_line: Reverted - the NULL checks were indeed unnecessary since free(NULL) is a no-op.
2. grid_move_lines: Fixed the logic. I had it inverted - source positions within the destination range contain valid moved data and should be skipped, not zeroed. Only positions outside the destination range need clearing.
3. grid_trim_history: Kept as-is per your feedback.

nicm:
Thanks but now your change to `grid_move_lines` is equivalent to what it was before, so you could just leave it unchanged? (If you want to tidy it up I think it would be better to remove the `continue` from the previous loop rather than adding it to this one.)

antonyr:
Good point, thanks. I've simplified the change to only fix grid_trim_history - the memset to zero stale entries after memmove is the actual fix. Reverted the other changes.

nicm:
OK I have applied this to OpenBSD now, will be in GitHub later, let me know if you see any problems. Thanks!

antonyr:
Thanks for the review and amazing software, Nicholas!

PR #4795, opened 2026-01-08
Add mouse position validation for double and triple clicks

shaobosong:
Store initial click positions to ensure double and triple clicks occur at the same location, and reset the click state if the mouse is moved.

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -311,6 +311,8 @@ server_client_create(int fd)
 	evtimer_set(&c->repeat_timer, server_client_repeat_timer, c);
 	evtimer_set(&c->click_timer, server_client_click_timer, c);
 
+	c->click_wp = -1;
+
 	TAILQ_INIT(&c->input_requests);
 
 	TAILQ_INSERT_TAIL(&clients, c, entry);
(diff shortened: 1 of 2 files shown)


nicm:
Thanks.

What is the reason for this? That is, what problem does it solve?

As far as the change goes (assuming it is needed):

- Tolerance can't be in pixels because the mouse position is in cells,  not in pixels, so the comment is misleading.

- Perhaps we should make the tolerance a small value, say 2 or 5 or 10? People with small fonts or shaky hands may find it hard to keep the mouse on the same cell.

shaobosong:
@nicm 
Thanks for the review.

The motivation for this change is to avoid treating mouse clicks at different locations as double- or triple-clicks solely based on timing and button.

For example, if I left-click in one pane and then quickly left-click in a different pane, tmux may interpret the second click as a second-click rather than a new single click. In my configuration, this results in a word being selected in the second pane, even though the user action was simply two independent clicks at different panes.

By additionally checking the click position, the intent is to ensure that multiple clicks are only combined when they occur at the same (or nearly the same) cell, which better matches user expectations.

You are absolutely right about the comment — the position is in cells, not pixels. That wording is incorrect and should be fixed.

I also agree that a zero tolerance is likely too strict in practice. Allowing a small tolerance in cells makes sense, especially for small fonts or less precise devices. This change was also influenced by usage with some modern terminal emulators (for example, Windows Terminal), where mouse input tends to be more sensitive and small, unintentional position changes between clicks are more common. I’m happy to update this to use a small default value (for example 5). However, this alone does not fully address the case where rapid clicks in different panes are still interpreted as multi-clicks, so some additional logic would be needed to distinguish clicks **across panes and windows(may include sessions)** from true multi-click sequences.

nicm:
Maybe instead of position it should remember where the click happened (`enum mouse_where`) and the pane ID if on a pane?

shaobosong:
@nicm 
A new option named `click-position-tolerance` is available now.
Additionally, check `where` and `wp` for multiple clicks.

nicm:
If you are checking the click is still in the same pane or not (or still on status line, scrollbar, etc), what is the need to also check it is within some x/y offset? Doesn't checking the pane solve the problem of straying into a different one by accident?

shaobosong:
I've noticed that in `Windows Terminal`, multi-click position checking is a strict tolerance behavior. However, in `Alacritty`, it behaves like the current `tmux`, with no offset checks.
In `Neovim` on `Linux`, it mirrors the behavior of `Windows Terminal`, but in `Neovim` on `Windows`, it aligns with `Alacritty`'s behavior.
I'm unsure which behavior is more user-friendly, so I've left the choice up to the users.

nicm:
I don't think we should add an option that barely anyone will change. Either it should be pane/status/etc check only or if it includes the x,y it should have a builtin tolerance of some sensible number say 5.

shaobosong:
@nicm 
That makes sense—very few users actually change this option—so I’ve now removed the x/y offset tolerance checks entirely.

nicm:
Great, I have applied this to OpenBSD now, it should be in GitHub later. Thanks!

PR #4797, opened 2026-01-08
Support select text and osc52 copy in popup window

gogongxt:
Support select text and osc52 copy in popup window.

https://github.com/user-attachments/assets/380962d6-fe0c-4642-81d8-4bc3fe7c3b53


nicm:
This looks OK but it adds a lot of duplicate code to `input_osc_52`, maybe a helper function would be better rather than copying the code?

gogongxt:
I have pushed a new commit that refactor duplicate code. Please review the code again. Thanks.

gogongxt:
@nicm Please review again, thanks.

nicm:

Issue #4784, opened 2026-01-02
Feature Request: allow argument interpolation in run-shell


jsmrcaga:
### tl;dr
Feature request to add argument interpolation after run-shell, allowing dynamic commands when using command-aliases.

### Issue description

When running `run-shell`, the shell-command parameter is not interpolated with arguments given _after_ the command, but allows for variable interpolation (FORMATS part of the manpages).

For example
```sh
run-shell 'echo #{client_tty}'
```
WIll draw inside the current pane, and display `/dev/ttys008`

However,
```sh
run-shell 'echo %1' "test"
```

Will not interpolate, drawing inside the current pane and end up only displaying `%1`

When using command-aliases, this feature would allow custom commands to receive arguments at runtime. A simple example is a layout selector, where we execute a script that allows arguments to setup/select a layout.
For example:
```sh
tmux set-option -s command-alias[100] "l=run-shell '/my/tmux/plugins/layout.sh %%'"
```

Would allow users to run `:l 3`, `:l -p 5` etc to choose a layout.

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

>tmux version (`tmux -V`).

* tmux 3.5a 
 
> Platform (`uname -sp`).

* Darwin arm

> Terminal in use (xterm, rxvt, etc).

* Ghostty

>  $TERM *inside* tmux (`echo $TERM`).
 
* tmux-256color
 
> $TERM *outside* tmux (`echo $TERM`).

*xterm-ghostty

nicm:
I like this idea but I'm not sure how difficult it is to do.

jsmrcaga:
@nicm i tried looking in the code but i'm very rusty in C, so could not find the exact place where arguments are parsed.
Copilot gave me some pointers but ended up confusing me more so i dropped it.

I did not look at how `command-prompt` does it though

nicm:
Aliases are expanded at parse time so in `cmd-parse.y`, so first thing I would try is having `cmd_parse_expand_alias` walk the arguments and perform any `%n` substitutions using `cmd->arguments`. You can see near the end of `cmd_parse_expand_alias` it creates a new argument list by using `TAILQ_CONCAT` to append the old arguments (given to the alias) in `cmd->arguments` on to the arguments from the last command generated by expanding the alias (in `last`).

[...]

PR #4802, opened 2026-01-09
feat(run-shell): add argument interpolation to shell command execution

deechtejoao:
- Update usage message to reflect support for multiple command arguments

I think this is sufficient to cover everything that was requested in https://github.com/tmux/tmux/issues/4784

### DIFF
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -133,14 +133,16 @@ cmd_run_shell_exec(struct cmd *self, struct cmdq_item *item)
 	if (!args_has(args, 'C')) {
 		cmd = args_string(args, 0);
 		if (cmd != NULL) {
-			char *interpolated_cmd = xstrdup(cmd);
+			struct format_tree *ft = format_create_from_target(item);
+			char key[16];
+
+			format_set_flags(ft, FORMAT_ARGS);
 			for (u_int i = 1; i < args_count(args); i++) {
-				char *new_cmd = cmd_template_replace(interpolated_cmd, args_string(args, i), i);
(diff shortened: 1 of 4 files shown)


nicm:
Oh yeah, you're right, they did just ask for `run-shell`. I read it as wanting to set any alias to `x=some-command -a %2 %1` and then do `x Y Z` which would be different.

I think this would be more flexible if it added the arguments to the format used to expand the command. Then format modifiers could be used.

So instead of calling `format_single_from_target` to expand the command, it would create its own `format_tree` and add the arguments to that. I'd probably add them as `argument1` etc then update the alias table in `format.c` so that `#1` could be used instead (you could even make the format code accept a flag to look for `%1` etc as well if you like that syntax - that would also mean we could convert `command-prompt` to do this as well later).

ghost:
Thanks for the feedback, I'll work on it.

nicm:
This looks good, a few comments:

- Declarations need to go at the top of the function with the others, not inside an inner block. This means `ft`, `key`, `i` in `cmd_run_shell_exec` and `value`, `valuelen`, `key` in `format_expand1`.

- The lines inside the `if (*fmt == '%')` block in `format.c` are too long.  You could shorten them by making it `if (*fmt == '%' && (ft->flags & FORMAT_ARGS))`. Or maybe a helper function.

- `goto` to a label inside an `if` is not ideal. I would either just copy the lines instead of using `goto`, or move the label to the end of the loop after the final `break` and have both `#` and `%` `goto` it.

Thanks!

nicm:

PR #4804, opened 2026-01-10
run-shell: arguments on the right of command are passed to the command

### LINKED ISSUE #4784: Feature Request: allow argument interpolation in run-shell

jsmrcaga:
This commit aligns run-shell with other commands, like new-window, split-window etc...

It converts the argument list to a vector and then formats them one by one before assigning them to cdata->cmd.

Behaviour works as expected:

* `run-shell -C 'display-message "message"' `
    This will run a tmux command, and display a message in the status bar
* `run-shell "echo" "something"`
    This will display "something" in a new window
* `run-shell './script.sh' arg1 arg2`
    This will pass arg1 and arg2 to script.sh
* `run-shell "echo #{client_tty}" something`
    This will pass "something" to echo and display something like "/dev/pts/0 something"

This has also been tested with command aliases, which Fixes #4784.

> [!NOTE]
> This was mostly done with an LLM, i reviewed and tested the code manually

> [!NOTE]
> 2. I also just saw https://github.com/tmux/tmux/pull/4802, which seems simpler as it interpolates from the command.
> I have not tested their solution, but both should have the same result by doing something different. I'll leave this PR open so we can discuss. I can also close it if that's what's decided.

### DIFF
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -44,9 +44,9 @@ const struct cmd_entry cmd_run_shell_entry = {
 	.name = "run-shell",
 	.alias = "run",
 
-	.args = { "bd:Ct:Es:c:", 0, 1, cmd_run_shell_args_parse },
+	.args = { "bd:Ct:Es:c:", 0, -1, cmd_run_shell_args_parse },
 	.usage = "[-bCE] [-c start-directory] [-d delay] " CMD_TARGET_PANE_USAGE
-	         " [shell-command]",
+	         " [shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_PANE, CMD_FIND_CANFAIL },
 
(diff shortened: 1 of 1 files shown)


nicm:
Thanks. The same comments wrt formats apply as to #4802. I'm going to close this in favour of that one for the moment.

Issue #4761, opened 2025-12-18
Feature req: command-prompt option to escape on backspace and empty prompt line


vbauerster:
I would like to propose to have an option which will affect `command-prompt` escape behavior. This option should trigger quit `command-prompt` on <kbd>backspace</kbd> press if and only if command line prompt was empty. This may seems to be useless at first, but may be handy sometimes due to pressing <kbd>backspace</kbd> instead of <kbd>esc</kbd> is presumably more ergonomic.

Some other console apps like `fzf` and `kakoune` for example have such option. If you have `fzf` installed you can quickly try it in order to see what I mean:

```
fzf --bind backward-eof:abort
```

Implementation wise I'm ok with session wide option something like `set -g command-prompt-bspace-eof on`, but you better know.

nicm:
I think this should be a flag to `command-prompt` not an option.

nicm:
Added to todo list, thanks.

daneofmanythings:
This feature has been added as of [#4812](https://github.com/tmux/tmux/pull/4812)!

PR #4812, opened 2026-01-14
feat(command-prompt): Adding Backspace as a cancel point for an empty command-prompt with flag.

daneofmanythings:
Here is a patch to add the feature from issue [#4761](https://github.com/tmux/tmux/issues/4761). 

A new flag '-e' has been added to 'command-prompt' which will enable Backspace/CTRL-h to exit the command if the prompt is empty. There are a few things I took liberty with: the flag name, the internal bit flag name/bit number, and the description in the man page.
I would love feedback on the quality of my work, and how I can improve it to get it merged. Thank you for your consideration!

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -6793,7 +6793,7 @@ executes the command every time the prompt input changes instead of when the
 user exits the command prompt.
 .Fl e
 makes
-.Em Bspace
+.Em BSpace
 cancel an empty prompt.
 .Pp
 .Fl T


nicm:
Looks good, but `e` should come after `b` in the usage (I would fix `i` as well while there, so `-1beFiklN`), and tmux calls the key `BSpace` so in `tmux.1` it should be like this:
~~~~
makes
.Em BSpace
cancel an empty prompt.
~~~~
Or if you prefer just say `backspace` without the `.Em` or a capital letter.

daneofmanythings:
Thank you for the feedback! I have made the requested changes.

nicm:

You might want to add a note to the issue to say it has been added.

daneofmanythings:
Done!

PR #4813, opened 2026-01-15
feat(list-x): Adding -O flag to list-x and switch-client commands

daneofmanythings:
This is an implementation of issue [#4811](https://github.com/orgs/tmux/discussions/4811) to add the -O sorting flag to the commands `switch-client`, `list-windows`, `list-sessions`, and `list-panes`.

**This pr is currently a draft and not ready to merge.**
 I wanted to show the approach I'm taking to make sure it is acceptable before committing to it. I also have some questions about how a change to the `switch-client` internals could/should affect the `detach-on-destroy` option.

I have ported over the sorting machinery from `choose-session` into `list-sessions`. Is this an acceptable solution? If so, would an abstraction around sorting be desired as this would spray similar logic over 4+ files.

As for `switch-client`, my thought was to alter the api of `session_next_session` and `session_previous_session` to take a sort_criterion parameter (or other relevant context) and do something similar to what i plan on doing in the `list-x` commands. A side effect would be that `detach-on-destroy` would basically inherit the -O flag. If it is not desirable to change the `session-x` api then im guessing i would have to abandon those calls in `switch-client` to do sorting in cmd_switch_client_exec.

I look forward to feedback. Thanks!

### DIFF
--- a/sort.c
+++ b/sort.c
@@ -197,10 +197,11 @@ sort_pane_cmp(const void *a0, const void *b0)
 	case SORT_CREATION:
 		result = a->id - b->id;
 		break;
+	case SORT_SIZE:
+		result = a->sx * a->sy - b->sx * b->sy;
 	case SORT_INDEX:
 	case SORT_NAME:
 	case SORT_ORDER:
-	case SORT_SIZE:
 	case SORT_END:
 		break;
 	}
@@ -258,8 +259,10 @@ sort_winlink_cmp(const void *a0, const void *b0)
 	case SORT_NAME:
 		result = strcmp(wa->name, wb->name);
 		break;
-	case SORT_ORDER:
 	case SORT_SIZE:
+		result = wa->xpixel * wa->ypixel - wb->xpixel * wb->ypixel;
+		break;
+	case SORT_ORDER:
 	case SORT_END:
 		break;
 	}
--- a/sort.c
+++ b/sort.c
@@ -235,6 +235,16 @@ sort_winlink_cmp(const void *a0, const void *b0)
 	case SORT_INDEX:
 		result = wla->idx - wlb->idx;
 		break;
+	case SORT_CREATION:
+		if (timercmp(&wa->creation_time, &wb->creation_time, >)) {
+			result = -1;
+			break;
+		}
+		if (timercmp(&wa->creation_time, &wb->creation_time, <)) {
+			result = 1;
+			break;
+		}
+		break;
 	case SORT_ACTIVITY:
 		if (timercmp(&wa->activity_time, &wb->activity_time, >)) {
 			result = -1;
@@ -248,7 +258,6 @@ sort_winlink_cmp(const void *a0, const void *b0)
 	case SORT_NAME:
 		result = strcmp(wa->name, wb->name);
 		break;
-	case SORT_CREATION:
 	case SORT_ORDER:
 	case SORT_SIZE:
 	case SORT_END:
--- a/sort.c
+++ b/sort.c
@@ -200,19 +200,18 @@ sort_pane_cmp(const void *a0, const void *b0)
 	case SORT_SIZE:
 		result = a->sx * a->sy - b->sx * b->sy;
 	case SORT_INDEX:
+		window_pane_index(a, &ai);
+		window_pane_index(b, &bi);
+		result = ai - bi;
 	case SORT_NAME:
+		result = strcmp(a->screen->title, b->screen->title);
 	case SORT_ORDER:
 	case SORT_END:
 		break;
 	}
+
 	if (result == 0) {
-		/*
-		 * Panes don't have names, so use number order for any other
-		 * sort field.
-		 */
-		window_pane_index(a, &ai);
-		window_pane_index(b, &bi);
-		result = ai - bi;
+		result = strcmp(a->screen->title, b->screen->title);
 	}
 
 	if (sort_crit->reversed)
@@ -307,7 +306,8 @@ sort_order_from_string(const char* order)
 			return (SORT_CREATION);
 		if (strcasecmp(order, "index") == 0)
 			return (SORT_INDEX);
-		if (strcasecmp(order, "name") == 0)
+		if (strcasecmp(order, "name") == 0 ||
+		    strcasecmp(order, "title") == 0)
 			return (SORT_NAME);
 		if (strcasecmp(order, "order") == 0)
 			return (SORT_ORDER);
--- a/tmux.1
+++ b/tmux.1
@@ -3092,7 +3092,9 @@ See the
 section.
 .Fl O
 specifies the sort order: one of
-.Ql name ,
+.Ql title
+(screen title),
+.Ql index ,
 .Ql size
 (area),
 .Ql creation
--- a/tmux.1
+++ b/tmux.1
@@ -3093,6 +3093,8 @@ section.
 .Fl O
 specifies the sort order: one of
 .Ql name ,
+.Ql size
+(area),
 .Ql creation
 (time), or
 .Ql activity
@@ -3125,6 +3127,8 @@ section.
 specifies the sort order: one of
 .Ql index ,
 .Ql name ,
+.Ql size
+(area),
 .Ql creation
 (time),
 or
--- a/tmux.1
+++ b/tmux.1
@@ -3125,6 +3125,8 @@ section.
 specifies the sort order: one of
 .Ql index ,
 .Ql name ,
+.Ql creation
+(time),
 or
 .Ql activity
 (time).
--- a/tmux.h
+++ b/tmux.h
@@ -1285,6 +1285,7 @@ struct window {
 	struct event		 offset_timer;
 
 	struct timeval		 activity_time;
+	struct timeval		 creation_time;
 
 	struct window_pane	*active;
 	struct window_panes 	 last_panes;
--- a/window.c
+++ b/window.c
@@ -328,6 +328,9 @@ window_create(u_int sx, u_int sy, u_int xpixel, u_int ypixel)
 	RB_INSERT(windows, &windows, w);
 
 	window_set_fill_character(w);
+
+	if (gettimeofday(&w->creation_time, NULL) != 0)
+		fatal("gettimeofday failed");
 	window_update_activity(w);
 
 	log_debug("%s: @%u create %ux%u (%ux%u)", __func__, w->id, sx, sy,


nicm:
Thanks. I agree the sort code would be better as common code if possible. Maybe a new file `sort.c` would be best?

I'm not sure how an internal sort API would work I guess maybe something like:

`enum sort_order sort_order_from_string(const char *)` - convert eg "time" to `SORT_TIME`

`void sort_sessions(enum sort_order, void (*)(struct session *,  void *), void *)` - fire callback for each session in sorted order? seems like this would be good for `list-sessions`

`void get_sorted_sessions(struct session **, u_int *)` - return list of sorted sessions. IIRC this is what formats do (build a list)

And so on for windows and panes.

Or maybe something else? I'm not sure if both callback and list variants would be needed? I'm not sure this would be usable for modes which does sorting differently IIRC, although it could use the same enums.

daneofmanythings:
It makes sense to me to at least try to unite sorting into a usable api for everything that uses a -O flag.

Mode trees already construct a context (`mode_tree_sort_criteria`) which is carried to the point of sorting where it is then handed off to a static variable to be accessible from the comparator functions. My instinct is to model the api off of what already exists and see if it is useful. At the very least, `x_sort_list` and `x_sort_type` can be more universal like you mentioned above. I'll whip up a sort.c and update cmd_list_sessions.c to use it. I'll update when its committed!

daneofmanythings:
**EDIT 3:** I have worked this issue more and found that the implementation below will not cleanly translate to all areas that would use this api. I will make another comment when I have something that will work for all sorting. Should have something by tomorrow.

Okay, first pass is done.

**Goals:**
- Create a sort.c which can be used by commands with -O flags to centralize common sorting logic.
- Keep the footprint small as not to disturb commands that currently implement -O.

sort.c implements two functions, `sort_criteria_create` and `sort_list`. `struct sort_criteria` holds the context for a sort which can be built up and passed around until the sort is ready to happen at which point it is passed into `sort_list` and the sort is executed. `sort_criteria_create` only really needs the `order` string passed to it as the lookup logic for the sort order is internal.

**Advantages:**
- Barely impacts existing code and will remove redundancies (enums, lookup tables).
- Centralizes the core sorting logic into one place.

**Drawbacks/Considerations:**
- I am using global state for the `struct sort_criteria` that each commands comparator functions will reference. Possible race condition if commands are executed in paralllel.
- The centralization of logic adds boilerplate to each comparator in that the default case needs to be handled in case an order option that is not supported sneaks by.
- I am passing all `struct sort_criteria` by value. It is small enough that this is likely preferred.
- I am sorting the lists in place.

Another approach i considered is to move all the comparators into sort.c and create a bespoke wrapper for each command that needs sorting. I personally like the locality of keeping the comparators with their commands, but I can totally see preferring moving more logic into sort.c. That would keep the cmd files looking a lot cleaner.

Let me know what you think! Cheers~

**EDIT 1:** I have made a new commit showing the alternate approach discussed above. I realized that the alternate approach was probably closer to what you were envisioning. Just more options to chew on.

**EDIT 2:** The current commit now compiles and can sort list-sessions by name.

nicm:
This looks mostly OK but some comments:

- Global variable is OK because tmux is single threaded, and you have no choice because `qsort_r` is not portable. But it doesn't need to copy the struct, it could be a pointer?

- It is more efficient to avoid pass-by-value for structs in C. So `struct sort_criteria sc` to `sort_list_sessions` can be `struct sort_criteria *sc`.

- I would consider just dropping `xsort` and `xcompar` completely and just doing it inline in `sort_list_sessions` etc since it would only add a couple of lines, for example below. But if you want to keep them they should have `sort_` names such as `sort_do_qsort` or something. 
~~~~
void
sort_list_sessions(struct session **list, u_int len, struct sort_criteria *sc)
{
        if (sc->order != SORT_NONE && sc->order != SORT_INVALID) {
                sort_criteria = sc;
                qsort((void**)list, len, sizeof *list, sort_cmp_sessions);
        }
}
~~~~

- Likewise `session_list_cmp_session` should be `sort_cmp_session`.

- Probably `enum sort_order` should be in `tmux.h` instead of hiding it behind `int`.

- Can you make the code style closer to tmux's? At the least, please use width-8 tabs instead of spaces for indentation.

- As well as `sort_list_sessions`, maybe it would be nice to have `sort_all_sessions` which creates and returns a sorted list of all sessions?

- The `log_debug` in `session_list_cmp_session` can be `fatalx`.

daneofmanythings:
Thank you for the feedback! I believe I have addressed all of it in this current commit. Current status has been pushed, note that there are still bugs. The two i know of are:
1. The sort order is not displaying properly upon opening a mode tree, It will display properly if 'O' is used to switch to the next sort ordering.
2. Switch-client has only just been implemented. It is failing under deterministic circumstances.
There's probably a lot more, I've just started looking at the files in ./regress to figure out how to write a testing script. If you have any idea why 1 is happening, let me know.

Some notable things:
- I have added a minimal api surface area to accommodate what functionality is needed, though I believe what is here should scale as any additional context can be plopped into `struct sort_criteria`.
- `sort_criteria_init` is a helper function. A `sort_criteria` can be assembled by hand, but half the time all the context is known early and can be initialized in the same place.
- I have decoupled the string arrays and the enum arrays. Previously the ordering of these arrays matched and that was used to convert between enum and string. Now that there is a single enum for all sort orders in a different file from the strings, the coupling seemed brittle to me. I've replaced it with two functions that do the conversions.
- At least one of the commands used its enum array to determine order when cycling through different sorting options. I have moved this logic to an optional `struct sort_ordering` which will house that logic for consumers that need this functionality.
- The sorting comparison logic was being used in multiple files for sessions, so I moved it to sessions.c. I also put it a function that returns a list of all sessions. I had written the same loop multiple tiles so i consolidated it. The forward declaration for this function in tmux.h is not properly aligned as it would shift like 20 lines over by 1 space and I'd like confirmation that it will stay before aligning it.
- `fatalx` seems like a drastic option for an unsupported sort order. The behavior prior to this api was to fail silently to a backup ordering/not sort the list. 
- I intend to be maximally compliant with all conventions of the project (formatting, coding, ect.) and had poorly formatted code as I did not understand the convention. I read style(9) and configured my editor, so my code should be correctly formatted going forward, but let me know if something slips past me!

nicm:
OK this looks on the right lines. A few more comments:

- I think this API is too open - there is too much public state. I think:
   - ~`sort_criteria` should be private to `sort.c` and callers must use `sort_criteria_init`. If we need a couple of different init functions for different purposes, or the caller has to call `sort_order_from_string` themselves, so be it.~ Edit: actually forget this because then it couldn't be on the stack.
   - There should be some specific functions for the use cases we have. If the caller wants a list of sorted sessions, there should be a function to do it, it should not need to call `sort_criteria_init` then `session_list_all` then `sort_run`, it should call `sort_get_sessions(order, reverse)`. That function can use the more specific API internally.
   - `session_sort_cmp` should be in `sort.c` rather than `session.c` (and named something like `sort_cmp_sessions`) and `session_default_sort_criteria`. (`sort_session_default_sort_criteria`). That is, it should be classes as "sorting" not as "sessions". `session.c` is already fairly big.  This also means `sort_criteria` in `sort.c.` can be static.

- You can add a licence header to `sort.c` but update the copyright line with your name and the year.

- Calling `fatalx` in cases where something should not happen is the right thing to do, think of it like an assert. We usually want to stop running in these cases, not carry on with a known bug. So the `log_debug` in `session_sort_cmp` in `session.c` should be a `fatalx`.

-  For `sort_order_to_string` you could `fatalx` for unknown cases, or just return "unknown" if you prefer.

- An active name like `sort_next_order` would be better than `sort_ordering_next`?

 - You could maybe avoid the need for `window_client_sort_ordering` if you use `SORT_INVALID` as an end marker or  added `SORT_END` or something? Or keep it if you prefer but I'd maybe call it `sort_orders` or `sort_order_list` since `sort_ordering` is not very clear to me.

nicm:
If you prefer instead of me looking at various draft versions, maybe you want to get it all working and tidied up the way you think it should look and then I can review the whole thing? I don't mind either way :-).

daneofmanythings:
I really appreciate how attentive you have been with this PR. It has been very helpful for me, and I think I now have enough to finished this feature. I'll update when it's ready for review, thanks again!

daneofmanythings:
Okay. This is now ready for review. I was unsure how I should implement the sorting for list-windows and list-panes given that they each have flags for expanded scopes, so i did a global sort on windows and a basic local sort of panes. Doing the global sort on panes doesn't seem trivial given that panes don't retain information about the sessions/winlinks that link to them, so the formatting information would be missing. Is it okay as is, and should i revert list-windows to the local sorting?

**EDIT 1**: Oh one more thing, I made an assumption that all pointers on supported platforms have word width in `sort_safe` for the `SORT_ORDER` branch.

daneofmanythings:
I cleaned up a few typos i found and squashed it.

nicm:
Great stuff. I have applied this to OpenBSD with some mostly minor style changes and code reordering.

I renamed `sort_safe` to `sort_qsort` because I think that is clearer, and I fixed a couple of small bugs:

- `w` is never initialized to `wl->window` in `sort_get_panes` so it will crash.

- The compare functions cannot in fact fatal, I think I misled you here - I thought this was a "should not happen" condition but in fact it can easily happen, for example `tmux lsw -aOcreation`.

This should be in GitHub as soon as it syncs up.

If you'd like to look at something else related - maybe it would be nice if all the objects had as many sort criteria as possible, for example there is no reason for windows to have an activity time but not a creation time; for panes name could be a proxy for title (I would probably allow both "title" and "name" for `-O`); and if clients can have a size why shouldn't panes and windows?

daneofmanythings:
Awesome, thanks! I will look through and see what I can add to be sorted by.

(I was rebasing so that the diffs for the new sort orders would have clean patches and accidentally overwrote the progress up until this point, so I'm leaving the [merged commit](https://github.com/tmux/tmux/commit/3c3d9ce3a93fc827f4f4eaff6cfc3317ed548413) here for my own book keeping)

daneofmanythings:
I have added some of the low hanging fruit for new sorting options. I do have a few questions about what the desired  outcome is. 

First, there are many things to add sort options for, but I am unsure how useful or desired they would be and could clutter the code. A good example is sessions, with the additional sorting options being `cwd`, `curw`, and possibly `flags`. Panes would be able to reuse `curw`, but it fees like noise to me. 

Second, you mentioned naming for panes. I made title an alias for name whilst taking the title from the pane's screen. Instead, should i add machinery for panes to have specifiable names and a command to set a pane's name?

nicm:
I don't think we need any new sort orders, just to see if the existing ones can apply where they do not now. No panes do not need names so there is no need to add them.

nicm:
Looks good, applied to OpenBSD with a couple of changes:

- `xpixel`/`ypixel` are the cell pixel sizes and may be zero, windows should be using `sx`/`sy` instead.

- You were missing some `break`s in `sort_pane_cmp`.

Should be in GitHub later, thanks!

PR #4818, opened 2026-01-19
Fix `window-size=latest` not resizing on `switch-client` in session groups

ilyagr:
When using session groups with window-size=latest (the default),
switching to a non-current window with `switch-client` from a larger
terminal did not resize the window immediately. The resize only happened
later when switching windows again.

Switching windows with `select-window` was not affected by this bug, but we use `switch-client` when the user clicks the mouse on the status bar to switch windows since 13707917.

See commit messages for more details.

The actual debugging was done by AI. If the fix is not correct, this could be treated as a bug report instead of a PR. (The provided test should serve as a reproduction; there's also [an asciicast demo](https://asciinema.org/a/ZvVD4SmbphmsvozW); ignore the weird mouse selection in the beginning, I think that's a [Ghostty bug](https://github.com/ghostty-org/ghostty/discussions/9312) ).

### DIFF
--- a/regress/session-group-resize.sh
+++ b/regress/session-group-resize.sh
@@ -3,6 +3,8 @@
 # Test that window-size=latest resizes windows correctly when switching
 # windows in session groups. When a client switches to a window, it should
 # resize immediately to match that client's size.
+#
+# Tests both switch-client and select-window, which use different code paths.
 
 PATH=/bin:/usr/bin
 TERM=screen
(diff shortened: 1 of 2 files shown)


nicm:
I've applied the fix to OpenBSD, will be in GitHub later (the sync seems to be stuck so whenever it is unstuck).

Will look at the test later.

Thanks!

nicm:
I added the test also. Thanks!

PR #4819, opened 2026-01-20
Overlay to visible ranges

mgrant0:
1. change overlay ranges to use visual ranges
2. add a debug function

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -181,15 +181,17 @@ menu_mode_cb(__unused struct client *c, void *data, u_int *cx, u_int *cy)
 }
 
 /* Return parts of the input range which are not obstructed by the menu. */
-void
+struct visible_ranges *
 menu_check_cb(__unused struct client *c, void *data, u_int px, u_int py,
-    u_int nx, struct overlay_ranges *r)
+    u_int nx)
 {
 	struct menu_data	*md = data;
 	struct menu		*menu = md->menu;
+	struct visible_ranges	*r;
 
-	server_client_overlay_range(md->px, md->py, menu->width + 4,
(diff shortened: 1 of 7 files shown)


mgrant0:
image


```
                  /* xxxx Breaking up the tty_draw_line like this isn't fully working. */
                  vr = tty_check_overlay_range(tty, x, y, width);

                  tty_default_colours(&defaults, wp);

                  for (r=0; r < vr->used; r++) {
                          if (vr->nx[r] == 0)
                                  continue;
                          /* Convert window coordinates to tty coordinates. */
                          px = vr->px[r];
                          /* i is px of cell, add px of region, sub the
                           * pane offset. If you don't sub offset,
                           * contents of pane shifted. note: i apparently unnec.
                           */
                          tty_draw_line(tty, s, /* i + */ vr->px[r] - wp->xoff, j,
                              vr->nx[r], px, y, &defaults, palette);
                  }
```

I tried to break up the call to tty_draw_line around the overlay but it's buggy.  When you right-click on the right of an overlay, it leaves an artifact.  I haven't been able to track this down yet.  Possibly related to calling or not calling the menu popup call to include its visual range somewhere.

mgrant0:
Ok after some tracing, i think i see what's going on and I think this issue may go away once we remove the other call to tty_check_overlay_range inside tty_draw_line.

mgrant0:
BTW, the idea of doing check of the visual range outside of tty_draw_line is so you only need to worry about the edge of the line being drawn (the character that might overflow the width).  This may now be easy to fix.

All calls to tty_check_overlay_range need to be done only once outside or it messes up the static array inside it.  This appears what is causing the above display bug.

PR #4822, opened 2026-01-21
fix slow escapetime on startup

jake-stewart:
tmux queries some stuff from the terminal and meanwhile `escape-time` is essentially replaced with 500ms. I find this annoying when opening a tmux popup running tmux for persistence and I hit escape in my prompt to edit vi normal mode.

```
# tmux.conf:
set -g escape-time 1

# start tmux
tmux -f tmux.conf new

# start nested tmux
tmux display-popup -E tmux -f tmux.conf new cat

# hit escape: noticeable delay
```

My proposed solution is to avoid applying the 500ms wait when escape is the only key pressed and instead apply a much lower 100ms (or escape-time if it is higher).

I am making a guess that the slow part of the response is silence and then once escape is sent it should not take very long for the rest of the terminal response to come through.

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -959,8 +959,8 @@ partial_key:
 	    (tty->flags & TTY_ALL_REQUEST_FLAGS) != TTY_ALL_REQUEST_FLAGS) ||
 	    !TAILQ_EMPTY(&c->input_requests)) {
 		log_debug("%s: increasing delay for active query", c->name);
-		if (delay < 500)
-			delay = 500;
+		int min_delay = len == 1 && *buf == '\033' ? 100 : 500;
+		delay = MAX(min_delay, delay);
 	}
 	tv.tv_sec = delay / 1000;
 	tv.tv_usec = (delay % 1000) * 1000L;


jake-stewart:
The 100ms I chose was arbitrary. It is far less annoying than the 500ms and easy enough to ignore. Ideally it would be escape-time but not sure.

nicm:
Hmm... I'm not sure this is a safe assumption to make.

nicm:
Does Windows Terminal respond to XDA (`\033[>q`)? It is usually the main problem terminal (not perhaps its fault). If it does, maybe we can use 500 for WT and 100 for others?

The other possibility would be to agitate for terminals to add [mintty's alternative escape sequence](https://github.com/mintty/mintty/wiki/CtrlSeqs#escape-keycode) which solves the problem completely but nobody has really picked up... if we knew a terminal supported it we could skip this special case entirely. I wouldn't hold my breath for anyone else to adopt it however...

ThomasDickey:
..more widely known as [XTVERSION](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h4-Functions-using-CSI-_-ordered-by-the-final-character-lparen-s-rparen:CSI-gt-Ps-q:Ps-=-0.1F0B).

In a quick check (Windows 11, Windows Terminal 1.23.13503.0), it doesn't respond.  The ones that I identified last year are noted in ncurses in references to [report+version](https://invisible-island.net/ncurses/terminfo.src.html#tic-report_version)

nicm:
OK thanks, so that idea is out... I'll think about it.

jake-stewart:
It is worth emphasising that this only happens for me with a popup running tmux inside tmux.

If I open a new terminal emulator window and start tmux, there is no escape delay. Is tmux not replying to some queries or something?

nicm:
Yeah popups are missing a lot of the reply stuff so that's probably why you always see the delay. This will be fixed when floating panes are done.

craigbarnes:
> The other possibility would be to agitate for terminals to add [mintty's alternative escape sequence](https://github.com/mintty/mintty/wiki/CtrlSeqs#escape-keycode) which solves the problem completely but nobody has really picked up... if we knew a terminal supported it we could skip this special case entirely. I wouldn't hold my breath for anyone else to adopt it however...

There was a similar [feature] in foot for a while, but it was [removed] in favor of just using the Kitty keyboard protocol (which makes <kbd>Esc</kbd> unambiguous, by virtue of encoding it as `CSI 27 u`).

I've been running [this] (somewhat related) dte commit for a while and it's been working fine.

[feature]: https://codeberg.org/dnkl/foot/commit/31c73f0cf01e1d0bdcd1435136b010c1c1ab442c
[removed]: https://codeberg.org/dnkl/foot/commit/e32707ffc04a5c4acee62680f080c927d96e1836
[this]: https://github.com/craigbarnes/dte/commit/f4b88e01f6e56735cb10626312b907e627b37d2e

nicm:
I'm going to close this because you should now use floating panes instead of popups.

PR #4825, opened 2026-01-21
Set PANE_STYLECHANGED when user options change

joshua-cooper:
When a user variable is modified, set PANE_STYLECHANGED on all panes so
that window-style and window-active-style are re-evaluated. This fixes
hot-reloading of window styles that reference user variables via format
strings.


nicm:

PR #4827, opened 2026-01-21
Fix theme hooks not firing when popup is open

joshua-cooper:


nicm:
Applied to OpenBSD now, will be in GitHub later, although the sync seems to be down so not sure when. Thanks!

PR #4828, opened 2026-01-21
Re-evaluate popup styles on each draw

joshua-cooper:
Popup styles were only applied once at creation time. When user variables  changed, the popup would not update its appearance.

### DIFF
--- a/popup.c
+++ b/popup.c
@@ -32,6 +32,8 @@ struct popup_data {
 	int			  flags;
 	char			 *title;
 
+	char			 *style;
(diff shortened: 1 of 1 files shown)


joshua-cooper:
I've been testing this via theme change hooks which requires this fix to be applied: https://github.com/tmux/tmux/pull/4827

nicm:
Applied to OpenBSD with a couple of trivial changes (line lengths), will be in GitHub later. Thanks!

PR #4829, opened 2026-01-21
Re-evaluate menu styles on each draw

joshua-cooper:
Menu styles (menu-style, menu-selected-style, menu-border-style) were not updating when user variables changed because styles were computed once at menu creation.

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -27,9 +27,13 @@ struct menu_data {
 	struct cmdq_item	*item;
 	int			 flags;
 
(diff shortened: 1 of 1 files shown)


joshua-cooper:
I've been testing this via theme change hooks which requires this fix to be applied: https://github.com/tmux/tmux/pull/4827

nicm:
Applied to OpenBSD with a couple of trivial changes (line lengths), will be in GitHub later. Thanks!

PR #4830, opened 2026-01-21
Re-evaluate copy-mode styles when user options change

joshua-cooper:
Copy mode styles (copy-mode-selection-style, etc.) were not updating when a user variable they referenced changed because the mode screen content was pre-rendered and not refreshed on style changes.

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -2684,13 +2684,25 @@ server_client_handle_key(struct client *c, struct key_event *event)
 void
 server_client_loop(void)
 {
-	struct client		*c;
-	struct window		*w;
-	struct window_pane	*wp;
+	struct client			*c;
+	struct window			*w;
+	struct window_pane		*wp;
+	struct window_mode_entry	*wme;
 
 	/* Check for window resize. This is done before redrawing. */
 	RB_FOREACH(w, windows, &windows)
 		server_client_check_window_resize(w);
+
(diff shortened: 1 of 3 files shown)


joshua-cooper:
This depends on https://github.com/tmux/tmux/pull/4825

nicm:
I don't think the check should be in `screen-redraw.c` but instead in `server-client.c` as part of the pane loop in `server_client_loop`?

nicm:
Looks good but any reason this can't be rolled into the "Send theme updates" loop rather than looping over the windows again separately?

I'm actually not sure if the theme updates loop could be part of the previous one as well...

joshua-cooper:
> Looks good but any reason this can't be rolled into the "Send theme updates" loop rather than looping over the windows again separately?

I don't think it can because it needs to run before PANE_STYLECHANGED gets cleared during the redraw. The theme updates loop runs after the redraw, so is too late.

It can appear to work when the redraw is deferred (e.g. due to a busy tty output buffer), but with a config like this, you won't see the changes until you leave and re-enter copy mode:

```
set -g @sel "red"
set -g copy-mode-selection-style "bg=#{@sel}"
bind t set -g @sel "blue"
bind T set -g @sel "red"
```

nicm:
OK makes sense.

nicm:

PR #4832, opened 2026-01-27
add left right mouse wheel 

john-peterson:
add left right mouse wheel 

emacs is a good mouse test because it will show mouse actions in status bar  

and log will show if wheel is working 

~~~
tmux -vv
ack writing tmux-server*log|ack wheel -i
~~~

# for apps with no wheels

I use it in tmux-better-mouse-mode to simulate mouse wheel with key strokes

I use a touch screen terminal 

I needed it for graphics that don't support wrap in pagers that don't support eheels

for instance weather on my phone small tall screen need horizontal scroll to see all

get weather

~~~
f=weather
curl  v2.wttr.in/$gps?T > $f
test -s $f || exit
curl  wttr.in/$gps?format=j1  > $f 
~~~

~~~
unset LESS
cat weather|less -S -#2
cat weather|les -w
~~~

here is an app that already support mouse wheels and don't need this 

~~~
cat weather|vim -MR - -c "set nowrap"
~~~

# check list

apps that already has it

emacs
vim

my patches 

termux
radare2 -c "VV"
tmux
tmux-better-mouse-mode

todo apps that don't have left right 

less --mouse

or any mouse wheel at all

les


nicm:
Looks fine but you will need to change tmux.1 as well.

nicm:

PR #4841, opened 2026-02-04
Also check PANE_STATUSREADY for pane_dead format to match pane_dead_status.

jpollack:
A pane is dead when its fd is closed (fd == -1) and its exit status is available
(PANE_STATUSREADY). This avoids a race where the fd closes before the SIGCHLD
handler has retrieved the exit status via waitpid(), which would cause pane_dead
to be 1 while pane_dead_status is still empty.


nicm:

PR #4845, opened 2026-02-06
feat(list-keys): Added F, O, and r flags to list-keys.

daneofmanythings:
This is a draft of my idea for adding the formatting flag to list-keys (sorting was a byproduct). I primarily use fzf as my discoverability tool so being able to format my delimiters in list keys will help me view them in fzf, similar to commands like list-sessions, and list-commands.

I did a few things, I refactored the code to use the internal formatting mechanism instead of doing it by hand, incorporated sort.c, and moved list-commands to its own file as I could not see a reason to have two commands occupy one file when there was no shared code.

I believe there are no regressions. Let me know what you think!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -106,6 +106,7 @@ dist_tmux_SOURCES = \
 	cmd-kill-window.c \
 	cmd-list-buffers.c \
 	cmd-list-clients.c \
+	cmd-list-commands.c \
 	cmd-list-keys.c \
 	cmd-list-panes.c \
 	cmd-list-sessions.c \
--- a/cmd-list-commands.c
+++ b/cmd-list-commands.c
@@ -1,0 +1,107 @@
+/* $OpenBSD$ */
+
+/*
+ * Copyright (c) 2007 Nicholas Marriott <nicholas.marriott@gmail.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF MIND, USE, DATA OR PROFITS, WHETHER
+ * IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
+ * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <sys/types.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "tmux.h"
+
+/*
+ * List all commands.
+ */
+
+#define LIST_COMMANDS_TEMPLATE					\
+	"#{command_list_name}"					\
+	"#{?command_list_alias, (#{command_list_alias}),} "	\
+	"#{command_list_usage}"
+
+static enum cmd_retval cmd_list_commands(struct cmd *, struct cmdq_item *);
+
+const struct cmd_entry cmd_list_commands_entry = {
+	.name = "list-commands",
+	.alias = "lscm",
+
+	.args = { "F:", 0, 1, NULL },
+	.usage = "[-F format] [command]",
+
+	.flags = CMD_STARTSERVER|CMD_AFTERHOOK,
+	.exec = cmd_list_commands
+};
+
+static void
+cmd_list_single_command(const struct cmd_entry *entry, struct format_tree *ft,
+    const char *template, struct cmdq_item *item)
+{
+	const char	*s;
+	char		*line;
+
+	format_add(ft, "command_list_name", "%s", entry->name);
+	if (entry->alias != NULL)
+		s = entry->alias;
+	else
+		s = "";
+	format_add(ft, "command_list_alias", "%s", s);
+	if (entry->usage != NULL)
+		s = entry->usage;
+	else
+		s = "";
+	format_add(ft, "command_list_usage", "%s", s);
+
+	line = format_expand(ft, template);
+	if (*line != '\0')
+		cmdq_print(item, "%s", line);
+	free(line);
+}
+
+static enum cmd_retval
+cmd_list_commands(struct cmd *self, struct cmdq_item *item)
+{
+	struct args		 *args = cmd_get_args(self);
+	const struct cmd_entry	**entryp;
+	const struct cmd_entry	 *entry;
+	struct format_tree	 *ft;
+	const char		 *template,  *command;
+	char			 *cause;
+
+	if ((template = args_get(args, 'F')) == NULL)
+		template = LIST_COMMANDS_TEMPLATE;
+
+	ft = format_create(cmdq_get_client(item), item, FORMAT_NONE, 0);
+	format_defaults(ft, NULL, NULL, NULL, NULL);
+
+	command = args_string(args, 0);
+	if (command == NULL) {
+		for (entryp = cmd_table; *entryp != NULL; entryp++)
+			cmd_list_single_command(*entryp, ft, template, item);
+	} else {
+		entry = cmd_find(command, &cause);
+		if (entry != NULL)
+			cmd_list_single_command(entry, ft, template, item);
+		else {
+			cmdq_error(item, "%s", cause);
+			free(cause);
+			format_free(ft);
+			return (CMD_RETURN_ERROR);
+		}
+	}
+
+	format_free(ft);
+	return (CMD_RETURN_NORMAL);
+}
--- a/cmd-list-keys.c
+++ b/cmd-list-keys.c
@@ -27,117 +27,41 @@
  * List key bindings.
  */
 
-static enum cmd_retval	cmd_list_keys_exec(struct cmd *, struct cmdq_item *);
+#define LIST_KEYS_TEMPLATE					\
+	"#{?notes_only,"					\
+	"#{key_prefix} "					\
+	"#{p|#{key_string_width}:key_string} "			\
+	"#{?key_note,#{key_note},#{key_command}}"		\
+	","							\
+	"bind-key #{?key_has_repeat,#{?key_repeat,-r,  },} "	\
+	"-T #{p|#{key_table_width}:key_table} "			\
+	"#{p|#{key_string_width}:key_string} "			\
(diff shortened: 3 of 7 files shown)


nicm:
This looks OK apart from style nits. The reason they are in the same file is because `list-commands` was tiny and there was no point in a file on its own, but it is OK to break it out now.

daneofmanythings:
I have cleaned up the implementation. The only behavior that differs is when using '-1' on a plain `lsk` command, even though the iteration logic is identical. I am unsure why this is happening, or how important that behavior is. '-1' Appears to be working as expected otherwise.

*EDIT 1*: I expect there to be some fixable formatting/styling. I'm still acclimating to the code style.

nicm:
OK this looks pretty good, a few minor comments and one major:

- `line` in `cmd_list_keys_exec` is `const char *` but should be `char *`.

- You need to handle ` SORT_MODIFIER` in all the `switch` in `sort.c` (clang is showing a warning).

- I think `cmd_list_keys_contains_repeatable_key_binding` would be better in `key-bindings.c` and called something like `key_binding_has_repeat`.

- There are a couple of things I don't like about the formats:
  - I don't like rewriting the format with `snprintf`. I think we should change `p` so that the argument can be a format and then pass the width in as part of the format (so you do like `p|#{max_key_width}:key_tablename`).
   - Instead of having `REPEAT_FMT`, `LIST_KEYS_TEMPLATE` and `LIST_KEYS_N_FLAG_TEMPLATE` it would be better to have one template and control it with the format, so do something like:
    ~~~~
        ft = format_create(cmdq_get_client(item), item, FORMAT_NONE, 0);
        format_defaults(ft, NULL, NULL, NULL, NULL);
        format_add(ft, "notes_only", "%d", args_has(args, 'N'));
        format_add(ft, "has_repeat", "%d", repeat);
        format_add(ft, "max_key_width", "%u", keywidth);
    ~~~~
    And then inside the format you can use those variables to pick the output. This means that anyone using `-F` can do the same, or pass in the same format as we are using with whatever changes they want. So your format would be something like:
    ~~~~
    #{?notes_only,#{key_prefix} #{p|#{max_key_width}}:key_string} #{?key_note,#{key_note},#{key_command}},bind-key #{?key_repeat,-r,  } -T #{p|#{max_table_width}:key_tablename} #{p|#{max_key_width}:key_string} #{key_command}}
    ~~~~
    If you like, you could probably roll the *whole* function of the `-N` and `-a` into this, so you could make it so that if the format expanded to an empty string, the line was omitted, then use that to skip lines without notes instead of doing it manually. This might be more tricky to do. 

- `cmd_skip_for_N_flag` would be better called something like `cmd_list_keys_should_skip`? Especially since it is also concerned with the `-T` flag.

daneofmanythings:
Thank you for pointing me in the right direction! I took care of the minor concerns and made big adjustments inline with your major concern regarding templating.

~~I did not like that after getting a list back from sort.c, there was additional logic and filtering that needed to be done, so I moved that filtering into the sort api. Now, the list you get back has exactly what is needed to display. I feel this simplifies the logic significantly overall with the drawback that `sort_get_key_bindings_table` got slightly more complex. This got rid of that leaky N flag checking function as well. For a raw N flag, I now squash together two calls to `sort_get_key_bindings_table` (the root and prefix tables) instead of some janky if/then checks. This expands the memory footprint of the runtime slightly, but again I feel this is much more clear.~~
I have moved all of the logic relevant to `list-keys` back into `cmd-list-keys.c` where it belongs. My decision was nagging on me. See the most recent commit.

The template has been unified to basically what you provided me above. I have added what I feel are the relevant format keys to the documentation so now users have access to the pretty formatting options.

I have squashed everything into a single commit. Let me know what you think!

**EDIT** Simplified part of the implementation and fixed a bug.

daneofmanythings:
I have fixed the conflicts!

***EDIT***: I goofed on the commit message :)

nicm:
OK I'll take another look, thanks.

nicm:
Looks good, applied now with a couple of tidy ups and one minor fix - if you do this:
~~~~
	format_add(ft, "key_table", "%s", xstrdup(bd->tablename));
~~~~
You are causing a memory leak because `format_add` copies the string (via `xvasprintf`) and the original pointer is lost. So just do:
~~~~
	format_add(ft, "key_table", "%s", bd->tablename);
~~~~
Should be in GitHub later. Thanks!

nicm:
Oh, one thing I noticed is that invalid sort formats do not show an error:
~~~~
$ tmux lsk -F '' -O foo
$
~~~~
It might be better to make this invalid? It's less confusing to catch typos immediately.

daneofmanythings:
From what I have found, the list commands that accept formats will print empty lines for each entry when given an invalid format. This is the line that is creating the divergent behavior in `list-keys`:
```c
			if (*line != '\0')
				cmdq_print(item, "%s", line);

```
Removing that if check will print blanks like the other list commands, and should probably be removed for consistency so that the command will print the blank lines. Do you find this to be sufficient? I agree that it should do more than nothing on a format failure.

daneofmanythings:
Something I could look into implementing is printing detailed error output in the place of the formatted line if there is a format lookup failure. Could be nice to help people debug formats in the case of typos or incorrect format keys.

nicm:
Sorry, I did not mean formats, I meant invalid sort *order*, so `-O` not `-F`.

`display -v` is meant for debugging formats.

daneofmanythings:
Ah okay my bad, there was a feeling in my bones that I misunderstood something. It's probably wise to remove that line so behavior is consistent across the list commands. I should have caught that before.

In regards to the invalid sort orders, I do feel that there should be a signal to the user that they tried to sort by an invalid order. We discussed this a bit when I was working on the `sort.c` pr. As it stands, there is a default order, usually `name` that sorting uses, but I don't like silent failures. I can open up a new pr to discuss options on how to give the api more feedback. I'll have to look more closely at what can be done, but I'd it rather spit out an empty list than hide a failure.

nicm:
It wasn't you, I wrote format but I meant to write order.

I think an invalid -O should report an error from the list commands. Not sure about choose, I guess they should be the same. I don't remember if there was a reason why not?

daneofmanythings:
I think there was some miscommunication between us in regards to how to handle invalid orders. I initially had it as a log output (that was a big guess on my part), Then we changed it to a fatalx. And in the final version, all reporting was stripped. 
I think the easiest way to accomplish passing back state to the caller is through either an additional out parameter or through an added result field in `struct sort_criteria`. I'd prefer the field approach as it keeps signatures the same, even though it muddles the structs meaning a bit. The passing of an invalid order to the sort api can be tracked by a static variable in `sort.c` and then can be referenced in each of the `sort_get_*` functions and set before returning. The caller can decide whether or not to do anything about it. How does that sound?

***EDIT1***: If the muddling of `struct sort_criteria`'s meaning is a problem, could a rename to `struct sort_context` be part of the solution? I'd be happy to make the  changes across the codebase if so.

daneofmanythings:
Oh, I should note that the default behavior before `sort.c` was to fail silently with a backup ordering, so there wasn't any regression, but I like the idea of giving the errors to the user.

nicm:
Won't you just do something like:
~~~~
sort_crit.order = sort_order_from_string(args_get(args, 'O');
if (sort_crit.order == SORT_END) {
    cmdq_error(item, "invalid sort order");
    return (CMD_RETURN_ERROR);
}

daneofmanythings:
Right, yes. That will error out when a typo or otherwise invalid order is given. The solution I was outlining would cover the additional case where a valid order is given, but that order is not supported/implemented by the `sort_*_cmp` that is called. For example: `SORT_ACTIVITY` in `sort_buffer_cmp`. The check for failure would then happen after getting the list of items from a `sort_get_*` and checking the added result enum field in the `sort_crit`. I'm realizing that this solution may be overkill and unnecessary. How do you feel about this?

For either method, I can add the checks to the list/choose commands.

nicm:
I had just thought of doing the former but if you want to do the latter as well, wouldn't it be simplest just to check the return value of `sort_order_from_string` against a list of the accepted values for each command? So something like:
~~~~
if (sort_crit.order != SORT_NAME &&
    sort_crit.order != SORT_ACTIVITY &&
    etc...) {
    cmdq_error(item, "invalid sort order");
    return (CMD_RETURN_ERROR);
}

daneofmanythings:
Yes that is an option. The reason I'm not preferential to that approach is that now the information about what each `sort_get_*` supports is duplicated in each location that calls it. The more locations that call the sort api, the more there will be duplicated code that checks for it in the same way. Each call site that wants to check will have a large chunk checking sort orders instead of just checking an enum specifying the result. Is that a valid concern at this time? A hybrid approach would be to make helper functions in the sort api that check the implemented sort orders for each `sort_get_*` so that it is duplicated, but in the same place, so that call sites can just use a helper to verify. I'm still preferential to adding a result field to `struct sort_criteria`, but ultimately I would like to defer to your judgement on the matter.

nicm:
Yes... but the only places that accept a user input are `-O` to commands? Everything else has a preset list, the user can't enter freeform text, so there is no room for picking an incorrect sort order.

But you could put the lists somewhere central and have the list commands validate against them.

daneofmanythings:
I may be trying to cross a bridge that hasn't been arrived at yet as there aren't that many call sites. 

My helper function suggestion is basically having the central list to validate against. Instead of the big conditional, a helper function would look something like `int sort_is_valid_order_buffers(enum sort_order)`. That function could then check for invalid and unimplemented orders. Or it could even look like `enum sort_order sort_session_order_from_string(const char *)` where the `sort_order_from_string` calls can be wrapped for validation if needed. Then one can just check for `SORT_END`.

nicm:
I don't know it seems a but premature to have common functions when they are only needed in one place, but if they aren't much bigger then could do that if you like.

daneofmanythings:
They are in more than one place if you want to support the `choose-` commands; `sort_get_panes_window`, `sort_get_winlinks_session` and `sort_get_clients` are all in an additional place, while `sort_get_sessions` is in 2 additional places (also in `cmd-switch-client`). This is what one of the functions would look like:
```c
enum sort_order
sort_session_order_from_string(const char* orderstr)
{
	enum sort_order	order;

	order = sort_order_from_string(orderstr);
	if (order != SORT_INDEX &&
	    order != SORT_CREATION &&
	    order != SORT_ACTIVITY &&
	    order != SORT_NAME)
	        order = SORT_END;

	return order;
}
```

I am hesitant to push this idea much further as just checking for `SORT_END` out of `sort_order_from_string` covers the most common error case, and the supported orders are documented in the man page. I'm beginning to think that this isn't worth it.

nicm:
I don't mind either way, I'm OK with the behaviour of error for orders that don't exist at all and default for those that exist but aren't supported in context, but if you want to check for those as well then that's fine, although I would probably not build out too much infrastructure just to detect it.

daneofmanythings:
I have convinced myself out of reporting unsupported orders. A user will likely look at the man page to learn about the `O` flag, and in doing so will see the supported orders, When debugging, they are likely to go back to that spot and check. Given that assessment, there isn't much to be gained practically, especially if it starts plopping 5 new nearly bespoke functions in the mix cluttering up the code. I agree with your initial assessment of checking for `SORT_END` after `sort_order_from_string`.

I'll address this in the `switch-client`, `list-*` and `choose-*` commands in a new pr if that works for you.

nicm:
Sure new PR is fine thanks.

PR #4848, opened 2026-02-07
Fix stability bugs and improve allocation performance

taylorconor:
A lot of somewhat unrelated small fixes and improvements in this PR -- I split it by commit to hopefully make it easier to review!

Stability fixes:
- Add NULL checks for grid_peek_line() in grid_string_cells and five sites in window-copy.c reachable during pane resize.
- Fix wrong TAILQ member in input_cancel_requests, use centry not entry.
- Move client keytable check before unref in key_bindings_remove_table.

Performance improvements:
- Use geometric (2x) growth for grid linedata and extended cell arrays instead of +1 reallocation.
- Batch runs of printable bytes in control mode output instead of per-byte evbuffer_add_printf.
- Do not call tty_attributes twice for single-width characters.
- Check log level before log_debug in tty_draw_line.
- Only set cursor position when needed in scrollbar drawing.
- Pre-grow extddata in grid_set_cells.
- Check event_pending before event_add in tty_add.
- Format layout change string once per window in control_notify_window_layout_changed instead of per client.
- Reuse format trees for subscriptions in control_check_subs_timer.
- Use offset tracking instead of strlcat in grid_string_cells.

### DIFF
--- a/control-notify.c
+++ b/control-notify.c
@@ -65,8 +65,11 @@ control_notify_window_layout_changed(struct window *w)
 		if (!CONTROL_SHOULD_NOTIFY_CLIENT(c) || c->session == NULL)
 			continue;
 		s = c->session;
-		if (winlink_find_by_window_id(&s->windows, w->id) != NULL)
-			control_write(c, "%s", cp);
+
+		if (winlink_find_by_window_id(&s->windows, w->id) == NULL)
+			continue;
+
+		control_write(c, "%s", cp);
 	}
 	free(cp);
 }
--- a/control.c
+++ b/control.c
@@ -1081,9 +1081,7 @@ control_check_subs_timer(__unused int fd, __unused short events, void *data)
 		case CONTROL_SUB_WINDOW:
 			control_check_subs_window(c, csub);
 			break;
-		case CONTROL_SUB_SESSION:
-		case CONTROL_SUB_ALL_PANES:
-		case CONTROL_SUB_ALL_WINDOWS:
+		default:
 			break;
 		}
 	}
@@ -1095,10 +1093,9 @@ control_check_subs_timer(__unused int fd, __unused short events, void *data)
 				ft = format_create_defaults(NULL, c, s, wl, wp);
 				RB_FOREACH_SAFE(csub, control_subs, &cs->subs,
 				    csub1) {
-					if (csub->type != CONTROL_SUB_ALL_PANES)
-						continue;
-					control_check_subs_all_panes_one(c,
-					    csub, ft, wl, wp);
+					if (csub->type == CONTROL_SUB_ALL_PANES)
+						control_check_subs_all_panes_one(
+						    c, csub, ft, wl, wp);
 				}
 				format_free(ft);
 			}
@@ -1111,10 +1108,9 @@ control_check_subs_timer(__unused int fd, __unused short events, void *data)
 			ft = format_create_defaults(NULL, c, s, wl, NULL);
 			RB_FOREACH_SAFE(csub, control_subs, &cs->subs,
 			    csub1) {
-				if (csub->type != CONTROL_SUB_ALL_WINDOWS)
-					continue;
-				control_check_subs_all_windows_one(c, csub, ft,
-				    wl);
+				if (csub->type == CONTROL_SUB_ALL_WINDOWS)
+					control_check_subs_all_windows_one(c,
+					    csub, ft, wl);
 			}
 			format_free(ft);
 		}
--- a/grid.c
+++ b/grid.c
@@ -939,10 +939,22 @@ grid_string_cells_us(const struct grid_cell *gc, int *values)
 	return (n);
 }
 
+static inline void
+grid_string_cells_cat(char *buf, size_t len, size_t *off, const char *s)
+{
+	size_t	slen = strlen(s);
+
+	if (*off + slen < len) {
+		memcpy(buf + *off, s, slen);
+		*off += slen;
+	}
+	buf[*off] = '\0';
+}
+
 /* Add on SGR code. */
 static void
-grid_string_cells_add_code(char *buf, size_t len, u_int n, int *s, int *newc,
-    int *oldc, size_t nnewc, size_t noldc, int flags)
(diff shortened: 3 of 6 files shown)


nicm:
Thanks.

440f0389e97b375ca3ae7cfc749aa0d929d4e2f9 - I don't see how the y can be out of range for most of these. Can you give more detail?

9308394ed70b40ed6280f2a455a06b833ad49fc9 - think this is right.

9dbb69fd6887b9d1e695e09a7b316135f7af05d5 - table is not dereferenced, so how is this a use-after-free?

a123ec2348fdfeaa778f832542141ce010309daf - screen_write_free_citem does not free ci so there is no use-after-free here (there is a case for renaming the function), and there is an unrelated change from fatal to log_debug here which we are not going to change.

1f7c96186c4325dfa5ca2840fd450f545ef0cd18 - no, unless you can show that the fatal are being hit in normal cases and why that is not a bug.

3ff557e7f462ca0d26be9c3b550ba3b71b45a1c6 - maybe, will look at this again

5215bc07a501d0e4f17dbd6bc2db06b0826828e7 - maybe

0755420dc45ebfed88a4ff9013ae00cd9464863a - think this is probably OK

ce8125b2f5eff1b1edf37751d5162aa2ed9b1ff4 - OK good idea

145ec9bd4387c058029dc2a8315d5be8eadf6ba4 - I don't think with an extra alloc is going to be enough of an improvement to be worth it

7282c7ea4acf69be7b12f7a56bfe7373134e5394 - maybe, will look at again

371845ec97dd74d1468ca80b18be29ba6ac8d133 - maybe

b74ae9d49e57b9a3ff5fdc6c1ca77a44e4f0c8ad - why is this better?

14edcfc683f5fd58d252145cabd59f4826fdc39a - I don't think this is worth it, format_draw is heavy enough this will probably be lost in noise  or have you numbers to show it is not?

db36df94fc9d57f8d7d0928098729719efd60f19 - probably yes

9ab1944baf38b6f84f4d3bd53b10ddc5201f50ff - maybe yes

3b882b75a80605126438964413fcb9f586ee76d0 - maybe, do you have numbers for this?

a7fba25456aa11e6243d2ab80aab31171be3a450 - I'll look at this again but my feeling is `grid_string_cells` is not a hot path so it's probably not worth the extra complexity

taylorconor:
Thanks for the review!
 
I've dropped the rejected commits (a123ec2, 1f7c961, 145ec9b) and I also dropped 14edcfc (lazy screen init) and 3b882b7 (compact threshold) since I don't have numbers for those and your intuition is a good enough signal to drop.

For the ones you had questions about:

440f038: grid_reflow during terminal resize rebuilds the grid and can change hsize+sy. Copy mode holds coordinates from before the resize: if a copy-mode operation runs afterward with the old y, it can be beyond the new hsize+sy and grid_peek_line returns NULL. The NULL return was already anticipated in grid_string_cells (the old code checked gl == NULL inside the loop), but the check was after the dereference a few lines above. The window-copy.c sites are the same pattern — stale coordinates after resize.

9dbb69f: Ah good point, "use-after-free" was wrong terminology; the pointer is compared not dereferenced. The real issue was that the TAILQ_FOREACH was outside the if (table != NULL) block. I've updated the commit message to just describe the code move.

b74ae9d: During a full redraw tty_add is called for every output chunk. After the first call the write event is already pending so subsequent event_add calls do redundant work inside libevent. event_pending is a cheaper check. Happy to drop if you don't think it's worth it!

nicm:
I've applied these to OpenBSD, they should be in GitHub later:

9308394ed70b40ed6280f2a455a06b833ad49fc9
9dbb69fd6887b9d1e695e09a7b316135f7af05d5
b74ae9d49e57b9a3ff5fdc6c1ca77a44e4f0c8ad
b515701172d16ba6b153479d2e196dff82ed511d
0755420dc45ebfed88a4ff9013ae00cd9464863a
5215bc07a501d0e4f17dbd6bc2db06b0826828e7

When they land can you please merge master into your MR so I can see what is left?

I don't think 7282c7ea4acf69be7b12f7a56bfe7373134e5394 will make much difference because `tty_cursor` shouldn't do anything if the cursor does not need to be moved. Is tmux actually emitting unnecessary cursor movement sequences? That seems like it would point to an error in `tty_cursor` or `tty_cell`.

taylorconor:
Done, thank you!

nicm:
OK I have applied:

https://github.com/tmux/tmux/pull/4848/commits/e85d74b482cb827142dbda90b12c1a4868730817

This one is already done:

https://github.com/tmux/tmux/pull/4848/commits/7ddcdbc3e18ca6e513076dfee6043d19120a15fa

I don't think this one is needed, `tty_cursor` should be a no-op if not needed and even if it does get through it should be cheap:

https://github.com/tmux/tmux/pull/4848/commits/1b0f5a1565b3ee27b32407dd112591418e046ad0

For https://github.com/tmux/tmux/pull/4848/commits/f2f35383125bbb28b3b1f2a7c220e3bd68188636 and https://github.com/tmux/tmux/pull/4848/commits/9984d4e4332882d5188d41be61feff9d9cc905f8:

- Firstly, they need an overflow check because we don't get it from `xreallocarray` anymore.

- Secondly, I think * 2 is maybe a bit aggressive. It seems like we could be wasting a lot of memory. Maybe it would be better just to add on 100 lines or so each time?

There are similar overflow errors in `cmd-list-keys.c` and `window_copy_stringify` which should both be using `xreallocarray`. `input_input` is OK on 64 bit platforms but could be wrong on 32 bit, it probably needs a check as well (which is realistically the only kind of platform where you will overflow before OOM anyway).

nicm:
I applied https://github.com/tmux/tmux/pull/4848/commits/d9b0022a4d8602cdcbb5bc807558b09ef5ab226d as well with some trivial changes, but I didn't test it because I don't have iTerm2 and I can't remember how control subs work and don't have time to figure it out :-/...

nicm:
Applied https://github.com/tmux/tmux/pull/4848/commits/17f33eef7c7c428233aacbefc39af7b3219b91ef as well.

PR #4858, opened 2026-02-15
Pass through which clipboard to set for OSC 52 paste

lindskogaxel:
Currently tmux will pass through the clipboard list on OSC 52 copy events, however it seems like paste events where hardcoded to reply without the clipboard that was used. This seems extra relevant with the introduction of the get-clipboard command where this feature might serve some purpose with the outer terminal, compared to just pasting from a tmux buffer. My primary reason for this PR however is that emacs expects this behavior and with how it is now, emacs can't paste with OSC 52 in tmux.

With this PR, tmux passes through the first clipboard found in the paste reply from the terminal if get-clipboard is not "buffer", otherwise it just uses the first one found in the paste query from the application.

There is no extra allocations here since the clipboard is not supposed to be a list in the reply (compared to the query, from how i understand it), so we just pass a char along to input_reply_clipboard.

Feedback is appreciated!

### DIFF
--- a/input.c
+++ b/input.c
@@ -3089,7 +3089,7 @@ input_osc_133(struct input_ctx *ictx, const char *p)
 
 /* Handle OSC 52 reply. */
 static void
-input_osc_52_reply(struct input_ctx *ictx)
+input_osc_52_reply(struct input_ctx *ictx, char clip)
 {
 	struct paste_buffer	*pb;
 	int			 state;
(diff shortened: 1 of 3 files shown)


nicm:
Looks good, thanks, but I think if the response *does* contain multiple clipboards in the first argument, tmux should probably ignore them rather than using the first?

Also I'd just change `flags` consistently to `clipboard` (or maybe just `clip` would be better because you have some long lines already) so in `input_osc_52_parse` and `input_osc_52` and `input_osc_52_reply` as well.

lindskogaxel:
Your suggestion makes sense, but would you also then validate the clipboard returned at that point? I pushed a new commit, now ignoring the clipboard if multiple was sent, along with some cleanups regarding the naming.

nicm:
Applied to OpenBSD now with some minor changes (mainly renaming other `flags` to `clip` and shortening lines), will be in GitHub later. Thanks!

PR #4861, opened 2026-02-16
Enable partial-width prompt overlay with additional message-* options

taylorconor:
## Overview

Allow prompts and messages to occupy only a portion of the status bar, overlaying the normal status content rather than replacing the entire line.

Currently, when the command prompt or a message is displayed, it overwrites the entire status line. This hides all window tabs and status information. With these changes, users can configure the prompt to occupy only part of the status bar (e.g. 20% right-aligned with a margin and minimum width), keeping the surrounding status content visible.

## New options

| Option              | Type    | Default      | Description                          |
|---------------------|---------|--------------|--------------------------------------|
| `message-format`    | string  | `#{message}` | Format string wrapping prompt content with decorations |
| `message-margin`    | number  | `0`          | Column offset between the message area and the status line edge determined by `message-position` |
| `message-min-width` | number  | `0`          | Minimum width of the message area in columns |
| `message-position`  | choice  | `left`       | Horizontal position (`left`, `centre`, `right`) |
| `message-width`     | string  | `100%`       | Width of prompt/message area (percentage, absolute, or `auto`) |

## Visual example

Before (current behavior) -- the entire status line is replaced by the prompt:

```
(rename-window) bash
```

After -- the prompt overlaysonly part of the status line, tabs and clock remain visible:

```
1:bash   2:bash   3:bash   4:bash                | (rename-window) bash    | Sun Feb 15  22:20
```

## Backwards compatibility

Default values (`message-width "100%"`, `message-position "left"`, `message-margin 0`, `message-min-width 0`, `message-format "#{message}"`) produce identical behavior to the current release. No visual change for users who do not configure the new options.

## Files changed

- **options-table.c** -- Register `message-format`, `message-margin`, `message-min-width`, `message-position`, and `message-width` options
- **status.c** -- Add `status_prompt_area()` geometry calculation and `status_prompt_draw_format()` decoration helper; modify `status_prompt_redraw()`, `status_message_redraw()`, `status_prompt_complete_list_menu()`, and `status_prompt_complete_window_menu()` to use constrained prompt area
- **tmux.1** -- Document the five new options

## Implementation approach

The rendering follows the existing "copy base status, then selectively overwrite" pattern. The only change is reducing the overwrite region from full-width to the computed prompt area. The `#{message}` marker in `message-format` is handled by splitting the raw format string before `format_expand_time()` so it survives expansion as a split point, then drawing the left/right decoration parts separately via `format_draw()`.

## Testing notes

- Default configuration (no new options set): identical output to current release
- `message-width 50%` + `message-position centre`: tabs visible on both sides of prompt
- `message-width 50%` + `message-position right` + `message-margin 20`: prompt inset from right edge
- `message-width 20%` + `message-min-width 30`: area clamps to 30 columns on narrow percentages
- Long input with constrained width: existing scroll logic works within the narrower area
- Tab completion menus: correctly offset by prompt area position
- `message-width auto`: sizes to content, rounds to 10-column increments
- `message-format` with `#[style]` directives: powerline arrows render correctly
- Works with `status-position top/bottom`, multi-line status bar, and `status off`

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -733,7 +733,10 @@ const struct options_table_entry options_table[] = {
 	{ .name = "message-format",
 	  .type = OPTIONS_TABLE_STRING,
 	  .scope = OPTIONS_TABLE_SESSION,
-	  .default_str = "#[#{E:message-style}]#{message}",
+	  .default_str = "#[#{?#{command_prompt},"
+		  "#{E:message-command-style},"
+		  "#{E:message-style}}]"
+		  "#{message}",
 	  .text = "Format string for the prompt and message area. "
 		  "The '#{message}' placeholder is replaced with the content."
 	},
--- a/options-table.c
+++ b/options-table.c
@@ -44,9 +44,6 @@ static const char *options_table_status_list[] = {
 static const char *options_table_message_line_list[] = {
 	"0", "1", "2", "3", "4", NULL
 };
-static const char *options_table_message_position_list[] = {
-	"left", "centre", "right", NULL
-};
 static const char *options_table_status_keys_list[] = {
 	"emacs", "vi", NULL
 };
@@ -726,7 +723,7 @@ const struct options_table_entry options_table[] = {
 	{ .name = "message-command-style",
 	  .type = OPTIONS_TABLE_STRING,
 	  .scope = OPTIONS_TABLE_SESSION,
-	  .default_str = "bg=black,fg=yellow",
+	  .default_str = "bg=black,fg=yellow,fill=black",
 	  .flags = OPTIONS_TABLE_IS_STYLE,
 	  .separator = ",",
 	  .text = "Style of the command prompt when in command mode, if "
@@ -736,7 +733,7 @@ const struct options_table_entry options_table[] = {
 	{ .name = "message-format",
 	  .type = OPTIONS_TABLE_STRING,
 	  .scope = OPTIONS_TABLE_SESSION,
-	  .default_str = "#{message}",
+	  .default_str = "#[#{E:message-style}]#{message}",
 	  .text = "Format string for the prompt and message area. "
 		  "The '#{message}' placeholder is replaced with the content."
 	},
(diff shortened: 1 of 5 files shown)


nicm:
I like the idea but I think this should be expressed through styles and not options, we are trying to move away from multiple options for styling the same thing. It would be better to make it so the message is drawn on top of the existing status line as standard and  `message-style` (or `message-format`) must contain a `fill` directive if it is to be cleared (the default can do this).

nicm:
What I'm suggesting would mean it could work like `status-format` and the default `message-format` could be something like:
~~~~
#[#{E:message-style}]#{message}
~~~~
And `message-style` something like:
~~~~
align=left fill=yellow bg=yellow
~~~~
If you want the ability to set the width I suggest adding a `width` directive for styles. This could take either a fixed width (<= max available) or a %.

taylorconor:
Thanks for the feedback! I've reworked the approach as suggested.

Here's the changes in the updated commit:

- **Removed** `message-width`, `message-position`, `message-margin`, and `message-min-width` options entirely. Width and position are now expressed through style directives on `message-style`.

- **Added percentage support to the `width` style directive** (`width=50%`), so `message-style` can set either a fixed width or a percentage of available space.

- **Changed the default `message-format`** to `#[#{E:message-style}]#{message}`, so the format references the style directly.

- **Added `fill` to the default `message-style`** (`bg=yellow,fg=black,fill=yellow`) and `message-command-style` (`bg=black,fg=yellow,fill=black`). The message is now always drawn on top of the existing status line, and the `fill` directive controls whether the background is cleared. Removing `fill` from the style would leave the status bar content visible behind the message text.

- **The `align` directive on `message-style`** controls horizontal position (`left`, `centre`, `right`) of the prompt area within the status line.

Example configuration for a centered half-width prompt:

```
set -g message-style "bg=yellow,fg=black,fill=yellow,align=centre,width=50%"
```

The `message-format` option is still available for adding decorations (powerline arrows etc.) around `#{message}`, but the geometry is entirely driven by the style.

nicm:
This looks good, but it has your neighbour stuff as well, could you take that out? (Need to take a look at that when I have time, sorry for delay...)

taylorconor:
Oops my bad. Rebased!

nicm:
The code looks good but I don't like how this is building the screen based on information that is not in the format.

If you want part of the text to be on the right you should have to say `#[align=right]`. I don't think having it mysteriously infer what is left and what is right based on where it is relative to `#{message}` is intuitive.

Similarly, why do you need to pull `fill` out of `message-style`? What happens if I set `message-format` to `#[fill=blue]#{message}` instead? Ideally `message-style` should not be checked in C code but instead embedded in `message-format` as something like `#[#{?command_mode,#{E:message-command-style},#{E:message-style}}]`.

I think there should only be one `format_draw` call and it should take care of everything, it already understands how to align to left and right. Take a look at `status.c:453` - it sets up the screen, then calls `format_draw` to draw the entire status line. Can we not do the same for this? Except instead of clearing the status line, we start by copying the existing one to draw the message on top.

It may need some tweaks to the formats to handle width - it may need to differentiate between "available width" (the total draw area) and the width the user wants (ie "width=X"). I can't remember how that is handled right now...

taylorconor:
Ok, I've reworked this to use a single `format_draw` call for both messages and prompts, following the same pattern as the status line drawing (copy existing screen, then format_draw on top).

The style is now embedded in message-format as a conditional `(#[#{?#{command_prompt},#{E:message-command-style},#{E:message-style}}]#{message})`, so `message-style` is not checked in C code for fill or style selection. `status_prompt_draw_format` and the left/right decoration inference are removed -- if you want right-aligned content, you use `#[align=right]` in the format. Width and align from `message-style` still control the area geometry (position and available width passed to `format_draw`), which I think is needed since `format_draw` doesn't currently distinguish between available width and desired width.

Let me know fi this satisfies the concern; I'm happy to adjust if you'd prefer it gets handled differently!

nicm:
I like this a lot. One thing - in `status_prompt_redraw` there is:
~~~~
        if (c->prompt_mode == PROMPT_COMMAND)
                stylename = "message-command-style";
        else
                stylename = "message-style";
~~~~
Is this still needed? Could the default always be `grid_default_style` since it will be replaced by the option?

taylorconor:
Ok! I removed the `message-command-style`/`message-style` selection in `status_prompt_redraw` (and the equivalent `style_apply` in `status_message_redraw`) and replaced both with `grid_default_cell`, since `message-format` already embeds the style via `#{?#{command_prompt},#{E:message-command-style},#{E:message-style}}`.

nicm:
Looks good and works well for me. Applied to OpenBSD now, will be in GitHub later. Thanks!

ccjmne:
Consider this `tmux` configuration:

```tmux
set -g message-style 'fg=white'
```

(that's all of it, nothing else needed for my example)

This previously valid entry would now cause the command prompt to be overlaid on top of the status bar, rather than (visually) replace it.

A solution is to add, for instance, `fill=black`.
**Is it *the most appropriate solution*?**

Issue #4859, opened 2026-02-15
Feature request: Configurable limit for memory usage for a pane's history


darlingm:
### Issue description

There is `history-limit`, which does not help for a program thrashing terminal rewrites that is not outputting new lines.  A `history-bytes` could be useful, for the maximum size in bytes/KB/MB kept in the history for each pane.

A bug in OpenAI Codex CLI ( openai/codex#11857 ) was causing 78KB/sec of redrawn characters for a blinking dot, which for 2-3 panes doing this ate about 48GB of memory for scrollback buffers!  A different fix in Codex CLI has indirectly suppressed this, but made me realize that although tmux offers `history-limit` for the number of lines, for situations like this which isn't caused by many lines but thrashing redraws, a `history-bytes` would be useful.  I see there's `memory-limit` which could be useful, but I still think a direct cap on the size of history would be useful, rather than tmux deciding what information to drop, which may unnecessarily include other panes' history.

### Required information

* tmux version: 3.6a (found on 3.4a, but after upgrading to 3.6a, history still grows uncontrolled and doesn't have a `history-bytes` type option for a maximum on a pane's history size in bytes.)
* Platform: `Linux unknown` (`Linux 6.18.8-arch2-1 x86_64 unknown`.)
* Terminal in use: Suckless terminal.
* $TERM *inside* tmux: `tmux-256color`.
* $TERM *outside* tmux: `st-256color`.
* Logs from tmux: Logs not needed for a feature request, but see this line in top, which was traced back to 2-3 panes of history which only had about 1,000 lines of scrollback each:

        PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
       2942 mdarling  20   0   48.2g  23.8g   3048 S   2.0  37.8  76:21.08 tmux: server

nicm:
I don't know what you mean. Where was this memory being held if not in the history?

nicm:
This sounds like either a leak in tmux or it is the usual poor glibc malloc, because tmux does not keep memory for a pane except what is in the history and unless you managed to have some spectacularly long lines, 1000 lines cannot take many gigabytes of memory.

If you patch tmux to call malloc_trim more often (so lines 160 and 183 of `server.c` change 3600 to say 300 for 5 minutes), does the memory use grow when the output is present but then drop around 5 minutes after it stops?

nicm:
You could also try building tmux with `--enable-jemalloc`.

[...]

PR #4862, opened 2026-02-16
Re-use extended entry when clearing RGB cell.

darlingm:
Previously, grid_clear_cell allocated a new extended slot on every RGB clear, leaving old slots unreachable until compaction, causing unnecessary extddata growth.  Under very high-rate EL/ED redraws, this can be a substantial amount of memory.  Fixes the underlying bug with #4859.

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -205,11 +205,17 @@ grid_clear_cell(struct grid *gd, u_int px, u_int py, u_int bg)
 	struct grid_line	*gl = &gd->linedata[py];
 	struct grid_cell_entry	*gce = &gl->celldata[px];
 	struct grid_extd_entry	*gee;
+	u_int			 old_offset = gce->offset;
+	int			 had_extended = (gce->flags & GRID_FLAG_EXTENDED);
 
 	memcpy(gce, &grid_cleared_entry, sizeof *gce);
 	if (bg != 8) {
 		if (bg & COLOUR_FLAG_RGB) {
-			grid_get_extended_cell(gl, gce, gce->flags);
+			if (had_extended && old_offset < gl->extdsize) {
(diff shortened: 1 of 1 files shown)


nicm:
Nice find. I have applied this to OpenBSD and it should be in GitHub later.

This is a good optimization and fixes clearing to RGB background, but I think you could still see a similar problem by repeatedly clearing to non-RGB to clear the extended flag then setting an extended cell (either RGB or UTF-8).

Compacting extended cells when they leave the visible screen seemed like a sensible compromise but maybe it is not enough. Perhaps we should do it at some other time as well? For example, we could track the number of unused extended cells and compact when the proportion becomes more than say 25%. Or maybe that is overcomplicated for what is problem that is dramatic but pretty rare...

darlingm:
You're right about the non-RGB clearing.  #4865 fixes this and a bug introduced by this PR, of reading uninitialized data that can happen through `grid_expand_line()`.

The only remaining path I can think of to unnecessarily growing `extddata` would be toggling a cell between styles that need extended storage and styles that don't, which would drop the `GRID_FLAT_EXTENDED` reference.

Compacting when it's over-allocated by some amount would work.  I wonder how much lag compaction could cause.  When lines are scrolling, I think the minimal lag would be less noticeable.  I wonder if it would look like a studder/jitter after a character if compaction kicked in.

Keeping a "freed" list to use before reallocating would work, but would itself increase memory usage to maintain.

nicm:
An option might be to compact periodically based on some other criteria, for example the number of cells written (every 100k or something?). This would only need an extra counter in `struct grid` and would avoid memory blowing up with super-frequent updates but probably wouldn't be noticeable in normal use (we only need to compact lines on the visible part of the grid after all).

PR #4865, opened 2026-02-18
Init new celldata entries, and re-use extended entry when clearing non-RGB cell.

darlingm:
#4862 fixed RGB clears by reusing an existing extended slot instead of creating a new one.  This PR has two follow-ups:

1. Fix a bug introduced by #4862: `grid_clear_cell()` may read `flags` and `offset` with uninitialized data to decide whether to reuse an extended slot, if `grid_expand_line()` grew `celldata`.  Zero the newly added `celldata` memory.
2. Reuse an existing extended slot for non-RGB clears as well.

Follows up on the underlying bug with #4859.


nicm:
Thanks... Could `grid_expand_line` just  use `xrecallocarray` instead of `memset`?

darlingm:
Yes, that makes sense.  I made the change.

nicm:

PR #4868, opened 2026-02-19
Add Sixel image rendering and pixel cleanup for display-popup

### LINKED ISSUE #4867: Sixel images not rendered in display-popup overlays; pixel artifacts persist after close
AI Assistant:

## Summary

When a program running inside a `tmux display-popup` outputs Sixel images, they are silently discarded and never rendered. Additionally, if sixel rendering is patched in, pixel artifacts ("ghosting") persist on screen after the popup closes.

### 1. DCS dispatch drops Sixel in popups

`input_dcs_dispatch()` returns early with `if (wp == NULL) return` before reaching the sixel parsing code. Popups don't have a `window_pane`, so all DCS sequences (including sixel) are dropped.

### 2. `popup_draw_cb()` ignores images

The popup draw callback only renders text lines via `tty_draw_line()` and skips images stored in the popup screen's image list (`pd->s.images`).

### 3. No sixel pixel cleanup on popup close

When a popup with sixel images closes, the text layer is repainted but pixel-level sixel data persists in the terminal emulator's graphics layer. Standard escape sequences (ED, EL) don't clear this.

## Proposed Fix

1. Move sixel handling in `input_dcs_dispatch()` before the `wp == NULL` guard, sourcing pixel dimensions from the client's tty when no window pane is available.
2. Add sixel image rendering in `popup_draw_cb()` by iterating `pd->s.images` and calling `tty_write_one(tty_cmd_sixelimage, ...)`.
3. Implement multi-strategy pixel cleanup in `popup_free_cb()`: alternate screen toggle (RMCUP/SMCUP), space overwrite, DECFRA rectangle fill, and blank sixel overlay.

Testing shows that the RMCUP/SMCUP alternate screen toggle accounts for ~90% of the cleanup effect on Konsole.

## Testing

Tested on Konsole with an automated test toolkit that:
- Generates red test images (PNG + Sixel)
- Opens popups (bordered and borderless), displays sixel, closes, captures screenshots
- Analyzes pixel residue programmatically

Test toolkit and methodology: [tmux Sixel Popup Cleanup Testing Toolkit](https://gist.github.com/gwpl/722cd9bab3b92b2cf4ae1bbcb16544d1)

## Screenshots

Image

Image

Image

Image

Image

Image

## Environment

- tmux: next-3.7 (`master` branch with `--enable-sixel`)
- Terminal: Konsole (KDE)
- OS: Arch Linux

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

gwpl:
## Summary

* Fix Sixel image rendering inside `display-popup` overlays (both bordered and borderless)
* Add multi-strategy pixel cleanup on popup close to prevent sixel ghosting artifacts

Fixes https://github.com/tmux/tmux/issues/4867

### Commit 1: Sixel rendering in popups

* Move sixel handling in `input_dcs_dispatch()` before `wp == NULL` guard, sourcing pixel dimensions from client tty when no window pane is available
* Add sixel image drawing in `popup_draw_cb()` via `tty_write_one(tty_cmd_sixelimage, ...)`
* Expose `tty_write_one()` (previously static) so popup.c can call it
* Add regression test `regress/display-popup-sixel.sh`

### Commit 2: Multi-strategy pixel cleanup

Four cleanup strategies in `popup_free_cb()`, selectable via compile-time bitmask:

| Strategy | Flag | Method |
|----------|------|--------|
| SCREEN | 0x1 | RMCUP/SMCUP alternate screen toggle + ED2/DECSED |
| SPACES | 0x2 | Overwrite popup area with spaces |
| DECFRA | 0x4 | VT420 rectangle fill per image |
| BLANK_SIXEL | 0x8 | Blank sixel with P2=2 background erase |

Testing shows SCREEN (RMCUP/SMCUP toggle) accounts for ~90% of the cleanup
effect on Konsole -- toggling the alternate screen buffer discards the
graphics layer.

## Testing

* Automated test toolkit: [tmux Sixel Popup Cleanup Testing Toolkit](https://gist.github.com/gwpl/722cd9bab3b92b2cf4ae1bbcb16544d1)
* Full test report: [Test Report: tmux Sixel Popup Rendering and Cleanup](https://gist.github.com/gwpl/69f78166f966286bb4e9df0c8de95702)

Regression results on Konsole (WhiteOnBlack, Hack 10pt, threshold=300):
```
baseline_no_sixel:       PASS (red=191, threshold=300)
sixel_visible_bordered:  PASS (red=32324, threshold=300)
cleanup_bordered:        PASS (red=255, threshold=300)
sixel_visible_borderless: UNSURE (red=211, threshold=300)
cleanup_borderless:      UNSURE (red=316, threshold=300)

REGRESSION: PASS
```

## Screenshots

*(Screenshots to be attached: sixel in popup, cleanup result on dark/light backgrounds)*

## Test plan

- [x] Sixel renders in bordered popup
- [x] Sixel renders in borderless popup (-B)
- [x] Pixel cleanup on bordered popup close (PASS)
- [x] Pixel cleanup on borderless popup close (PASS/UNSURE)
- [x] Tested with WhiteOnBlack and BlackOnWhite backgrounds
- [x] Regression test passes (`regress/display-popup-sixel.sh`)
- [ ] Test on other sixel-capable terminals (foot, kitty, WezTerm)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/input.c
+++ b/input.c
@@ -2559,28 +2559,45 @@ input_dcs_dispatch(struct input_ctx *ictx)
 	int			 p2;
 #endif
 
-	if (wp == NULL)
-		return (0);
-	oo = wp->options;
-
 	if (ictx->flags & INPUT_DISCARD) {
 		log_debug("%s: %zu bytes (discard)", __func__, len);
 		return (0);
 	}
 
 #ifdef ENABLE_SIXEL
-	w = wp->window;
+	/*
+	 * Handle sixel images for both panes (wp != NULL) and popups
+	 * (wp == NULL, but ictx->c is set). Get pixel dimensions from
+	 * the window if available, otherwise from the client's tty.
+	 */
 	if (buf[0] == 'q' && ictx->interm_len == 0) {
-		if (input_split(ictx) != 0)
-			return (0);
-		p2 = input_get(ictx, 1, 0, 0);
-		if (p2 == -1)
-			p2 = 0;
-		si = sixel_parse(buf, len, p2, w->xpixel, w->ypixel);
-		if (si != NULL)
-			screen_write_sixelimage(sctx, si, ictx->cell.cell.bg);
+		u_int	xpixel = 0, ypixel = 0;
+
+		if (wp != NULL) {
+			w = wp->window;
+			xpixel = w->xpixel;
+			ypixel = w->ypixel;
+		} else if (ictx->c != NULL) {
+			xpixel = ictx->c->tty.xpixel;
+			ypixel = ictx->c->tty.ypixel;
+		}
+		if (xpixel != 0 && ypixel != 0) {
+			if (input_split(ictx) != 0)
+				return (0);
+			p2 = input_get(ictx, 1, 0, 0);
+			if (p2 == -1)
+				p2 = 0;
+			si = sixel_parse(buf, len, p2, xpixel, ypixel);
+			if (si != NULL)
+				screen_write_sixelimage(sctx, si,
+				    ictx->cell.cell.bg);
+		}
 	}
 #endif
+
+	if (wp == NULL)
+		return (0);
+	oo = wp->options;
 
 	/* DCS sequences with intermediate byte '$' (includes DECRQSS). */
 	if (ictx->interm_len == 1 && ictx->interm_buf[0] == '$') {
--- a/popup.c
+++ b/popup.c
@@ -26,29 +26,6 @@
 
 #include "tmux.h"
 
-#ifdef ENABLE_SIXEL
-/*
- * Strategies for clearing stale sixel pixels when a popup with images closes.
- * Different terminal emulators respond to different clearing methods, so
- * multiple strategies can be combined (bitwise OR).
- *
- * SCREEN (RMCUP/SMCUP toggle) is the most effective on Konsole and most
- * modern terminals -- toggling the alternate screen buffer discards the
- * graphics layer. The other strategies provide additional coverage for
- * terminals where the alternate buffer toggle alone is insufficient.
- *
- * Override at compile time: make CPPFLAGS="-DSIXEL_CLEAR_STRATEGIES=0x1"
- */
-#define SIXEL_CLEAR_SCREEN	0x1	/* RMCUP/SMCUP + ED2/DECSED/CLEAR */
-#define SIXEL_CLEAR_SPACES	0x2	/* Overwrite popup area with spaces */
-#define SIXEL_CLEAR_DECFRA	0x4	/* DECFRA rectangle fill (VT420+) */
-#define SIXEL_CLEAR_BLANK_SIXEL	0x8	/* Blank sixel with P2=2 bg erase */
-#define SIXEL_CLEAR_ALL		0xf	/* Enable all strategies */
-
-#ifndef SIXEL_CLEAR_STRATEGIES
-#define SIXEL_CLEAR_STRATEGIES	SIXEL_CLEAR_ALL
-#endif
-#endif /* ENABLE_SIXEL */
 
 struct popup_data {
 	struct client		 *c;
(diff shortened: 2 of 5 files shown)


nicm:
This looks OK, thanks. A few comments:

- Since you wrote this using AI, it is large enough to be copyrightable and they are claiming authorship, I need to see a statement from the Claude website showing they do not assert any actual ownership or copyright over the code (meaning that you are free to submit it to tmux with the ISC license tmux uses). I have a feeling someone may have shown me this before for Claude but I can't keep track of all the different AI tools so if you could dig it out that would be great...

- I don't think we need a compile time option and different strategies for removing SIXELs. I would be inclined just to always clear the whole area with spaces. Compared with the volume of data sent to most terminals a few spaces is not much. But if you want to do multiple things (could definitely try DECFRA if the terminal supports it - check for `TERM_DECFRA`) that is OK (but remember that DECFRA cannot clear to nondefault background colour).

nicm:
OK you can forget the first point since Claude has a good statement in their ToS, I have linked to it here since this is likely to come up again: https://github.com/tmux/tmux/wiki/Contributing#use-of-ai

gwpl:
AI Assistant:

Thanks for the review! Addressing both points:

## 1. Claude/Anthropic Copyright & Licensing

Anthropic explicitly assigns all rights in Claude outputs to the user and asserts no ownership. This is safe for ISC (and any OSS license):

* **Consumer Terms:** *"we assign to you all our right, title, and interest (if any) in Outputs"* -- [source](https://www.anthropic.com/legal/consumer-terms)
* **Commercial Terms:** *"Customer (a) retains all rights to its Inputs, and (b) owns its Outputs."* -- [source](https://www.anthropic.com/legal/commercial-terms)
* **Copyright Indemnity (API):** Anthropic defends customers from copyright claims on outputs and pays settlements -- [source](https://www.anthropic.com/news/expanded-legal-protections-api-improvements)

Additionally, Anthropic actively supports open source: the [Claude for Open Source](https://claude.com/contact-sales/claude-for-oss) program offers 10,000 free Claude Max 20x subscriptions to OSS maintainers.

Full reference with all links: [Anthropic Claude: Code Ownership & OSS Licensing Compatibility](https://gist.github.com/gwpl/2a7161d8d1ff47bbdc48d5feec25e838)

## 2. Simplify cleanup strategies

Agreed. Will push a revision that:

* **Removes** the compile-time bitmask, strategy defines, and RMCUP/SMCUP toggle
* **Always clears** the whole popup area with spaces (simple, effective)
* **Also tries DECFRA** when `TERM_DECFRA` is set (respecting that DECFRA cannot clear to non-default background colour)
* Drops the blank sixel strategy

Will update shortly.

j4james:
> remember that DECFRA cannot clear to nondefault background colour

@nicm Are you sure about that? In my experience, terminals with support for both sixel and `DECFRA` were able to overwrite images using both default and non-default colours. `DECERA` also worked.

The only issue I'm aware of with failure to overwrite was with Konsole, which couldn't erase sixel images with _default_ colours (nondefault colours did work), and that was just using spaces, since Konsole didn't support rectangular area operations. I also think that bug was supposed to have been fixed, but I'm not certain of that.

nicm:
Sorry, I misremembered - DECFRA cannot clear *to* default background, it clears the text but leaves attributes in place. Or at least, it did in xterm 327.

I asked Thomas Dickey about it in 2017 and he said that was expected, but I just tried my tests in xterm 390 and they worked, so maybe it has changed, although the code looks the same...

Not sure that it would affect SIXEL anyway.

ThomasDickey:
> Sorry, I misremembered - DECFRA cannot clear _to_ default background, it clears the text but leaves attributes in place. Or at least, it did in xterm 327.
> 
> I asked Thomas Dickey about it in 2017 and he said that was expected, but I just tried my tests in xterm 390 and they worked, so maybe it has changed, although the code looks the same...
> 
> Not sure that it would affect SIXEL anyway.

It wouldn't affect SIXEL (the graphics stuff doesn't mesh with character stuff).  Glancing at the xterm code, it appears to work as I described (perhaps some details on what and how you tested the feature would clarify things).

nicm:
Sorry, yes, it does work the same with xterm 390 - nothing has changed. This was my original email:

>printf '\033[42m\033[2Jxxx\033[41m\033[32;1;1;100;100$x'
>
>Works OK, the terminal is cleared as red.
>
>But if I change the background colour back to default instead of red:
>
>printf '\033[42m\033[2Jxxx\033[0m\033[32;1;1;100;100$x'
>
>The terminal is cleared as green - the DECFRA cleared the "xxx" text, but did not reset the background colour.

So you can clear to a background colour with DECFRA but not to the default colour.

ThomasDickey:
That's essentially because the "background color" is absence of color as set by the application.   (If someone supposes DEC terminals supported default color, I'd have to see the page and exact quote to discuss it).

j4james:
This is documented in VT520/VT525 Video Terminal Programmer Information, page 5-55:
> DECFRA replaces the rectangular area's character positions and attributes with the specified fill character. The fill character assumes the visual character attributes set by the last select graphic rendition (SGR) command.

Also in DEC STD 070-5, page 5-170:
> the terminal fills the rectangular area of character positions defined by Pt, Pl, Pb, and Pr, replacing both the character and the rendition present in those character positions with the character defined by the decimal value of Pch and the current renditions set through the SGR command.

In other words, for every position in the defined range, it's essentially doing:
```
display[y][x].code = Pch;
display[y][x].rendition = current_rendition;
```

There's nothing in the documentation suggesting anything more complicated than that. There's no mention of it selectively replacing some of the rendition attributes, and keeping others, dependent on the value of the current rendition.

ThomasDickey:
I think you already said that.  I was mentioning points which have little or no coverage in the documentation:
(a) default colors aka SGR 39/49 versus the color support in VT525
(b) clarifying what happens to SIXEL images overwritten by text messages.

(if it's obvious, then DEC 070 is overlong)

ThomasDickey:
Now... VT520RM lists SGR 39/49, but provides no description.  And SIXEL images aren't a feature of that model.  So it appears to me that discussing DEC's documented behavior for combining the two features relies on a lot of interpretation (i.e., not likely to get agreement).  With xterm, we can verify what it does, and occasionally suggest that it wasn't the best set of choices (but that's true of DEC terminals as well).

j4james:
> default colors aka SGR 39/49 versus the color support in VT525

The default colors on the VT525 are controlled by the `DECAC` control sequence, also configurable in the settings. See page 5-16: "The normal text colors are used in ANSI SGR color mode as the default foreground and background colors", and page 2-29: "The normal text background and foreground selections determine the writing color used when all SGR attributes are turned off." So when you write out text using the default colors, you're just getting a user-configured (or possible application-defined) pair of color indices.

> clarifying what happens to SIXEL images overwritten by text messages.

That's kind of obvious, because there's really only one thing that can happen (unless you're using a terminal with overstrike mode like the VK100). When writing out text, the terminal has to fill the background of the cell it's writing to, otherwise it would keep accumulating layers of text whenever you wrote to a cell that already had text in it.

And overwriting an image is no different from overwriting anything else, because it's just pixels in the display. It's not like the terminal would be recording the source of each pixel so it can somehow treat "image pixels" differently from "text pixels".

And anyway we know exactly how the DEC graphics terminals work, because there are videos and screenshots from people that have tested them, demonstrating this behavior. You can even test some of them yourself if you want using MAME.

> And SIXEL images aren't a feature of that model. 

I'm not sure how that matters. The behavior of text overwriting an image isn't going to somehow change because the terminal uses a different kind of escape sequence to set the text color attributes.

nicm:
Actually, what is the point in clearing the SIXEL explicitly with DECFRA or spaces? Why not just do a full redraw?

gwpl:
AI Assistant:

@nicm Re "why not just do a full redraw?" — the risk is that `tty_invalidate()` + redraw may use ED/EL to clear before redrawing text, and ED/EL don't clear sixel pixels on most terminals. The explicit spaces guarantee a glyph is written to every cell in the popup area, which is the one universally reliable way to overwrite the graphics layer.

That said, if you'd prefer a simpler approach, happy to test whether `tty_invalidate()` + `CLIENT_REDRAWWINDOW` alone is sufficient on Konsole/xterm — it's possible the full redraw does write to every cell.

<details>
<summary>DECFRA + sixel interaction (re the thread above)</summary>

Whether DECFRA clears sixel is **terminal-dependent, not standardised**:

- **VT420/VT510 spec**: DECFRA is a text-plane character-cell operation only — [no mention of graphics](https://vt100.net/docs/vt510-rm/DECFRA.html)
- **Real VT340 hardware**: DECERA did *not* affect sixel ([hackerb9/vt340test#16](https://github.com/hackerb9/vt340test/issues/16)), though EL *did*
- **xterm ≥370 (2021)**: Extended DECFRA/DECERA to also clear sixel graphics ([changelog](https://invisible-island.net/xterm/xterm.log.html)), driven by [notcurses#1740](https://github.com/dankamongmen/notcurses/issues/1740)
- **Konsole**: No DECFRA support at all

So spaces remain the only universal method. DECFRA is a bonus on xterm ≥370.
</details>

<details>
<summary>Offer: empirical testing on Linux terminals</summary>

Happy to run empirical tests (Konsole, xterm, foot) to verify:

1. Does redraw-only (no explicit spaces) leave sixel artifacts?
2. Does DECFRA clear sixel pixels on each terminal?
3. Do spaces alone suffice everywhere?

We have an [automated test toolkit](https://gist.github.com/gwpl/722cd9bab3b92b2cf4ae1bbcb16544d1) with pixel-level screenshot analysis that can answer these definitively.
</details>

Full research notes: https://gist.github.com/gwpl/e231915fce3c5ee37a6cad9dffc984a0

nicm:
I think instead of treating popup clear specially maybe it would be better to check for SIXELs at a lower level and if there are any use a redraw mechanism that clears them, so skip ED/EL and use DECFRA/spaces at that point? Maybe in `tty.c` set a flag in `struct tty` when a SIXEL is drawn and then check it to avoid ED/EL, would that work?

gwpl:
So @nicm I encountered in KDE Konsole redraw issues especially when switching tmux windows or scrolling amd Sixel drawing often staying "ghosting effect", however I thought its bigger and separate issue... if you think we could improve how tmux handles sixel in general , I am glad to help with some of my Agentic Engineering to make quality assurance empirical tests (that analize screenshots during interactions), debug, or make even Pull Requests following requests/guidelines what would be desired implementation (and testing ofc, to include proper testing... end2end integration testing in frame buffer in dockerfiles ?) ... anyway, I use Sixel often to display e.g. remderings of graphs from graphviz or other visuals , kind of accepted fact that I love with half working support (for which I need to overerite TERM var for tools like "viu" etc to work), still glad to try to run Agentic Workflows like I did for PR I attached to this issue if it would be any help (even as inspiration or proof of concept for further iterations of design). Also we peolple in thread would consider few options, I can ask Agents to make POC for each if helpful in decision making. 

Just looking for a way to be helpful.

My bottom line is following:

* Sixel currently in Tmux under KDE Konsole half works anyway, so adding imperfect support for popup would help me a lot as I could use mainstream tmux instead own patched
* Because it is partially working anyway then adding imperfect popup should be extra value added and step forward
* If in future better solution comes along , this commits can be amended, reverted, refactored, code removed, if better solution comes later along to replace this one.

Just sharing some thoughts. Thank you for taking care of tmux so many use!

nicm:
OK well if clearing SIXEL is already broken outside of some terminals, I think it would be better to not fix it just for popups - meaning I would leave the special case stuff for popups out for now and just do a full redraw.

Then if you would like to fix redraw properly to correctly clear SIXELs separately that would be great, but I think it needs to be done across the board, not just for popups.

gwpl:
No no, in Pop Ups it is completely not supported, 

zero, 

null,

I get zero null Sixel output in KDE Konsole in Popup,

that's why a PR, that seems to work.

nicm:
That's not what I mean. I mean it is OK to add support for popups, but we do not need to support special ways of clearing them if tmux does not use the same ways of clearing them for non-popups.

nicm:
I see what you mean about Konsole, it seems a bit broken with SIXEL:

The main thing is that scrolling an image appears to redraw it one line higher without removing the old image, so you end up with stuff like this:
image

I don't know if this is a tmux or Konsole bug.

Clearing SIXEL images also doesn't work:

- Part of this may be the same issue because tmux will scroll to clear in some circumstances. If you do `cat tools/image.sixel; clear` it doesn't clear the image but `cat tools/image.sixel; clear; clear` does, so I suspect in one case tmux is scrolling and in the other it is not.

- But splitting the window does not clear the images, and that should not be scrolling.

I tried a couple of naive changes to remove different ways of clearing but it didn't seem to help, so it needs to be investigated properly.

There is also some odd stuff with weird background colours appearing with `C-b Space` (so after reflow maybe?), but I suspect this is unrelated to SIXEL and more likely the `tty-draw.c` changes.

nicm:
I expect all the above may be one single problem in that tmux doesn't know how to correctly clear SIXEL images in Konsole.

nicm:
I don't think using spaces clears SIXELs either in Konsole (at least 23.08.5), try this:
~~~~
printf '\033[H\033[2J'; cat tools/image.sixel
sleep 2
printf '\033[H';perl -e 'print " "x50, "\n" for 1..20'
~~~~
I don't think we can do much to support it...

j4james:
> I don't think using spaces clears SIXELs either in Konsole

This is a known issue with Konsole, which I mentioned previously. You can see the bug report here:
https://bugs.kde.org/show_bug.cgi?id=456354

It was supposed to have been fixed, but I've seen a lot of people complaining that it still doesn't work. The last version I tested was 24.12.3, which I think should have had the fix, and that definitely didn't work for me.

You can overwrite an image with spaces if you use a non-default background color, but that doesn't actually erase it, so if you write to that area again with the default background the image will reappear.

The only reasonable method of deleting an image on Konsole is with `\e[2J`. But if you're willing to write Konsole specific code, you can trick it into removing the image without clearing the _entire_ screen with something like `\e7\e[2;3r\e[100S\e[r\e8` (the scroll operation will push the image offscreen, regardless of whether it's in the margin area or not, and this way you're only losing two rows of text). Not that I'm recommending this - just thought it was interesting.

> The main thing is that scrolling an image appears to redraw it one line higher without removing the old image, so you end up with stuff like this:

I think tmux is partly to blame for this, because it seems to scroll sixel images by repainting the whole screen, instead of just scrolling them the way you would scroll text (there may be a good reason for that though). But it's Konsole's fault that the old image isn't erased when you write over it.

nicm:
OK well I don't think there is any point in doing any work to support Konsole until this shown to be fixed.

I don't know why we do a full redraw on scrolling, probably something to do with panes, or maybe just because it was easier.

PR #4869, opened 2026-02-19
Add session grid with mirror panes, GNUmakefile bootstrap, and clean build layout

jordanhubbard:
## Summary

- **Session grid view** (`C-b G` / `show-session-grid`): Displays all sessions simultaneously in a 2D tiled grid. Each grid cell is a real "mirror pane" that shows a live preview of a target session's active pane and forwards keyboard input to it. Standard tmux pane navigation (`C-b arrow`, `C-b o`, etc.) works natively within the grid. `C-b (` and `C-b )` cycle between grid panes; `C-b Enter` (`select-grid-session`) switches to the mirrored session; `C-b G` toggles back. The grid auto-updates when sessions are created or destroyed. The `-G` command-line flag starts tmux directly in grid mode.

- **GNUmakefile bootstrap wrapper**: A `GNUmakefile` that auto-detects the OS, installs missing build dependencies (Homebrew on macOS, lists packages on Linux/BSD), runs `autogen.sh` and `configure`, and builds tmux — all from a single `make` invocation on a fresh git clone. Supports macOS, Linux, FreeBSD, OpenBSD.

- **Clean build layout**: The GNUmakefile performs a VPATH (out-of-tree) build so the source tree stays clean. Object files go into `obj/`, the final binary into `bin/tmux`. `make clean` removes build products; `make distclean` removes both directories entirely.

### New commands
- `show-session-grid` (alias `gridview`) — toggle the grid view
- `select-grid-session` (alias `gridsel`) — switch to the session in the active mirror pane

### New format variables
- `pane_mirror` — `1` if the pane is a mirror pane, `0` otherwise
- `pane_mirror_session_name` — name of the session a mirror pane targets

### Key design decisions
- Mirror panes are real panes in the layout system (no PTY, `PANE_MIRROR` flag) whose screen content is refreshed from the target session via `screen_write_preview()` on a 1-second timer
- The grid session is marked with `SESSION_GRID` and excluded from `switch-client -n/-p` cycling
- `switch-client -n/-p` cycles panes instead of sessions when the client is in the grid session

## Test plan
- [ ] `make` on a fresh clone bootstraps and builds cleanly, objects in `obj/`, binary in `bin/tmux`
- [ ] `make clean` / `make distclean` work correctly
- [ ] Start multiple sessions, press `C-b G` to enter grid view — all sessions appear as tiled panes with border labels
- [ ] Navigate with `C-b arrow`, `C-b o`, `C-b (`, `C-b )` — moves between grid panes
- [ ] `C-b Enter` switches to the selected session
- [ ] `C-b G` again toggles back to previous session
- [ ] Type in a grid pane — input reaches the real session
- [ ] Create/destroy sessions while grid is active — grid rebuilds automatically
- [ ] `tmux -G` starts directly in grid mode
- [ ] Grid session does not appear in `C-b s` (choose-tree) cycling with `C-b (`/`)`

Made with [Cursor](https://cursor.com)


nicm:
Thanks a few comments:

- What is the use case for this mode?

- Please remove the GNUmakefile change, we do not need this. If you want it you will need to maintain it privately.

- How to use features does not go in `README.md` or `README`, only in  `tmux.1`, so please remove those changes.

- Adding  `tmux -G` flag is not necessary when there is a command can already be used for the same purpose, so please remove that also.

- Two new commands is excessive for the same feature. Roll them into one with flags please, or into existing commands if it makes sense.

- You cannot copyright something as "tmux contributors". Use your name please.

- Since you wrote this with AI and it is large enough to be copyrightable, I will need to see a statement on the publisher's website saying they do not assert any copyright over generated code, so that you are able to ISC license it.

Thanks!

nicm:
I feel like this is something might be better done as a mode rather than as actual panes. Also maybe there is no reason to limit it to sessions, a window grid might also be nice.

jordanhubbard:
1. OK, I can separate out the GNU Makefile - the reason I added it in the first place was to simply aid the testing process, because on any new OS install (mac or Linux) the auto* tools are not installed and bootstrapping it becomes "an exercise for the reader" so I figured there was no point in putting a new user through that UX given that there was already no Makefile at all.

2. The -G flag can and will be removed.  It was just a convenience to launch into this mode from startup.

3. The use case is to see all windows at once without having to manually create each pane - I have as many as 10 sessions open at once and I wanted to simply create an instant "tiled view" which replicated a window manager's tiled layout when that mode is selected.

Legitimately no concern or insult taken if this is not useful to anyone else - I can just keep it in my fork if there is no upstream interest.  The copyright line is a typo - it's 90% vibe coded and not really *anyone*'s code, least of all mine.  It was just coded from a spec.  This is likely to be the new norm.

I say 90% because the other issues were found by clang's static analyzer and a simple pass through the code looking for best practices.

jordanhubbard:
OK, all suggestions made.  Review and accept or close with comments.  I achieved what I set out to do. :)

nicm:
OK I see where you are coming from on `GNUmakefile` but I don't think it is worth maintaining it just to allow truly new users (those few who want to use `master` instead of a release) to figure out they need to run `gmake` or `make` than to figure out they have to run `sh autogen.sh; ./configure; make`. Most new Linux installs don't have `make` or `gcc` installed either, so no matter what we do there will be a bit of learning to be done. autotools is perhaps not as common as it once was but it is still not unusual, and everything is documented here: https://github.com/tmux/tmux/wiki/Installing (I thought that link was in `README` but I notice now it is not).

This looks good and works well. I do think it is a useful feature. Thanks for making those changes.

It would be nice if it showed the `1`, `2` etc like `display-panes`, so you can quickly go a session, instead of having to navigate to it. I think it'd be OK to only do the first 10, or perhaps the 8 or 10 nearest the selected session or window, so it moves as you navigate?

I'm not wild about the command name, I think it'd be better as either `display-grid` or `grid-mode` rather than a `show` command which are mostly about listing things like like options, messages, etc.

It should probably automatically zoom when you enter and leave the mode, like the choose modes do.

It's also be nice if had a `-O` flag and `O` key and use the sorting code so you can change the order of the sessions or windows.

However, I really won't be able to take this without seeing something saying that the owner of the AI you used releases the copyright. I think we can get away with that for the small fixes which are probably not copyrightable, but `window-grid.c` is over 500 lines. I don't see anything obvious on the Cursor website but they should have something, so if you could dig it out it would be great. If you can't or don't want to do that, that's OK, we'll have to leave this aside for the moment.

I'll take a look at the smaller changes later, although most look obvious enough.

Thanks!

jordanhubbard:
OK, I can add the other simplifications, but just to clarify one thing - I am the owner of the AI.  :). There is no 3rd party involved here.  I used only OSS tools and OSS models to do this.  There is no 3rd party involved to sign anything.

That's why I noted that clang static analyzer and address sanitizer were responsible for about 10% of the commits on their own.  The LLM I run on my machine is just one tool out of many that I can use to improve software, add features, scan for security holes or bad practices, etc and so forth.  An LLM is just a tool.  Sometimes the trainers of the tool like to be a little mischievous and add themselves to the list of credits but those can be removed again - it's mostly a "Kilroy was here" kind of thing, it has no legal weight.

nicm:
OK, thanks. I don't think that AI tools have been around long enough that we can he sure that is the case, so if the code was actually generated by AI rather than AI used for example for analysis or for interpretation, I need to see a statement that the authors of the AI program don't assert any copyright over it. If you trained the LLM yourself or are the owner of contour AI then your statement is here is enough.

jordanhubbard:
By the way, I looked through your commit logs and see that there are already AI contributions that have been merged, so in a sense you've already crossed the rubicon!

nicm:
Yes, unlike some projects I'm fine with taking AI contributions, I just think we should be confident (as far as we can) that the person contributing the code owns it and can license it acceptably. If they did not write it, and in fact it was written by an AI model, then that is unclear, but it seems that a minimum, practical approach is to at least check that the person or company who trained the model are releasing the copyright, which is an indication that *they* are happy the inputs to the model were unencumbered, even though perhaps you and I can't really know.

nicm:
If you are telling me you wrote it and AI was used only to assist and did not generate substantial amounts of code directly then I'm happy enough with that :-).

I'm more concerned about cases where the entire diff is written by AI and then slightly tweaked.

PR #4871, opened 2026-02-20
Fix broken fuzzing harness

DavidKorczynski:
This is to fix the current breaking OSS-Fuzz build: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#tmux


nicm:
Applied, thanks!

Issue #3488, opened 2023-03-05
Scrolling in read-only mode?


Andlon:
Hi. I am using `tmux` as replacement for `screen` because it supports read-only sessions, and I'm really only using the most basic feature set (detach and attach), so forgive me if my question does not make sense in a more general context.

I have the problem that if I attach to a session with "read only" mode, I cannot seem to scroll at all. [Following this Stackoverflow post](https://unix.stackexchange.com/questions/435742/tmux-scroll-in-readonly), if I try `Ctrl + B, [`, I only get the message that `Client is read-only`.  So it appears as if you cannot scroll in read only mode. Is this intended behavior? Perhaps my understanding of what "read only" means is off.

I've observed this behavior on tmux 3.1c (Debian 11) and 3.2a (Ubuntu 22.04). I have not yet tested the latest version, but it's not mentioned in the change log, and I suspect I'm not running into a bug, but rather a problem of my understanding. `$TERM` is `xterm-256color` outside of `tmux` and `screen` inside `tmux`.

Is there a way to obtain the behavior I want? I want to have something like a read-only session, in the sense that I cannot accidentally write to `stdin` or press `Ctrl + C` or similar, but still be able to scroll the output? I'm sorry if this is covered somewhere, but all sources I found so far just referred back to the `Ctrl + B, [` shortcut that I mentioned up above.

nicm:
`copy-mode` is not allowed in read only mode.

It could be allowed but checks would have to be added to make it read only (disable the copy commands).

nicm:
Added to todo list.

afh:
I'd like to understand more about this feature request, @Andlon, could you elaborate a bit?

In my tests where I simply added `CMD_READONLY` to the `.flags` of `struct cmd_entry` for the copy or list-keys command for a quick test, yet entering copy mode from the read-only client did also enter copy mode for the session observed.

In my mind copy mode should be "local", i.e. only affecting the read-only client, but not the session it observes.

@Andlon is this also how you envision the feature?

@nicm is a "local" copy mode easily possible in tmux currently or does it go against (core) concepts?

nicm:
No, it is not possible. Just make copy mode  allowed for read only clients but don't let them copy. An option `allow-read-only-copy-mode` would probably be appropriate (you will need to add a member to `struct cmd_entry` with the option to check for read only commands).

afh:
[...]

PR #4881, opened 2026-02-23
feat(copy-mode): Enabled read-only copy mode restricted to navigation commands.

daneofmanythings:
This is an implementation of the feature request from issue [#3488](https://github.com/tmux/tmux/issues/3488).

I also tried to enable the search commands, but ran into an issue and as such is not part of the PR. It may be something of interest, so I will detail what I did and the issue. 

I can enable `command-prompt` to work in the new `CMD_READONLY_COPYMODE` with the current technique as well as adding logic on `server-client.c:2673` with something like:
```c
...
	if (~c->flags & CLIENT_READONLY ||
	    options_get_number(global_options, "allow-read-only-copy-mode")
	    == 1) {
...
```
This enables the copy mode commands that use the command prompt to function, but it also allows for arbitrary execution of other commands. I found some logic that I would think prevents this in `server_client_dispatch_command` in `server-client.c:3566`, but it doesn't seem to in this case. I do not understand how commands are filtered by `CMD_READONLY` outside of `key_bindings_dispatch` in `key-bindings.c:679`, and why they may be filtered in more than one place. I'd be happy to put more work into this aspect of the feature with a nudge in the right direction if desired.

Let me know what you think, thanks!

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -36,7 +36,7 @@ const struct cmd_entry cmd_copy_mode_entry = {
 	.source =  { 's', CMD_FIND_PANE, 0 },
 	.target = { 't', CMD_FIND_PANE, 0 },
 
-	.flags = CMD_AFTERHOOK,
+	.flags = CMD_AFTERHOOK|CMD_READONLY,
 	.exec = cmd_copy_mode_exec
 };
 
--- a/cmd-send-keys.c
+++ b/cmd-send-keys.c
@@ -39,7 +39,8 @@ const struct cmd_entry cmd_send_keys_entry = {
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
-	.flags = CMD_AFTERHOOK|CMD_CLIENT_CFLAG|CMD_CLIENT_CANFAIL,
+	.flags = CMD_AFTERHOOK|CMD_CLIENT_CFLAG|CMD_CLIENT_CANFAIL|
+		 CMD_READONLY,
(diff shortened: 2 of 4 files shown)


nicm:
Thanks. I don't think we need a new option for this, so long as the copy commands are turned off, the rest can be permitted as part of the existing read only flag.

daneofmanythings:
Good note, I have made the change! Entering and navigating in copy mode is now permitted from a read-only client as baseline behavior.

nicm:
This looks good but it does not apply for me:
~~~~
$ git apply ./4881.diff
error: patch failed: window-copy.c:2722
error: window-copy.c: patch does not apply
$ patch -sp1 -l <4881.diff
1 out of 2 hunks failed--saving rejects to window-copy.c.rej
~~~~

daneofmanythings:
Oh shoot. I rebased to the current master HEAD and squashed the commits. Let me know if it works now!

nicm:
This looks good, but do we think `send-keys` is OK without `-X` (such as `send-keys xxx` or `send-keys -R`) for readonly clients?

nicm:
I guess maybe it is - how would they trigger the command without `command-prompt`?

daneofmanythings:
Total oversight on my part; the vulnerability can be triggered with a key bind. It was not my intention to allow such behavior. I have added a guard at the top of `send-keys` to check for the client's read-only status and error out if `-X` is not present.

nicm:
Looks good, applied to OpenBSD now.

Thanks!

PR #4884, opened 2026-02-23
Add scroll-exit-on, scroll-exit-off, and scroll-exit-toggle copy_cmd

xcdnlgd:
This allows me to temporarily disable scroll-exit when I'm dragging to select text so that tmux won't accidentally quit copy-mode when I'm also scrolling down at the same time.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -2230,6 +2230,18 @@ Toggle rectangle selection mode.
 .Xc
 Refresh the content from the pane.
 .It Xo
+.Ic scroll-exit-on
+.Xc
+Turn on exiting copy mode when scrolling to the end of the buffer.
+.It Xo
+.Ic scroll-exit-off
+.Xc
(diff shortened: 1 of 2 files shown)


nicm:
Thanks... If you added a `scroll_exit` format to `window_copy_formats`, you would only need `scroll-exit-toggle`?

xcdnlgd:
I'm mimicking the existing `rectangle-on/off/toggle`.

And I actually don't use scroll-exit-toggle in my `\.tmux.conf`

```
bind -T copy-mode-vi MouseDrag1Pane send -X begin-selection \; send -X scroll-exit-off
bind -T copy-mode-vi MouseUp1Pane send -X clear-selection \; send -X cursor-move \; send -X scroll-exit-on
```

I think keep the `on/off` can simplify the config file

nicm:
OK, applied to OpenBSD now, will be in GitHub later. Thanks!

PR #4885, opened 2026-02-23
feat(copy-mode): Added presistent cursor scroll state through refreshing in copy mode.

daneofmanythings:
Added the feature from this point of the todo list:
```
In copy mode, refresh (r) will update the content but not move the scroll position, 
so the text appears to jump. It would be better to also move the position so there 
was no apparent movement.
```

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -2714,11 +2714,10 @@ window_copy_cmd_refresh_from_pane(struct window_copy_cmd_state *cs)
 	data->backing = window_copy_clone_screen(&wp->base, &data->screen, NULL,
 	    NULL, wme->swp != wme->wp);
 
+	data->oy = screen_hsize(data->backing) - oy_from_top;
 	if (data->oy > screen_hsize(data->backing)) {
 		data->cy = 0;
 		data->oy = screen_hsize(data->backing);
-	} else {
-		data->oy = screen_hsize(data->backing) - oy_from_top;
 	}
 
 	window_copy_size_changed(wme);
--- a/window-copy.c
+++ b/window-copy.c
@@ -2702,9 +2702,12 @@ window_copy_cmd_refresh_from_pane(struct window_copy_cmd_state *cs)
 	struct window_mode_entry	*wme = cs->wme;
 	struct window_pane		*wp = wme->swp;
 	struct window_copy_mode_data	*data = wme->data;
+	u_int				 oy_from_top;
 
 	if (data->viewmode)
 		return (WINDOW_COPY_CMD_NOTHING);
+
+	oy_from_top = screen_hsize(data->backing) - data->oy;
 
 	screen_free(data->backing);
 	free(data->backing);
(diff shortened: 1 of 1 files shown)


nicm:
I think you need a check that `oy_from_top <= screen_hsize(data->backing)` since the latter could have changed?

daneofmanythings:
Careless oversight on my part! I believe this flow should work. I don't know why I had an `else` in the overflow catch.

I'm unsure if overflow is possible when `oy_from_top` is assigned, so the second commit is there in case that state can happen, and can be ignored otherwise.

nicm:
Your change is not quite right. You are checking `oy_from_top > screen_hsize(data->backing)` before calling `window_copy_clone_screen` but at that point the check cannot be true, because you just set `oy_from_top`.

The check needs to go *after* calling `window_copy_clone_screen` which may have changed the history size. I've applied it but changed it to this:
~~~~
@@ -2732,15 +2732,20 @@ window_copy_cmd_refresh_from_pane(struct
        struct window_mode_entry        *wme = cs->wme;
        struct window_pane              *wp = wme->swp;
        struct window_copy_mode_data    *data = wme->data;
+       u_int                            oy_from_top;

        if (data->viewmode)
                return (WINDOW_COPY_CMD_NOTHING);
+       oy_from_top = screen_hsize(data->backing) - data->oy;

        screen_free(data->backing);
        free(data->backing);
        data->backing = window_copy_clone_screen(&wp->base, &data->screen, NULL,
            NULL, wme->swp != wme->wp);

+       if (oy_from_top > screen_hsize(data->backing))
+               oy_from_top = 0;
+       data->oy = screen_hsize(data->backing) - oy_from_top;
        if (data->oy > screen_hsize(data->backing)) {
                data->cy = 0;
                data->oy = screen_hsize(data->backing);
~~~~
Thanks!

nicm:
Actually my change is slightly wrong too still, I changed it to this instead:
~~~~
@@ -2743,10 +2743,9 @@ window_copy_cmd_refresh_from_pane(struct
        data->backing = window_copy_clone_screen(&wp->base, &data->screen, NULL,
            NULL, wme->swp != wme->wp);

-       if (oy_from_top > screen_hsize(data->backing))
-               oy_from_top = 0;
-       data->oy = screen_hsize(data->backing) - oy_from_top;
-       if (data->oy > screen_hsize(data->backing)) {
+       if (oy_from_top <= screen_hsize(data->backing))
+               data->oy = screen_hsize(data->backing) - oy_from_top;
+       else {
                data->cy = 0;
                data->oy = screen_hsize(data->backing);
        }

daneofmanythings:
Thanks for troubleshooting that. It's more fiddly than I would have thought. Out of curiosity, did you notice the error you made through a test, or from looking at it again and realizing it wasn't quite right?

nicm:
I just noticed when I looked at it again.

PR #4887, opened 2026-02-24
Add DECRPM response for mode 2026 (synchronised output)

dsturnbull:
Programs that use synchronised output (e.g. to avoid tearing during redraws) may send DECRQM to probe whether the terminal actually supports mode 2026 before relying on it.  Without a DECRPM reply, the query times out or falls back to a default assumption, which can lead to unnecessary flicker or missed optimisation.

The `INPUT_CSI_QUERY_PRIVATE` handler in `input.c` responds to DECRQM (`CSI ? Ps $ p`) for several private modes but not for mode 2026 (synchronised output). tmux already handles the set/reset side of mode 2026 (`SM`/`RM` → `screen_write_start_sync` / `screen_write_stop_sync`), so the missing query response appears to be an oversight.

### Change

Add a `case 2026` to the DECRQM switch that replies with:

    ESC [ ? 2026 ; Ps $ y

where Ps = 1 if `MODE_SYNC` is active, Ps = 2 if reset.

This follows the same pattern used by every other mode in the block.

### Test

`regress/decrqm-sync.sh` exercises both cases end-to-end:

1. **Default (reset)** — queries mode 2026 immediately after session creation, expects Ps = 2.
2. **After SM** — sends `CSI ? 2026 h` to enable synchronised output, then queries, expects Ps = 1.

### DIFF
--- a/input.c
+++ b/input.c
@@ -1642,6 +1642,10 @@ input_csi_dispatch(struct input_ctx *ictx)
 		case 2031:
 			input_reply(ictx, 1, "\033[?2031;2$y");
 			break;
+		case 2026: /* synchronized output */
+			n = (s->mode & MODE_SYNC) ? 1 : 2;
+			input_reply(ictx, 1, "\033[?2026;%d$y", n);
+			break;
 		}
 		break;
 	case INPUT_CSI_DSR:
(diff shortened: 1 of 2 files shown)


nicm:
Applied now, thanks!

PR #4889, opened 2026-02-25
fix: added `enum sort_order` validation to -O flag locations

daneofmanythings:
This PR is waiting for [#4845](https://github.com/tmux/tmux/pull/4845) to hit the master branch for the last change.


daneofmanythings:
Okay, this is ready for review. I added validation checks for `enum sort_order` to the commands that implement the -O flag.

nicm:

PR #4892, opened 2026-02-26
Fix memory leak in file.c (#4877)

### LINKED ISSUE #4877: Memory leak in `file.c` (tmux 3.6a)
### Issue description

`file_get_path()` allocates `path` but does not free it when the input is a relative path.

The function allocates `path` (lines **47–48**):

    if (strncmp(file, "~/", 2) != 0)
        path = xstrdup(file);

If `path` is absolute, it returns `path` directly (lines **55–56**):

    if (*path == '/')
        return (path);

However, if `path` is relative, it allocates `full_path` and returns it without freeing `path` (lines **57–58**):

    xasprintf(&full_path, "%s/%s", server_client_get_cwd(c, NULL), path);
    return (full_path);

Since `path` is heap-allocated and no longer needed after constructing `full_path`, it should be freed before returning on this branch.

### Required information
tmux version: tmux 3.6a

Platform: Linux x86_64

Chris-LewisI:
Fixes #4877

This PR fixes a memory leak in file.c where "path" 
was not being freed when "full_path" was being leveraged.

Changes:
- Added free() call in file_get_path to properly free allocated memory

### DIFF
--- a/file.c
+++ b/file.c
@@ -55,6 +55,7 @@ file_get_path(struct client *c, const char *file)
 	if (*path == '/')
 		return (path);
 	xasprintf(&full_path, "%s/%s", server_client_get_cwd(c, NULL), path);
+	free(path);
 	return (full_path);
 }
 


nicm:

Issue #4800, opened 2026-01-09
Floating panes issues


mgrant0:
_[Feel free to try this branch and report problems here but note this feature is development and is not ready for use]_

Issue to track the dev issues in the [floating_panes](https://github.com/tmux/tmux/tree/floating_panes) branch.

1. ~Manage padding in tty.c~
2.  ~Floating panes should be able to be "minimised" or "lowered" to make them invisible without closing them.~
3.  Scrollbar slider/offset calculation.  It seems that the height calculation of the scrollbar is based on the window size and not the pane size and thus when it calculates pages per scollbar scroll, it is not fully scrolling to the bottom of the screen when the scrollbar slider is at the bottom (in copy-mode).
4. ~Address/remove all xxxx flagged comments.~
5. ~Fix crashing issue when opening an overlay popup with floating pane.~
6. Start tmux with empty desktop or so you can start it with only floating panes from .tmux.conf
7. A way to tile a floating pane and float a tiled pane. Get join-pane and break-pane to work with floating panes. Should not be difficult.  Code already supports tiling and floating panes, just need to wire things up.
8. Add some buttons to the border: - ^ x to minimise, zoom, and close a pane.
9. ~Modify status line 2 to be the name of the window pane and not width x height.~
10. ~There needs to be a format variable if a pane is floating so that formats can distringuish them.~
11. ~Test/verify that the window offsets work properly, i.e. setting the window to larger than the tty size.Create sub-issues under this issue for related issues.~
12. ~Add -s border-style and -S style args to the new-pane command.  Probably also need a default style option variable for these.~
13. ~Fix the sixel code which is currently broken with respect to visible ranges.~ Fixed it for floating panes but sixel images still need a lot of work in general, not just for floating panes.  They need to be integrated into the screen grid struct.
14. ~Add an option to new-pane to wait before closing after cmd exits.~
15. ~Update tmux man page.~
16. ~cmd-minimise-pane has a -a option to minimise all panes, this should unminimise all panes if they were just minimised.  There needs to be some way to minimise/unminimise panes via the keyboard one by one somehow, maybe cycling through minimised pains?~ `PREFIX _` minimises a pane, `PREFIX w`, use `choose-tree` to unminimise window.  single-cicking pane in status line minimises/unminimises the pane.

wyvernbw:
[...]

PR #4893, opened 2026-02-27
feat(format variable): added format variable for floating panes.

daneofmanythings:
I implemented `10` from the [floating panes issues](https://github.com/tmux/tmux/issues/4800).
```
There needs to be a format variable if a pane is floating so that formats can distinguish them.
```
Let me know how it looks!

### DIFF
--- a/format.c
+++ b/format.c
@@ -1004,6 +1004,20 @@ format_cb_pane_fg(struct format_tree *ft)
 	return (xstrdup(colour_tostring(gc.fg)));
 }
 
+/* Callback for pane_floating_flag. */
+static void *
+format_cb_pane_floating_flag(struct format_tree *ft)
+{
+	struct window_pane	*wp = ft->wp;
+
+	if (wp != NULL) {
+		if (wp->flags & PANE_FLOATING)
+			return (xstrdup("1"));
+		return (xstrdup("0"));
+	}
+	return (NULL);
+}
+
 /* Callback for pane_bg. */
 static void *
 format_cb_pane_bg(struct format_tree *ft)
@@ -3283,6 +3297,9 @@ static const struct format_table_entry format_table[] = {
 	},
 	{ "pane_fg", FORMAT_TABLE_STRING,
 	  format_cb_pane_fg
+	},
+	{ "pane_floating_flag", FORMAT_TABLE_STRING,
+	  format_cb_pane_floating_flag
 	},
 	{ "pane_format", FORMAT_TABLE_STRING,
 	  format_cb_pane_format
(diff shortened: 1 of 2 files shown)


nicm:
@mgrant0 --^ :-)

mgrant0:
Cool thanks will take a look at this.  It was on my todo list, thanks for doing this!

daneofmanythings:
@mgrant0 Thanks! Let me know if you'd like me to try and tackle more of what is on the floating panes issue list, and, if so, what the best way to contribute to the branch is.

mgrant0:
I have been tracking some open issues in this ticket: https://github.com/tmux/tmux/issues/4800

Several of the things on that list I have in my head what to do, just need to get to them.  There are definitely a few that i can spin off easily to you if you want.

Can you hit me on irc (libera)?  We can discuss.

mgrant0:
I would rename your flag to `pane_floating_flag` since it is a boolean flag.

daneofmanythings:
> I have been tracking some open issues in this ticket: #4800
> 
> Several of the things on that list I have in my head what to do, just need to get to them. There are definitely a few that i can spin off easily to you if you want.
> 
> Can you hit me on irc (libera)? We can discuss.

Your issue tracking is where I found that the format variable needed implementation! I've not used irc before, but I just got it set up and registered on libera. Do we simply exchange nicks now?



> I would rename your flag to `pane_floating_flag` since it is a boolean flag.

I do think you are right that the naming was incomplete, but the suffix doesn't fit the established convention. There are many other format keys implemented which are boolean flags and none of them use the `_flag` suffix. Closest I could find to the name is `cursor_blinking                 1 if the cursor is blinking`, so it feels that `pane_floating` is the right place to be. The most recent commit contains that adjustment.

mgrant0:
I'm ok with `pane_floating`.  Another possible direction is `pane_layout` which would return tiled or floating.  Look in the man page, there're definitely format variables with _flag that return 0/1.  Maybe nic has a strong opinion.  

My nic on libera is mgrant.  Or we can add comments to the ticket which is public.  Whichever you prefer is fine.

daneofmanythings:
Oh, you're totally right. I should have searched `_flag`. It is unclear to me why only some boolean format keys contain that suffix. The only one i can find that is avoiding a naming conflict with the suffix is `session_activity` and `session_activity_flag`. There are also commands which return actual flag/s ending with the `_flag` suffix, not 0/1. My preferred convention for boolean naming is the form `pane_is_floating`, but introducing a new paradigm may not make things more clear. I would like to hear what @nicm thinks too.

I like the idea of `pane_layout` though I worry that it may make formats more verbose than they need to be at the moment. I'm not very knowledgeable about formats yet, but I'm envisioning `#{?pane_floating, ... }` vs `#{?#{==:#{pane_layout},floating}, ... }`. 

I'll send you a message on irc about the rest of the todo list to see if I may be helpful in completing some more tasks!

nicm:
I would stick `_flag` on the end personally but it doesn't really matter.

daneofmanythings:
Name has been fixed.

mgrant0 (approved):

mgrant0:
@daneofmanythings merged, thanks!

PR #4895, opened 2026-02-28
fix(docs): misplaced '.It Xo' macro in tmux.1

daneofmanythings:


nicm:

Issue #4874, opened 2026-02-21
Memory leak in `format.c` (tmux 3.6a)


huanghuihui0904:
### Issue description

`format_loop_sessions()` allocates `all` (and possibly `active`) but does not free them before returning.

When `format_choose(es, fmt, &all, &active, 0) != 0`, the function explicitly allocates `all` (lines **4416–4419** in current master; **4430–4433** in tmux 3.6a):

    if (format_choose(es, fmt, &all, &active, 0) != 0) {
		all = xstrdup(fmt);
		active = NULL;
	}

However, `all` (and `active`, when non-NULL) are not freed on any return path. The function returns `value` directly (line **4450**  in current master: line **4472** in tmux 3.6a):

    return (value);

As a result, `all` is always heap-allocated but never freed before returning. When `format_choose()` succeeds, `active` is also heap-allocated and not freed. 

A possible fix is to free `all` and `active` before returning, similar to how `format_loop_windows()` handles its corresponding `all`/`active` strings.

### Required information
tmux version: tmux 3.6a

Platform: Linux x86_64

PR #4898, opened 2026-02-28
Fix: memory leak in format_loop_sessions() in format.c (#4874)

huanghuihui0904:
Free allocated memory for `active` and `all` variables. Refer to how `format_loop_windows()` handles its corresponding `all`/`active` strings.


nicm:

Issue #4878, opened 2026-02-21
Memory leak in `menu.c` (tmux 3.6a)


huanghuihui0904:
### Issue description

`menu_add_item()` allocates `s` via `format_single()`/`format_single_from_state()` but does not free it when the formatted item name expands to an empty string.

The string is allocated here (lines **88–91** in the current master branch; **83–86** in the tmux 3.6a version):

    if (fs != NULL)
		s = format_single_from_state(qitem, item->name, c, fs);
	else
		s = format_single(qitem, item->name, c, NULL, NULL, NULL);

If the expanded string is empty, the function returns early without freeing `s` (lines **92–95** in the current master branch; **87–90** in the tmux 3.6a version):

    if (*s == '\0') { /* no item if empty after format expanded */
		menu->count--;
		return;
	}

Since `s` is heap-allocated and ownership remains within this function, it should be freed before returning on this path (for example, `free(s);` before `menu->count--`).

### Required information
tmux version: tmux 3.6a

Platform: Linux x86_64

kantnero:
I'd like to work on this if it's not already assigned

nicm:
Go for it just make a PR or attach a diff to this issue.

nicm:
Thanks both!

PR #4899, opened 2026-03-01
memory fix leak in menu.c issue #4878

kantnero:
Free allocated memory for s after function early return
link to issue description 
[#4878](https://github.com/tmux/tmux/issues/4878)


nicm:

PR #4900, opened 2026-03-01
Memory leak in file.c (tmux 3.6a) #4877

kantnero:
#4877 
Free path when full_path is return from file_get_path(struct client *c, const char *file)


nicm:

Issue #3407, opened 2022-12-05
Feature Proposal: Allow send-keys to read keys to send from a file


nmeum:
I am occasionally (ab)using tmux's `capture-pane` and `send-keys` commands for unit testing of interactive shell programs (e.g. REPLs with readline support). For this purpose, it would be useful if it would be possible to have `send-keys` read keys that should be sent to a pane from a file. Otherwise, one has to deal with shell word-splitting etc. and that is very annoying. A useful format might be having each line, in the file read by `send-keys`, terminated by an `Enter` unless it is escaped with a backslash. So for example, an input file of the form:

```
foo bar
C-d \
baz
```

Would cause the following `send-keys` invocation: `tmux send-keys <…> "foo bar" Enter C-d baz Enter`.

Is there an upstream interest in this feature? If so I could look into creating a patch for it, if not just close this issue.

nicm:
I don't think this is worth the complexity when this should be easy to script. If sh won't do it nicely, use python or perl or one of the million other scripting languages.

PR #4903, opened 2026-03-01
feat(target flag): -t expands formats

daneofmanythings:
This is an implementation of the todo list item:
```Expand targets as formats. Should this be done by default or require a prefix (maybe a leading + or just look for #{)?```
which is feature request from issue #3407. The code treats targets as formats, though this behavior can easily be adjusted to look for a prefix like '+'. The new function `args_get_target` has the same behavior as `args_get` and is a drop-in replacement for calls to `args_get(args, 't')`.

One part that I'm not sure about is in `cmdq_find_flag` where I expand both targets and sources to keep the code footprint light. I believe that this would amount to a no-op on sources, but I am unsure of potential side effects. If there are unwanted side effects, I would either add if checks or split it into two functions.

Am I on the right track?

### DIFF
--- a/arguments.c
+++ b/arguments.c
@@ -683,6 +683,32 @@ args_get(struct args *args, u_char flag)
 	return (TAILQ_LAST(&entry->values, args_values)->string);
 }
 
+/* Get arguement target value. Will be NULL if it isn't present */
+const char *
+args_get_target(struct args *args, struct cmdq_item *item)
+{
+	char			*target;
+	u_int			 len;
+	static char		*t = NULL;
+	static u_int		 tsz = 0;
+
+	target = format_single_from_target(item, args_get(args, 't'));
+	if (*target == '\0') {
+		free(target);
+		return (NULL);
+	}
+
+	len = strlen(target);
+	if (tsz <= len) {
+		tsz = len + 100;
+		t = xreallocarray(t, sizeof *t, tsz);
+	}
+	memcpy(t, target, len);
+	t[len] = '\0';
+	free(target);
+	return (t);
+}
+
 /* Get first argument. */
 u_char
 args_first(struct args *args, struct args_entry **entry)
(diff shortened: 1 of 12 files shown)


nicm:
I do think a prefix might be better or we are going to risk breaking configs? What do you think?

daneofmanythings:
Oh, of course. I wasn't sure what the motivation behind a prefix was. You had mentioned in the todo list that '+' was a possible prefix, is this still the case, so we would then be looking at target formats being of the form `-t+"#{...}"`?

nicm:
`+` is maybe not a good choice since it is already used for window offsets, although that is maybe not very common. There are not many special characters left though, so not sure what to use.

Maybe we just bite the bullet and expand by default. How many people are going to have `#{` in their targets at the moment?

Bit of a pity we couldn't use `-T`, but it is already used by a few commands.

daneofmanythings:
I initially did not consider that format expansion may break some configs probably because I couldn't imagine a use case for a target looking like a format expansion. I would also estimate that the risk is very low. 

I am a bit smitten with the idea of -T being the flag for target format expansion, and have compiled a list (the most recent commit) of the current commands that contain a -T flag which would need to be reworked. Some of them seem very low risk, while others WILL  break a LOT of configs, those being key binding (key-table specification) and menu display (titles) commands. I really like the idea, but I don't think it is worth it. What is tmux's history with config breaking releases? Take a look at the list and see what you think.

If we were to do a prefix, I would suggest '#' because it is on theme, so a target with expansion would look like `-t#"#{...}"`. Makes the most sense to me. 

I find it so unlikely that configs would break that I am comfortable shipping default expansion, though I wouldn't say my intuition is calibrated yet.

nicm:
People hate it when we break configs and it will cause a lot of confusion and complaining. Realistically we cannot change `-T` for `bind-key`.

TBH I would be inclined just to do it by default. How many people are really using `-t` with `#{}` in it?

daneofmanythings:
That works for me! Ready for review then.

nicm:
Thanks, looks good, I'm probably going to wait for a while see how I like the idea of doing it by default in a few more days...

PR #4906, opened 2026-03-02
Make -c start-dir work with `new-session -A`

jody-frankowski:
I've noticed that `new-session` when combined with its attach capability `-A` didn't take into account the `-c start-directory` flag.


nicm:

PR #4912, opened 2026-03-05
Add kitty keyboard protocol support

sundbp:
Implement the kitty progressive keyboard enhancement protocol, allowing tmux to negotiate richer key encoding with supporting terminals (ghostty, kitty, foot, etc.) and forward it to applications that opt in.

New server option 'kitty-keys' (off/on/always):
- off: disabled (default)
- on: negotiate kitty protocol with the outer terminal via CSI ? u auto-detection; inner applications receive kitty encoding only when they explicitly push CSI > N u
- always: force kitty disambiguate mode on all inner applications, similar to extended-keys always

Protocol support:
- Input: full CSI u / CSI ~ / CSI letter parser for kitty-encoded keys, including modifier colon subfields and event types (release events are discarded)
- Output: kitty CSI u encoder with reverse mapping tables for functional keys and legacy arrow/F-key sequences
- Inner application negotiation: CSI > N u (push), CSI < N u (pop), CSI ? u (query), CSI = N u (set) handled via input.c state machine with flag stack and alternate screen save/restore
- Auto-detection: CSI ? u query sent to outer terminal; response auto-adds the kitkeys terminal feature and activates the parser

Disambiguate mode (flag 1) follows the kitty spec for legacy fallback:
- Tab, Enter, Backspace without modifiers send raw bytes
- Ctrl+key combinations that map to C0 control codes (Ctrl+C, Ctrl+D, Ctrl+Z, etc.) send the raw C0 byte when Ctrl is the sole modifier
- Unmodified printable ASCII uses legacy encoding
- Caps Lock and Num Lock modifiers are stripped on both input and output since they are only meaningful with report-all-keys (flag 8+)

New key definitions: F13-F35, keypad keys, media keys, modifier keys, Super/Hyper/RealMeta modifier bits, Caps Lock/Num Lock modifier bits. Key string parsing and display supports Super-, Hyper-, and RealMeta- prefixes.

Terminal feature 'kitkeys' adds Enkitk/Dskitk capabilities. Foot included in default features; other terminals auto-detected via the CSI ? u query or configurable via terminal-features.


nicm:
Thanks. Will take a closer look some time soon. Only thing is I think this should be rolled into `extended-keys`/`extended-keys-format` somehow rather than adding a new option.

nicm:
A few other comments:

- `input-keys.c` is getting kind of big, so perhaps make `input-kitty.c` and move your kitty-specific functions in there?

- Please remove references to "legacy" key codes, it is unclear, use either VT10x or extended like the existing comments, depending which you want to refer to.

- I think `Super-`, `Hyper-`, `RealMeta-` are too long. I'd go with `Sp-`, `Hy-`, `Rm-`, although I don't like the name "Real Meta" much anyway so I'd probably just drop it for now, how many keyboards have both Alt and Meta?

- You can remove `(kitty keyboard protocol)` from the comments for `KEYC_` since these keys may exist in other encodings.

- Similar to above, I think this is making `tty-keys.c` too big and a new file `tty-kitty.c` would be better.

- I think "function key" is the more usual term rather than "functional key".

- 8 seems a bit excessive for `KITTY_KBD_STACK_MAX` and it is making `struct screen` a good bit bigger. TBH I would only keep two (current and saved).

It looks good otherwise, apart from a few style nits which can be easily fixed later. I didn't test it yet though.

Thanks!

sundbp:
I'm having a second run at it. I've been testing it - mostly good, found a couple of things. Incorporating these and will then attack your points before others can take a 2nd look.

sundbp:
Ok, I've had a 2nd pass at it now. I've gone over it, fixed a few small things and added some tests and so forth. I've also incorporated the feedback - apart from somehow rolling it into extended-keys / extended-keys-format. That one I don't feel very certain around an approach for. Let me know what you think @nicm 

btw, the python test script is there as I didn't manage to produce a bash version. It's feel like it should be possible but I failed at it.

nicm:
Code looks good to me, a couple of minor things:

- In the new files, `input-kitty.c` and `tty-kitty.c`, you should change the copyright line at the top of the file to your name and email address (if you don't want to include the email address, you can leave it out), and to 2026 instead of 2007.

- In `input_csi_dispatch`, variable declarations should go at the top of the file, not inside inner blocks. So that is `int flags, raw_flags, dropped, count, mode, flagset, raw_flagset;`. If the code or number of variables is getting unwieldly, use helper functions.

- Is there a rationale for making this a separate `kitty-keys` option? Why don't we just try to turn it on if we think the terminal supports it and `extended-keys` is `on`? It falls back sensibly to the standard extended keys, doesn't it?

- I don't know there is any need for `Dskitk`/`Enkitk`. This is not something different terminals are going to implement in a different way so users or terminals will need to redefine. I think we could just hardcode the escape sequences.

It's a bit of a pity they chose to occupy four modifiers to CSI u for this one feature when one would have been fine (or even extending CSI m further), but there is little to be done about that now.

I tried it lightly in ghostty, at least the normal extended keys seem to still work (such as `C-BSpace`, `C-1` etc). My keyboard doesn't really have any media keys or similar, and Super/Hyper are taken by the window manager, so I'll have to take your word for it that that sort of thing works.

ilyagr:
With this PR, if I run the following on Kitty 0.46 or Ghostty 1.3, most keys do not register in the resulting window (including normal letters). I'm using macOS Sequoia. I'm also using fish 4.5.0 (though I'm not sure where this matters).

    ./tmux -Ltest -f /dev/null set -s kitty-keys always \; set -s extended-keys always \; new

I think kitty-keys do not play well with extended-keys.

I am also getting some separate problems with keys like ^D. When I ran

    ./tmux -Ltest -f /dev/null set -s kitty-keys always  \; new dash

pressing ^D results in `^[[100;5u` being printed. 

Here is a log where I start tmux with `dash` with *both* extended-keys and kitty-keys and press `abc^d`, then have to use ^bx to quit (but the bug with ^d is present even without extended-keys):

[tmux-server-85653.log](https://github.com/user-attachments/files/25957254/tmux-server-85653.log)


According to AI:

> The interesting part starts at line **33149**. Lines 26029-33079 are terminal detection/startup. Here's the summary:
> 
> - **Line 33149-33737**: `a`, `b`, `c` received as plain bytes, written to pane, routed through `input_key_mode1` (which falls through to extended-keys encoding instead of raw bytes)
> - **Line 33747-33762**: `C-d` received as `\033[100;5u`, written to pane via `input_key_kitty` (sends CSI u to dash instead of raw `0x04`)
> - **Line 33833-33834**: `C-b` received as `\033[98;5u` (tmux prefix)
> - **Line 34079**: `x` — the quit
> 
> So lines ~33149–33762 are the key part — about 600 lines. You could trim the log to roughly lines 33100–34100 for a focused excerpt.

(but I didn't trim it :)


Here is AI's theory on what the bug with ^D is:

> Look at the key flow for `C-d`:
> 
> 1. Ghostty sends `\033[100;5u` to tmux (100='d', 5=Ctrl+1)
> 2. tmux parses this and creates key code `0x200000000064` — that's `d` with a `KEYC_CTRL`-like flag (the `0x2000000000` prefix)
> 3. In `input_key_kitty` at line 197, `mod` is computed from this key and is non-zero (Ctrl)
> 4. At line 263, `onlykey` is `0x64` ('d'), which is `<= 0x7f`, so we enter that block
> 5. At line 268, `mod != 0`, so the early return is skipped
> 6. Lines 273-275: it encodes as `\033[100;5u` and sends that to dash
> 
> **The missing logic**: in disambiguate mode (flag 1 only), `C-d` should be sent as the raw byte `0x04` (C0 control), not CSI u. The kitty spec says that Ctrl+letter combinations that produce C0 codes (0x01-0x1a) should use the legacy byte in disambiguate mode. There's no check here for "Ctrl is the only modifier and the key maps to a C0 code → send the C0 byte instead."


----


I was also sorry to discover that "release events" are discovered. My primary use-case for this was a terminal app that uses those (it's a weird one, Ki editor, it has so-called "momentary layers"). But, according to the same AI, fixing that would be a large project.

ilyagr:
Actually, I'm experiencing the bug where typing "abc" doesn't work even without turning on kitty-keys, just with this PR and

     ./tmux -vv -Ltest -f /dev/null set -s extended-keys always \; new fish

It works fine without this PR, at d32088b.

This does only become a problem with fish, though (I have fish 4.5.0), using dash here avoids the bug.

But this is a bigger problem because this breaks existing users with `extended-keys` set to `always` in their config, of which I am one (whether that's wise is a different question, I don't remember what didn't work without that setting).

ilyagr:
Oh, and thank you for working on it!

sundbp:
@nicm thanks for the input! I'll have a look at the smaller adjustments. In terms of folding into extended-keys it seemed "safer" to make it an opt-in for this specific case and being able to tell that mode and this mode apart. But I think you're right it could be combined. Up for attempting that.

@ilyagr thanks for giving it a spin. And it draws out that I just tried this in ghostty on linux (minor testing in kitty), with kitty-keys on. I have completely glossed over any manual testing without kitty-keys being on. I think kitty-keys always is generally a bad idea - you'll get key codes sent to apps that don't support them etc. First port of call is some manual testing (outside the existing regression tests) with kitty-keys off, and with both extended-keys and kitty-keys on. I put the PR forward mostly to gage interest - I'm aware it needs a lot more testing!

sundbp:
My quick test (zsh user, don't belive foot comes into this), on ghostty 1.2.3:
1. kitty-keys always and extended-keys always = gridlock, no input works
2. kitty-keys on and extended-keys always = works ok in my limited testing
3. kitty-keys off and extended-keys always = works ok in my limited testing
3. kitty-keys always and extended-keys on = works ok in my limited testing
3. kitty-keys on and extended-keys on = works ok in my limited testing

So the path that's (definitely) broken is always+always. In general testing I'm not sure it's good to ever use kitty-keys always - it'll send the wrong key codes to various applications that don't know how to handle them. I haven't taken the time to understand what extended-keys really does - but as a first step I'll fix the gridlock case.

nicm:
I'd be inclined to remove the `always` for `kitty-keys` if it doesn't make sense.

It's OK to keep `kitty-keys` for now and we can think about how/if to integrate it with the other options later.

sundbp:
This removes the conflict as an intermediate step. I think I agree that it does not make sense to allow always for kitty-keys - I can't see a case where it makes sense. I'm far from an expert here though - very much open to bigger brains with more experience in this field opining!

sundbp:
Addressed the items raised in your previous comment @nicm . Unifying kitty-keys and extended-keys is the item left out, and the one I feel a bit nervous around attempting.

I'd love for people to test this more, and for more knowlegable people to comment if it makes sense to remove kitty-keys = always. I think we need to build a bit of confidence via more wide spread testing before confirming all is well. In my daily usage I haven't found any issues, but there's a lot of corner cases out there I'm sure given the rich domain of tty interactions..

ilyagr:
You can't reproduce the issue where ^D doesn't work with `kitty-keys off and extended-keys always`? Nor the one where nothing works with fish + `kitty-keys off and extended-keys always`?

Yamagi:
> You can't reproduce the issue where ^D doesn't work with kitty-keys off and extended-keys always? Nor the one where nothing works with fish + kitty-keys off and extended-keys always?

It reproduces for me with alacritty + zsh. It's even somewhat stranger. ^D doesn't work right after tmux was started. But it starts working after Vim (which enabled the Kitty Keyboard Protocol) exited...

sundbp:
So this is why I'm thinking that it basically never makes sense to set kitty-keys = always. With this setting you will ALWAYS have the kitty protocol active, meaning that anything running that doesn't speak it, will not get the inputs it expects. In this mode, Ctrl-D works fine if the thing you run inside of of your session support the protocol, otherwise it'll get "garbage" keys. If you run with kitty-keys = on then you will only send the extended protocol keys when the running application actually supports the protocol, and if it doesn't support it the expected codes are sent. The sane path.

I strongly suggest testing with kitty-keys = on. And as I suggested before, I think it makes sense to not even support = always and I remove that from the PR. I only added it by symmetry to how the extended-keys was done, but now I think it likely doesn't make sense.

Yamagi:
Ah, okay. I set `kitty-keys = always` in the first place because Vim wasn't working with `kitty-keys = on`. So this warrants a second look.

Vim requests the Kitty Keyboard Protocol by emitting `[=1;1u`. If I read https://sw.kovidgoyal.net/kitty/keyboard-protocol/#ctrl-mapping right this should enable disambiguate escape codes, the same way as setting `kitty-keys = always` does unconditionally.

Let's take *ctrl-3* as example key combination. When I request disambiguate escape codes in plain Alacritty I get:

```
% printf '\033[=1;1u'
% cat -v
^[[51;5u
```

This is a CSI escape code as specified by the Kitty Keyboard Protocol. Key 51 is "3", modifier is 1 + 4 for ctrl.

But when I do the same in tmux I get:

```
% printf '\033[=1;1u'
% cat -v
^[[27u
```

This is a legacy escape code. ctrl combines with "3" maps to Key 27 which is Escape.

So even if disambiguate escape codes where requested, a legacy escape is send. This looks like a bug in tmux. Or am I missunderstanding something?

sundbp:
Probably a bug! I'll look at it.

sundbp:
@Yamagi I've attempted to deal with the outer vs inner kitty mode with this latest commit. Any feedback welcome!

cippaciong:
Hello @sundbp and thanks a lot for your efforts on this feature!

I did a quick test on your latest commit with `./tmux -Ltest -f /dev/null set -s kitty-keys on \; set -s extended-keys on \; new` and it looks good to me. I didn't do a thorough test to be honest, just a few key combinations that I knew were not working before and they all give the expected results in and out of tmux

```
$ printf '\e[>1u'

# Pressing in sequence: Ctrl-/ Ctrl-. Shift-Enter Ctrl-3
$ cat -v
^[[47;5u ^[[46;5u ^[[13;2u ^[[51;5u
```

sundbp:
Great to hear. I'm not sure what the appropriate path to getting enough confidence to merge this looks like.. I gave it a go as it seemed that missing kitty keyboard mode is a growing "gap" in tmux, but I'm far from an expert in this space. As it touches on fairly central key code interpretation paths, even in off mode we still touch new code (that should do nothing :)) in the main input paths - hence it it does seem high impact in case I've messed something up. Review by others and real world testing is only thing I can think of.

Yamagi:
I can comfirm that Vim is now working fine with `kitty-keys on`.

> I gave it a go as it seemed that missing kitty keyboard mode is a growing "gap" in tmux, but I'm far from an expert in this space. As it touches on fairly central key code interpretation paths, even in off mode we still touch new code (that should do nothing :)) in the main input paths - hence it it does seem high impact in case I've messed something up. 

For me this is one of the most important additions to tmux in the last years. While I can work with legacys keys and modifyOtherKeys work around the most annoying problems, the Kitty Keyboard Protocol is another level. Reliable mappings for ctrl-$number alone are a huge advantages.

I'll use this branch on my work laptop over the next few days. I'll report any bugs I encounter and give feedback if there are no remaining problems for me.

cippaciong:
> I'll use this branch on my work laptop over the next few days. I'll report any bugs I encounter and give feedback if there are no remaining problems for me.

I've been doing the same and I think I found an issue with compose key and accent letters.

This is what I get outside of tmux when I type ```RightAlt + ` + e``` (RightAlt is my compose key):
```
printf '\e[>1u'; cat -v
è
```

And this is what I get in tmux (`kitty-keys on`, `extended-keys on`) with the same combination:
```
printf '\e[>1u'; cat -v
^[[1107339459u
```

sundbp:
Great - I don't type those kinds of characters so hadn't tried this. I pushed a fix for this case - please see if it behaves as expected @cippaciong

cippaciong:
Awesome, works now!

```
printf '\e[>1u'; cat -v
èéẁç°ü
```

Thanks @sundbp for the quick fix

hovsater:
Hey! 👋 I'm super excited to try this out as I'm using [dte](https://gitlab.com/craigbarnes/dte) which supports both the legacy mode (`modifyOtherKeys`) as well as Kitty's keyboard protocol.

I don't think it makes sense to ever set `kitty-keys = always` as it would require all applications you interact with to also support and understand the escape sequences emitted by the protocol. Normally, the running application asks the terminal to enable the protocol for it specifically and then it's turned off again once you quit the program.

I'm going to report back soon!

hovsater:
I tested the following combinations:

- `kitty-keys = on` and `extended-keys = on`
- `kitty-keys = off` and `extended-keys on`

And it seems to be working great at first glance! I'm going to use this branch for work tomorrow and see if I'm running into anything unexpected.

sundbp:
I pushed the change to remove the always mode. It's was kind of useful during debugging, but I don't think it's a sensible foot gun to keep around (as this comment thread illustrates).

nicm:
Where is this now? Still any changes to come? I can take another look. Thanks

sundbp:
Nothing outstanding from me - and haven't had any reports of new issues found from the few people testing so far.

Yamagi:
I've used this branch as daily driver on my work laptop and on my personal desktop since last Wednesday. I didn't encounter any further problems with `extended-keys = on` and `kitty-keys = on`. From my side everything is working as expected.

GladioFeng:
PixelSnap 2026-03-26 at 11 16 57@2x
I encountered an error with the invalid option: kitty-keys.

kohane27:
I'm also using this fork and solved some real problems for me. Before using the fork, when pasting in Neovim it sometimes becomes ``[27;5;106~Key binding Action[27;5;106~Space`[27;5;106~`` and hence I can't use `extended-keys on`. But this fork allows me to use `extended-keys`! I will continue to test and see how it goes. Thanks again!

GladioFeng:
It works for me now, but how can I bind Cmd + \ in tmux? D-\ is not working.

ilyagr:
I also tried out the latest version of this PR briefly, and didn't have any problems this time, even with `extended-keys = always` (I briefly tired both `kitty-keys=on` and `off`).

craigbarnes:
~I just tried a few key combinations off the top of my head that tend to be problematic in new Kitty keyboard protocol implementations and found the following differences:~

<details>
  <summary>Original text hidden, to avoid confusion (see <a href="https://github.com/tmux/tmux/pull/4912#issuecomment-4182175213">below</a>)</summary>

**Command:**

```sh
printf "\033[>1u" && cat -v
```

**Results:**

| Input                                                          | Kitty      | foot       | tmux        |
|----------------------------------------------------------------|------------|------------|-------------|
| <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>Tab</kbd> | `^[[9;8u`  | `^[[9;8u`  | `m`         |
| <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>\\</kbd>                  | `^[[92;4u` | `^[[92;4u` | `^[[124;4u` |
| <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>9</kbd>                   | `^[[57;4u` | `^[[57;4u` | `^[[40;4u`  |
| <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>a</kbd>                  | `^[[97;6u` | `^[[97;6u` | `^[[65;6u`  |

**Command:**

```sh
printf "\033[>5u" && cat -v
```

**Results:**

| Input                                                          | Kitty          | foot           | tmux        |
|----------------------------------------------------------------|----------------|----------------|-------------|
| <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>Tab</kbd> | `^[[9;8u`      | `^[[9;8u`      | `m`         |
| <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>\\</kbd>                  | `^[[92:124;4u` | `^[[92:124;4u` | `^[[124;4u` |
| <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>9</kbd>                   | `^[[57:40;4u`  | `^[[57:40;4u`  | `^[[40;4u`  |
| <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>a</kbd>                  | `^[[97:65;6u`  | `^[[97:65;6u`  | `^[[65;6u`  |

**Command:**

```sh
git clone https://gitlab.com/craigbarnes/dte.git
cd dte
make
./dte -K
```

**Results:**

| Input                                                          | Kitty        | foot           | tmux     |
|----------------------------------------------------------------|--------------|----------------|----------|
| <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>Tab</kbd> | `C-M-S-tab`  | `C-M-S-tab`    | `m`      |
| <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>\\</kbd>                  | `M-S-\`      | `M-S-\`        | `M-S-\|` |
| <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>9</kbd>                   | `M-S-9`      | `M-S-9`        | `M-S-(`  |
| <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>a</kbd>                  | `C-S-a`      | `C-S-a`        | `C-S-a`  |

 - - -

I haven't done an exhaustive set of tests, but other <kbd>Alt</kbd>+<kbd>Shift</kbd> combos with e.g. punctuation keys exhibit the same kinds of differences as those shown above.

The Kitty documentation/spec also contains the [following statement]:

> Note that the codepoint used is always the lower-case (or more technically, un-shifted) version of the key. If the user presses, for example, <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>a</kbd> the escape code would be `CSI 97;modifiers u`. It *must not* be `CSI 65; modifiers u`.

...which this PR currently doesn't conform to.

Note that the only reason the `./dte -K` tests above show `C-S-a` instead of `C-S-A` in tmux is because dte does some basic normalization of keys `a..z`. This is done due to quirks in XTerm's `modifyOtherKeys` protocol and iTerm2's old, Kitty-like protocol, both of which could emit the same style of CSI sequences as the Kitty protocol.


[following statement]: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#:~:text=Note%20that%20the%20codepoint,65%3B%20modifiers%20u.
</details>

craigbarnes:
Oops, disregard everything I said above. Turns out I made a blunder and was accidentally testing `extended-keys` instead of `kitty-keys`. Sorry for the noise.

With `kitty-keys` enabled and the config properly reloaded, I haven't been able to find any problematic key combos.

sundbp:
(I'll rebase and resolve the tmux.h clash later today when I'm at my computer)

sundbp:
I've rebased this branch and resolved the conflict that had crept in versus master. I'd appreciate if anyone testing could give it an all clear again post rebase.

@nicm I think we're probably about as good as we'll get in terms of confidence of merging.

nicm:
Thanks, I am away this week and have a lot on when I get back but I'll try to look again at this again soon.

GladioFeng:
When `set -s kitty-keys on` is enabled, the clipboard does not work in Yazi.

I have confirmed that it works in tmux 3.6, so this may be a bug.

I am using branch d4d91732731691bd2bddb8c194b756ddaf1b5148.

sundbp:
Thanks for the report - I'm not familar with yazi clipboard functionality. Are you able to put down some more detailed observations etc to allows us to try to address it more effectively? Thanks @GladioFeng

GladioFeng:
> Thanks for the report - I'm not familar with yazi clipboard functionality. Are you able to put down some more detailed observations etc to allows us to try to address it more effectively? Thanks @GladioFeng

I will provide a minimal reproducible example (MRE) shortly.

nicm:
OK a couple more comments:

I don't understand why you change `KEYC_MASK_KEY`. A Unicode key is 32 bits and then we have 8 bits for the type. But you have added an extra 4 bits to `KEYC_MASK_KEY` above the type which don't seem to be used? You only need to change the modifiers, and since you are only adding three there are enough bits for them already.

Mixing a terminal that supports Kitty keys with one that only supports X10 keys seems OK (although of course some of the keys don't generate what a Kitty-expecting application might want, but that can't be helped - extended keys is the same). But mixing extended keys with Kitty keys still seems a bit funny, for example:

- I run tmux in ghostty (which supports Kitty keys but does not seem to support extended keys, or if it does it does so in a very odd way) and attach the same session from WezTerm (which has extended keys but not Kitty keys):
- I have `kitty-keys on` and `extended-keys on` in `.tmux.conf` - I do `printf '\033[>1u'` in the pane to turn Kitty keys on.
- `C-2` shows `^[[50;5u` when pressed in ghostty, but `^[[27;5u` when pressed in WezTerm which is clearly wrong (WezTerm emits `\033[27;5;50~`, so it seems like the key is being truncated instead of translated to the Kitty form).
- Other keys like `S-Tab` are similarly incorrect.

The other way round is similarly a bit off: 

- Instead of sending `\033[>1u` I send `\033[>4;2m` to turn on extended keys
- Then for `C-2` in WezTerm I correctly get `^[[27;5;50~`, but in ghostty I get `^[[27;5;32~` which is `^@` - which means Kitty keys are not being turned on, even though ghostty supports them and I have `kitty-keys on`. I would expect tmux to receive `^[[50;5u` from ghostty but map this to the extended keys form `^[[27;5;50~`.

sundbp:
I think you are right about the key mask. Early on I thought it seemed a good idea but turned out to be more over engineering than useful. I can address that.

Regarding extended-keys and kitty-keys: I think the example with the multiple clients supporting different modes is great for teasing out assumptions and design choices! The simplest model would be to make these two modes mutually exclusive. There may be valid use cases where one would like to connect different clients with different capabilities and have both modes work - but I think the pragmatic path is to get the mutually exclusive mode to work well to begin with as it'll likely be enough for the vast majority of use cases.

Making the modes mutually exclusive should be pretty straight forward.

nicm:
Another problem:

I run tmux in ghostty with `kitty-keys on`, then do `tmux bind -n Pause lsk` - but pressing `Pause` (`Fn-b` on my keyboard for some reason) does nothing. If I then do `printf "\033[>1u` in the pane, it starts working. tmux needs to turn the keys on immediately if the terminal supports them so that users can bind them.

nicm:
OK, mutually exclusive is fine. I imagine most people will be using one main terminal which will have one or the other, and if they attach from any different ones it will be temporary and they can manage with X10 keys alone.

So Kitty terminal -> X10 pane and X10 terminal -> Kitty pane need to work as sensibly as they can, but its OK if mixing extended and Kitty keys is not allowed.

I think the best thing would be (as I sort of proposed before) to just roll `kitty-keys` into `extended-keys` or `extended-keys-format` so the user can select one or the other but not both.

hovsater:
> I run tmux in ghostty (which supports Kitty keys but does not seem to support extended keys, or if it does it does so in a very odd way) [...]

Ghostty supports both `kitty-keys` and `extended-keys` and I don't think there's anything odd about it. If there is, I would love to know what that is. I've spent a fair amount of time together with Mitchell to get that right.

> tmux needs to turn the keys on immediately if the terminal supports them so that users can bind them.

Unless I misunderstand you, that's not something we want by default? When `set kitty-keys on`, you would expect the underlying application to turn the mode on/off. With `set kitty-keys always` it would be always on. Unconditionally enabling `kitty-keys` would lead to a lot of compatibility issues.

> I think the best thing would be (as I sort of proposed before) to just roll kitty-keys into extended-keys or extended-keys-format so the user can select one or the other but not both.

I kind of like having `extended-keys` and `kitty-keys` set to `on`. For instance, if I run `dte` inside a terminal without support for the Kitty keyboard protocol, I would still get the benefits of `modifyOtherKeys 1` (i.e., `extended-keys`). If I run `dte` in a terminal that supports the protocol, it would enable it instead.

This will still cause issues (like you described) if you connect two different terminals that does not have the same support. But I would rather improve upon that use case than making them mutually exclusive.

nicm:
>Ghostty supports both kitty-keys and extended-keys and I don't think there's anything odd about it. If there is, I would love to know what that is. I've spent a fair amount of time together with Mitchell to get that right.

Turn on mode 2 (`\033[>4;2m`) in ghostty and then press `^C`. It sends `^C` when it should send `^[[27;5;99~`.

Or try `C-1` - it should send `^[[27;5;49~` but it sends `1`. This is the same in mode 1 and mode 2.

This is ghostty 1.3.1.

> Unless I misunderstand you, that's not something we want by default? When set kitty-keys on, you would expect the underlying application to turn the mode on/off. With set kitty-keys always it would be always on. Unconditionally enabling kitty-keys would lead to a lot of compatibility issues.

It should not require an application running inside tmux to do anything for key bindings to work. tmux needs to turn on whatever it needs and then map that to whatever the application inside tmux wants.

>This will still cause issues (like you described) if you connect two different terminals that does not have the same support. But I would rather improve upon that use case than making them mutually exclusive.

When someone fixes it we can turn it on but for the moment it is broken so it should be turned off.

sundbp:
I've started on the 2 refactors - removing the extra 4 bits, and incorporating kitty into extended-keys mode.

One thing that's been a pain through in terms of confidence is that the `regress/input-keys.sh` tests seem to be a bit unstable and failing also on master branch. I just wanted to check if that's a "me" thing or the general situation?

nicm:
It works for me  but it is a bit flaky. If you add a sleep before the `send-keys` as well as after it seems to work better. There may be a better way to do this TBH.
````
--- a/regress/input-keys.sh
+++ b/regress/input-keys.sh
@@ -18,8 +18,9 @@ assert_key () {
        expected_code=$2

        W=$($TMUX new-window -P -- sh -c 'stty raw -echo && cat -tv')
+       sleep 0.1
        $TMUX send-keys -t$W "$key" 'EOL' || exit 1
-       sleep 0.2
+       sleep 0.1

        actual_code=$($TMUX capturep -pt$W | \
                              head -1 | \

nicm:
This works much better for me, there is no real need to recreate the window each time:
~~~~
--- a/regress/input-keys.sh
+++ b/regress/input-keys.sh
@@ -13,18 +13,18 @@ $TMUX set -g escape-time 0

 exit_status=0

+W=$($TMUX new-window -P -- sh -c 'stty raw -echo && cat -tv')
+
 assert_key () {
        key=$1
        expected_code=$2

-       W=$($TMUX new-window -P -- sh -c 'stty raw -echo && cat -tv')
+       $TMUX send-keys -t$W -R \; clear-history -t$W || exit 1
        $TMUX send-keys -t$W "$key" 'EOL' || exit 1
-       sleep 0.2

        actual_code=$($TMUX capturep -pt$W | \
                              head -1 | \
                              sed -e 's/EOL.*$//')
-       $TMUX kill-window -t$W || exit 1

        if [ "$actual_code" = "$expected_code" ]; then
                if [ -n "$VERBOSE" ]; then

nicm:
I've pushed this and a few other minor changes now if you want to try it again, it passes for me every time.

sundbp:
@nicm I've made an attempt at folding kitty keys into extended-keys here, and looked at the 4 bits. The fix to the input-keys.sh test works nicely for me - thanks!

I haven't looked at the binding issue. I'm not sure I see that one clearly enough to have a solution in mind!

nicm:
Looks good. I would probably have added them to `extended-keys` so it is `on`, `off`, `always`, `kitty`, but I don't mind using `-format` instead.

The point with the key bindings is that I should be able to do for example  `bind Sp-Left swapp -L` or `bind VolumeDown kill-pane` in tmux and if the terminal supports Kitty keys, the bindings should work even if no pane has requested them.

This means tmux needs to enable enough for its own use immediately - you can see this with the existing extended keys, tmux turns them on in `tty_update_features` and then "downgrades" them to X10 keys if the application in a pane has not turned them on

In an ideal world, any combination of terminal and application would work as far as possible, so tmux would request either Kitty keys or extended keys (depending on the `extended-keys` and `extended-keys-format` options) from the terminal and if the key was being passed to a pane, translate depending what the application inside the pane wants (so X10, extended mode 1, mode 2 or Kitty). But I am OK with only a subset working - perhaps only support these combinations:

Terminal|Pane
--|--
X10|X10
Kitty|X10
Kitty|Kitty
Extended|X10
Extended|Extended

That is, if tmux is configured for standard extended keys, but the application requests Kitty, the request would be ignored and X10 keys only would be sent. And similarly if tmux is configured for Kitty keys and the application requests standard extended keys. I think that is fine.

sundbp:
Thanks for the feedback. I'll try to take that onboard and see what I can do. 

Worth noting I haven't done a ton of manual testing post the re-organization. Anyone following along is very welcome to refresh and give it a spin!

sundbp:
@nicm had a  first stab at the remaining parts. not done too much manual testing yet - but thought I'd share to sense check direction.

nicm:
Only did a quick test but works much better for me, I can bind keys only supported by Kitty like `Pause` and they work immediately.

The idea looks right, tmux always turns on Kitty keys. I guess you will need to mediate between what tmux needs and what the active pane wants in terms of flags? Like we do for SGR mouse mode (and extended keys eg we always use mode 2).

craigbarnes:
> I think `Super-`, `Hyper-`, `RealMeta-` are too long. I'd go with `Sp-`, `Hy-`, `Rm-`, although I don't like the name "Real Meta" much anyway so I'd probably just drop it for now, how many keyboards have both Alt and Meta?

FWIW, there's some prior art for the format of these modifiers [in Emacs](https://www.gnu.org/software/emacs/manual/html_node/efaq/Binding-combinations-of-modifiers-and-function-keys.html). I [opted](https://gitlab.com/craigbarnes/dte/-/commit/05c1c0adfaee832d21c9854ff083d063d5a22458) to use the same `s-` (Super) and `H-` (Hyper) prefixes in dte, since the existing `C-`, `M-` and `S-` ones already matched. Not sure if the convention matters much, but just thought I'd mention it here.

sundbp:
No strong opinion from me. Happy to go with emacs conventions if that has general support.

nicm:
I don't really mind, I think `s` is a bit of a poor choice for Super but if you want to match emacs go for it.

allmazz:
Great work! Really miss this feature. Hope it will cover Ctrl+RET too

nicm:
Is this now ready to go or what is left to do? Thanks

musjj:
Is there a reason why release events are discarded? Is the support for the protocol intended to be partial at the moment?

sundbp:
@nicm I dropped the ball. The last thing I had in mind was the modifiers.

@musjj I don't t think there was a reason. My memory is now pretty rusty - any particular situations and code locations you have in mind?

musjj:
@sundbp

It's this line (also mentioned in your original PR description):

https://github.com/tmux/tmux/pull/4912/changes#diff-fd5d00eeffbec5a58a55c15bca5c5a0677d099df7999049afe30d42604228bb5R853-R856

I'm working a TUI that needed to detect key release events, so I'm wondering why it's getting dropped in this PR.

sundbp:
@musjj I can't recall a reason right now.. I'm guessing I may have found issues with it and "punted" on it as not a super common use case. Any chance you could experiment with your use case to see if you can get support for it working?

musjj:
Ok tested your PR and key release events are not reported, unfortunately. Modifier key events *are* reported though, so your PR is already partially working.

I tried to tweak your PR here: https://github.com/musjj/tmux/commit/5ddf17d2993c6c89df0e9e16021c8ffb1373c7f5

But I'm still not getting any key release events. Do you have any idea why it's not working?

sundbp:
@musjj thanks for doing the leg work. I need to re-engage with this PR as I've forgotten a lot of it by now  :/ I'll try to have a look at it and wrap this one up asap.

nicm:
I'm going to close this for the moment but it is clearly a feature that people would like so if you return to it (or anyone else picks it up) do please let me know to reopen or open a new PR. Thanks!

sundbp:
Understood. Life came in the way :(

ilyagr:
> Is there a reason why release events are discarded?

Tmux internals have no concept of release events, and i think you'd need to add one. You'd need to figure out if and when to synthesize/mask one of somebody either switches to tmux or away from tmux with a button held.

I think a simpler useful thing would be for tmux to be able to process release events itself, e.g. so you could bind copy mode selection to arrows with shift key pressed, or make a tmux ui with "momentary layers" like https://ki-editor.org/docs/category/momentary-layers--1/

nicm:
There is an updated PR [here](https://github.com/tmux/tmux/pull/5405) if anyone wants to continue to test.

PR #4920, opened 2026-03-13
Fix issue where popup window gets overwritten by background updates.

taylorconor:
This fixes an issue I've noticed where the pane behind a popup window occasionally draws over the popup window. This is especially noticeable when `htop` is running in the background pane, and a something like an empty emacs buffer is in the popup window.

### DIFF
--- a/regress/popup-overlap.sh
+++ b/regress/popup-overlap.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+# Test for popup getting overwritten by background updates
+# We use a nested tmux to capture what the inner tmux actually draws to the terminal.
+
+PATH=/bin:/usr/bin
+TERM=screen
+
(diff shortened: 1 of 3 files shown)


nicm:
Thanks - this looks sensible to me but:

- What's the `nx >= 10` change for in `tty_draw_line_clear`?

- `r` and `rr` declarations need to go to the top of the function in `tty_draw_line_clear`, not be in the middle.

taylorconor:
Woops my bad, updated `r` and `rr` declarations.

Ad for `nx >= 10`, it's pre-existing code. It appears to be some kind of optimization threshold for how tmux decides to clear the screen.

nicm:
Oh yeah sorry I thought you added it. I'll look at this again later, thanks.

nicm:

PR #4922, opened 2026-03-14
Added clock-mode-timezone option and added per pane clock options

UntoastedVoice:
Added option to change timezone in clock mode with the `clock-mode-timezone` option. I also added arguments `12-with-timezone`, `24-with-timezone`, `24-with-seconds-and-timezone` and `12-with-seconds-and-timezone` to the `clock-mode-style` option.

The options `clock-mode-color`, `clock-mode-style` and `clock-mode-timezone` now also works per pane, such that it is possible to display different time zones with different styles and colors in different panes using the `-p` flag.

image

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -36,7 +36,7 @@ static const char *options_table_mode_keys_list[] = {
 	"emacs", "vi", NULL
 };
 static const char *options_table_clock_mode_style_list[] = {
-	"12", "24", "12-with-seconds", "24-with-seconds", NULL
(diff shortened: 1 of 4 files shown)


charlesrhoward:
Looks super cool - hope it gets merged!

nicm:
Thanks but I don't think this is worth it, this is adding a good bit of code for a feature that not all that many people will want. `clock-mode` was added as simple test mode and really should have been removed long ago, it is not intended to have all sorts of bells and whistles. If you want more features there are external programs - a quick search turns up `tty-clock` and `peaclock`, so maybe try one of those.

PR #4925, opened 2026-03-17
fix: hyperlink loss in popup redraw

AntoineGS:
Linux unknown // It's Arch Linux / Omarchy
tmux 3.6a
tmux-256color

I was running into an issue with https://github.com/tmuxpack/tpack/issues/8 (tpm replacement) where the hyperlinks (OSC-8) would disappear after a few seconds. It uses the display-popup functionality of tmux and when running directly in a normal terminal it would not be an issue.

To replicate simply do `:display popup` and run `printf "\e]8;;https://example.com\aClick\e]8;;\a\n";`.
Then move the mouse over the hyperlink (holding whatever keys you need to in your terminal) and you will see the hyperlink disappear.

I tracked it down to the redraw of the popup, which would not copy over the hyperlinks.
I have accompanied the change with a test!

Let me know if any adjustments are needed.

### DIFF
--- a/popup.c
+++ b/popup.c
@@ -279,6 +279,10 @@ popup_draw_cb(struct client *c, void *data, struct screen_redraw_ctx *rctx)
 	popup_reapply_styles(pd);
 
 	screen_init(&s, pd->sx, pd->sy, 0);
+	if (pd->s.hyperlinks != NULL) {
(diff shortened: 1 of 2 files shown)


nicm:

PR #4926, opened 2026-03-17
systemd: ensure pane scopes survive during server shutdown

dtor:
Previously, systemd transient scopes for tmux panes were sibling units to the tmux server process. During system shutdown, systemd would often terminate these pane scopes in parallel with the server, causing state-saving plugins (like tmux-resurrect) to fail properly save sessions state because the panes would be already gone. Typical symptoms of this situation are 'no server running' error messages in the journal, a failed tmux service, and 0-length resurrect state on disk.

Fix the problem by ensuring that panes are not killed too early by discovering the systemd unit that started tmux service (if any), favoring user-level units over system-level ones, and attaching 'Before=' and 'PartOf=' properties to the transient pane scopes.

These annotations establish a strict dependency: systemd is now explicitly barred from stopping the panes until the server process's unit has finished its shutdown sequence (including all ExecStop hooks).

This ensures that state-saving scripts can reliably query all panes before the server is finally killed.

### DIFF
--- a/compat/systemd.c
+++ b/compat/systemd.c
@@ -109,7 +109,7 @@ systemd_move_to_new_cgroup(char **cause)
 	sd_bus_message		*m = NULL, *reply = NULL;
 	sd_bus 			*bus = NULL;
 	sd_bus_slot		*slot = NULL;
-	char			*name, *desc, *slice;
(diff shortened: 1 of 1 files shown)


nicm:
Applied, thanks!

PR #4927, opened 2026-03-17
Fix get-clipboard 'both' option in manpage

sethgirvan:
Looks like `buffer` was written a second time when it was meant to be `both`.


nicm:

PR #4929, opened 2026-03-18
feat: Panes' status format now displays title/flag information

daneofmanythings:
This is an implementation of point 9 from issue #4800:
'Modify status line 2 to be the name of the window pane and not width x height.`

I have added two new flags, `pane_zoomed_flag` and `pane_minimised_flag` to be utilized by formats. The implementation borrows the technique and styling used by `window_flags` for formatting panes on the status line.

@mgrant0

### DIFF
--- a/format.c
+++ b/format.c
@@ -1004,6 +1004,15 @@ format_cb_pane_fg(struct format_tree *ft)
 	return (xstrdup(colour_tostring(gc.fg)));
 }
 
+/* Callback for pane_flags. */
+static void *
+format_cb_pane_flags(struct format_tree *ft)
+{
+	if (ft->wp != NULL)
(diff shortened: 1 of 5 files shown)


mgrant0:
This needs some more work, not fully working for me yet.  I opened 2 windows but only the first showed in the status line 2.

Here's what I see: 

image

Secondly, though it's tempting to use the pane's title, it can be quite long (I have my cwd in there as you see).  Maybe that's fine.  Let's see how it goes once this works.  Might need to shorten things up.

mgrant0:
wait, i spoke too soon! sorry, it does seem to work.  I opened 2 panes but in the other window.

mgrant0 (commented):
Looks good.

PR #4930, opened 2026-03-18
utf8: support ranges in codepoint-widths option

15935648a:
Currently, codepoint-widths only accepts individual codepoints, requiring one entry per codepoint. With Nerd Fonts v3 "Propo" variant having 10741 wide icons, setting them one-by-one causes ~10 second delay on tmux start.

This adds support for ranges in the U+XXXX-U+YYYY format, allowing the same configuration with just a few entries:

      set -sa codepoint-widths "U+E000-U+E09F=2"
      set -sa codepoint-widths "U+E0C0-U+F8FF=2"
      set -sa codepoint-widths "U+F0001-U+FFFFF=2"

Requested by users configuring Nerd Fonts v3 "Propo" variant, which requires setting ~10k wide codepoints individually.

### DIFF
--- a/regress/codepoint-widths-range.sh
+++ b/regress/codepoint-widths-range.sh
@@ -1,0 +1,51 @@
+#!/bin/sh
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(cd .. && pwd)/tmux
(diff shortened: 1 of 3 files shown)


nicm:
Applied to OpenBSD now, with a small change to break the `RB_INSERT` bits into a helper function - will be in GitHub later. Thanks!

PR #4935, opened 2026-03-18
Fix division by zero and null dereference in grid.c

dantte-lp:
## Summary

Fix a division by zero in `grid_reflow_split()` and a null pointer
dereference in `grid_string_cells()`.

## Changes

**grid_reflow_split() — division by zero (line ~1374):**

`(gl->cellused - 1) / sx` divides by `sx` (target screen width)
without checking for zero.  If a reflow is triggered with a
zero-width screen (degenerate resize), this causes SIGFPE.  Add an
early return when `sx == 0`.

**grid_string_cells() — null dereference (line ~1121):**

`lastgc` (pointer-to-pointer parameter) can be NULL when
`GRID_STRING_WITH_SEQUENCES` is set.  The initialization guard at
line 1095 checks `lastgc != NULL` but the dereference further down
at line 1120 did not, leading to a null pointer dereference.  Add
the missing NULL check.

## Detection

Found by Clang scan-build with `core.DivideZero` and
`core.NullDereference` checkers.

## Testing

- Builds clean with GCC 14, GCC-15 (`-fanalyzer`), Clang 22
- No new warnings on grid.c from GCC-15 `-fanalyzer`

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -1117,7 +1117,8 @@ grid_string_cells(struct grid *gd, u_int px, u_int py, u_int nx,
 		if (gc.flags & GRID_FLAG_PADDING)
 			continue;
 
-		if (flags & GRID_STRING_WITH_SEQUENCES) {
+		if (lastgc != NULL &&
+		    (flags & GRID_STRING_WITH_SEQUENCES)) {
 			grid_string_cells_code(*lastgc, &gc, code, sizeof code,
 			    flags, s, &has_link);
 			codelen = strlen(code);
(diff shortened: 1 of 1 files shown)


nicm:
> grid_reflow_split() — division by zero (line ~1374):

Can this happen? How do you reproduce?

> grid_string_cells() — null dereference (line ~1121):

Applied, thanks!

dantte-lp:
You are right — I traced the full call chain and `sx == 0` cannot reach
`grid_reflow_split()` through any current path.

`screen_resize_cursor()` clamps `sx` to >= 1 (screen.c:313–316) before calling
`screen_reflow()`, and all callers of `screen_resize()` also enforce minimum
sizes (`PANE_MINIMUM`, TTY floor of 80, popup validation).

The check was flagged by Clang scan-build (`core.DivideZero`) on the division
at line ~1377.  Per C11 6.5.5p5, division by zero is undefined behavior, so
the checker flags it regardless of reachability.  It is a defensive guard
(CERT C INT33-C), not a fix for a reproducible bug.

Happy to drop this hunk since the null dereference fix is already applied.

Thank you for the review.

kecleon:
these AI slop PRs waste so much dev time

PR #4936, opened 2026-03-18
Fix use-after-free and null dereference in mode-tree.c

dantte-lp:
## Summary

Fix a use-after-free in `mode_tree_key()` and a null pointer
dereference in `mode_tree_display_help()`.

## Changes

**mode_tree_key() — use-after-free (line ~1231):**

`mode_tree_display_menu()` can trigger callbacks that free or
modify the mode tree data structure.  Reading `mtd->preview` after
this call accesses potentially freed memory.  Save the preview
value into a local variable before the call.

**mode_tree_display_help() — null dereference (line ~1176):**

The `lines` pointer returned by `helpcb` can be NULL, but the loop
`for (line = lines; *line != NULL; line++)` dereferences it
unconditionally.  Add `line != NULL` to the loop condition.

## Detection

Found by Clang scan-build (`unix.Malloc`, `core.NullDereference`)
and GCC-15 `-fanalyzer` (`CWE-416 use-after-free`).

## Testing

- Builds clean with GCC 14, GCC-15 (`-fanalyzer`), Clang 22
- No new warnings on mode-tree.c

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -1173,7 +1173,7 @@ mode_tree_display_help(__unused struct mode_tree_data *mtd, struct client *c)
 	}
 	for (line = mode_tree_help_start; *line != NULL; line++)
 		h++;
-	for (line = lines; *line != NULL; line++)
+	for (line = lines; line != NULL && *line != NULL; line++)
 		h++;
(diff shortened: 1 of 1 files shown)


nicm:

PR #4937, opened 2026-03-18
Fix unsigned underflow in tty clear operations

dantte-lp:
## Summary

Fix unsigned integer underflow in `tty_cmd_clearendofline()` and
`tty_cmd_clearendofscreen()` that can cause massive TTY clear
operations when the cursor position is beyond the screen edge.

## Changes

**tty_cmd_clearendofline():** `nx = ctx->sx - ctx->ocx` wraps when
`ocx >= sx` (cursor beyond screen width, possible with wide
characters). Add early return guard.

**tty_cmd_clearendofscreen():** Two underflows:
- `ny = ctx->sy - ctx->ocy - 1` wraps when `ocy + 1 >= sy`
- `nx = ctx->sx - ctx->ocx` wraps when `ocx >= sx`
Wrap both clear operations in bounds checks.

## Testing

- No new warnings from GCC-15 -fanalyzer or scan-build on tty.c

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -1703,7 +1703,11 @@ tty_cmd_clearline(struct tty *tty, const struct tty_ctx *ctx)
 void
 tty_cmd_clearendofline(struct tty *tty, const struct tty_ctx *ctx)
 {
-	u_int	nx = ctx->sx - ctx->ocx;
+	u_int	nx;
+
+	if (ctx->ocx >= ctx->sx)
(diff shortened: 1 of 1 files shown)


nicm:
It feels like this shouldn't happen, because the cursor should not be beyond the end of the line, but it would be interesting to see a case where it does and if so we can revisit this.

PR #4938, opened 2026-03-18
Fix unsigned underflow in window-copy scroll calculations

dantte-lp:
## Summary

Fix unsigned integer underflow in multiple window-copy.c functions
where `gd->hsize - data->oy` is computed without checking that
`oy <= hsize`, causing out-of-bounds grid access during resize
races.

## Changes

Affected functions:
- `window_copy_visible_lines()` — loop start/end from wrapped value
- `window_copy_search_mark_at()` — array index from wrapped value
- `window_copy_match_start_end()` — `(sy * sx) - 1` wraps when grid empty
- `window_copy_write_line()` — grid line offset from wrapped value
- `window_copy_cursor_prompt()` — grid line index from wrapped value

Fix: compute base offset into a local variable with guard
`if (oy > hsize) base = 0; else base = hsize - oy`.

## Testing

- No new warnings from GCC-15 -fanalyzer on window-copy.c

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -4237,13 +4237,19 @@ window_copy_visible_lines(struct window_copy_mode_data *data, u_int *start,
 {
 	struct grid		*gd = data->backing->grid;
 	const struct grid_line	*gl;
+	u_int			 base;
 
-	for (*start = gd->hsize - data->oy; *start > 0; (*start)--) {
+	if (data->oy > gd->hsize)
+		base = 0;
+	else
+		base = gd->hsize - data->oy;
+
+	for (*start = base; *start > 0; (*start)--) {
 		gl = grid_peek_line(gd, (*start) - 1);
 		if (gl == NULL || ~gl->flags & GRID_LINE_WRAPPED)
 			break;
 	}
-	*end = gd->hsize - data->oy + gd->sy;
+	*end = base + gd->sy;
 }
 
 static int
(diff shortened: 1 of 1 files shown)


nicm:
I don't know what "resize races" means, how do you reproduce this?

dantte-lp:
You are right to push on this — I went back and traced every path through
`window_copy_resize()` and the scroll functions more carefully.

In `window_copy_resize()`, `oy` is reclamped at lines 1038–1044 *before*
`window_copy_size_changed()` calls the drawing functions at line 1046.
The scroll functions (`pageup1`, `scroll_down`, `scroll_up`) all clamp `oy`
to `screen_hsize(backing)` at the point of modification.

I was not able to construct a concrete sequence where `oy > hsize` at the
point these functions are called.  The "resize races" description in the PR
was wrong — I apologize for the inaccuracy.

The guards are defensive: `gd->hsize - data->oy` is an unsigned subtraction
(`u_int - u_int`).  Per C11 6.2.5p9, if `oy` ever exceeds `hsize` — through
a future code change or an edge case I missed — the result wraps to ~4 billion
silently (defined behavior, but wrong).  CERT C INT30-C recommends guarding
unsigned subtractions where the operand relationship is not immediately obvious
from local context.

I understand if you prefer not to add guards for a condition that does not
currently occur.  Happy to close this PR if so.

nicm:
Thanks, will take another look at all of these at some point. I am in two minds about defensive checks, on the one hand yes they can prevent a bug having an effect, but OTOH they can hide bugs in a caller that would otherwise be detected. An assert might be better but we haven't really used them much... and then people go and build with `NDEBUG` anyway. Maybe these should be `fatal`s.

PR #4939, opened 2026-03-18
Fix unsigned underflow in window pane buffer offset

dantte-lp:
## Summary

Fix unsigned integer underflow in `window_pane_get_new_data()` and
`window_pane_update_used_data()` where `wpo->used - wp->base_offset`
wraps when `base_offset > used`, causing out-of-bounds buffer access.

## Changes

Add guards to return early (with `*size = 0` and `NULL` for the
getter, plain return for the updater) when `base_offset > used`.

## Testing

- No new warnings from GCC-15 -fanalyzer on window.c

### DIFF
--- a/window.c
+++ b/window.c
@@ -1729,7 +1729,13 @@ void *
 window_pane_get_new_data(struct window_pane *wp,
     struct window_pane_offset *wpo, size_t *size)
 {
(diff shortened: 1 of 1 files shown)


nicm:
I don't really think this should happen either... if it turns out it does we can add these :-).

PR #4943, opened 2026-03-22
feat: Add border buttons to panes

daneofmanythings:
This is a WIP implementation of point 8 from #4800:
`Add some buttons to the border: - ^ x to minimise, zoom, and close a pane.`

The wiring is mostly done. The current progress is at linking behavior to the button press. 

Some context: button "faces" are loaded from an options array. Upon a successful mouse event on a button, that index is stored in `struct mouse_event` in a new field `btn`.

A goal of the implementation is to allow uses to not only customize the number and faces of buttons, but the behavior as well. In service of that, I was thinking about adding another options string array which holds tmux commands. Each command is "linked" to its button by its index in the array (There will be some bookkeeping). Those commands can then be passed directly to `run -C` or through an intermediary which gets the tmux command string from the options array indexed at `event->mouse_event->btn`.

I also factored out around 1k lines in `server-client.c:server_client_check_mouse` through a macro. It was faster to factor it out than to add a new check to everything. More could be done, but I think it's a happy medium at the moment.

Let me know how it looks! @mgrant0

### DIFF
--- a/format-draw.c
+++ b/format-draw.c
@@ -50,6 +50,7 @@ format_is_type(struct format_range *fr, struct style *sy)
 	case STYLE_RANGE_LEFT:
 	case STYLE_RANGE_RIGHT:
 		return (1);
+	case STYLE_RANGE_BORDER:
 	case STYLE_RANGE_PANE:
 	case STYLE_RANGE_WINDOW:
 	case STYLE_RANGE_SESSION:
@@ -1064,6 +1065,8 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 		case STYLE_RANGE_USER:
 			log_debug("%s: range user|%u at %u-%u", __func__,
 			    sr->argument, sr->start, sr->end);
+			break;
+		case STYLE_RANGE_BORDER:
 			break;
 		}
 		format_free_range(&frs, fr);
--- a/format.c
+++ b/format.c
@@ -107,6 +107,7 @@ format_job_cmp(struct format_job *fj1, struct format_job *fj2)
 #define FORMAT_NOT 0x80000
 #define FORMAT_NOT_NOT 0x100000
 #define FORMAT_REPEAT 0x200000
+#define FORMAT_BUTTONS 0x400000
 
 /* Limit on recursion. */
 #define FORMAT_LOOP_LIMIT 100
@@ -1336,6 +1337,8 @@ format_cb_mouse_status_range(struct format_tree *ft)
 		return (xstrdup("left"));
 	case STYLE_RANGE_RIGHT:
 		return (xstrdup("right"));
+	case STYLE_RANGE_BORDER:
+		return (xstrdup("border"));
 	case STYLE_RANGE_PANE:
 		return (xstrdup("pane"));
 	case STYLE_RANGE_WINDOW:
@@ -4280,7 +4283,7 @@ format_build_modifiers(struct format_expand_state *es, const char **s,
 		}
 
 		/* Now try single character with arguments. */
-		if (strchr("mCLNPSst=pReqW", cp[0]) == NULL)
+		if (strchr("mBCLNPSst=pReqW", cp[0]) == NULL)
 			break;
 		c = cp[0];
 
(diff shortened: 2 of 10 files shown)


daneofmanythings:
pinging again because of a typo @mgrant0

mgrant0:
You  created a new mouse button named `MouseDown1BorderDefault`.  What is the intent here?  It kinda looks like you may have gotten confused with the `push-default` format command which is used to save the way things are formatted so you can output something and then pop them back.  But ahh, now that I look at it more, I think your intent with "Default" was to mean not clicking in the button area.  Hmm, this word Default is confusing in my opinion.   

The way the status line works is:
```
bind -n MouseDown1Status { switch-client -t= }
```
The `floating_pane` branch defined this to front a pane if clicked in the border:
```
bind -n MouseDown1Border { select-pane -M }
```
so no problem changing this if needed.  

Also, I note that in the status line, it did not require a "Default" section to handle clicking elsewhere in the line, so please have a look at this.  At least, I think so...  Call me out if I'm wrong!

The dragging of a pane is handled in `cmd-resize-pane.c`.  Here, I special cased the dragging the top border of a floating pane to move the pane around rather than resize it.  To be precise, you need to drag the top border of the pane staying away from the 2 characters in each of the corners because those are used as resize handles.  I really had to be 2 characters because otherwise positioning the mouse on the corner elbow was just next to impossible. 

What if you create a new button MouseFooNBorderTop (where Foo is Up, Down, Drag but those are all automatically created, and N is the button number).  This BorderTop area would be the top area as I described above (excluding the corner and character next to the corner).  Then, we could use this new button event and assign it to a new pane drag function instead of special caseing it in the resize.

To drag you would add something like:
```
bind -n MouseDrag1BorderTop { move-pane -t= }
```

Then you could treat this BorderTop much more like the status line buttons I think.  Ok let's muse about how one might bind buttons and faces...

I think you are right there needs to be an array of buttons somehow.  For simplicity, I would probably have 2 arrays, one for the faces and one for the actions.  But, hmm, this means setting some action rather than binding something.  I do not know if there is a more tmux-like way of binding things to an array, maybe Nic has some ideas?

Another possible approach is to create a fixed number of buttons up there.  If you did that, I'd probably create 5 or so but if they are not set, they are not displayed, and don't take up space  I'm not a big fan of fixed length things like that though, but i don't know if there's another way.

At some point it gets down to that it may just be simple enough for now to put the action in an array variable and not be something you bind.

But in any case, you are going to have to bind something to MouseDown1BorderButtons or MouseDown1BorderTop.

Some other ideas, you know the way the menu is bound, maybe you could put an if clause in the format and if it's in the button B: area do something, if not, then drag the pane (or just frankly just use the existing pane-resize which does this already).  This would be kinda thinking of that area like a menu that's always displayed in the border.  I say menu but it's presented as these little buttons, not a drop-down menu, though logically it is in fact a sort of menu.  Could something like that work?

nicm:
Maybe I don't really follow how this works but I don't think this should have new options. I think you are on the right lines with adding more ranges for this, but I don't know what the purpose of the options are.

Would it not be simpler to follow how the status line works and add a new range type (I wouldn't call it `button` because that is easily confused with mouse buttons), then a way to bind keys to that type? You could either use an argument to the range and pass that to the key binding so it could pick which command to execute, or - probably more intuitive for users - just define say 5 or 8 fixed ones?

So you would do something like:
~~~~
set -g pane-status-format '#[range=control1]x#[norange]'
bind MouseDown1Control1 kill-pane
~~~~

nicm:
BTW I think this should definitely be supported for non-floating panes also so it would be good to write it against master so we can get it in before floating panes are complete.

Thanks!

nicm:
We already have `user` ranges which work exactly how I'm describing but I think they work on the status line only.

daneofmanythings:
Thank you all for the feedback! I will redo this PR against the master branch; the implementation has no floating pane specific logic. I'll get started on that now.

> Maybe I don't really follow how this works but I don't think this should have new options. I think you are on the right lines with adding more ranges for this, but I don't know what the purpose of the options are.
> 
> Would it not be simpler to follow how the status line works and add a new range type (I wouldn't call it `button` because that is easily confused with mouse buttons), then a way to bind keys to that type? You could either use an argument to the range and pass that to the key binding so it could pick which command to execute, or - probably more intuitive for users - just define say 5 or 8 fixed ones?
> 
> So you would do something like:
> 
> ```
> set -g pane-status-format '#[range=control1]x#[norange]'
> bind MouseDown1Control1 kill-pane
> ```

I initially was going to implement it in this manner, but I noticed that the current implementation for the status line doesn't work like that and I figured I should try to keep the methodologies aligned. Currently, a click on an actionable range all gets funneled into a `STATUS` enum even though there are checks with high enough fidelity to behave as you  described above. Here is an abbreviation of the responsible code (server-client.c:846):
```c
			case STYLE_RANGE_PANE:
                               ...
				where = STATUS;
				break;
			case STYLE_RANGE_WINDOW:
                               ...
				where = STATUS;
				break;
			case STYLE_RANGE_SESSION:
                               ...
				where = STATUS;
				break;
			case STYLE_RANGE_USER:
				where = STATUS;
				break;

```
Then, the responsibility of determining what happens is passed to a single command through this binding: `"bind -n MouseDown1Status { switch-client -t= }",`. This is what I was trying to model it after, but I prefer the solution you detailed above. As such, How would you feel about also refactoring the current system to work similarly so that we end up with a few more binds like: `"bind -n MouseDown1StatusPane { select-pane -t= }",`?

It would be possible to generalize this fully to where everything is a 'Control' bind, but I don't think that is necessary at the moment.

nicm:
I don't know that there is any point in having separate Pane, Window, and Session keys - they all set a target, so unless I am missing something it would break configs without much benefit?

The difference for the buttons/controls/whatever is you can't use the target so need some other way to indicate which was clicked. I would probably just go for a fixed 10, so Control0 to Control9.

But in any case ideally the displayed text and clickable area should be set with `range` inside the `-format` and the commands bound to keys. That way one can use the same ranges to add buttons to the status line as well as the pane status line.

daneofmanythings:
Yes, that makes sense! I'm closing this PR and will open a new one against the master branch.

PR #4944, opened 2026-03-22
Add -s/-S for new_pane

0xC0FFEE:
This PR implements `12. Add -s border-style and -S style args to the new-pane command. Probably also need a default style option variable for these.` (#4800)

This works in my testing. The style is correctly changed, when a floating pane becomes a regular pane. And updates to the global options are applied correctly for existing panes without custom options.


nicm:
Was this applied? @mgrant0 ?

mgrant0:
@nicm applied now.

PR #4946, opened 2026-03-23
feat: Add new set of style ranges and keycodes for pane and status buttons

daneofmanythings:
This is the refactor of PR #4943 against the main branch. It is not finished yet, but I believe the big parts are done. I need to do some more rigorous testing. Buttons on the pane border are mostly working. I would like to investigate an offset I have to correct for. I have not yet tested buttons on the status line.

I did copy over work done by @mgrant0 in the floating_panes branch (mainly in `server-client.c:server_client_check_mouse_in_pane`). He has worked on mouse event border detection, so i ported that over without the logic for floating panes.

Here is the basis for what I have been testing with:
```
set -g pane-border-format '#[range=control|0]x#[norange]'
bind -n MouseDown1Control0 kill-pane
```
It is nearly identical to what @nicm wrote in the old PR with the exception to how the control number is given to the range and assigning the mouse event to the root keytable.

Let me know how it looks to you guys!

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -630,7 +630,7 @@ server_client_check_mouse_in_pane(struct window_pane *wp, u_int px, u_int py,
 	if (((pane_status != PANE_STATUS_OFF &&
 	    (int)py != pane_status_line && py != wp->yoff + wp->sy) ||
 	    (wp->yoff == 0 && py < wp->sy) ||
-	    ((int)py >= wp->yoff && (int)py < wp->yoff + (int)wp->sy)) &&
+	    (py >= wp->yoff && py < wp->yoff + wp->sy)) &&
 	    ((sb_pos == PANE_SCROLLBARS_RIGHT &&
 	    (int)px < (int)wp->xoff + (int)wp->sx + sb_pad + sb_w) ||
 	    (sb_pos == PANE_SCROLLBARS_LEFT &&
@@ -687,35 +687,6 @@ server_client_check_mouse_in_pane(struct window_pane *wp, u_int px, u_int py,
 			return (KEYC_MOUSE_LOCATION_BORDER);
 	}
 	return (KEYC_MOUSE_LOCATION_NOWHERE);
-}
-
-static enum mouse_where
-server_client_control_where(u_int n)
-{
-	switch (n) {
-	case 0:
-		return CONTROL0;
-	case 1:
-		return CONTROL1;
-	case 2:
-		return CONTROL2;
-	case 3:
-		return CONTROL3;
-	case 4:
-		return CONTROL4;
-	case 5:
-		return CONTROL5;
-	case 6:
-		return CONTROL6;
-	case 7:
-		return CONTROL7;
-	case 8:
-		return CONTROL8;
-	case 9:
-		return CONTROL9;
-	default:
-		return NOWHERE;  /* unreachable */
-	}
 }
 
 /* Check for mouse keys. */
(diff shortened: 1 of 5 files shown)


nicm:
Looks good to me. How about just having `CONTROL` in `mouse_where` and using a separate variable to store which control it actually is?

Similarly maybe the same for `STYLE_RANGE_CONTROL*`? You could put the range number in the `argument`.

The only place you will actually have to list 0-9 is for key bindings where you will actually need multiple `KEYC_MOUSE_KEY` etc.

daneofmanythings:
Good idea. I consolidated the style range to a single range, `STYLE_RANGE_CONTROL`, which carries the id in the arg. I was having trouble eliminating the need for the enumerated `CONTROL` in `mouse_where` because it doesn't fit into the macro given that it is a runtime value. I am not that good at macros though, Is there a way to accomplish this?

nicm:
My thought was that if you only have one entry you can do without the macros. I don't really know that big multiline macros improve things here. Instead of macros it might be better to make it use tables instead of switches, but in either case (macros or tables) it would be better as a separate MR

daneofmanythings:
Ah, okay. I have reverted the keycode detection in server-client.c and added logic for the control keycodes in a consistent way.

daneofmanythings:
The PR is rebased to the current master with the new key code generation.

nicm:
Looks good on a quick look but there seems to be a bug of some sort.

Try this:
~~~~
tmux set -g pane-border-status top
tmux set -g pane-border-format '#{?pane_active,#[reverse],}#{pane_index}#[default] \"#{pane_title}\" #[align=right range=control|0](X)#[norange] #[range=control|1](Z)#[norange]'
tmux bind -n MouseDown1Control0 list-keys
tmux bind -n MouseDown1Control1 resize-pane -Z
tmux splitw
~~~~
Then click on the `(Z)` to zoom one of the panes. Both the `(X)` and `(Z)` no longer work.

daneofmanythings:
Good catch! I forgot to add over the pane zoom detection logic from `floating_panes`. It is working now.

nicm:
OK I have applied this with some minor tidying, thanks.

There is one issue, but I'm not sure what the cause is. If you make the controls `zoom-pane` or `kill-pane` they work fine, but if you make them `list-keys` or `display-menu` they either do not work or work in the wrong pane (active pane, not mouse pane). If you do `-t=` then they don't do anything so clearly the mouse target is not available. I'm not sure why this is. It doesn't matter for `list-keys` but it is annoying for `display-menu`, although not sure how much it is worth looking into even for that.

daneofmanythings:
Yes, I believe the issue is that list-keys does not take a target, it just grabs the active pane. I was able to get it working with this: `bind -n MouseDown1Control0 "select-pane -t=; list-keys"`. I am currently investigating why the menu does not work. It absolutely should from both an implementation and a UX perspective.

daneofmanythings:
~~Okay, I have some idea of what is going on with menus, and I'm not confident it is related to this PR. I think it has to do with parsing the menu contents from a bind assigned in a config. `cmd_display_menu_exec` is failing on this line of `cmd-display-menu.c:334`:~~
```
		if (count - i < 2) {
			cmdq_error(item, "not enough arguments");
			goto fail;
		}
```
~~because the count is 1. It seems as though the segments of the menu aren't being recognized in the argument parsing and it is just a single string. This is not unique to mouse events. Here are the three examples where binds fail in the config:~~
```
bind -n MouseDown1Control2 display-menu -t= -xM -yM -MO -T'test' " 'Next' 'n' {switch-client -n}"
bind -n MouseDown3Status "display-menu -t= -xW -yW -T '#[align=centre]#{window_index}:#{window_name}' \" 'Kill' 'X' {kill-window}\""
bind -T prefix < display-menu -xW -yW -T '#[align=centre]#{window_index}:#{window_name}' " 'Kill' 'X' {kill-window}"
```
~~The second two are minimal forms of binds that exist in the default key table. When I alter them directly in the key table, they are parsed correctly and the binds work, but not so when assigned in a config. My search led me to `cmd.c:cmd_parse` which is only called in `cmd-parse.y:849`, and I am unfortunately entirely unfamiliar with yacc.~~

~~There is a chance that I am trying to construct these binds incorrectly in the config. Am I doing it right? If so, do you have an idea as to what is going on with the parsing?~~

Okay, I got it working. I indeed was binding things incorrectly in the config. Here is what I got to work:
```tmux
bind -n MouseDown1Control2 "display-menu -t= -xM -yM -MO -T'test'  'Kill' 'k' {kill-pane}"
```
It is a minimal example. For me, the menu appears, correctly gets the target, and executes the command against that target.

daneofmanythings:
I have added the control ranges and mouse events to the documentation, but am not sure how to get the section headers to italicize in their contexts. They are showing up as quoted.

nicm:
`.Sx` is correct for a section reference.

nicm:
Applied but I changed `'ControlN` to `.Ql ControlN`, new-sentence-new-line, and changed to use `N` instead of `X` consistently for the argument. Thanks!

Issue #4941, opened 2026-03-20
Hyphens in man page may be rendered incorrectly


Keith-S-Thompson:
### Issue description

In some configurations, `man tmux` renders '-' (U+002D, HYPHEN-MINUS) characters as '‐' (U+2010, HYPHEN).

This makes search and copy-and-paste from a displayed man page awkward. 

For example, if I search for `default-shell` in the man page, it's not found, and copying the name `default‐shell` yields something that isn't syntactically valid.

This problem was exposed by a change in GNU groff 1.23.0, such that HYPHEN-MINUS characters are rendered as HYPHEN.

Debian and Ubuntu have a custom version of groff that doesn't do this, so the tmux man page is displayed correctly. (I'm using a copy of groff built from source, so I don't have that patch.)

One workaround is to set LANG=C, so the man page is rendered in plain ASCII.

I have a tentative patch for `tmux.1`, based on the current `master` branch, that fixes this problem, replacing `-` with `\-`, which is rendered properly as U+002D. Before I submit a pull request, I need decisions on a few points, particularly point 5:

1. I've changed `-` to `\-` for everything that a user might want to enter as text: command names, option names, etc. (such as `select-window` or status-style`.
2. I've changed `-` to `\-` for references to shift, control, and meta characters, such as `C-b`.
3. I've changed `~` to `\[ti]` and `^` to `\[ha]` in a few places.
4. I've changed `UTF-8` to `UTF\-8`.
5. For now, I have *not* changed argument names that are not literal text. For example, in  
  `select-window [-lnpT] [-t target-window]`  
  I've changed `-` to `\-` in "select-window" but not in "target-window". I'm not at all sure this is the way to go, and I tend to think both should be rendered as U+002D. Using a Unicode HYPHEN in `target-window` is arguably correct, but using a plain ASCII HYPHEN-MINUS feels more consistent. I'll wait for feedback before making this change.
6. I've made one small editorial change, changing "shell start-up files" to "shell startup files" (I happened to run across this while searching for `-` characters).

References:
- <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1052675>
- <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1041731>
- `man -s 7 man-pages` and search for "Generating optimal glyphs" (on Ubuntu)

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

[...]

PR #4948, opened 2026-03-24
Update man page to render hyphens correctly (#4941)

Keith-S-Thompson:
Render '-' characters as ASCII hyphen-minus rather than Unicode hyphen where appropriate.


nicm:
Looks good to me, the only changes in `mandoc` output are where you made a couple of obvious corrections and where it word wraps slightly differently. Applied to OpenBSD now, should be in GitHub later. Thanks!

PR #4951, opened 2026-03-25
Add bracket_paste_flag format variable

gnachman:
Add `#{bracket_paste_flag}` to query whether a pane has bracketed paste mode enabled (`ESC[?2004h`). Returns 1 if set, 0 if not.

This is needed by iTerm2's tmux integration.


nicm:

PR #4953, opened 2026-03-25
refactor: Updating key code representations to  simplify mouse key codes.

daneofmanythings:
This PR is currently a proof of concept that a lookup table could work. I have not fully incorporated all keycodes. I was hoping to get some guidance as to whether or not this is an acceptable method. I have synchronized and organized the keycodes and mouse_where/type so that I can use arithmetic with strides to calculate the offset into the table for a specific keycode. If any of the above orderings change it will break keycode detection, so I made bossy comments. It is brittle regardless.

I will finagle the dragends into the function once I have an established method.

### DIFF
--- a/key-string.c
+++ b/key-string.c
@@ -402,7 +402,7 @@ key_string_lookup_key(key_code key, int with_flags)
 		s = "MouseMovePane";
 		goto append;
 	}
-	if (key == KEYC_MOUSEMOVE_STATUSLINE) {
+	if (key == KEYC_MOUSEMOVE_STATUS) {
 		s = "MouseMoveStatus";
 		goto append;
 	}
--- a/key-string.c
+++ b/key-string.c
@@ -402,7 +402,7 @@ key_string_lookup_key(key_code key, int with_flags)
 		s = "MouseMovePane";
 		goto append;
 	}
-	if (key == KEYC_MOUSEMOVE_STATUS) {
+	if (key == KEYC_MOUSEMOVE_STATUSLINE) {
 		s = "MouseMoveStatus";
 		goto append;
 	}
--- a/server-client.c
+++ b/server-client.c
@@ -1292,7 +1292,7 @@ try_again:
 	 * application requests, so don't let them exit the prefix table.
 	 */
 	if (key == KEYC_MOUSEMOVE_PANE ||
-	    key == KEYC_MOUSEMOVE_STATUS ||
+	    key == KEYC_MOUSEMOVE_STATUSLINE ||
 	    key == KEYC_MOUSEMOVE_STATUS_LEFT ||
 	    key == KEYC_MOUSEMOVE_STATUS_RIGHT ||
 	    key == KEYC_MOUSEMOVE_STATUS_DEFAULT ||
--- a/server-client.c
+++ b/server-client.c
@@ -1292,7 +1292,7 @@ try_again:
 	 * application requests, so don't let them exit the prefix table.
 	 */
 	if (key == KEYC_MOUSEMOVE_PANE ||
-	    key == KEYC_MOUSEMOVE_STATUSLINE ||
+	    key == KEYC_MOUSEMOVE_STATUS ||
 	    key == KEYC_MOUSEMOVE_STATUS_LEFT ||
 	    key == KEYC_MOUSEMOVE_STATUS_RIGHT ||
 	    key == KEYC_MOUSEMOVE_STATUS_DEFAULT ||
--- a/server-client.c
+++ b/server-client.c
@@ -922,6 +922,7 @@ have_event:
 		 * End a mouse drag by passing a MouseDragEnd key corresponding
 		 * to the button that started the drag.
 		 */
+		type = KEYC_TYPE_MOUSEDRAGEND;
 		c->tty.mouse_drag_flag = 0;
 		c->tty.mouse_slider_mpos = -1;
 	}
@@ -960,24 +961,24 @@ have_event:
 	}
 
 	if (key == KEYC_UNKNOWN) {
-		/* Adjust the button number */
-		if (b == MOUSE_BUTTON_1)
+		/* Adjust the button number. */
+		if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_1)
 			bn = 1;
-		else if (b == MOUSE_BUTTON_2)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_2)
 			bn = 2;
-		else if (b == MOUSE_BUTTON_3)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_3)
 			bn = 3;
-		else if (b == MOUSE_BUTTON_6)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_6)
 			bn = 6;
-		else if (b == MOUSE_BUTTON_7)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_7)
 			bn = 7;
-		else if (b == MOUSE_BUTTON_8)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_8)
 			bn = 8;
-		else if (b == MOUSE_BUTTON_9)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_9)
 			bn = 9;
-		else if (b == MOUSE_BUTTON_10)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_10)
 			bn = 10;
-		else if (b == MOUSE_BUTTON_11)
+		else if (MOUSE_BUTTONS(b) == MOUSE_BUTTON_11)
 			bn = 11;
 		else
 			bn = 0;
--- a/tmux.h
+++ b/tmux.h
@@ -242,7 +242,7 @@ enum key_code_mouse_location {
          ((unsigned long long)(l) << KEYC_MOUSE_LOCATION_SHIFT))
 #define KEYC_MOUSE_STRING(name, s)				      \
 	{ #s "Pane", KEYC_ ## name ## _PANE },			      \
-	{ #s "Status", KEYC_ ## name ## _STATUSLINE },		      \
+	{ #s "Status", KEYC_ ## name ## _STATUS },		      \
 	{ #s "StatusLeft", KEYC_ ## name ## _STATUS_LEFT },	      \
 	{ #s "StatusRight", KEYC_ ## name ## _STATUS_RIGHT },	      \
 	{ #s "StatusDefault", KEYC_ ## name ## _STATUS_DEFAULT },     \
--- a/tmux.h
+++ b/tmux.h
@@ -293,7 +293,7 @@ enum {
 	C0_US
 };
 
-/* Special key codes */
+/* Special key codes. */
 enum {
 	/* User key code range. */
 	KEYC_USER = KEYC_SHIFT_TYPE(KEYC_TYPE_USER),
--- a/tmux.h
+++ b/tmux.h
@@ -242,7 +242,7 @@ enum key_code_mouse_location {
          ((unsigned long long)(l) << KEYC_MOUSE_LOCATION_SHIFT))
 #define KEYC_MOUSE_STRING(name, s)				      \
 	{ #s "Pane", KEYC_ ## name ## _PANE },			      \
-	{ #s "Status", KEYC_ ## name ## _STATUS },		      \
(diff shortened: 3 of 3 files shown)


nicm:
I think your instinct is right and this is too brittle. Maybe it would be better to have multiple tables. We are converting (type, where, buttons) -> key, right? So maybe something like:
~~~~
struct mouse_key_map {
    u_nt from;
    key_code key;
};
struct mouse_where_map {
  u_ int from;
  struct mouse_key_map *map;
};
struct mouse_type_map {
   u_int from;
  struct mouse_where_map *map;
};

#define MOUSE_BUTTON_MAP(type, where) \
 struct mouse_key_map button_map_ ## type ## _ ## where ## [] =  { [MOUSE_BUTTON_1] = KEYC_ ## type ## 1_ ## where, \
     and so on
};

MOUSE_BUTTON_MAP(MOUSEDOWN, PANE);
and so on

#define MOUSE_WHERE_MAP(type) \
struct mouse_where_map where_map_ ## type ## [] = {
[MOUSE_WHERE_PANE] = &button_map_## type ## _PANE,
and so on
};

MOUSE_WHERE_MAP(MOVE);
and so on

struct mouse_type_map type_map[] = {
[MOVE] = &where_map_move,
and so on
};

key_code
mouse_map_to_key(type, where, button)
{
// range checking 
return type_map[type].map[where].map[button];
}
~~~~
Of course would need them to be const, static etc and some variant macros probably for stuff that doesn't care about buttons, but you get the idea. Id say it should be a new source code file.

The other option is to just build a big RB tree at startup, but that costs a lot of mem for something that could be done statically.

Or we could stick with big switches generated with macros but pull that all into a new function in a new file so it doesn't make the code where it is used so hard to follow.

nicm:
Or... there is nothing magic about mouse key codes, we could change to represent them in a different way like maybe we could pick a range for mouse key codes and say 8 bits for where, 8 bits for buttons, 8 bits for type. Then they could just be built directly. So long as we make it so they can't be confused with other key codes or Unicode.

key_code is 64 bits so should be possible?

nicm:
Currently we have key codes like this:

0x00000000000 - 0x0000010dfff = Unicode keys
0x0000010e000 - 0x0000010efff = builtin KEYC_* (not all 0xfff are used)
0x0000010f000 - 0x0000010f3e8 = user keys

The KEYC_* keys and user keys are in PUA B which is 0x100000 to 0x10fffd.

We could just allocate a new range in the same PUA for mouse keys, but I don't think we need to - there are only around 50 KEYC_* keys, so they don't need much space.

Instead we could use the space differently. 0xfff gives us 12 bits. For each mouse keys we have type, "where" (location), and button. So we could give them four bits each like this:

~~~~
ba9876543210
111100000000 type
000011110000 where
000000001111 button

enum key_code_type {
    KEYC_TYPE_KEY = 0,
	KEYC_TYPE_MOUSE_MOVE = 1,
	KEYC_TYPE_MOUSE_DOWN = 2,
	KEYC_TYPE_MOUSE_UP = 3,
	KEYC_TYPE_MOUSE_DRAG = 4,
	KEYC_TYPE_MOUSE_DRAG_END = 5,
	KEYC_TYPE_MOUSE_WHEEL_UP = 6,
	KEYC_TYPE_MOUSE_WHEEL_DOWN = 7,
	KEYC_TYPE_MOUSE_SECOND_CLICK = 8,
	KEYC_TYPE_MOUSE_DOUBLE_CLICK = 9,
	KEYC_TYPE_MOUSE_TRIPLE_CLICK = 10,
};

enum key_code_where = {
    KEYC_WHERE_NONE = 0,
    KEYC_WHERE_PANE = 1,
    KEYC_WHERE_STATUS = 2,
    KEYC_WHERE_STATUS_LEFT = 3,
    KEYC_WHERE_STATUS_RIGHT = 4,
    KEYC_WHERE_STATUS_DEFAULT = 5,
    KEYC_WHERE_BORDER = 6,
    KEYC_WHERE_SCROLLBAR_UP = 7,
    KEYC_WHERE_SCROLLBAR_SLIDER = 8,
    KEYC_WHERE_SCROLLBAR_DOWN = 9,
};
~~~~

And there are only 12 buttons, so that would fit too.

KEYC_BASE could stay as 0x10e000 since it has the top 4 bits zero already so would match KEYC_TYPE_KEY, only the mouse definition and the test macros would need to change.

Possibly we could roll user keys into the same scheme (KEYC_TYPE_USER?) and use the whole PUA which would give us 15 bits (up to 0x7fff) but I don't know if we need to do that.

This would all mean that `server_client_check_mouse` could build the keys pretty much directly without needing either switches or lookup tables.

What do you think? Am I missing anything here?

nicm:
Well, I forgot you are adding more `where`s for control buttons, so maybe we would need 5 or 6 bits for that? So we would need to roll in user keys and use the entire range? That would not be much more of a change though?

daneofmanythings:
Thanks for putting some time into this! Yes this is what I have been prototyping. I really like the idea of widths corresponding to the constituent parts of a keybind. I have been wrestling with the preprocessor and I have something to show for it. It is again just proof of concept and things can be tweaked (ex: widths are 8bit at the moment, should probably be 5/6 like you said to divide the 3 available bytes and leave room for another key parameter. 6 makes sense to me), but this method looks like a scalable way to generate the binds. I am attaching the file I've been playing in to this message. It does over-generate, but that doesn't really matter given that the keyspace would otherwise not be used. How does it look?

[test.c](https://github.com/user-attachments/files/26317324/test.c)

daneofmanythings:
Oh, and as for the ranges for different types of binds, I thought we put non-mouse keys above the PUAs, user keys above that, then start the 3ish bytes for mouse keycodes leaving the top two bytes for flags as they are. I noticed that there are only 7 defined bytes in `key_code` atm. if another parameter for mouse binds aren't needed, the extra space can be co-opted by the flags; the 3 bytes down to 15/18 bits.

nicm:
Nothing can go above the PUAs because those are valid Unicode characters, all tmux internal keys need to fit in the PUAs.

So if we use the full range of PUA B that is 15 bits.

We could use PUA A as well in theory but I don't think it is necessary - 5 bits is 32 which should be more than enough for each component of a mouse key. And the two non-mouse ranges (user keys and function keys) can have 10 bits each (keeping 5 for the type) which will be 1024, more than enough for both

nicm:
Well actually we have 64 bits now and Unicode characters are only 32, so there is room to put tmux's keys above 0xffffffff - they don't need to fit in the PUAs anymore. In fact that might be better.

So you could use do something simple like:

16 bits - flags
8 bits - modifiers
8 bits - type (0 = unicode key, 1 = function key, 2 = user key, 3 = mouse move, 4 = mouse drag, etc)

That then leaves 32 bits for each of the types, plenty of room for any of them

nicm:
Your test.c looks the right idea but I would definitely use enums over defines, and we will need to assign room for everything not just mouse keys.

daneofmanythings:
I have a "working" prototype. I have not fully tested this, but typing works, as does clicking around windows/panes/sessions. A couple things to note:
1. There is a naming conflict for `STATUS` in `arpa/nameser_compat.h:100`. I changed the enum name to STATUSLINE as a placeholder to see if the implementation would work. I like how clean the construction of the mouse key codes is, but have enums that aren't scoped doesn't feel good.
2. I do a little bit of dispatch right before constructing the key to correctly assign the mouse button. Are the `#defines` for `MOUSE_BUTTON_N` in `tmux.h` important? My gut says they are as I don't see the reason for the values. Ideally I would change them to represent the associated number. I'm guessing they come from some other api up the chain.

How does this look to you?

nicm:
Because these are global enums they should have fully qualified names for clarity and to avoid conflicts.

I would probably also rename `mouse_where` to match the `key_code`/`KEYC` scheme:
~~~~
enum key_code_type {
    KEYC_TYPE_UNICODE,
    KEYC_TYPE_USER,
    ...
};

enum key_code_mouse_where {
    KEYC_MOUSE_WHERE_PANE,
    KEYC_MOUSE_WHERE_STATUS,
    ....
};
~~~~
I would probably consider changing `where` to `location` at this point because `where` is a bit funny...

daneofmanythings:
That was simpler than I thought it would be. Enum names are now fully qualified.

nicm:
This looks good, a few trivial things:

- Comments on a line on their own should be full sentences so they need a period at the end, for example `/* Special key codes. */`.

- Don't see a need for `KEYC_TYPE_SHIFT` when it is only used twice in lines right together, just replace it with `24`.

- `KEYC_VALUE_MAX` is not very enlightening... and its only used for user keys, so why not leave it as `KEYC_NUSER`?

- I think it would be clearer if `KEYC_MASK_TYPE` was with the other `KEYC_MASK_*` and in the same hex form, then looking at the four `KEYC_MASK_*` lines will show the key layout?

- Remove the comma from `KEYC_TYPE_NOTYPE, /* end */` the only reason to keep a comma is to avoid a diff having to change two lines to add one item, but we will never want to add an item after `KEYC_TYPE_NOTYPE`.

- I think I would replace `KEYC_TYPE_SPECIAL` with `KEYC_TYPE_FUNCTION`? I know there was a comment saying "special keys" but it doesn't seem quite right.

- Maybe change `KEYC_MOUSE_KEYCODE` to `KEYC_MAKE_MOUSE_KEY` or something since `KEYC....KEYCODE` seems a bit funny and is not very different from `KEYC_MOUSE_KEY`.

- The `/* Mouse key code ranges */` comment should probably point out that these must remain at the end.

daneofmanythings:
I have made the requested changes, though i would like to bikeshed slightly about a few of them.
1. `KEYC_TYPE_*` is in the 5th byte, so i believe it should be a 32 shift like i had it. That is how it currently is in the pr. `KEYC_TYPE_SHIFT` is gone.
2. `KEYC_TYPE_FUNCTION` would sound good to me if there wasn't already function keys (F1 - F12). I personally find this confusing. How does `KEYC_TYPE_FUNCTIONAL` sound? That is how i have it in the pr currently.
3. I have put `KEYC_NUSER` back and it its original value of 1000. I feel like this should just be removed though. All types have 4 bytes to play with, and from what I saw, this old limit was an implementation detail that is no longer relevant.

nicm:
1) yes apparently i can't multiply properly.

2) functional doesn't really mean the same thing. typically function keys covers other keys as well. or leave it as special if you don't like function, i don't really mind

3) that's fine but you need some value to check against to make sure someone does not try to set a user key that is outside the range?

daneofmanythings:
2. I will change it to FUNCTION. I'd rather go with your instinct.
3. That is why I had VALUE_MAX. All keys in this new paradigm have the same max. It wasn't super relevant though as it covered a 32bit value and that is the width of the types it was checking against, so it ended up being basically a no-op.

nicm:
Right but it was only used for user keys. I think it is better to have NUSER and an explicit check rather than relying on the type to bound it.

daneofmanythings:
Understood! I believe all points have been addressed. I rebased against master as well.

nicm:
Looks good, I have applied this to OpenBSD now, it will be in GitHub later on. Let me know when you've been able to update the other MR. Thanks!

jmtysonjr:
On macos using iterm or ghostty (w/ empty config file), tmux built from today's current master(1081876) now requires one to press the ENTER key when left-click-selecting text with the mouse in order to copy to clipboard.

In prior master branch builds one only had to left-click-select-release in order to copy selection to clipboard.

view ~/.tmux.conf
set-option -g mouse on
~

mckellygit:
Hi,

Since this commit C-MouseDrag1Pane and M-MouseDrag1Pane and C-SecondClick1Pane and M-SecondClick1Pane no longer seem to work ok.

.tmux.conf:

set -g mouse on
bind-key -T root M-MouseDrag1Pane display-message M-Drag1
bind-key -T root C-MouseDrag1Pane display-message C-Drag1

thx,
-m

daneofmanythings:
Thank you both for the regression reports! I'll look into these today.

daneofmanythings:
Both regressions should now be fixed. I had missed assigning DRAGEND type for the key code to properly be recognized, and to filter out the button assignment when there is a modifier.

daneofmanythings:
> Looks good, I have applied this to OpenBSD now, it will be in GitHub later on. Let me know when you've been able to update the other MR. Thanks!

I have updated #4946 to the current master. When these fixes go through, I'll rebase and push!

nicm:
I've applied this to OpenBSD now, let us know if there are still problems. Thanks all!

mckellygit:
Works well, thanks so much!

jmtysonjr:
Left-click-selecting text with the mouse now works with the current master build. Thanks!

PR #4954, opened 2026-03-26
Add OSC 9;4 progress bar support

### LINKED ISSUE #4833: Add support for OSC9;4 progress bars
Documented here: https://learn.microsoft.com/en-us/windows/terminal/tutorials/progress-bar-sequences

This is supported by a number of terminals including gnome-terminal and iTerm. I'm not sure what the UI for this would be for this exactly, but you could imagine putting the current progress into a variable and then adding it to the status bar somehow.

zedinosaur:
Exports two format variables, `pane_progress_bar_state` and `pane_progress_bar_progress`. `pane_progress_bar_state` possible values are `hidden`, `normal`, `error`, `indeterminate` and `paused`. `pane_progress_bar_progress` is a numerical value between 0-100.

Fixes #4833.

Documented here: https://learn.microsoft.com/en-us/windows/terminal/tutorials/progress-bar-sequences#progress-bar-sequence-format.

### DIFF
--- a/format.c
+++ b/format.c
@@ -2250,6 +2250,39 @@ format_cb_pane_pipe_pid(struct format_tree *ft)
 	return (value);
 }
 
+/* Callback for pane_pb_progress. */
+static void *
+format_cb_pane_pb_progress(struct format_tree *ft)
+{
+	char    *value = NULL;
+
+	if (ft->wp != NULL) {
+		xasprintf(&value, "%d", ft->wp->base.progress_bar.progress);
+	}
+	return (value);
+}
+
+/* Callback for pane_pb_state. */
+static void *
+format_cb_pane_pb_state(struct format_tree *ft)
+{
+	if (ft->wp != NULL) {
+		switch (ft->wp->base.progress_bar.state) {
+			case PROGRESS_BAR_HIDDEN:
+				return xstrdup("hidden");
+			case PROGRESS_BAR_NORMAL:
+				return xstrdup("normal");
+			case PROGRESS_BAR_ERROR:
+				return xstrdup("error");
+			case PROGRESS_BAR_INDETERMINATE:
+				return xstrdup("indeterminate");
+			case PROGRESS_BAR_PAUSED:
+				return xstrdup("paused");
+		}
+	}
+	return (NULL);
+}
+
 /* Callback for pane_right. */
 static void *
 format_cb_pane_right(struct format_tree *ft)
@@ -3342,6 +3375,12 @@ static const struct format_table_entry format_table[] = {
 	},
 	{ "pane_pipe_pid", FORMAT_TABLE_STRING,
 	  format_cb_pane_pipe_pid
+	},
+	{ "pane_pb_progress", FORMAT_TABLE_STRING,
+	  format_cb_pane_pb_progress
+	},
+	{ "pane_pb_state", FORMAT_TABLE_STRING,
+	  format_cb_pane_pb_state
 	},
 	{ "pane_right", FORMAT_TABLE_STRING,
 	  format_cb_pane_right
--- a/input.c
+++ b/input.c
@@ -164,6 +164,7 @@ static void	input_reset_cell(struct input_ctx *);
 static void	input_report_current_theme(struct input_ctx *);
 static void	input_osc_4(struct input_ctx *, const char *);
 static void	input_osc_8(struct input_ctx *, const char *);
+static void	input_osc_9(struct input_ctx *, const char *);
 static void	input_osc_10(struct input_ctx *, const char *);
 static void	input_osc_11(struct input_ctx *, const char *);
 static void	input_osc_12(struct input_ctx *, const char *);
@@ -2673,6 +2674,9 @@ input_exit_osc(struct input_ctx *ictx)
 	case 8:
 		input_osc_8(ictx, p);
 		break;
+	case 9:
+		input_osc_9(ictx, p);
+		break;
 	case 10:
 		input_osc_10(ictx, p);
 		break;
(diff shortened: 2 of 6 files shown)


nicm:
Looks OK but some comments:

- I'd prefer it if the function was `input_osc_9` and all the cases were handled inside it, like the other functions. That is, the check for `4` is inside the function.

- Comments like `/* maybe deal with indeterminate state differently */` are not helpful, either do it or remove the comment (or add one saying why you made that choice).

- Some code style:
  - Please use p[1] which is more readable than *(p+1).
  - Remove trailing spaces.
  - Line up variables in declarations in `input_osc_9_4` and `struct progress_bar`.
  - Single line comments need to be properly punctuated sentences.
  - Newline before `{` on function `screen_set_progress_bar` declaration.

zedinosaur:
Thank you for the thorough review! I think I've addressed all your points, please take another look.

nicm:
This looks OK to me now, thanks.

I tried this in GNOME Terminal 3.52.0 and I don't see anything, what other terminals support it?

I'd like to see it because I am wondering if it would be better if tmux supported this itself instead of passing it on to the outside terminal, then it would work even if the terminal doesn't support it. Also I don't know how intrusive it is or if there potential for misuse - does it need to be ignored by default and have a way to turn it on if you want it?

I also don't know if we should implement OSC 9, which would be trivial to do on top of this... but nobody has asked for it, and I feel like it would definitely need to be optional.

zedinosaur:
This CL doesn't pass through to enclosing terminal, it just makes them available to tmux as format strings. I do have a followup CL to exactly that, which has a terminal-features knob to enable/disable it. I can add it to this PR if you want. Also Gnome terminal does support it.

Probably worth punting on the rest of OSC 9 if no one is asking for it.

nicm:
OK makes sense. This looks good but a couple more minor things:

- You have state 4 as "paused" but the link you gave has it as "warning", as does iTerm2's docs.

- I don't mind if state 1 is "normal" or "default" but `PROGRESS_BAR_DEFAULT` and `format_cb_pane_progress_bar_state` should agree (both "normal" or both "default").

- Maybe call the variables (and associated functions) just `pane_pb_state`/`pane_pb_progress` since the `progress_bar` is a bit long?

- When the progress bar is changed in `input.c` it should probably do this to redraw the status lines otherwise you can't usefully use the format variables in them:
~~~~
server_redraw_window_borders(w);
server_status_window(w);
~~~~

This works quite well for me (albeit leaving out indeterminate), so I wonder if we should put it somewhere by default (don't do that yet):
~~~~
set -g status-right '#{?#{==:#{pane_progress_bar_state},hidden},,#[bg=#{?#{==:#{pane_progress_bar_state},error},red,#{?#{==:#{pane_progress_bar_state},paused},yellow,green}}]#{R:o,#{e|/:#{pane_progress_bar_progress},10}}#{R:_,#{e|-:10,#{e|/:#{pane_progress_bar_progress},10}}}#[default]}'
~~~~

zedinosaur:
> OK makes sense. This looks good but a couple more minor things:
> 
> You have state 4 as "paused" but the link you gave has it as "warning", as does iTerm2's docs.

Yes there's some disagreement about this one. The ConEmu documentation (where this originated AFAIK) says paused: https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC. So does the ghostty docs: https://ghostty.org/docs/vt/osc/conemu. Paused seems a little more useful to me as a concept here but I'm not married to it.

> I don't mind if state 1 is "normal" or "default" but PROGRESS_BAR_DEFAULT and format_cb_pane_progress_bar_state should agree (both "normal" or both "default").

Fixed.

> Maybe call the variables (and associated functions) just pane_pb_state/pane_pb_progress since the progress_bar is a bit long?

Done.

> When the progress bar is changed in input.c it should probably do this to redraw the status lines otherwise you can't usefully use the format variables in them:
> 
> server_redraw_window_borders(w);
> server_status_window(w);

Done.

> This works quite well for me (albeit leaving out indeterminate), so I wonder if we should put it somewhere by default (don't do that yet):
> 
> set -g status-right '#{?#{==:#{pane_progress_bar_state},hidden},,#[bg=#{?#{==:#{pane_progress_bar_state},error},red,#{?#{==:#{pane_progress_bar_state},paused},yellow,green}}]#{R:o,#{e|/:#{pane_progress_bar_progress},10}}#{R:_,#{e|-:10,#{e|/:#{pane_progress_bar_progress},10}}}#[default]}'
> 

I was thinking about as a followup to do something fancier with some built in progress bars/spinners. There's some interesting things you can do with unicode to "draw" them, https://www.npmjs.com/package/cli-spinners has a lot of ideas.

nicm:
OK I'll take another look. The defaults can't use Unicode because not all terminals are Unicode, only ACS characters.

nicm:
OK, I have applied this to OpenBSD with a few minor changes. It should be in GitHub later.

- The renamed format variables were in the wrong place in the table now, so bsearch was not finding them.

- Renamed a couple of things for consistency.

- Added a common `goto bad` in `input_osc_9` for all the error cases.

- Added an overflow check when multiplying so it can't wrap around.

Let me know if you see anything wrong.

I'd quite like the progress bar to appear in `status-right` by default and I think it would be harmless to add, if you want to do that. It has to use ASCII or ACS though, it can't be Unicode.

Thanks!

zedinosaur:
Thanks! Let me send you that other PR first to be able to pass the progress bar to the enclosing terminal and then I'll look at adding it to status-right by default. Indeed I know we'll need it to ASCII by default, I was getting ahead of myself.

PR #4955, opened 2026-03-26
Fix integer underflow in mode_tree_set_current when line_size is zero.

### LINKED ISSUE #4950: SIGBUS crash in mode_tree_get_current on choose-tree navigation (next-3.7 regression)
## Summary

`choose-tree` crashes with SIGBUS (`EXC_BAD_ACCESS`) on the first navigation key press (Down arrow) in tmux next-3.7 on macOS ARM64. Works fine on 3.6a.

## Root Cause

`mode_tree_get_current()` in `mode-tree.c` has no bounds check:

```c
void *
mode_tree_get_current(struct mode_tree_data *mtd)
{
    return (mtd->line_list[mtd->current].item->itemdata);
}
```

During tree rebuild after a key press, `mtd->line_size` becomes 0 transiently. The code at line ~435 then sets:
```c
mtd->current = mtd->line_size - 1;  // 0 - 1 = UINT_MAX (0xFFFFFFFF)
```

This causes `mode_tree_get_current` to compute:
```
line_list + UINT_MAX * sizeof(mode_tree_line)
= line_list + 0xFFFFFFFF * 24
= line_list + 0x17FFFFFFE8
```

Resulting in SIGBUS at address `0x17FFFFFFE8` (GPU carveout reserved region on macOS ARM64).

## Disassembly Confirmation

```
_mode_tree_get_current:
+0:   ldr  x8, [x0, #0xa0]     ; x8 = mtd->line_list
+4:   ldr  w9, [x0, #0xc0]     ; w9 = mtd->current (u_int)
+8:   mov  w10, #0x18           ; sizeof(mode_tree_line) = 24
+12:  umull x9, w9, w10         ; x9 = current * 24
+16:  ldr  x8, [x8, x9]        ; CRASH: x9 = 0x17FFFFFFE8
+20:  ldr  x0, [x8, #0x8]      ; (never reached)
```

## Reproduction

100% reproducible. Crash occurs on every `choose-tree` navigation, not just with grouped sessions.

```
tmux next-3.7 (HEAD-b101ffa, also HEAD-08779aa)
macOS 26.3.1 (25D2128), Apple M1 Pro (ARM64)
```

Steps:
1. `tmux new -d -s test -n win0 && tmux new-window -t test -n win1`
2. Attach: `tmux attach -t test`
3. Open tree: `Ctrl+A w` (or `Ctrl+A s`)
4. Press Down arrow → **SIGBUS, server dies**

Automated repro (Python, requires real PTY):
```python
import pty, os, time, subprocess, signal

subprocess.run(['tmux', 'kill-server'], capture_output=True)
time.sleep(0.5)
subprocess.run(['tmux', 'new', '-d', '-s', 'test', '-n', 'win0'])
subprocess.run(['tmux', 'new-window', '-t', 'test', '-n', 'win1'])

pid, fd = pty.fork()
if pid == 0:
    os.execvp('tmux', ['tmux', 'attach', '-t', 'test'])
else:
    time.sleep(1)
    os.write(fd, b'\x01')       # Ctrl+A
    time.sleep(0.2)
    os.write(fd, b'w')          # choose-tree
    time.sleep(1)
    os.write(fd, b'\x1b[B')    # Down arrow → crash
    time.sleep(0.5)
    r = subprocess.run(['tmux', 'ls'], capture_output=True)
    print("CRASHED" if r.returncode != 0 else "OK")
    os.kill(pid, signal.SIGTERM)
    os.waitpid(pid, 0)
```

**Does NOT crash** with `tmux send-keys` — only through real terminal client input (`server_client_key_callback` path).

## Crash Report (macOS)

```
Exception: EXC_BAD_ACCESS / SIGBUS
Subtype: KERN_PROTECTION_FAILURE at 0x00000017ffffffe8

Thread 0 Crashed:
  mode_tree_get_current     +16
  window_tree_key           +64
  window_pane_key           +100
  server_client_key_callback +12396
  cmdq_next                 +416
  server_loop               +80
```

Faulting address `0x17FFFFFFE8` is identical across 9 crashes, 2 different binary builds. Register `x9 = 0x17ffffffe8` = `UINT_MAX * 24` confirms the underflow.

## Suggested Fix

Add bounds check to `mode_tree_get_current` and guard `line_size - 1` when `line_size == 0`:

```c
void *
mode_tree_get_current(struct mode_tree_data *mtd)
{
    if (mtd->line_size == 0)
        return (NULL);
    return (mtd->line_list[mtd->current].item->itemdata);
}
```

And at line ~435:
```c
if (mtd->current >= mtd->line_size) {
    if (mtd->line_size == 0)
        return (0);  /* or handle empty tree */
    mtd->current = mtd->line_size - 1;
```

## Regression

- **3.6a**: No crash (tested with same setup, same repro script)
- **next-3.7** (HEAD-08779aa, Feb 25 build): Crashes
- **next-3.7** (HEAD-b101ffa, Mar 24 build, includes PR #4936 fix): Still crashes

PR #4936 (use-after-free in `mode_tree_display_menu`) does not fix this — different bug.

15935648a:
When line_size is 0, mtd->line_size - 1 underflows to UINT_MAX for the u_int type, causing out-of-bounds access. Guard against this in mode_tree_up, mode_tree_down, and mode_tree_set_current.

fix #4950

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -301,6 +301,8 @@ mode_tree_clear_tagged(struct mode_tree_list *mtl)
 void
 mode_tree_up(struct mode_tree_data *mtd, int wrap)
 {
+	if (mtd->line_size == 0)
+		return;
 	if (mtd->current == 0) {
 		if (wrap) {
 			mtd->current = mtd->line_size - 1;
@@ -317,6 +319,8 @@ mode_tree_up(struct mode_tree_data *mtd, int wrap)
 int
 mode_tree_down(struct mode_tree_data *mtd, int wrap)
 {
+	if (mtd->line_size == 0)
+		return (0);
 	if (mtd->current == mtd->line_size - 1) {
 		if (wrap) {
 			mtd->current = 0;
@@ -433,6 +437,8 @@ mode_tree_set_current(struct mode_tree_data *mtd, uint64_t tag)
 		return (1);
 	}
 	if (mtd->current >= mtd->line_size) {
+		if (mtd->line_size == 0)
+			return (0);
 		mtd->current = mtd->line_size - 1;
 		if (mtd->current > mtd->height - 1)
 			mtd->offset = mtd->current - mtd->height + 1;
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -367,6 +367,8 @@ mode_tree_swap(struct mode_tree_data *mtd, int direction)
 void *
 mode_tree_get_current(struct mode_tree_data *mtd)
 {
+	if (mtd->line_size == 0)
+		return (NULL);
 	return (mtd->line_list[mtd->current].item->itemdata);
 }
 


creatornader (commented):
Thanks for picking this up! The guards on `mode_tree_up`, `mode_tree_down`, and `mode_tree_set_current` look right.

One suggestion: `mode_tree_get_current()` itself has no bounds check either:

```c
void *
mode_tree_get_current(struct mode_tree_data *mtd)
{
	return (mtd->line_list[mtd->current].item->itemdata);
}
```

If any other code path calls `mode_tree_get_current` when `line_size == 0`, it will still dereference a garbage index. A defense-in-depth guard here would make it robust against future callers:

```c
void *
mode_tree_get_current(struct mode_tree_data *mtd)
{
	if (mtd->line_size == 0)
		return (NULL);
	return (mtd->line_list[mtd->current].item->itemdata);
}
```

(This is the function that was at the top of the crash stack in issue #4950 — the SIGBUS at `mode_tree_get_current +16`.)

15935648a:
Thanks for the suggestion! I've added the bounds check to `mode_tree_get_current()` as you suggested.

nicm:

PR #4957, opened 2026-03-27
Add three new fuzzers

DavidKorczynski:
The goal is to extend fuzzing coverage by way of OSS-Fuzz. Latest report available here:
https://storage.googleapis.com/oss-fuzz-coverage/tmux/reports/20260327/linux/src/report.html


nicm:
Good one, applied now, thanks.

It feels like `input-fuzzer.dict` could be added to a bit but I don't really know how it works, maybe what is there is enough...

nicm:
I think there is something wrong with the format fuzzer, because ossfuzz has opened this issue: https://oss-fuzz.com/testcase-detail/5348870392709120

image

The test file is a single `#`: 
[clusterfuzz-testcase-minimized-format-fuzzer-5348870392709120.txt](https://github.com/user-attachments/files/26431989/clusterfuzz-testcase-minimized-format-fuzzer-5348870392709120.txt)

However, neither Valgrind nor ASAN show any problem with this, and as far as I can tell from code inspection, a single `#` should not be a problem.

Let me know if you can see an issue.

nicm:
Hold on, think I have found it...

nicm:
OK ignore me, I found the problem:
~~~~
--- a/format.c
+++ b/format.c
@@ -5463,7 +5463,7 @@ format_expand1(struct format_expand_state *es, const char *fmt)
                        buf[off++] = *fmt++;
                        continue;
                }
-               if (*fmt++ == '\0')
+               if (*++fmt == '\0')
                        break;

                ch = (u_char)*fmt++;

nicm:
This one, however, does not appear to be a tmux problem:
image
This seems to be a pessimistic case in `regcomp`, the command parses correctly for me but it takes five seconds in `regcomp` and the memory use jumps.

[testcase.txt](https://github.com/user-attachments/files/26433043/testcase.txt)

~~~~
$ time ./tmux run -C "$(cat testcase.txt)"
syntax error

real    0m6.496s
user    0m0.011s
sys     0m0.015s

PR #4958, opened 2026-03-27
window-copy: clamp data->oy to prevent underflow after history shrinks

futpib:
When in copy mode with a large scroll offset and the window is resized so that history shrinks, `data->oy` can exceed `screen_hsize` causing an unsigned integer underflow in the `py` computation. Clamp `data->oy` in `window_copy_resize` and `window_copy_cmd_refresh_from_pane` before the subtraction.

Fixes: #4956


nicm:

PR #4965, opened 2026-04-01
Add WAYLAND_DISPLAY to default update-environment list

WGH-:
This environment variable serves the same role as `DISPLAY` on X.

Just like with `DISPLAY`, this helps fixing the problem where the tmux server was initially started in Wayland session, but later on used through SSH, and programs incorrectly inherit this env var, displaying their GUI _somewhere_ instead of falling back to TUI or failing explicitly. Or vice versa.

 And just like with DISPLAY and other variables in the update-environment list, this change is safe if you don't use Wayland. Only harmless `-WAYLAND_DISPLAY`/`unset WAYLAND_DISPLAY` is emitted.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -1024,7 +1024,8 @@ const struct options_table_entry options_table[] = {
 	  .scope = OPTIONS_TABLE_SESSION,
 	  .flags = OPTIONS_TABLE_IS_ARRAY,
 	  .default_str = "DISPLAY KRB5CCNAME MSYSTEM SSH_ASKPASS SSH_AUTH_SOCK "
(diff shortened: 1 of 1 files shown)


nicm:
I wonder if we should set the default `update-environment` via `configure`/`Makefile.am` like we do `TERM`/`TMUX_CONF`/`lock-command`, because this is really a Linux thing, and there are some others like `MSYSTEM` that is Cygwin-only.

nicm:
Maybe it isn't worth the hassle for the same of a dozen or so environment variables...

nicm:

PR #4967, opened 2026-04-02
fix: NULL deref in sort.c

daneofmanythings:


nicm:

PR #4968, opened 2026-04-02
feat: Added responsive status redraws for `pane_current_command`

daneofmanythings:
The issue this PR solves is delayed redraws of the status line when using the format variable `pane_current_command`. Before, the status wouldn't be triggered to redraw when the shell had a new command ran in it. This left the format variable feeling laggy, often displaying outdated information until manual intervention.

I have added a new stateful field to `struct window_pane` to store the last "seen" foreground pid running at the pane's file descriptor. A new check is make across all clients in `server-client.c:server_client_loop`, to determine if there is a new fg pid at each of the panes being displayed by the client. Finding one such pane will exit early and trigger a status redraw.

The added complexity is minimal, but not nothing considering currently this only affects a single format variable. I find this to be the most useful format variable for panes on the status line, so I may be biased towards this feature.
@mgrant0

### DIFF
--- a/names.c
+++ b/names.c
@@ -91,13 +91,13 @@ check_window_name(struct window *w)
 		evtimer_del(&w->name_event);
 
 	w->active->flags &= ~PANE_CHANGED;
+	server_status_window(w);
 
 	name = format_window_name(w);
 	if (strcmp(name, w->name) != 0) {
 		log_debug("@%u new name %s (was %s)", w->id, name, w->name);
 		window_set_name(w, name);
 		server_redraw_window_borders(w);
-		server_status_window(w);
 	} else
 		log_debug("@%u name not changed (still %s)", w->id, w->name);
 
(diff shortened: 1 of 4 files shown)


nicm:
Hmm not sure about this.

I think this should maybe `automatic-rename` which is the default way of using `pane_current_command`... but we already slowed that down because working out the name was so slow on some platforms.

daneofmanythings:
I completely missed that somehow. I have reworked things to use `automatic-rename`.

Tracking state is still the same. The big difference is that the `fg_pid` is checked in the existing pane evaluation logic. Then in `names.c:check_window_names`, the status redraw logic is pulled out of the conditional that checks for a window name change. Unchanged active panes have been filtered out by that point, so a status redraw is warranted on clients displaying the window.

daneofmanythings:
I don't feel strongly about this as a feature. I'm closing it for now, and it can be revisited later if the need arises.

PR #4972, opened 2026-04-03
Add a capability for OSC 9;4 progress bars and set in the client

### LINKED ISSUE #4833: Add support for OSC9;4 progress bars

zedinosaur:
Fixes #4833

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -41,6 +41,7 @@ static void	server_client_check_redraw(struct client *);
 static void	server_client_check_modes(struct client *);
 static void	server_client_set_title(struct client *);
 static void	server_client_set_path(struct client *);
+static void	server_client_set_progress_bar(struct client *);
 static void	server_client_reset_state(struct client *);
 static void	server_client_update_latest(struct client *);
 static void	server_client_dispatch(struct imsg *, void *);
@@ -2056,6 +2057,7 @@ server_client_check_redraw(struct client *c)
 			server_client_set_title(c);
 			server_client_set_path(c);
 		}
+		server_client_set_progress_bar(c);
 		screen_redraw_screen(c);
 	}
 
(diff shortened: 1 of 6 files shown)


nicm:
This looks OK, thanks, a few comments:

- I think it should check if the state is changed - the is no point in setting the progress bar every time even if it has not changed. You have it in `struct client` but don't check if it has changed. I'd probably move it to `struct tty` and do the check in `tty_set_progress_bar`.

- `/* Terminal has sixel capability. */` -> `/* Terminal supports the OSC 9;4 progress bar. */`

- `struct tty* tty` -> `struct tty *tty` (space before `*`) in the `tty_set_progress_bar` definition in `tty.c`.

zedinosaur:
Thanks for the review! Comments inline.

> This looks OK, thanks, a few comments:
> 
> I think it should check if the state is changed - the is no point in setting the progress bar every time even if it has not changed. You have it in struct client but don't check if it has changed. I'd probably move it to struct tty and do the check in tty_set_progress_bar.

I now check to make sure it changed but I didn't move it into struct tty since I was following the pattern for title and path, where they're stored in client. Still happy to make that change but it seems like those two at least should change as well?

> /* Terminal has sixel capability. */ -> /* Terminal supports the OSC 9;4 progress bar. */

Done.

> struct tty* tty -> struct tty *tty (space before *) in the tty_set_progress_bar definition in tty.c.

Done.

zedinosaur:
Let me know if you still want me to move this into struct tty. Happy to do it, just wanted to double check it was the right move.

nicm:
This looks good to me, thanks. The OpenBSD tree is currently locked for release but I'll apply this once it is open again.

nicm:
I've applied this now with some minor changes (I changed `tty_set_progress_bar` so `pb` is pointer since passing structs by value is not really the done thing in C and tends to cause confusion). Should be in GitHub later. Thanks!

jamesharris-garmin:
when might this end up in a release?

PR #4973, opened 2026-04-03
Cache getpwuid results in format callbacks

bmaurer:
## Summary

On Linux with nscd, each `getpwuid()` call creates a new Unix domain socket connection to the nscd daemon. The format callbacks `format_cb_user()` and `format_cb_client_user()` call `getpwuid()` on every format expansion with no caching. With many control mode clients attached (e.g. stale iTerm2 `-CC` sessions), each maintaining subscriptions that trigger periodic format expansion, the uncached calls can saturate the single-threaded event loop and make the server unresponsive.

- Cache the server username in a static variable in `format_cb_user()` since `getuid()` never changes
- Cache the peer username in `struct client` for `format_cb_client_user()` since the peer uid is fixed at connection time

This follows the same pattern already used by `find_home()` in `tmux.c`.

### DIFF
--- a/format.c
+++ b/format.c
@@ -1605,9 +1605,13 @@ format_cb_client_user(struct format_tree *ft)
 	struct passwd	*pw;
 
 	if (ft->c != NULL) {
+		if (ft->c->peer_user != NULL)
+			return (xstrdup(ft->c->peer_user));
 		uid = proc_get_peer_uid(ft->c->peer);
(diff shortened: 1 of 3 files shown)


nicm:
Applied to OpenBSD now with some trivial changes (I renamed `peer_user` to just `user` since it is just the client's user really), will be in GitHub later. Thanks!

Issue #4977, opened 2026-04-06
`command-prompt -N` forwards terminating non-digit key to underlying pane


barrettruth:
### Issue description

`command-prompt -N` accepts on the first non-digit key but that same key is still processed afterwards.

### Reproduction

Reproducible on minimal tmux config + latest local build.

```bash
tmux -Ltest kill-server
tmux -Ltest -f/dev/null new-session 'bash -lc '\''i=0; while :; do printf "tick %04d\n" "$i"; i=$((i+1)); sleep 0.2; done'\''
```

Then inside tmux run:

```tmux
command-prompt -N "display-message -d 3000 'number=%1'"
```

Type `123` and press `<cr>`.

- <u>Actual Behavior</u>: `<cr>` is consumed by the numeric prompt and correctly exits. *However*, the `<cr>` "reaches the pain" and a newline is inserted into the pane outputs, interrupting the stream of `tick <n>` printed characters in the above script.
- <u>Expected Behavior</u>: the numeric prompt accepts and consumes non-numeric chars, leaving the pane contents unaltered.

### Required information

* tmux version (`tmux -V`): tmux 3.6a, and also reproduced with a local build from current Git master.
* Platform (`uname -sp`): Linux x86_64
* Terminal in use (xterm, rxvt, etc): Ghostty
* $TERM *inside* tmux (`echo $TERM`): tmux-256color
* $TERM *outside* tmux (`echo $TERM`): xterm-ghostty
* Logs from tmux (`tmux kill-server; tmux -vv new`): not attached yet

***NOTE***: I am for sure willing to work on this (I already have a solution in mind).

Thanks for your consideration.

nicm:
If you want this it work differently I think there would need to be a flag so copy mode works as it does now.

But a nicer thing to do might be to add a flag for a "validation" format, so you would do something like:
~~~~
command-prompt -F '{#!:#{m/r:[^0-9],#{input}}'
~~~~
And it would not let you hit `Enter` if the format was false.

Or you could do one where the format was evaluated per character and it would not add the character to the input if the format is false. Depends what use case you have in mind.

PR #4978, opened 2026-04-06
Add unfrozen mode for command-prompt

barrettruth:
Add an unfrozen mode for `command-prompt`.

While testing `command-prompt` on top of continuously updating panes, I noticed that it always took the frozen path. `display-message` already has `-C` for the unfrozen case, so this follows the same idea for prompts instead of changing the default behaviour globally.

Implementation-wise, `-C` just gives prompts an unfrozen path so prompt setup can skip `TTY_FREEZE` instead of taking the normal frozen path. Without `-C`, `command-prompt` keeps the current behaviour. `tmux.1` documents the new option.

## Testing

Since this is my first tmux PR I tested the crap out of it with a local tmux build.

- compared plain `command-prompt` with `command-prompt -C` over continuously updating pane output
- checked `-1`, `-k`, `-N`, and multiple prompts
- checked cancel behaviour
- checked resize while the prompt was open
- checked prompt from a popup
- checked two attached clients and verified the prompt remained local to the client where it was opened

I also noticed #4977 in my testing! (This PR does not change that behaviour)

***NOTE***: This is my first PR to tmux.

### DIFF
--- a/cmd-command-prompt.c
+++ b/cmd-command-prompt.c
@@ -42,8 +42,8 @@ const struct cmd_entry cmd_command_prompt_entry = {
 	.name = "command-prompt",
 	.alias = NULL,
 
-	.args = { "1beFiklI:Np:t:T:", 0, 1, cmd_command_prompt_args_parse },
-	.usage = "[-1beFiklN] [-I inputs] [-p prompts] " CMD_TARGET_CLIENT_USAGE
+	.args = { "1CbeFiklI:Np:t:T:", 0, 1, cmd_command_prompt_args_parse },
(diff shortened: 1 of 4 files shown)


nicm:
Looks good - applied to OpenBSD now, will be in GitHub later. Thanks!

barrettruth:
@nicm sorry, I'm new to tmux. wdym "applied to openbsd now?"

nicm:
This is explained here: https://github.com/tmux/tmux/wiki/Contributing#source-code-and-development-process

barrettruth:
Thanks!

PR #4979, opened 2026-04-07
refactor: Combined the code for `split-window` and `new-pane`

daneofmanythings:
I have combined the code for `split-window` and `new-pane` as requested in #4800. I have merged the `split-window` code into `cmd-new-pane.c` because it is a more general name. Here is an overview of what I have done.

- The command now operates off of two modes, creating a window that is `FLOATING` or `TILED`. `split-window` defaults to creating a `TILED` window and `new-pane` a `FLOATING` one.
- I have added a new flag `-T` which takes an argument. Right now the options are `floating/f` or `tiled/t`. This determines the mode for the command run.
- Factored out the respective code for creating the `struct layout_cell` for each mode. There was no overlap. (the `struct spawn context` construction is the same.)
- Put some style specific code for `FLOATING` mode into its own function. Its presence broke up the logical flow in the main `*_exec` callback.

**This PR has not been exhaustively tested yet!** I did a cursory smattering of commands and it appears to function fine. Most of what I have done is copy-pasting out certain parts of code, so I don't expect things to break. I would like some feedback on the structure of what I have done before I test all the options to make sure they function.

What do you yall think? @nicm @mgrant0


daneofmanythings:
Okay, I have tested the options for this. It appears to be working with one exception. I found a null deref in `layout.c:265:layout_fix_offsets1` when using the `-f` option. This crash is also in the `floating-panes` branch. I will look into fixing it with a new PR. As for this PR, I am ready for feedback!

nicm:
Looks good, although I didn't test it yet :-). A few things:

- You will need to check with @mgrant0 about the `cmd-find.c` change.

- There are two spaces between `type]` and `[shell` in `cmd_new_pane_entry` and `cmd_split_window_entry`, but should only be one.

- In `cmd_new_pane_get_type`, the string comparisons - `strncmp(val, "floating", (sizeof "floating") - 1)` will allow anything starts with `floating`, so `floating-NOT`, and the same for the `f` one. And I think using `strcasecmp` would be nice. So probably these should just be `strcasecmp(val, "floating") == 0 || strcasecmp(val, "f") == 0)` to allow exactly `floating` or `f` but case insensitive.

- I actually think it would be better to leave the file as `cmd-split-window.c` and delete `cmd-new-pane.c` since that is what it is in `master`, so we will not lose the CVS history. I know this is probably a bit of a pain to do now so if it is too much faff leave it and we will just live with it...

- In the man page, convention is that each new sentence gets a new line. So new line after every `.` (just a newline in the mdoc source, not a displayed new line with `.Pp`). This means later changes are likely to be easier to read, particularly small ones.

daneofmanythings:
I have implemented your suggestions. A few notes:
- I spoke briefly with @mgrant0 about `cmd-find.c` and have reverted the change. It isn't relevant in this PR anyway.
- I have retooled how to specify the pane "type". The option flag is now `M` (mode) because there isn't really a "type" of pane. The command is spawning a pane in a "mode". The option now only accepts `f` and `t`. This simplified the code and the man page now more clearly describes the behavior.
- It didn't take long to change things back to `cmd-split-window.c`. I didn't consider the loss of history.

nicm:
This looks good but can you keep `SPLIT_WINDOW_TEMPLATE` rather than changing it.

daneofmanythings:
Oh, of course. I missed that one. Fixed.

PR #4981, opened 2026-04-07
feat: Default behavior for pane border buttons

daneofmanythings:
I have added a simple format variable to track whether or not the use has set the mouse on, and a default `[x]` button to kill the pane to `pane-border-format`. 

Is this aesthetic appropriate for the default behavior?

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -468,6 +468,7 @@ key_bindings_init(void)
 
 		/* Mouse button 1 down on default pane-border-format */
 		"bind -n MouseDown1Control9 { display-menu -t= -xM -yM -O -T 'Kill pane #{pane_index}?' 'Yes' 'y' { kill-pane -t= } 'No' 'n' {}}",
+		"bind -n MouseDown1Control8 { resize-pane -Z }",
 
 		/* Mouse wheel down on status line. */
 		"bind -n WheelDownStatus { next-window }",
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -467,7 +467,7 @@ key_bindings_init(void)
 		"bind -n C-MouseDown1Status { swap-window -t@ }",
 
 		/* Mouse button 1 down on default pane-border-format */
-		"bind -n MouseDown1Control9 { display-menu -t= -xM -yM -O -T 'Kill pane #{pane_id}?' 'Yes' 'y' { kill-pane -t= } 'No' 'n' {}}",
+		"bind -n MouseDown1Control9 { display-menu -t= -xM -yM -O -T 'Kill pane #{pane_index}?' 'Yes' 'y' { kill-pane -t= } 'No' 'n' {}}",
 
 		/* Mouse wheel down on status line. */
 		"bind -n WheelDownStatus { next-window }",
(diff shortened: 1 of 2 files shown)


nicm:
Is there any need for `mouse_on` when you can use `mouse`? Albeit slightly more verbosely - `#{?#{==:#{mouse},on},...`. (Maybe it would be nice to have something like  `?==`/`?!=` as a shorthand).

I'm not really wild about close without a prompt... maybe it should use `confirm-before` like `C-b &` and `x`?

daneofmanythings:
I wasn't aware options could be put into formats! That works just fine. Those shorthands could reduce some nesting, but I'm fine with where it is at now. 

I have added a confirmation, but I have put it in the form of a menu. It feels more ergonomic when using a mouse for the confirmation to pop up right by the mouse. Let me know how that feels to you. I can certainly change it to using `confirm-before`.

nicm:
I like this but I think `Confirm?` should be in the menu title and the options should be `Yes (y)` and `No (n)`, so it looks like this:
~~~~
display-menu -T '#[align=centre]Confirm?' Yes y '' No n ''
~~~~
Or maybe `Kill pane #{pane_id}?` would be better than `Confirm?`?

daneofmanythings:
Oh yes. I like the `Kill pane #{pane_id}` title. I have it set to that now. There is a `%` being displayed with the pane title though, not sure why.

nicm:
Oh I forgot that `mouse` would evaluate to 1. I think you could just do `#{?mouse,...` then.

nicm:
The `%` is part of the pane ID, you could use `pane_index` instead to use the pane index instead of ID.

daneofmanythings:
Ah, very nice. I have changed to `pane_index` and cleaned up the conditional logic.

nicm:
Looks good, do we want any more buttons? Maybe `z` or something for zoom/unzoom?

daneofmanythings:
I have added zoom. This idea comes from the `floating_panes` branch. If i remember right, the current thinking for symbols on the pane is `_` for minimize, `~` for floating, `^` for zoom, and `x` for close. I figured I'd add what is relevant to `master` before finishing the feature on `floating_panes`. How do you feel about the choice of symbols? @mgrant0 Did I get them right? I'm struggling to find the relevant correspondence.

nicm:
Personally I think random ASCII characters for icons look a bit funny, why not `[z]` for zoom?

mgrant0:
Could we use ACS unicode characters to put some nice glyphs there if available and fall back to these 7-bit alternatives otherwise?  I know the user could modify the format to use ACS if they wanted to.  But what do you think about a reasonable default to use ACS and if not use these other characters we've chosen?

nicm:
I thought of that but the ACS character set is pretty limited and nothing jumps out as being a great match:

image

daneofmanythings:
Something to to consider: I added condition that changes the face of zoom from `[z]` to `[u]` when a pane is zoomed.

nicm:
This looks good to me, I will apply it when the OpenBSD tree unlocks and we can tweak the format/characters later if needed (we will want to add the other icons for floating panes anyway later).

nicm:

Issue #4987, opened 2026-04-08
Add Choose a session and Detach to session mouse menu


psztoch:
### Issue description

Please add:
- Choose a session (CTRL-B s)
- Detach (CTRL-B d)
to mouse menu available from session name.

P.S. `Rename (n)` shortcut conflict with `Next (n)` in session menu.

daneofmanythings:
I have submitted a PR to fix the key bind conflict. As for the feature you want, I can't speak to whether or not it will be added. However, you can customize your config to achieve the desired behavior. Give this a try in your `tmux.conf`:
```tmux
bind -n MouseDown3StatusLeft "display-menu -t= -xM -yW -T '#[align=centre]#{session_name}' 'Next' 'n' {switch-client -n} 'Previous' 'p' {switch-client -p} '' 'Renumber' 'N' {move-window -r} 'Rename' 'r' {command-prompt -I \"#S\" {rename-session -- '%%'}} '' 'New Session' 's' {new-session} 'New Window' 'w' {new-window} 'Choose Session' 'c' { choose-tree -Zs } 'Detach Session' 'd' {detach-client}"
```
I took the default session menu string and added the fields you wanted.

psztoch:
The proposed functionality only makes sense if it is included in the default configuration. The goal was to provide access to *detach* command in a simple way (accessible via the mouse) so that a typical-low-tech user on generic plain linux, who executes remote instructions from a helpdesk, could easily use tmux.

Thank you for the example — it taught me a bit more, and I will use it in another case.

nicm:
I think we should add detach but not choose. The menu is `DEFAULT_SESSION_MENU` at the top of `key-bindings.c`.

nicm:
I've added detach, should be in GitHub later on. It would be nice there was a way to show a submenu of some sort to switch session instead of fully entering choose mode, but I couldn't get it to work nicely (eg to open in the same place as the session menu).

PR #4989, opened 2026-04-08
fix: changed 'Rename' keybind in session menu to avoid conflict.

daneofmanythings:
Addresses part of #4987.


nicm:

PR #4990, opened 2026-04-08
bug: validate pd->job for NULL ref

### LINKED ISSUE #4964: SIGSEGV in popup_display: missing NULL check on job_run return
### Issue description

`popup_display()` in `popup.c` does not check for NULL return from `job_run()`. When `fork()`/`forkpty()` fails, `job_run()` returns NULL via its `fail:` label (`job.c:238`), and the subsequent `job_get_event(pd->job)` dereferences a NULL pointer, crashing the tmux server with SIGSEGV.

```c
// popup.c:857-861 (commit dc12da4)
pd->job = job_run(shellcmd, argc, argv, env, s, cwd,
    popup_job_update_cb, popup_job_complete_cb, NULL, pd,
    JOB_NOWAIT|JOB_PTY|JOB_KEEPWRITE|JOB_DEFAULTSHELL, jx, jy);
pd->ictx = input_init(NULL, job_get_event(pd->job),  // <-- pd->job is NULL
    &pd->palette, c);
```

Crash stack from macOS crash report (`.ips`):

```
job_get_event+0              ← KERN_INVALID_ADDRESS at 0x40 (offset of ->event in struct job)
popup_display+756
cmd_display_popup_exec+648
cmdq_next+844
server_loop+80
```

**Reproduction:** Intermittent — requires `fork()` to fail, which happens under moderate process pressure. Using `display-popup` heavily (e.g., via plugins that spawn a separate tmux server for popups and accumulate detached sessions) increases the likelihood. I have 8 crash reports over 2 days, all with the identical stack trace.

The bug is confirmed present in the current source at commit `dc12da4` (next-3.7) by code inspection — no NULL check exists between the `job_run()` call and the `job_get_event()` dereference.

**Potential fix:** Check `pd->job` for NULL after `job_run()` and handle the error (e.g., clean up `pd` and return an error code), similar to how other callers handle allocation failures.

### Required information

- **tmux version:** `next-3.7` (built from source, commit `dc12da4`). Also reproduced on `3.6a` (Homebrew).
- **Platform:** `Darwin arm` (macOS 26.3, ARM64, Apple Silicon)
- **Terminal:** Ghostty 1.3.1
- **$TERM inside tmux:** `ghostty`
- **$TERM outside tmux:** `xterm-ghostty`
- **Logs:** The crash is intermittent and depends on `fork()` failing under resource pressure, so it cannot be reliably triggered with `tmux -vv`. Full macOS `.ips` crash report with register state and memory layout available on request.

nsavvide:
This PR introduces a precheck validation on `pd->job`, and freeing the resources if `fork()` failed due to system exhaustion.

closes #4964

### DIFF
--- a/popup.c
+++ b/popup.c
@@ -858,10 +858,19 @@ popup_display(int flags, enum box_lines lines, struct cmdq_item *item, u_int px,
 		    popup_job_update_cb, popup_job_complete_cb, NULL, pd,
 		    JOB_NOWAIT|JOB_PTY|JOB_KEEPWRITE|JOB_DEFAULTSHELL, jx, jy);
 
-    if(pd->job == NULL) {
-      popup_free_cb(c, pd);
-      return (-1);
-    }
+		if (pd->job == NULL) {
+			server_client_unref(pd->c);
+			screen_free(&pd->s);
+			colour_palette_free(&pd->palette);
+			free(pd->or[0].ranges);
+			free(pd->or[1].ranges);
+			free(pd->r.ranges);
+			free(pd->title);
+			free(pd->style);
+			free(pd->border_style);
+			free(pd);
+			return (-1);
+		}
 
 		pd->ictx = input_init(NULL, job_get_event(pd->job),
 		    &pd->palette, c);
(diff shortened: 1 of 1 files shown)


nicm:
Thanks but this should either call `popup_free_cb` or add a new function `popup_free` which can be called in both places.

nicm:
This is the same as my fix I think which I have applied now.

PR #4991, opened 2026-04-09
fix: plural "windows" when session has only one window

afterfiveguy:
### Summary
Fix plural "windows" string output displayed when session has only one window 

### The bug
Simple plural "windows" is displayed instead of single "window"

### How i found it
I just created a new session and when listing it it should have displayed "1 window" instead I saw the plural "windows"

### Fix

Use conditional format to choose what display in list-sessions choose-tree output (instead of "1 windows" it's now "1 window")

### Platform
I'm on macos but Its the same on other platforms

### Before Fix / After Fix
Screenshot 2026-04-09 at 8 15 19 AM

### DIFF
--- a/cmd-list-sessions.c
+++ b/cmd-list-sessions.c
@@ -29,7 +29,7 @@
  */
 
 #define LIST_SESSIONS_TEMPLATE				\
-	"#{session_name}: #{session_windows} windows "	\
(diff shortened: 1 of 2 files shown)


nicm:
No this has come up before and it is too much hassle to do everywhere for no real purpose.

afterfiveguy:
Fair enough. This can be closed then.

PR #4993, opened 2026-04-09
Sanitize strings embedded in OSC escape sequences

### LINKED ISSUE #4992: set-titles with #{pane_title} corrupts outer terminal when title contains control characters
## System info

```
Linux x86_64
tmux 3.6a
tmux-256color (inside tmux), xterm-256color (outside)
```

- Platform: Linux (RHEL 9)
- Terminal: Windows Terminal via SSH (Remote-SSH)
- $TERM outside tmux: `xterm-256color`
- $TERM inside tmux: `tmux-256color`

## Description

When `set-titles` is `on` and `set-titles-string` references `#{pane_title}` (or `#T`), the expanded string is written raw into the OSC 0 sequence sent to the outer terminal via `tty_set_title()`. If the pane title contains C0 control characters, they can:

- Inject new escape sequences (ESC, 0x1B)
- Prematurely terminate the OSC sequence (BEL, 0x07)
- Corrupt the outer terminal's escape sequence parser

This causes the outer terminal to desync — mouse SGR sequences appear as literal text (e.g. `0;145;20M0;145;20m`), the prefix key stops working, and clicking panes produces garbage output.

## Steps to reproduce

Minimal config (`tmux -Ltest -f/dev/null new`, then set options):

```
set -g set-titles on
set -as terminal-features ',xterm-256color:hyperlinks'
set -g set-titles-string "#S / #W / #{pane_title}"
```

Then run a program that sets a pane title containing characters whose presence triggers the issue. In my case, the program was Claude Code which sets its pane title via OSC 0.

## Root cause

`tty_set_title()` in `tty.c` writes the title string raw between TSL (`\033]0;`) and FSL (`\a`):

```c
tty_putcode(tty, TTYC_TSL);
tty_puts(tty, title);
tty_putcode(tty, TTYC_FSL);
```

The title comes from `format_expand_time()` in `server_client_set_title()` which expands `#{pane_title}` to the raw stored pane title. No sanitization is applied before embedding the string in the OSC sequence.

By contrast, `screen_set_path()` already sanitizes paths using `utf8_stravis()` with `VIS_OCTAL|VIS_CSTYLE|VIS_TAB|VIS_NL` before storing them, but `screen_set_title()` only validates UTF-8 via `utf8_isvalid()`.

## Fix

I've submitted a PR with a fix that adds `utf8_sanitize_osc()` to `utf8.c` — a function that strips C0 control characters (0x00-0x1f) and DEL (0x7f) while preserving valid UTF-8 multibyte characters. `tty_set_title()` and `tty_set_path()` call this function before writing between TSL/FSL.

The approach follows the same `utf8_open()`/`utf8_append()` iteration pattern used by the existing `utf8_sanitize()` function.

chris-monardo:
## Summary

Fixes #4992

When `set-titles` is on and `set-titles-string` includes `#{pane_title}`, the expanded title is written raw into the OSC 0 sequence sent to the outer terminal. C0 control characters in the title can inject escape sequences or prematurely terminate the OSC, corrupting the outer terminal's parser.

## Changes

- Call `utf8_stravis()` in `tty_set_title()` and `tty_set_path()` before writing between TSL/FSL, using `VIS_OCTAL|VIS_CSTYLE|VIS_TAB|VIS_NL` — the same flags used for session names, window names, and paths elsewhere in tmux.

## Test plan

- Set `set-titles on` and `set-titles-string "#S / #W / #{pane_title}"`
- Open a pane where a program sets a title containing control characters
- Verify: no mouse corruption, title displays in outer terminal, prefix key works
- Verify: titles with plain ASCII and Unicode display correctly

### DIFF
--- a/tmux.h
+++ b/tmux.h
@@ -3533,6 +3533,7 @@ int		 utf8_strvis(char *, const char *, size_t, int);
 int		 utf8_stravis(char **, const char *, int);
 int		 utf8_stravisx(char **, const char *, size_t, int);
 char		*utf8_sanitize(const char *);
+char		*utf8_sanitize_osc(const char *);
 size_t		 utf8_strlen(const struct utf8_data *);
 u_int		 utf8_strwidth(const struct utf8_data *, ssize_t);
 struct utf8_data *utf8_fromcstr(const char *);
--- a/tmux.h
+++ b/tmux.h
@@ -3533,7 +3533,6 @@ int		 utf8_strvis(char *, const char *, size_t, int);
 int		 utf8_stravis(char **, const char *, int);
 int		 utf8_stravisx(char **, const char *, size_t, int);
 char		*utf8_sanitize(const char *);
-char		*utf8_sanitize_osc(const char *);
 size_t		 utf8_strlen(const struct utf8_data *);
 u_int		 utf8_strwidth(const struct utf8_data *, ssize_t);
 struct utf8_data *utf8_fromcstr(const char *);
--- a/tty.c
+++ b/tty.c
@@ -730,7 +730,7 @@ tty_set_title(struct tty *tty, const char *title)
 	    !tty_term_has(tty->term, TTYC_FSL))
 		return;
 
-	sanitized = utf8_sanitize_osc(title);
+	utf8_stravis(&sanitized, title, VIS_OCTAL|VIS_CSTYLE|VIS_TAB|VIS_NL);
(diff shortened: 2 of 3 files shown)


nicm:
Thanks, I think this should use utf8_stravis not a new function.

chris-monardo:
> Thanks, I think this should use utf8_stravis not a new function.

I think it should be ready now.

PR #4996, opened 2026-04-09
Translate keypad keys to text in prompt input

barrettruth:
Translate keypad keys to text in prompt input.

While testing `command-prompt` with keypad input, I noticed prompt entry bypasses the normal pane key translation path. This PR is related to #4977 but doesn't fully fix it.

Keypad keys arrive as `KEYC_KP_*` rather than actual printable chars. Plain prompts, `-N` and `-1` do not handle keypad entry either.

## Implementation

just normalize unmodified keypad keys` before prompt processing, so prompts then see keypad input as its printable equivalent. For example, consider `KPEnter`, which now reaches the existing prompt accept path as an ordinary `Enter` instead of being an unhandled key.

I tested this a bunch with a local tmux build, including `-1`, `-N` `-k`, keypad punctuation (`KP{.,/,*,+,-}`)

### DIFF
--- a/regress/tty-keys.sh
+++ b/regress/tty-keys.sh
@@ -60,6 +60,31 @@ assert_key () {
 
 }
 
+assert_prompt () {
+	flags=$1
+	keys=$2
+	expected=$3
+
+	$TMUX2 command-prompt $flags 'display-message -pl "%1"' > "$TMP" &
+	sleep 0.05
+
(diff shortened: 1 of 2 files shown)


nicm:
Thanks, this looks good to me. The OpenBSD tree is currently locked for release but I'll apply this once it is open again.

nicm:

PR #4997, opened 2026-04-09
Fix pane border corruption with popup overlays

peripherium:
## Problem

When a window is split and a popup is displayed, pane borders may not be
redrawn correctly. In particular, stale borders can remain visible next
to the popup.

`tty_check_overlay_range()` is invoked multiple times during line drawing,
including from `tty_draw_line_clear()`.

This causes previously computed overlay range data to be overwritten,
leading to inconsistent rendering when popups overlap pane borders.

This is a regression introduced by PR #4920.

## Reproduction

1. split a window into two panes
2. open a popup
3. resizep from the popup
2026-04-09_0x30000cx436

---

I was not able to reproduce the original issue fixed by PR #4920, so it
would be good if @taylorconor could confirm that this change does not
interfere with that fix.

### DIFF
--- a/screen-redraw.c
+++ b/screen-redraw.c
@@ -950,7 +950,7 @@ screen_redraw_draw_pane(struct screen_redraw_ctx *ctx, struct window_pane *wp)
 	struct grid_cell	 defaults;
 	struct visible_ranges	*r;
 	struct visible_range	*rr;
-	u_int			 i, j, k, top, x, y, width;
+	u_int			 i, j, k, top, x, y, width, used;
 
 	if (wp->base.mode & MODE_SYNC)
 		screen_write_stop_sync(wp);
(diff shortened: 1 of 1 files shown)


nicm:
Thanks but I don't think this is quite right. You limit the ranges to 2 but what if the popup menu is open? There could be 3 ranges - look at `popup_check_cb`.

I think it is OK for now to have 3 and a `fatalx` if there are more.

peripherium:
You're right, I overlooked the case `pd->md != NULL` in the original implementation.

Given the many code paths in `popup_check_cb()`, it's probably more sensible to switch from a static array to `xcalloc()`. This should make the code less error-prone with further modifications.

`screen_redraw_draw_pane()` isn't on a particularly frequently called path, so the memory allocation overhead of `xcalloc()` should be negligible.

nicm:
I've applied this to OpenBSD now but I changed it to call `xreallocarray` inside the loop and `free` once outside, rather than allocating and freeing for every loop.

This could be a frequently called path, tmux will very often redraw the entire pane for various reasons, but I think it will be OK.

Fix should be in GitHub later on.

Thanks!

PR #4998, opened 2026-04-10
Do not forward Enter after `command-prompt -N`

### LINKED ISSUE #4977: `command-prompt -N` forwards terminating non-digit key to underlying pane

barrettruth:
Fixes #4977.

While testing `command-prompt -N` in my previous pr I noticed the prompt accepts on `<cr> as expected but that same `<cr>` reaches the underlying pane afterwards and erroneously inserts a newline.

## The Change

Make the numeric prompt consume Enter/Return once the prompt accepts instead of being forwarded afterwards.

***Only consumes `\r, \n`. I didn't know what to do about other control sequences, such as `C-m`, which still goes through as a newline as of this PR. Input highly appreciated.***

I tested this with a local tmux build as follows:

- ran `command-prompt -N "display-message -d 3000 'number=%1'"` over a continuously updating pane
- `123<cr>` accepted without leaking newline
- other non-digit terminators pass through same as before (for example `123j` accepts then sends `j`)

### DIFF
--- a/regress/tty-keys.sh
+++ b/regress/tty-keys.sh
@@ -60,6 +60,35 @@ assert_key () {
 
 }
 
+assert_numeric_prompt () {
+	keys=$1
+	expected_prompt=$2
+	expected_pane=$3
+
+	W=$($TMUX2 new-window -P -- sh -c 'stty raw -echo && cat -v')
+	$TMUX2 command-prompt -N 'display-message -p "%%"' > "$TMP" &
+	sleep 0.05
+
+	$TMUX send-keys $keys
+
+	wait
+	sleep 0.05
+
+	actual_prompt=$(tr -d '[:space:]' < "$TMP")
+	actual_pane=$($TMUX2 capturep -pt$W | tr -d '[:space:]')
+	$TMUX2 kill-window -t$W || exit 1
+
+	if [ "$actual_prompt" = "$expected_prompt" ] &&
+	    [ "$actual_pane" = "$expected_pane" ]; then
+		if [ -n "$VERBOSE" ]; then
+			echo "[PASS] $keys -> $actual_prompt / $actual_pane"
+		fi
+	else
+		echo "[FAIL] $keys -> $expected_prompt / $expected_pane (Got: '$actual_prompt' / '$actual_pane')"
+		exit_status=1
+	fi
+}
+
 assert_key 0x00 'C-Space' # -- 'Escape 0x00' 'C-M-Space'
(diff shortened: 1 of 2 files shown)


nicm:
Why should `Enter` be special here?

barrettruth:
This was something that I was unsure about - what *other* keys to include.

However, consider the following behavior that is *definitely* wrong:

1. open numeric prompt
2. type a few digits
3. press `<cr>`

then, the `<cr>` functions to both make the prompt accept the typed digits but ALSO sends a `<cr>` to the underlying pane TOO. I think this is definitely not desired behavior.

So there are two things:

1. should `<cr>` be handled in the way this pr suggests?
2. what about other non-numeric keys?

I don't know (and seek advice) for #2 - however, I am very convinced that this PR only makes tmux "more correct."

nicm:
But the use case for the numeric prompt is for copy mode, where you do say `M-3` then press a key and the key is repeated 3 times. Why would we want to exempt `Enter` from this? I know by default `Enter` is not bound to anything it is sensible to repeat, but someone could rebind it?

barrettruth:
I misunderstood the feature here.

Issue #4837, opened 2026-02-03
clock-mode is still invisible


Keith-S-Thompson:
### Issue description

Commit 66011fe4 is supposed to fix an invisible clock problem.

I've built tmux from source, from the current tip of the master branch, commit 615c27c1 (which includes commit 66011fe4), and the problem is still there.

When I run `tmux` under `xterm +cm` (enable ANSI color-change escape sequences), `tmux clock-mode` correctly shows a clock.

When I run `tmux` under `xterm -cm` (disable ANSI color-change escape sequences), `tmux clock-mode` does not show a clock.

(This is not a problem for me. I don't use `clock-mode`. I just happened to see the fix in the git log and was curious about it.)

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

* tmux version (`tmux -V`).
  `tmux -V` reports "tmux next-3.7". Built from master branch, commit 615c27c1 .
* Platform (`uname -sp`).
  Linux x86_64 (Ubuntu 24.04.3 LTS)
* Terminal in use (xterm, rxvt, etc).
  xterm version 406
* $TERM *inside* tmux (`echo $TERM`).
  `tmux-256color`
* $TERM *outside* tmux (`echo $TERM`).
  [`xterm`](url)
* Logs from tmux (`tmux kill-server; tmux -vv new`).

Keith-S-Thompson:
Logs:

[tmux-client-281795.log](https://github.com/user-attachments/files/25355272/tmux-client-281795.log)
[tmux-server-281797.log](https://github.com/user-attachments/files/25355271/tmux-server-281797.log)

nicm:
Please try this again after #5001 and let me know if it is still a problem. Thanks!

Keith-S-Thompson:
The clock is displayed correctly in an xterm without ANSI color support, with version 3.6a-303-g9a94f486.

With ANSI color support, it's rendered with block characters. Without color support, it uses '#' characters:
```
# #####       ##### #####
#     #   #       # # 
# #####       ##### #####
# #       #   #         #   
# #####       ##### #####
```
Is there some clean way to show block characters in non-color mode?

nicm:
Might be able to use reverse if the terminal supports it but it's probably not worth the hassle for this so long as it is not completely invisible.

Keith-S-Thompson:
[...]

PR #5001, opened 2026-04-11
fix clock-mode rendering without colour-change capability

peripherium:
Ensures clock-mode works in terminals without ANSI colour-change support. Fixes issue #4837.


nicm:
Thanks. The OpenBSD tree is currently locked for release but I'll apply this once it is open again.

nicm:

PR #5004, opened 2026-04-12
Avoid crashing when a control client disconnects before identify completes

atomicstack:
## Context
A control-mode client can set `CLIENT_CONTROL` from `MSG_IDENTIFY_LONGFLAGS` and then disconnect before `MSG_IDENTIFY_DONE`. In that state `server_client_lost()` still calls `control_stop()`, but `control_start()` has not run yet so `c->control_state` is NULL and tmux crashes dereferencing it.

## Changes
Return early from `control_stop()` when `control_state` was never initialized, and clear `c->control_state` after normal teardown. This is safe because all control-specific allocations cleaned up by `control_stop()` are rooted in `control_state` and are only created after `control_start()`: read and write bufferevents, pane offset tracking, queued output blocks, subscriptions, and the subscription timer.

Add a regression script that connects directly to the tmux socket, sends only `MSG_IDENTIFY_LONGFLAGS` with `CLIENT_CONTROL`, and disconnects before `MSG_IDENTIFY_DONE`. That reproduces the crash before the fix and verifies the server survives afterward.

## Checks

- `make -j4`
- `sh regress/control-client-lost-before-identify.sh`
- `sh regress/control-client-sanity.sh`

## Related PRs
#4980 seems to cover similar ground, but doesn't address this particular scenario.

### DIFF
--- a/control.c
+++ b/control.c
@@ -828,6 +828,9 @@ control_stop(struct client *c)
 	struct control_block	*cb, *cb1;
 	struct control_sub	*csub, *csub1;
 
+	if (cs == NULL)
+		return;
+
 	if (~c->flags & CLIENT_CONTROLCONTROL)
 		bufferevent_free(cs->write_event);
 	bufferevent_free(cs->read_event);
@@ -841,6 +844,7 @@ control_stop(struct client *c)
 		control_free_block(cs, cb);
 	control_reset_offsets(c);
 
+	c->control_state = NULL;
 	free(cs);
 }
 
--- a/regress/control-client-lost-before-identify.sh
+++ b/regress/control-client-lost-before-identify.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+# Regression test for control clients that disconnect before MSG_IDENTIFY_DONE.
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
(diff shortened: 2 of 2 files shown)


atomicstack:
I hit this repeatedly in a program using control mode before applying this fix. I have 7 local macOS crash dumps from April 12, 2026, and they all show the same signature:

- macOS 15.7.4 on Apple Silicon
- `EXC_BAD_ACCESS (SIGSEGV)` / `KERN_INVALID_ADDRESS at 0x0000000000000038`
- crash on the main thread in `control_stop + 36`, called from `server_client_lost + 720`
- the crash report also says `crashed on child side of fork pre-exec`

Representative stack from `tmux-crash-dump-01.txt`:

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   tmux   control_stop + 36
1   tmux   server_client_lost + 720
2   libevent_core-2.1.7.dylib   event_process_active_single_queue + 948
3   libevent_core-2.1.7.dylib   event_base_loop + 960
4   tmux   proc_loop + 52
5   tmux   server_start + 600
6   tmux   client_main + 1040
7   tmux   main + 2032
```

The 1st and 7th dumps are about 2 hours apart (`2026-04-12 17:52:43 +0200` and `2026-04-12 19:46:26 +0200`) and still match exactly, so this was not a one-off memory smash; it was a stable teardown crash in the control-mode path.

nicm:

PR #5005, opened 2026-04-13
Add vertical window list column (`status-position left`/`right`)

mkschreder:
### Summary

`status-position` currently accepts only `top` and `bottom`. This PR extends it with `left` and `right` values that replace the horizontal status bar with a narrow vertical column on the left or right edge of the terminal, listing session windows one per row — like a sidebar tab strip.

### New options

| Option | Type | Default | Description |
|---|---|---|---|
| `status-position` | choice | `bottom` | Extended to accept `"left"` and `"right"` |
| `status-column-format` | string | `#{window_index}:#{window_name}#{window_flags}` | Format string for each row of the vertical column |
| `status-column-width` | number | `0` (auto) | Maximum column width; `0` auto-fits to the longest rendered entry |

### How it works

**Layout reservation** — `status_column_size()` computes the column width per-client by expanding `status-column-format` for every window and taking the widest result (capped by `status-column-width` when set). This width is subtracted from the usable terminal width in `resize.c` (`clients_calculate_size`, `default_window_size`) and in `tty.c` (`tty_window_bigger`, `tty_window_offset1`) so windows are sized to fit the remaining area.

**Drawing** — `status_redraw_vertical()` renders a `col_w × tty.sy` screen with one row per window, styled with `window-status-current-style` / `window-status-style` and `status-style` as background. `screen_redraw_draw_status()` blits this screen to the left or right TTY edge each frame. Pane, border, scrollbar, and pane-status drawing all shift their x-coordinates by `statuscols` when the column sits on the left.

**Mouse click-through** — Each window row is registered in a new `vertical_ranges` list on `struct status_line`. `status_get_range()` and the mouse-event path in `server-client.c` use this list to translate a click anywhere on the row into a window-select action, matching the behaviour of the horizontal status bar.

**Message and prompt overlay** — When the session has a message or command prompt active, a single row is reserved at the bottom of the terminal (overlaid on top of the last window row) for the text, while the window list column remains visible on the side.

**Popup menu positioning** — `cmd-display-menu.c` detects vertical mode and positions the window popup to the left or right of the column rather than above or below a horizontal status line.

### Bug fixed

`winlink_add()` inserts a new winlink into `s->windows` *before* the backing `struct window` is created. `status_column_size()` and `status_redraw_vertical()` iterate that tree, so they now skip any winlink with `wl->window == NULL` to avoid a null-pointer dereference in `format_defaults()` that caused an immediate server crash on every `new-window` invocation with a vertical status position.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -131,7 +131,34 @@ cmd_display_menu_get_pos(struct client *tc, struct cmdq_item *item,
 	 * status line position.
 	 */
 	top = status_at_line(tc);
-	if (top != -1) {
+	if (status_at_column(tc) != -1) {
+		/*
+		 * Vertical column mode: find this window's row in the column
+		 * and expose its y-coordinate for popup positioning.
+		 */
+		TAILQ_FOREACH(sr, &tc->status.vertical_ranges, entry) {
+			if (sr->type == STYLE_RANGE_WINDOW &&
+			    sr->argument == (u_int)wl->idx)
+				break;
+		}
+		if (sr != NULL) {
+			int	cxstart = status_at_column(tc);
+			u_int	cw = status_column_size(tc);
+
+			if (cxstart == 0) {
+				/* Left column: popup to the right of it. */
+				format_add(ft, "popup_window_status_line_x",
+				    "%u", cw);
+			} else {
+				/* Right column: popup to the left of it. */
+				format_add(ft, "popup_window_status_line_x",
+				    "%u", (cxstart > (int)w) ? cxstart - w : 0);
+			}
+			format_add(ft, "popup_window_status_line_y", "%u",
+			    sr->start);
+		}
+		top = 0;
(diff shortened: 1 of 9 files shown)


nicm:
Good start, but this is too static. What if I want to show sessions or panes or something else instead of windows?

I think there should be some mechanism to say "next line" in styles and it should then be one option like `status-format`.

The easiest thing would probably be to make `format_draw` consume from `expanded` until a new line was reached and then return a pointer to the start of the next line (that is, the character after the marker), allowing the caller to move the cursor to the next line and call `format_draw` again.

Of course it would be *nicer* to support it within `format_draw` and implement `valign` similar to `align`. But that would be much more complicated.

barrettruth:
can there be a screenshot of what this would look like on the left and right?

Think this would be informative and allow for immediate aesthetic critiques.

nicm:

PR #5007, opened 2026-04-14
Add window context to pane notification

### LINKED ISSUE #4985: Feature Request: Preserve originating context in hooks
### Issue description

Currently, when a pane is destroyed, pane-level hooks (`pane-exited`, `after-kill-pane`) lose their original window context if the destroyed pane was the last pane in that window.

Because Tmux destroys the empty window and shifts focus to the next available window before executing the hook queue, the hook implicitly runs in the context of the newly focused window. This makes it impossible to safely execute scoped layout commands (like `select-layout -E`) on pane death without resorting to `remain-on-exit` hacks.

Hook variables exist `hook_{pane,window,session,...}` but they are only sparsely populated: if a `pane-exited` hook is fired, only `hook_pane` has information.

### Desired Behavior

There should be a way to programmatically identify the full context where a hook was fired, without dropping the _upper_ context. That is, 

- `pane` hooks should populate `hook_{pane,window,window_name,session,session_name}`
- `window` hooks should populate `hook_{window,window_name,session,session_name}`
- `session` hooks should populate `hook_{session,session_name}`

The logic of this may be implemented in:

https://github.com/tmux/tmux/blob/31d77e29b6c9fbb07d032018da78db3a8a38d979/notify.c#L179-L227

maybe with something like
```c
if (w != NULL) {
		format_add(ne->formats, "hook_window", "@%u", w->id);
		format_add(ne->formats, "hook_window_name", "%s", w->name);
	} else if (wp->window != NULL) {
		format_add(ne->formats, "hook_window", "@%u", wp->window->id);
		format_add(ne->formats, "hook_window_name", "%s", wp->window->name);
	}
```

### Required information

* tmux version `3.6a`
* Platform `Linux x86_64`
* Terminal in use: `Gnome Terminal`
* $TERM *inside* tmux `tmux-256color`
* $TERM *outside* tmux `xterm-256color`

Susensio:
Given that a pane correspond to only one window, populate the appropriate fields when possible. Fixes #4985

### DIFF
--- a/notify.c
+++ b/notify.c
@@ -210,6 +210,8 @@ notify_add(const char *name, struct cmd_find_state *fs, struct client *c,
 	}
 	if (wp != NULL)
 		format_add(ne->formats, "hook_pane", "%%%d", wp->id);
+		format_add(ne->formats, "hook_window", "@%u", wp->window->id);
+		format_add(ne->formats, "hook_window_name", "%s", wp->window->name);
 	format_log_debug(ne->formats, __func__);
 
 	if (c != NULL)


nicm:
Applied to OpenBSD now with a minor fix (needed {}s), will be in GitHub later. Thanks!

PR #5013, opened 2026-04-18
Support starting copy mode directly in rectangular selection

schweikert:
Add -r flag to copy-mode and begin-selection to allow keybindings that immediately start a rectangular selection.

I use this to implement Ctrl-Click and drag for a rectangular selection:

```
bind-key -T root C-MouseDrag1Pane copy-mode -M -r
bind-key -T copy-mode-vi C-MouseDrag1Pane send-keys -X begin-selection -r
bind-key -T copy-mode-vi C-MouseDragEnd1Pane send-keys -X copy-selection-and-cancel
```

Disclosure: implemented with the help of AI (Gemini)

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -30,8 +30,8 @@ const struct cmd_entry cmd_copy_mode_entry = {
 	.name = "copy-mode",
 	.alias = NULL,
 
-	.args = { "deHMqSs:t:u", 0, 0, NULL },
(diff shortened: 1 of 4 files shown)


nicm:
Can't you bind `copy-mode -M \; send -X rectangle-on` and `send-keys -X rectangle-on \; send -X begin-selection` instead?

schweikert:
Hmmm, you are right :-) I just tested your proposed config change and it works.

I did attempt to do that before writing a patch, but somehow I must have made a mistake. Thank you!

PR #5023, opened 2026-04-22
Clarify that set-environment -r is a fork-time exclusion flag.

josephtingiris:
The current documentation for set-environment -r states that variables are "removed from the environment before starting a new process." This is ambiguous as it implies the variable is deleted from the tmux environment block entirely. In practice, the variable remains in the session/global environment but is filtered out only during the creation of child processes.

Additionally, this flag is cleared if the variable is subsequently updated (e.g., via update-environment).  This distinction is critical for users debugging why variables 'reappear' after a client attachment triggers an update-environment event, which clears the exclusion mark.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -6807,8 +6807,7 @@ The
 .Fl u
 flag unsets a variable.
 .Fl r
-indicates the variable is to be removed from the environment before starting a
-new process.
+marks the variable to be excluded from the environment passed to a forked child process; it does not delete the variable from the tmux session.
 .Fl h
 marks the variable as hidden.
 .Tg showenv


nicm:
I don't think there is much wrong with the existing text, it is pretty clear the flag is about a future action, not removing the variable immediately. And it follows right after `-u` which *is* what you want to unset it immediately.

josephtingiris:
Thanks for your time.   I agree, there's not much wrong with the existing text and it's pretty clear.  But, as a user and IMHO, the man page still isn't super clear to me.  When, exactly, or under what conditions are the set-environment flags reset?

As well, the words 'hidden', 'remove', and 'unset' are ambiguous even among programmers.  I'm not suggesting a change to the flags themselves, only to better clarify their meanings so they can be more easily understood by others.  Yes, I could continue to expand my scope and see '-u', then 'show-environment', or ultimately (thankfully) look at the source code to find my answers.  However, that's all very time consuming.

A lot can be said about brevity.  I was hoping that reading man page would save me some time.  This is what helped.  Here are tables of my findings regarding `set-environment` and `show-environment` behaviors for tmux 3.5a:

# set-environment (aka setenv)

| Flags | Requires Value? | Valid Example | Output | Result / Behavior |
|------|-----------------|--------------|--------|-------------------|
| -g | Yes | `set-environment -g VAR value` | none | Sets global variable `VAR=value` |
| -g | No | `set-environment -g VAR` | `no value specified` | Error; value required |
| -g | Invalid format | `set-environment -g VAR=value` | `variable name contains =` | Error; must pass value separately |
| -gh | Yes | `set-environment -gh VAR value` | none | Sets **hidden** global variable (not shown unless `-h`) |
| -gh | No | `set-environment -gh VAR` | `no value specified` | Error; existing value unchanged |
| -gr | No | `set-environment -gr VAR` | none | Marks variable as removed for new processes; appears as `-VAR` |
| -gr | Yes | `set-environment -gr VAR value` | `can't specify a value with -r` | Error; variable unchanged |
| -gu | No | `set-environment -gu VAR` | none | Fully unsets variable (removed entirely) |
| -gu | Yes | `set-environment -gu VAR value` | `can't specify a value with -u` | Error; variable unchanged |
| -F | Yes | `set-environment -F VAR "#{pane_id}"` | none | Expands value as a format string before setting |
| -t session | Yes | `set-environment -t mysession VAR value` | none | Sets variable for a specific session instead of global |

# show-environment (aka showenv)

| Flags | Example | Output Behavior | Notes |
|------|---------|----------------|-------|
| (none) | `show-environment` | Lists all session variables | Default: current session |
| -g | `show-environment -g` | Lists global variables | Includes active variables only |
| -g (with removed vars) | `show-environment -g` | `-VAR` prefix | Indicates variables marked removed via `-r` |
| -h | `show-environment -h` | Includes hidden variables | Hidden vars omitted by default |
| -gh | `show-environment -gh` | Shows global + hidden vars | Needed to see variables set with `-h` |
| -s | `show-environment -s` | Outputs shell commands | Format: `VAR=value; export VAR` |
| -t session | `show-environment -t mysession` | Shows variables for that session | Overrides current session |
| variable name | `show-environment VAR` | Shows only that variable | If omitted, shows all |

Issue #4838, opened 2026-02-03
Feature Request: Hybrid/Relative Line Numbers in Copy Mode


pattames:
# Feature Request: Hybrid/Relative Line Numbers in Copy Mode

## Summary

Add support for displaying relative line numbers (and optionally hybrid line numbers) in copy mode, similar to Vim's `relativenumber` and `number relativenumber` options.

## Motivation

When navigating in copy mode with vi-style keybindings, users frequently use motion commands with counts (e.g., `5j`, `12k`) to jump to specific lines. Currently, calculating the correct count requires mental arithmetic based on the `[current/total]` indicator or manually counting lines.

Vim solved this ergonomic problem with relative line numbers, which display each line's distance from the cursor. This makes count-based navigation immediate and intuitive—you see "7" next to a line, you type `7j` to get there.

Given that tmux's copy mode already embraces vi-style navigation, relative line numbers would be a natural enhancement.

### Option 1: Relative line numbers
Display each line's distance from the cursor position:

```
  3 │ some output above
  2 │ more output
  1 │ line just above cursor
  0 │ ← cursor line
  1 │ line just below cursor
  2 │ more output
  3 │ some output below
```

### Option 2: Hybrid line numbers (preferred)
Show the absolute line number at the cursor, with relative numbers elsewhere (mirrors Vim's `set number relativenumber`):

```
  3 │ some output above
  2 │ more output  
  1 │ line just above cursor
847 │ ← cursor line (absolute line number)
  1 │ line just below cursor
  2 │ more output
  3 │ some output below
```

This provides both context (where am I in the buffer?) and efficient navigation (how far to jump?).

## Suggested Configuration

```tmux
# Possible new options
set -g copy-mode-line-numbers off        # default, current behavior
set -g copy-mode-line-numbers relative   # relative only
set -g copy-mode-line-numbers hybrid     # absolute at cursor, relative elsewhere
set -g copy-mode-line-numbers absolute   # absolute line numbers for all lines
```

### Required information

* tmux version (`tmux -V`): tmux 3.6a
* Platform (`uname -sp`): Linux x86_64
* Terminal in use (xterm, rxvt, etc): Alacritty
* $TERM *inside* tmux (`echo $TERM`): tmux-256color 
* $TERM *outside* tmux (`echo $TERM`): xterm-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`).

nicm:
[...]

PR #5025, opened 2026-04-23
Add copy mode line numbers

leohenon:
Implements #4838.

Adds `copy-mode-line-numbers` with these modes:

- `off`
- `absolute`
- `relative`
- `hybrid`

Also adds:

- `copy-mode-line-number-style`
- `copy-mode-current-line-number-style`

The gutter is rendered during pane redraw, so this does not resize panes or depend on a split-pane workaround. Cursor and mouse coordinates are adjusted to account for the gutter.

When copy mode is entered with the mouse, line numbers stay off.

Default is unchanged: line numbers are off unless enabled.

Tested in local shells, remote shells, and full-screen TUIs, including cursor movement, scrolling, search, selection, keyboard and mouse copy/paste, narrow panes, zoomed panes, and copy-mode exit.

Did some follow-up optimization work to keep the performance overhead of rendering line numbers as low as possible.

Benchmarked `master` vs this branch using the same local build setup and the same window size (`72x25`), in copy mode over a large scrollback buffer. For this branch, `copy-mode-line-numbers` was enabled. The table shows three
 runs for each case. “Vertical cursor movement” measures repeated `cursor-up`; “Page scrolling” measures repeated `page-up`.

| Benchmark | master | this branch |
|---|---:|---:|
 | Vertical cursor movement | 0.34s, 0.34s, 0.36s | 0.34s, 0.36s, 0.35s |
| Page scrolling | 0.04s, 0.04s, 0.04s | 0.04s, 0.04s, 0.04s |

 I did not observe a measurable slowdown relative to `master`.

I also built this locally both with and without `utf8proc`, and ran `regress/copy-mode-test-vi.sh` and `regress/copy-mode-test-emacs.sh`.

Open to feedback and adjustments.

### DIFF
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -63,6 +63,7 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 	struct client		*c = cmdq_get_client(item);
 	struct session		*s;
 	struct window_pane	*wp = target->wp, *swp;
+	struct args		*new_args = NULL;
 	int			 line_numbers;
 
 	if (args_has(args, 'q')) {
@@ -87,8 +88,12 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 	else
 		swp = wp;
 	line_numbers = 1;
-	if (event != NULL && KEYC_IS_MOUSE(event->key))
+	if (event != NULL && KEYC_IS_MOUSE(event->key)) {
 		line_numbers = 0;
+		new_args = args_copy(args, 0, NULL);
+		args_set(new_args, 'm', NULL, 0);
+		args = new_args;
+	}
 	if (!window_pane_set_mode(wp, swp, &window_copy_mode, NULL, args)) {
 		window_copy_set_line_numbers(wp, line_numbers);
 		if (args_has(args, 'M'))
@@ -102,8 +107,12 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 	if (args_has(args, 'S')) {
 		window_copy_scroll(wp, c->tty.mouse_slider_mpos, event->m.y,
 		    args_has(args, 'e'));
+		if (new_args != NULL)
+			args_free(new_args);
 		return (CMD_RETURN_NORMAL);
 	}
 
+	if (new_args != NULL)
+		args_free(new_args);
 	return (CMD_RETURN_NORMAL);
 }
--- a/cmd-copy-mode.c
+++ b/cmd-copy-mode.c
@@ -63,6 +63,7 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 	struct client		*c = cmdq_get_client(item);
 	struct session		*s;
 	struct window_pane	*wp = target->wp, *swp;
+	int			 line_numbers;
 
 	if (args_has(args, 'q')) {
 		window_pane_reset_mode_all(wp);
@@ -85,10 +86,15 @@ cmd_copy_mode_exec(struct cmd *self, struct cmdq_item *item)
 		swp = source->wp;
 	else
 		swp = wp;
+	line_numbers = 1;
+	if (event != NULL && KEYC_IS_MOUSE(event->key))
+		line_numbers = 0;
 	if (!window_pane_set_mode(wp, swp, &window_copy_mode, NULL, args)) {
+		window_copy_set_line_numbers(wp, line_numbers);
 		if (args_has(args, 'M'))
 			window_copy_start_drag(c, &event->m);
-	}
+	} else
+		window_copy_set_line_numbers(wp, line_numbers);
 	if (args_has(args, 'u'))
 		window_copy_pageup(wp, 0);
 	if (args_has(args, 'd'))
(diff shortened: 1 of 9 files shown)


nicm:
Thanks. I like the feature but I don't think this should be implemented in `screen-redraw.c`. This is a copy mode feature and it should be writing to the copy mode screen in `window-copy.c` like the indicator in the top-right.

leohenon:
Thanks, that makes sense. I’ll move it into `window-copy.c`.

leohenon:
Pushed the refactor. Re-ran builds, regress tests, and benchmark.

nicm:
Thanks, looks good, a few comments:

- Not all terminals have 256 colours so can you change `colour243` to an ANSI colour for the default `copy-mode-line-number-style`? Maybe `fg=white,dim`?

- The selection currently includes the line number. It might be nicer if it didn't, or if it was selected if any part of the line was selected? It doesn't matter too much if this is a pain to do.

- In the man page and in `options-table.c`, `current-line-number-style` should come before `copy-mode-line-numbers` so they are sorted alphabetically.

- Most of the helper functions like `window_copy_line_number_width` are now only called from callers who already have a valid  `wme` and will only be called in copy mode, so they don't need to check that `wme` is non-`NULL` and that it is set to copy mode. It would probably be better if they took the `wme` as an argument instead of `wp`. If they need the pane they can use `wme->wp`.

- In `window_copy_update_cursor`, `width` and `content_sx` should be declared at the head of the function with the other variables, not in an inner block.

- Are you intentionally turning off copying with the mouse view mode? I think we want to keep it allowed.

leohenon:
Thanks for the feedback.

 - Changed the default `copy-mode-line-number-style` to ANSI `fg=white,dim`.
 
 - Line numbers are now excluded from the visual selection. This needed a small selection clipping change so the highlight cannot extend into the gutter.
 
 - Reordered the new options in `options-table.c` and `tmux.1` alphabetically.
 
 - Passed `wme` to the helpers instead of `wp`.
 
 - Moved `window_copy_update_cursor()` declarations to the function head.
 
 - Re-enabled mouse copying in view mode;  that was unintentionally disabled while I was adjusting mouse coordinate mapping for the gutter
 
 - I also fixed an issue I noticed where scrolling could leave stale gutter values with `relative` or `hybrid` by falling back to a full redraw on that path.

nicm:
This works well now.

The only thing I can see now is that the line numbering does not match how tmux numbers lines, either for the indicator in copy mode or for `capture-pane`. Normally, line 0 is the top visible line and then lines are numbered upwards into the history so if you have 4 lines on screen and 3 in history it would be:
~~~~
3
2
1
0
1
2
3
4
~~~~
That's why the indicator always says `[0/211]` or similar when you enter copy mode.

This is a bit confusing especially if you use `goto-line` (`g` in copy mode). If I have 295 lines total and type `g 100` it goes to absolute line number 195 because this is 100 lines into the history.

I'm not really sure what we want to do here.

I'd be slightly tempted to just drop the existing line numbering scheme and change both `goto-line` and the indicator to use absolute line numbers. But we can't realistically change `capture-pane` (if we wanted it to use a different line numbering scheme we'd have to add a flag) and it is nice to be able to match up the line number you see in the indicator with what you pass to `capture-pane` and what you enter with `goto-line`. I think there are probably tools that rely on this.

The other option may be to add `default` or `history` as another choice for `copy-mode-line-numbers` and then make `goto-line` and the indicator follow the same scheme (that is, they stay as they are for `default` and anything else uses `absolute`).

leohenon:
Yes, the mismatch is an issue, but I think the second option is better here. If copy mode switches to absolute numbering entirely, `goto-line` and the indicator stop matching `capture-pane`, so we just move the inconsistency elsewhere and also change the current behaviour.

 Adding a default/history mode seems cleaner, it keeps compatibility with `capture-pane`, while letting the other modes use one consistent numbering scheme. I can try that and see how it works.

nicm:
Sounds good, thanks.

leohenon:
Added `default` mode and made the indicator follow the numbering for the other modes. `copy-mode-position-format` now uses copy mode position formats, so `scroll_position` and `history_size` stay unchanged.

nicm:
Not sure this is quite right, when I turn on `default` line numbers it shows 0-83 for the visible lines correctly but when I scroll up the line numbers all change (maybe they are meant to?) but the 0 seems in the wrong place - if I do `g 55` it goes to line 137. Funny if I go to the top and scroll down the seem to descend properly.

I think the visible lines should go 0-N and the history should go M-0. Even if that means there are two lines 1s.

nicm:
Oh maybe `g` is OK, I forgot it doesn't move the cursor only the top line (maybe it should, I dunno).

The line number indicator and `g` seem to work well in `absolute` mode.

leohenon:
I see the issue. In `default` the line numbers should repeat 0-N upward. I have fixed that.
 
For `g`, yes, it sets the top line. That should already be consistent between all modes. default keeps the existing `capture-pane` numbering, and the other modes use absolute numbering. After `g` N in any mode, the indicator and the top line should match N.

nicm:
Looks good and works well, I have applied this now with some trivial style changes (mostly wrapping long lines, some of which may have already been there). It is in OpenBSD and will be in GitHub later on.

Thanks!

PR #5027, opened 2026-04-24
feat: Added `new-pane`

daneofmanythings:
This is the second half of the request to combine `new-pane` and `split-window` in the `floating_panes` branch from #4800.

I have left in some of the code for floating panes to hopefully make it easier to merge things in the future, and noted in the code when I have done so. If this is not the right approach, let me know how it should be done and I will update the PR!

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -27,21 +27,23 @@
 #include "tmux.h"
 
 /*
- * Split a window (add a new pane).
+ * Create a new pane.
  */
 
 #define SPLIT_WINDOW_TEMPLATE "#{session_name}:#{window_index}.#{pane_index}"
 
-static enum cmd_retval	cmd_split_window_exec(struct cmd *,
-			    struct cmdq_item *);
+static enum cmd_retval	cmd_split_window_exec(struct cmd *, struct cmdq_item *);
 
-const struct cmd_entry cmd_split_window_entry = {
-	.name = "split-window",
-	.alias = "splitw",
+const struct cmd_entry cmd_new_pane_entry = {
+	.name = "new-pane",
+	.alias = "newp",
 
-	.args = { "bc:de:fF:hIl:p:Pt:vZ", 0, -1, NULL },
-	.usage = "[-bdefhIPvZ] [-c start-directory] [-e environment] "
-		 "[-F format] [-l size] " CMD_TARGET_PANE_USAGE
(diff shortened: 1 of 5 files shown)


nicm:
Thanks looks good just one thing - I don't think `last_x`/`last_y` should be `static` variables in `cmd_split_window_get_floating_layout_cell`. Maybe put them in `struct window`? I'd call them something different like `last_new_pane_x` and `y` maybe.

nicm:
Oh I think `enum new_pane_mode` could be merged into the declaration since its only used inside the function, so:
~~~~
enum { NONE, TILED, FLOATING } pane_mode = NONE;
~~~~

daneofmanythings:
I will make those changes! I have changed `new-pane` a little in my work combining `float/tile-pane`. I will include this feedback into the `floating_panes` branch with that commit, and then once that commit is accepted, I will port all the changes into this PR so there is a point of consistency between the two branches once the changes are done.

nicm:
Great thanks, let me know when this is ready to go again.

daneofmanythings:
Okay, this is ready for review again @nicm. I have removed the (currently) irrelevant floating code, I think i was wrong to include it.

nicm:
Is `window_pane_tile_geometry` going to be used in other files? If not, just leave it in `cmd-split-window.c`.

In `cmd.c`, `new_pane` should come before `new_session` to keep the list sorted.

Otherwise looks good, thanks.

nicm:
In the man page can you change `waits for a keypress (any non-mouse key)` to `waits for a key to be pressed` - `keypress` is not really a word, and I don't think we need to say mouse is excluded. Also this section could maybe benefit from being a couple more paragraphs, maybe one before `An empty ...` since this is really a different topic than the style of the pane.

daneofmanythings:
I have touched up the organization and removed "keypress."

nicm:
thanks did you see this comment as well https://github.com/tmux/tmux/pull/5027#issuecomment-4466939464

will merge this tomorrow probably

daneofmanythings:
Oh, I somehow missed that comment! Yes, `window_pane_tile_geometry` will be used by `move/join-pane`. I am working on centralizing the geometry calculations. The same thing is being done for floating geometries.

I have moved the `new_pane` entries to the correct location in `cmd.c`.

nicm:
I've applied this to OpenBSD now with some trivial line length changes. Thanks!

PR #5028, opened 2026-04-24
fix: Applying fix proposed in #4800 for `break-pane` to `join-pane`

daneofmanythings:
I have discovered that `join-pane` has the same fatal issue that `break-pane` had. I have applied the same fix to `join-pane`.


daneofmanythings:
uhg, tried to apply against the wrong base again...

PR #5029, opened 2026-04-24
fix(floating_panes): Z-index related crashing in `join-pane`

daneofmanythings:
I have discovered that `join-pane` has the same fatal issue that `break-pane` had. I have applied the fix from @nicm to `join-pane`.


daneofmanythings:
@mgrant0

mgrant0 (approved):

mgrant0:
Looks good, merged it.

PR #5034, opened 2026-04-24
Do not leak socketpair fds in `pipe-pane` when `fork()` fails.

barrettruth:
see [cmd-pipe-pane](https://github.com/tmux/tmux/blob/bcd17cf99afd564c559d04d1cb01ed7fc7e2749c/cmd-pipe-pane.c#L127-L132)

`fork()`, when failing doesn't close the pipe on both ends

(this one is a bit stupid but I'm trying to familiarize myself with the codebase :sweat_smile:)


nicm:

PR #5035, opened 2026-04-25
Have `Ctrl-[` behave like `Esc` in interactive UIs

ccjmne:
Without any extended keys protocol, `Ctrl-[` would be effectively indistinguishable from `Esc`.
In terminals that support extended keys, I'd still like to be able to emulate that "quirk" that's become traditional.

I propose to add `C-[` alongside `Esc` in interactive UIs where `Esc` is already treated as a standalone cancel/exit key.

I essentially added an extra `case '['|KEYC_CTRL` wherever we already had a `case '\033'`, as well as bound `C-[` the same way `Escape` is in `copy-mode`.

The impacted UIs are:

- `command-prompt` with `vi` and `emacs` keys,
- `copy-mode` with `vi` and `emacs` keys,
- `display-menu` mode,
- the choosers (`choose-buffer`, `choose-tree`...)

## Prior art

`C-[` is the traditional equivalent of `Escape`, and is recognised in several tools, such as Zsh's ZLE, Vi and its descendants, less; `ncurses`-powered TUIs.


nicm:
This only makes any difference when `extended-keys` is turned on, right?

ccjmne:
Hmm...
If I understand it correctly, `tmux` may negotiate for extended keys both upstream with the terminal, and downstream with the pane's active process, and its `extended-keys` option only concerns itself with the downstream protocol. The changes in this PR only affect `tmux` itself, the buck stops there.

I gathered as much from our manual:

```txt
    tmux will always request extended keys itself if the terminal supports them.  See also
    the extkeys feature for the terminal-features option, the extended-keys-format option
    and the pane_key_mode variable.
```

Therefore, I think the answer would be no.

I believe it would extend the behaviour regardless of the configuration of `extended-keys`, seeing as the interpretation of these is constrained to `tmux` itself, which *always requests extended keys* with the terminal emulator, correct?

nicm:
But how would `Escape` and `C-[` be any different without extended keys?

ccjmne:
The answer is a bit incomplete, I intend to *verify* rather than surmise, but I can't do the former right away... hopefully later tonight :-)

I think that we *do* have "extended keys" even without setting the `extended-keys` option in `tmux`, because `tmux` always requests them *for itself* with the terminal emulator, and the `extended-keys` option is only to decide how `tmux` would talk to the processes that run under it, in its panes... Is that not right?

Well, I think you're correct, we can only distinguish between `Ctrl-[` and `ESC` when we use extended keys. However, `tmux` always requests those for itself: the changes I propose here become effective:

> whenever the terminal emulator support extended keys

not:

> whenever the `tmux` option `extended-keys` is `on` (or `always`).

nicm:
I just mean that without extended keys there is no such thing as `C-[`, only `Escape`, so this should have no effect. You didn't mention extended keys so I was wondering if this was to make `C-[` work with them or you had some obscure terminal where there was a difference.

ccjmne:
Okay okay, *yes*, it's precisely that, to make `Ctrl-[` work with extended keys :-)
Currently using mostly `foot` and `ghostty`, both of which communicate by default with `tmux` via extended keys, as far as I can tell.

nicm:
OK I don't think this can do any harm, we aren't likely to use `C-[` for anything else.

nicm:
Oh, I just noticed you *did* mention extended keys, I must be blind...

Anyway this is applied to OpenBSD now, will be in GitHub later. Thanks!

ccjmne:
Nonono, you're not blind, I just amended the original message so it's a clearer introduction, for posterity.
I had never mentioned extended keys to begin with ;-)

Thank you very much, have a nice day!

PR #5036, opened 2026-04-25
Do not deref NULL job in `window_copy_pipe_run` when `job_run()` fails.

barrettruth:
this one is pretty self-explanatory


nicm:

PR #5037, opened 2026-04-25
Do not hang `run-shell` when `job_run()` fails.

barrettruth:
i caught in the same audit pass that turned up #5034 / #5036.

foreground `run-shell` hangs forever if `job_run` fails... which realistically only happens if your server is right up against its fd cap (this will almost certainly never happen).

this is the same shape of contract violation as in `67e6f824` though, so better safe than sorry!


nicm:

PR #5038, opened 2026-04-25
fix: typo when using `-p` in `join-pane`

daneofmanythings:


nicm:

PR #5039, opened 2026-04-25
fix: fixed site of null dereference in layout.c

daneofmanythings:
Here is the command to reproduce the crash: `splitw; splitw -hf`.


daneofmanythings:
@mgrant0

PR #5042, opened 2026-04-28
fix(floating_panes): null dereference in `screen_redraw_two_panes` from `screen_redraw_border_arrows`

daneofmanythings:
The border arrows regression test was crashing due to passing `NULL` as an argument to a function that dereferences that argument. The regression test now passes.


daneofmanythings:
@mgrant0

PR #5043, opened 2026-04-28
refactor(floating_panes): Merged `un/minimise-pane`, `float-pane`, and `tile-pane` into existing commands.

daneofmanythings:
This is an implementation of what was requested in #4800, to use existing commands for the new functionality.

This PR contains the following:
- `un/minimise-pane` has been absorbed by `resize-pane`.
- `float-pane` and `tile-pane` have been absorbed by `move/join-pane`.
- Some generalized functions put into `window.c`.

The original proposal was to also incorporate `break-pane`, but when testing and implementing, it didn't make much sense. `move/join-pane` is a much more usable home for floating and tiling a pane. This PR does have some bugs, but they are bugs present in the base `floating_panes` branch.

I'd love to get some feedback! @nicm @mgrant0


mgrant0:
I am looking at this.

mgrant0:
```
	.usage = "[-bdfFhv] [-l size] [-x width] [-X x-position] [-y height] "
		 "[-Y y-position] " CMD_SRCDST_PANE_USAGE,
```
Though I think I understand the `-l`, `-x`, `-X`, `-y`, `-Y` args (and if I look in the code there's a `-p` flag that you are missing in the arg list, I find these args confusing.  And I feel like if you have `-h` there should be a `-v` for vertical but -v is already in use.  Are you missing `-M type`?

If I understand correctly:

- Tiled panes handle -M t -l | -p, -h
- Floating panes handle -M f -x, -y, -X, -Y

I am having a difficult time explaining to myself how these options all work by looking at the usage message.  I think I would minimally group things differently like -x width -y height -X pos -Y pos -l size.  

I am not crazy about -M f / -M t.  It's combersome.  It feels very untmux like.  The word 'mode' is already used in another context and certainly you are not setting the 'mode' of the pane, if anything you are setting the 'layout'.  Perhaps -L is a better choice?

I think you guys might have been discussing this when I was occupied the last couple weeks.

I know X-windows has a standard way of specifying a geometry with `XxY±x±y`.  This works for floating panes.  Could something like this be extended to cover tiled panes I wonder?  For tiled panes, maybe the syntax of geometry could be something like `±s` where s is number or number%, if no + or - defaults to +, so a size arg of 5 or +5 makes the pane 5 lines or columns, -5 makes the other pane 5 and gives the new pane the rest (as in take 5 away from this one).  Similarrly for percentages -25% means give this pane 75% of the area .  So any of the numbers in the geometry could be percentages (the X, the Y, the x or y).  Then the question becomes how might one extend this syntax to say a tiled window's geometry is relative to the full window?  The letter 'w' as in `[w]±s` (where s is a size or size%) or maybe the number `0` as in `0±s`? (afterthought: I think I have this backwards, to be consistent the 'relative to window' should be +0 and that means relative to top or left depending on how it is split and -0 relative to bottom or right of entire window depending on how it is split, hmm this has issues too with -5+5 being ambiguous, let me think about this a little...).

With a standard geometry specifier, this frees up lots of letter args and maybe just -g for geometry could be used.  A geometry in the `XxY±x±y` format implies a floating pane and an geometry in the `0±s` format implies a tiled one, so you wouldn't necessarily need a -M (or rather -L) option in that case.

Just some thoughts, is this overcomplicating things or does it solve an issue before we move this to master where these args become more set in stone?

daneofmanythings:
Thank you for taking a close look @mgrant0! I will try to clarify the thought process for the new flags. I have two main goals: 1. Don't change any existing flags, and 2. Have a unified flag set for any command that interacts with pane geometry and layouts. The `-l`, `-p`, `-f`, `-b`, `-h`, and `-v` options are currently in use by `split-window`, and the `-x` and `-y` options are in use by `resize-pane`, and notably control pane width and height, either as a number of lines or percentage. This is the starting point. The remaining functionality needed is pane position for floats, which I have as `-X` and `-Y`, and pane layout (I had it as creation "mode" but I like the "layout" semantics you mentioned a lot better), which I have as `-M` with appended `t` or `f`. I would prefer to have unprefixed flags that represent the pane layout. Do you have any suggestions? Here are the flags currently used in master by `split-window` and `resize-pane`:
```
"bc:de:fF:hIl:p:Pt:vZ"
"DLMRTt:Ux:y:Z"
```
Maybe it is sufficient to have a single flag which specifies that it is meant to be floating, and have tiling be the implicit default?

I think trying to adapt the X methodology for pane geometry is more confusing than adding a few new flags.

I also use a prefix modifier for flags specifying un/minimising in `resize-pane` I'll think about how to flatten that out too. I was borrowing the method from `display-menu`, but the prefix flags in that command have a lot of options. It may not be the best fit here.

mgrant0:
The X methodology was just an idea I was toying with.  It has 2 advantages in that it's already well known and it saves args.  But it is confusing for tiled panes.  I spent some time trying to fix my issues with that but I haven't found a good way.  

Things like there is a `-h` flag and a `-v` flag so for this sort of consistency, I feel like there should probably be a tiled flag (which is the default) and a floating flag.  We didn't need that before because these were separate commands.  I need to think about this a little more tomorrow but I think there's a case that when there are options which are dependent on other options (like whether it's a tiled or floating pane) that there should be separate commands.  It feels overloaded.

Backwards compatibility is very important.  Let's  not lose that.

It's a shame we can't use -F and -T for floating and tiled because they are taken.  -L floating (or f), -L tiled (or t) would work if we can't find any single unused options or if we don't end up splitting things up again.  But really other than that, your choices are reasonable.  I would prefer a clearer usage string and group -X, -Y and -x, -y next to one another rather than keeping things strictly alphabetic.

daneofmanythings:
I see what you mean about the overloading. This wouldn't be the first command set to have flags dependent on others though. I find that grouping functionality into commands by use-case to be intuitive. Want a new pane, then use `new-pane`. Want to move a pane into a tiled layout, use `move-pane`. I tried to have the documentation clearly denote what flag works with other flags. That being said, I am open for whatever and am happy to make changes! I did my best to retain the functionality you added into these commands while incorporating them like @nicm requested.

I agree about not being able to use `-F` and `-T`. Finding semantically meaningful flags to represent what we need given what is available isn't obvious. If we do end up sticking with a prefixed flag, `-L` is a much better choice than `-M`.

After thinking more about it, I do think I can streamline what I currently have for un/minimising in `resize-pane`. Unfortunately `new-pane` and `move-pane` have different needs so I can't apply the same method.

I'll refrain from making more adjustments to this PR until we have a consensus regarding direction!

nicm:
FWIW I am happy to go either way here. The flags are already not completely consistent I think. It is nice if they are the same across commands but I don't think it matters too much, so long as they kind of make sense.

Adding `geometry` is a reasonable idea and if you want to go that way I think it is fine for new commands to have only `-g geometry`. Existing commands can have `-g` added and the existing `-x`/`-y`/etc can be left for compatibility.

I don't think I would make geometry imply whether the pane is tiled or floating - I think that wouldn't be very user friendly. An explicit "I want this tiled/floating" and an error if the wrong type of geometry is used would be better.

I suggested some "mode" or "type" for tiled or floating but yes "mode" and "type" are both used elsewhere for a different purpose. "layout" is too but perhaps it does make most sense - so maybe `-L` for layout would be best? Or if a combination flag doesn't seem right, I don't mind dropping that idea and just having maybe `-L` for floating and `-T` for tiled instead.

mgrant0:
Let me reason things through here...

`break-pane` breaks a pane into a new window but it it could also break a pane out of the layout transforming a tiled pane into a floating pane.  Calling this break-pane make sense.  I would say it needs a simple arg to say "don't put this in a new window, put it in this window but floating." (-L float). Without any geometry args, the pane's geometry should be what the next new pane would be which can be overriden with -X, -Y, -x, -y.

`join-pane` would indeed be the opposite of `break-pane`.  It shouldn't seem to make any difference if the src is a floating or tiled pane.  The dst must be a tiled pane, splitting tiled panes (as in 2 panes glued together) isn't implemented., or must be a window.

`join-pane` is synonymous with `move-pane`.  This command can be used to move a pane to another window as well.  I guess it does make sense for it to be able to _move_ it from a tiled pane in one window to a floating pane in another.  Here's where things things start to break down.  It makes no sense to join a pane to a floating pane.  Nor does not make sense to join a tiled pane to another window and have it turn into a floating pane--this actually implies that join-pane can act as break-pane if the target window is the current window!

It does make sense to move a floating pane to another window and I suppose it could make sense to move a tiled pane to another window and float.  And conversly move a floating pane to another window and tile it which is really what join-pane does.  I have an issue with joining a tiled pane to another window and floating it.

So, you might say, why do i care?  Why not just allow all these different permutations even if they defy logic no matter how counter-intuitive they are?  Let's play with some examples.  Let's use `-L float` and `-L tile` (though I did find one edge case where I needed -L tile and used -I for tIle but it's superfulous).

| Command | Description |
| --- | --- |
|join-pane -t 2|Move active pane to window 2.  If it's floating, should the default be leave it floating or join it to a tiled pane?  Based on the name of the command, I would say it should tile it.|
|join-pane -t 2 -L float|Let's say a tiled pane.  Move it to window 2 and float it.  Not very join like.|
|join-pane -t 2 -L tile|If floating, this should tile it in window 2, otherwise the -L tile arg is superfluous.|
|join-pane -L floating|On a floating pane, destination is the same window, so this would do nothing.  On a tiled pane would act like break-pane.  Very cofusing.|
|join-pane -L tile|On a floating pane it would tile it, weird.  On a tiled pane, I would geuss would do nothing.|
|move-pane -t 2|Move active pane to window 2.  If it's floating, I would expect it to remain floating.  If it's tiled, I would expect it to remain tiled.|
|move-pane -t 2 -L floating|A tiled pane becommes floating in window 2.  A floating pane just moves.|
|move-pane -t 2 -L tiled|If floating, this should tile it and move it to window 2.  If tiled, just move it.|
|move-pane -L floating|On a floating pane in active window do nothing.  On a tiled pane would float it.|
|move-pane -L tiled|On a floating pane, it would tile it.  On a tiled pane do nothing.|

It feels like `move-pane` should behave differently than `join-pane`, or rather it feels like it should have different defaults for floating panes.  In fact, it might be that `break-pane`, `join-pane`, and `move-pane` are all the same underlying function with different defaults.

`join-pane` I would say should not need the -L arg (though it could support it hidden, nor any geometry args).  The goal of `join-pane` is to join a pane to another pane, as in, split and insert.  If `join-pane` is run on a floating window, it joins it to a pane (in current or target window).  `join-pane` is the opposite of `break-pane` and `float-pane`.

`break-pane` does the reverse of `join-pane`, it takes a pane out of the layout and pushes it to a new window.  A floating pane is pushed to a new empty window as the only pane in that window.  It could have a -L option to float it in the new window I suppose.

`float-pane` is needed to do something like `break-pane` except the resulting pane is floating and the default target is always the active window.  `join-pane` would then be the opposite of `float-pane`.

`move-pane` moves panes to other windows. `move-pane` should move a floating window and keep it floating, not join it.  Here is where it diverges from `join-pane`.  I would resist trying to add a -L arg here because it becomes for full consistency you also need something like a -I flag for tIle to convert a floating pane to a tiled pane on the move (this is what `join-pane` does so it's not needed here!)  And the -L flag is the same as `break-pane`.  So I wouldn't over complicate `move-pane`, leave (or hide) those args.

This keeps `break-pane`, `move-pane` and `join-pane` symantics the same for tiled panes so backwards compatible and does the similar thing one might expect if run on a floating pane.  The documentation would then change to describe them as separate commands rather than one referring to the other.  And the new `float-pane` command.

I think simply -L as the arg meaning fLoat.  I don't see a case of needing -L float and -L tile. (or -L and -I)

I am less convinced that `resize-pane` should imply minimising it.  I would keep those as separate commands. 

Have I missed any cases?  Is there a better suggestion than -L meaning fLoat?  Pitty -f and -F are not available!

nicm:
I don't think we should add more commands, the four we have (join, break, move, resize) all don't have many flags so I think we should fit all the possibilities into them even if they don't quite match exactly.

nicm:
Most people will use the commands via key bindings, menus or status line controls so if name vs behaviour only "kind of" makes sense I think that is OK.

daneofmanythings:
@mgrant0 Thank you for the time you put in to your feedback! 

I agree that we should convert to using single flags. I would like to suggest `-O` for floating. The dominant vowel sound makes sense to me if the starting letter is taken. This also gives runway for using `I` for tiling if there ever is a need. 

I agree that the semantics and behavior around `join-pane` and `move-pane` are dubious. When I merged `new-pane` and `split-window` I aliased the commands with sensible defaults, and then completely overlooked that here and should have at least done that. Your ideas for how they should function are spot on. I can try to fence off behavior instead of aliased commands with sensible defaults to see what the code turns into. I fear that it may add more complexity than it is worth, but I would need to implement it to find out. It may be that the added behavior can be controlled by restricting flags so that the code remains basically the same and retains backwards compatibility as you pointed out! Shall I give it a go?

`break-pane` is a tricky one. The behavior is different enough that I don't think it is a good idea to dispatch to it from `join-pane`, nor do I think it should handle converting a pane to a float. This was what I first tried to do with this PR, and what I found myself doing was re-implementing major parts of `join-pane` inside of `break-pane` as they don't really share enough code. Semantically it is tempting, but I think it will add more complexity than it is worth. The case that convinced me of this is breaking a tiled pane to a new window and floating it. This shouldn't be supported behavior in my opinion.

I hope I addressed everything; if not, please point it out!

mgrant0:
-O or -L, I'm ok with either or if you can find a lower-case letter, if @nicm has an opinion, go with that.  I'll use -O in this comment to mean "the resulting pane should be floating".

Let me focus on just folding `float-pane` into `break-pane`.  I think the only material difference between float-pane and break-pane is that by default `break-pane` takes a pane and sends it to a new window.  

`break-pane -O` could do more or less what`float-pane` did: float the pane and target the active window instead of opening a new window:

| Command|Args|Target pane|Behaviour|
| --- | --- | --- | --- |
|break-pane| |tiled|move pane to new window|
|break-pane| |floating|move pane to new window, tiled full window|
|break-pane|-O|tiled|float pane into current window|
|break-pane|-O|floating|does nothing|

Is there any reason to have a way to make `break-pane` float a pane into a new window?  Meaning the pane would be pushed to a new empty window and would be floating in a field of background dots?  (background dots are fine, they represent the "desktop", I have no problem with these dots showing when there's no tiled windows).  This is the same as creating a new empty window, then using `move-pane`.

The above should eliminate `float-pane` from my previous comment.

nicm:
We use -O for sort order a lot so -L might be better... although it is used for "left" in the likes of refresh-client and resize-pane. There is probably no perfect flag so either is fine.

It is OK to do things in steps so break/join first and then think about minimize/unminimize.

Ideally I'd like to get new-pane sorted out first because it is needed before you can do anything with floating panes at all, so maybe it'd be good to add whatever flag is decided to it first.

mgrant0:
-W and -V are not used anywhere I could find.  What do you guys think of -W to mean that the resulting pane should be floating?  And tiling explicit by default as you say.  No "prefixed" flags.

No X-style geometry. -x, -y, -X, -Y are fine.

Ok don't displatch `break-pane` from `join-pane`.  I do like very much that one undoes what the other does, so running `join-pane` on a pane that was previously floated or sent to another window puts it back where it came from.

I don't want to hold this up anymore.  Making resize-pane minimise (-m) and unminimise (-u) is fine for now.

@daneofmanythings Can you please clean up the args and the arg doc string?  Make sure nothing missing.  Also you just need the u and m in the .args, you don't need "-u unminimise".  

Thanks!

daneofmanythings:
I believe I have satisfied most of what you wanted @mgrant0!

Big picture on what I have done with the flags: 
- `-V` now makes a command that defaults to floating, tile instead. `-V` can be anything, but I chose it because it kinda looks like a down arrow, this can be revisited in the future.
- Since all the new functionality is mainly in "new" commands (`move-pane` and `new-pane`), these default to floating.
- The established commands can no longer interact with floating windows. It didn't make sense for `join-pane` to float a pane, and likewise for `split-window` to create a floating pane.
- `move-pane` will float a tiled pane by default now.
- `-m` now handles all minimisation actions in `resize-pane` and is context aware. `-a` is the "all" flag to un/minimise everything.

> Ok don't displatch break-pane from join-pane. I do like very much that one undoes what the other does, so running join-pane on a pane that was previously floated or sent to another window puts it back where it came from.

I'm not sure that we have the plumbing to intelligently restore a pane to its previous layout, at least I haven't seen it in the codebase yet. I see that you did some "best fit" work for `tile-pane`, but it does not truly restore the layout. I want a way to do this too, but it may be out of scope for this PR. Let's talk about getting this done though! I  even see that there is a "layout rework" on the TODO list.

I should have a good chunk of time this week to make this PR satisfactory, so hit me with the feedback!

mgrant0:
Great! I'll look at this in detail over the next day or so. 

The `float-pane` and `tile-pane` commands saved and restored the layout.  I'm hoping that still works.

daneofmanythings:
I did not include all of the layout restoration logic. The floating geometry is cached and applied when no dimensions are given, but not the tiling logic. When tiling, one can specify exactly where in a layout to put the pane with the `join-pane` machinery. I would like to save the layout logic for tiling, but I'm not sure how to do it in a rigorous way. Here are three proxy example where what was previously implemented fails to restore the layout:
```tmux
splitw -l20%; splitw -h; float-pane; tile-pane
splitw; splitw; float-pane; tile-pane
splitw -h; float-pane; tile-pane
```
It appears as though the layout turns into vertically stacked panes with equal y-dimensions when calling `tile-pane`. The logic to pull from a stored layout when tiling can be added to `join-pane`, but I haven't done it yet for the above reason. How would you like to handle this?

mgrant0:
I definitely see the issues with the 3 examples you provided.  It put the pane back in the layout but always stacked vertically there needs to be some last_layout flag telling it how it was layed out.  The pane proportions in my mind starts to be a headache.  What if the size of the window or pane changed?  I think I wouldn't go down that road.  If we can minimally save the last place in the tree the floated pane was and put it back there, for me, that's fine.

I think we may have a slight misunderstanding on `move-pane`.  I don't think `move-pane` should float a pane.  `break-pane` with some arg should float a pane.  

I don't see a -V option to `break-pane`.  

If I start tmux and then `move-pane` and get a single floating pane, then ctrl-d close the floating pane, tmux crashes.  It should gracefully exit.

I know we go around and around on the option, but could we use -W?  for me, -v or -V seems like some sort of verbose mode, though i see your arrow, an arrow doesn't seem to make sense to me.  -W could be for window (though ya i know it's a pane in tmux but it's a window like thing in a window system).

So `break-pane -W` would float a pane and `join-pane` with no args would attempt to rejoin the pane whence it came.  This makes it possible to add back the PREFIX @ command I had to toggle a floating pane.  I chose @ because it was available, not because it had any great pnemonic, we can find a better one if you like.

daneofmanythings:
There isn't a minimal way to save a layout that is fault tolerant to changes. Maybe we don't allow a pane to automatically rejoin a layout if the layout has changed?

I'll change the flag to `-W`.

I don't think we should allow a user to float a pane if it is the only pane in the window. What is the use case? It was my intention to disallow this behavior.

The `@` key binding can be added back once `join-pane` has default layout joining behavior implemented using `... { move-pane } { join-pane }` without moving behavior into `break-pane`.

I will make `break-pane` float a pane with a flag. Little nit about flag choice though: currently the `-W` flag is used for new commands to behave like their established tiling counterparts. Making `-W` float a pane in `break-pane` would be the opposite, an established command using `-W` to float a pane instead. Is this incongruity acceptable?

mgrant0:
The last "position" of a pane is which window's layout tree it was in and which pane (if any) it was after.  `join-pane` defaults to `-v` vertical stacking.  So if there is any ambiguity you stack vertically.  If the window is empty (all dots desktop) it becomes the only active window.  If there's only a single window, it splits it and becomes the lower one.  If there is a different tree in the window then when it was last split, and the old pane id is no longer there, i would do as `join-pane` does today which looks like it joins it to the last pane in the list.  Have I missed anything?  In general, I am trying to get the command to do what I think any user would reasonably expect.

I think you should definitely be able to float a pane even if it is the only pane in a window.  It pops the pane out of the layout and you see it floating on the desktop of dots.  You can now open other floating panes if you want to work that way.  Please do not inhibit this behaviour.

I meant `-W` (capital W) to mean float the pane like it was a window in all cases.  Do we still need something that says behave like established tiling counterpart if no `-W`?  Sorry, not following what you are saying here.

daneofmanythings:
Thank you for the clarifications! 

I'll try to explain what I meant about the flag a little better. `join-pane` and `split-window` cannot make a pane float. They don't accept the geometry flags nor do they accept a flag to float a pane. As you pointed out, `join-pane` having this functionality makes it behave contrary to its semantics and too much like `break-pane`. As it stands right now, `move-pane` and `new-pane` are "super" commands of these. `new-pane` can split a window if it is given the `-W` option, otherwise it defaults to creating a floating pane. `move-pane` can join a pane into a tiled layout if given the `-W` option, otherwise it defaults to floating the pane. `break-pane` currently has no notion of floating panes; its default is tiling. To move the default floating behavior out of `move-pane` and into `break-pane` would invert the meaning of the `-W` option; instead of tiling the pane, it would specify to float it.

The behavior for `new-pane` that nick requested was to float by default. That is the invariant I am building from. Is is okay with you that `-W` means "do the opposite of default" instead of being specific about tiling/floating?

mgrant0:
I think we are still not quite on the same page yet but hopefully we are converging.

Let's not make -W to mean two different things in related contexts.  Use a different letter, perhaps -L for the other context.

-W for when the resulting pane should be floating, as in, it's going to look like a window in the UI sense.  Whenever you have -W, you also have -x, -y, -X, -Y to specify the geometry of the floating pane.

Currently `move-pane` does the same as `join-pane` but we are talking about breaking that symetry.  If you `move-pane` on a tiled pane to a window (without -W), it should continue to be like `join-pane`.  Backwards compatibility is preserved.  If you `move-pane` on a floating pane to another window, it remains floating.  That seems totally expected. You should also be able to specify a geometry of where to move the floating pane to.  Moving a floating pane to a different location in the current window is perfectly reasonable.  None of these actions involve a -W arg.

`move-pane -W` is the same as `break-pane -W`.  Essentially move the pane to another window (or maybe this window) and turn it into a floating pane.   `break-pane` on a floating pane (with or without -W) should move it to another window.

Now the opposite, should one be able to move a pane to another window and tile it?  Do not reuse -W for that, that's a different option, let's use -L for that to mean put it in the layout.  But here is where `move-pane` and `join-pane` diverge.  The -L option only make sense if the _src_ pane is floating and the desired result is a tiled pane somewhere.  So the new option is only really needed in `move-pane` because `join-pane` does that normally if the src is floating  (You could add it for consistency I suppose).  Conversly, `join-pane` does not get a -W option because that simply doesn't make sense.  (`break-pane` does that).   Therefore, `move-pane -L` is the same as `join-pane` when the source is a floating pane.  (-L ignored if the src is tiled).

Is this clear now the differences and inter-relationships between `move-pane`, `join-pane`, and `break-pane`?

Nic asked if we could fold `new-pane` into `split-pane`.  Therefore `split-pane -W` is the same as `new-pane`.

Unless I misremember, Nic wanted my `new-pane` to go away and the only way to make a new floating pane would be via the `split-pane` command.  `new-pane` is something i wrote for this branch.  To be honest, one does not naturally think of splitting a window when opening a new window in window system parlance.  I personally would prefer if we kept `new-pane` and added `split-window -W`.  I am ok with `new-pane -L` being the same as `split-pane`, as in, create a new pane in the layout vs floating.  Let's go with this for now.  

Summary:

- ~`move-pane -W` --> same as `break-pane -W`, floats a pane~
- ~`move-pane -L` --> same as `join-pane`, tiles a pane~
- `split-pane -W` --> same as `new-pane`, new floating pane
- `split-pane -L` unnecessary, don't add -L
- `new-pane -W` unnecessary, don't add -W
- `new-pane -L` --> redundent since `split-pane` will split a pane, but add it if you want to be consistent
- `break-pane -W` --> same as `move-pane -W`, float pane in current window
- `break-pane -W -a (or A|b|B)` --> same as above but in new window after (or at end, before, start)
- `break-pane -a (or A|b|B)` --> on floating window, moves floating pane to new window after (or before...)
- `break-pane -L [a|A|b|B]` --> on floating pane, like `break-pane` into new window but tiled
- `join-pane -L` --> redundent since if src is floating it will tile it, but add it if you want to be consistent
- `join-pane -W` --> nonsense

Everything stays backwards compatible.  All existing tmux commands do the same as before on tiled panes.  Noone's .tmux.conf should break because of these changes.  Again, please let me know if I have missed something or created some thing unworkable.

daneofmanythings:
In discussing this PR, I have felt that it should be broken up into several smaller ones to track changes in a more detailed manner. I will be closing this, but referencing it in related PRs.

PR #5047, opened 2026-05-01
Fix memory leak when freeing control-mode format subscriptions

aarcamp:


aarcamp:
👋 to all the OpenBSD slackers, just checked and see my last commit was back in 2006.. now that's slackin' :)

nicm:

PR #5049, opened 2026-05-01
Fix control client hang on exit after toggling no-output

aarcamp:
A control-client exit wedge is possible, reproducible by:

1. queue some pane %output blocks
2. execute `refresh-client -f no-output`
3. perform a control-client exit

On an unfixed tmux, control_reset_offsets() frees the control_pane objects without removing their queued control_block entries from all_blocks. The control client then waits forever in server_client_check_exit() because control_all_done() never becomes true.


nicm:

PR #5051, opened 2026-05-01
fix: check wp for NULL before dereferencing in window.c

alexarama:
## Problem

`wp` is dereferenced at line 2012 before the null check at line 2017.
If `wp` is NULL, the program crashes before reaching the guard.

Detected by Cppcheck static analysis [nullPointerRedundantCheck].

## Fix

Declared `w` and `wo` without initialization, moved the null check 
for `wp` before any dereference, then initialized `w` and `wo` safely.

## Verification

Verified with Cppcheck — warning no longer appears after this fix.


nicm:

PR #5053, opened 2026-05-02
Add Emacs-style recenter-top-bottom

### LINKED ISSUE #4689: Feature Request: add recenter-top-bottom to copy mode
### Issue description

tmux is wonderful, but I miss [`recenter-top-bottom`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Recentering.html) (usually bound to `C-l`) from emacs quite a bit when I'm in copy mode.

If it's reasonable to provide pointers on how that might be implemented, I'd be interesting in giving it a shot.

### Required information

* tmux version (`tmux 3.5a`).
* Platform (`Darwin arm`).
* Terminal in use (`iterm2`).
* $TERM *inside* tmux (`xterm-256color`).
* $TERM *outside* tmux (`xterm-256color`).

0xycn:
Fixes #4689

Adds an Emacs-style recenter-top-bottom implementation to copy-mode

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -570,6 +570,7 @@ key_bindings_init(void)
 		"bind -Tcopy-mode M-Down { send -X halfpage-down }",
 		"bind -Tcopy-mode C-Up { send -X scroll-up }",
 		"bind -Tcopy-mode C-Down { send -X scroll-down }",
+		"bind -Tcopy-mode C-l { send -X recenter-top-bottom }",
 
 		/* Copy mode (vi) keys. */
 		"bind -Tcopy-mode-vi '#' { send -FX search-backward -- '#{copy_cursor_word}' }",
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -570,7 +570,7 @@ key_bindings_init(void)
 		"bind -Tcopy-mode M-Down { send -X halfpage-down }",
 		"bind -Tcopy-mode C-Up { send -X scroll-up }",
 		"bind -Tcopy-mode C-Down { send -X scroll-down }",
-		"bind -Tcopy-mode C-l { send -X recenter-top-bottom }",
+		"bind -Tcopy-mode C-l { send -X recentre-top-bottom }",
 
 		/* Copy mode (vi) keys. */
 		"bind -Tcopy-mode-vi '#' { send -FX search-backward -- '#{copy_cursor_word}' }",
--- a/tmux.1
+++ b/tmux.1
@@ -2224,6 +2224,11 @@ Turn off rectangle selection mode.
 .Xc
 Toggle rectangle selection mode.
 .It Xo
+.Ic recentre\-top\-bottom
+(emacs : C-l)
+.Xc
+Cycles the current line between centre, top, and bottom.
+.It Xo
 .Ic refresh\-from\-pane
 (vi: r)
 (emacs: r)
--- a/window-copy.c
+++ b/window-copy.c
@@ -2789,6 +2789,32 @@ window_copy_cmd_refresh_from_pane(struct window_copy_cmd_state *cs)
 	return (WINDOW_COPY_CMD_REDRAW);
 }
 
+static enum window_copy_cmd_action
+window_copy_cmd_recenter_top_bottom(struct window_copy_cmd_state *cs)
+{
+	struct window_mode_entry	*wme = cs->wme;
+	struct window_copy_mode_data	*data = wme->data;
+	
+	u_int cy = data->cy;
+	u_int sy = screen_size_y(&data->screen) - 1;
+	u_int sm = sy / 2;
+	
+	if (cy == 0) {
+		window_copy_scroll_down(wme, sy - cy);
+		data->cy = sy;
+	}
+	else if (cy == sy) {
+		window_copy_scroll_up(wme, sy - sm);
+		data->cy = sm;
+	}
+	else {
+		window_copy_scroll_up(wme, cy);
+		data->cy = 0;
+	}
+	
+	return (WINDOW_COPY_CMD_REDRAW);
+}
+
 static const struct {
 	const char			 *command;
 	u_int				  minargs;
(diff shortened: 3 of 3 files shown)


nicm:
Looks reasonable, can please change `recenter` to `recentre` to match the other spellings of centre, and can you add it to tmux.1 as well (there is a list of commands under `WINDOWS AND PANES`)?

nicm:
@joegallo do you want to try this and see if it behaves as you expect?

0xycn:
@nicm done. renamed to recentre and fixed some issues related to centring the line first.

joegallo:
Sure thing! I'll take a look today.

joegallo:
I don't see any difference in the ordinary expected case, where (for example) you're `recentre-top-bottom`ing the beginning of some function declaration in the middle of a 2000 line file. 🚀 

However, it seems like the behavior might differ in the edge case of `recentre-top-bottom`ing content that's very near the beginning of the buffer/file and very near the end of the buffer/file.

Here's what I'm seeing:

In tmux, if I `cat tty.c; echo '---------'; cat osdep-unknown.c` then I get a small amount of output below a `---------` line. Then if I put my cursor at the line that starts with `osdep_get_name` and hit `ctrl-l` several times, I'll end up with my cursor moving to some other line of code (it looks like it tries to move around the line where my cursor was, but fails, and gives up and moves my cursor instead).

By way of contrast, if I open `osdep-unknown.c` in emacs and move my cursor to the same line, then when I repeatedly hit `ctrl-l` it keeps the cursor and display where they are (basically "this is as close I can get to the middle"), then moves the `osdep_get_name` to the top (displaying more blank space than exists at the end of the file), then moves it as close as it can to the middle.

That is, in emacs:
1. ctrl-l -- noop, "this is as close as I can get to the middle". (call the current screen state 'state 1')
2. ctrl-l -- add empty space at the end, move `osdep_get_name` to the top.
3. ctrl-l -- back to state 1, "this is as close as I can get to the bottom".
4. ctrl-l -- noop, still state 1, "this is as close as I can get to the middle".
(etc)

No amount of `ctrl-l` key presses will move my *cursor* to a different line, all it will do is move (or not move) the currently selected line (and cursor) around the screen.

joegallo:
How about something like this?

```diff
diff --git a/window-copy.c b/window-copy.c
index 4cdee2fa..ed3d46d0 100644
--- a/window-copy.c
+++ b/window-copy.c
@@ -2794,31 +2794,43 @@ window_copy_cmd_recentre_top_bottom(struct window_copy_cmd_state *cs)
 {
 	struct window_mode_entry	*wme = cs->wme;
 	struct window_copy_mode_data	*data = wme->data;
+	u_int				 cy, sy, sm, oy;
 
-	u_int cy = data->cy;
-	u_int sy = screen_size_y(&data->screen) - 1;
-	u_int sm = sy / 2;
+	cy = data->cy;
+	sy = screen_size_y(&data->screen) - 1;
+	sm = sy / 2;
 
 	if (cy == sm) {
+		oy = data->oy;
 		window_copy_scroll_up(wme, cy);
-		data->cy = 0;
-	}
-	else if (cy == 0) {
+		data->cy = cy - (oy - data->oy);
+	} else if (cy == 0) {
+		oy = data->oy;
 		window_copy_scroll_down(wme, sy);
-		data->cy = sy;
-	}
-	else if (cy == sy) {
+		data->cy = cy + (data->oy - oy);
+	} else if (cy == sy) {
+		oy = data->oy;
 		window_copy_scroll_up(wme, sy - sm);
-		data->cy = sm;
-	}
-	else {
+		data->cy = cy - (oy - data->oy);
+	} else {
 		if (cy < sm) {
+			oy = data->oy;
 			window_copy_scroll_down(wme, sm - cy);
-			data->cy = sm;
-		}
+			if (data->oy != oy)
+				data->cy = cy + (data->oy - oy);
 			else {
+				window_copy_scroll_up(wme, cy);
+				data->cy = cy - (oy - data->oy);
+			}
+		} else {
+			oy = data->oy;
 			window_copy_scroll_up(wme, cy - sm);
-			data->cy = sm;
+			if (data->oy != oy)
+				data->cy = cy - (oy - data->oy);
+			else {
+				window_copy_scroll_down(wme, sy - cy);
+				data->cy = cy + (data->oy - oy);
+			}
 		}
 	}
 
```

I think this'll fix the cursor-drift. That is, it appears to keep the cursor on the same line no matter which edge cases I put it through.

The behavior here isn't _exactly_ the same as the emacs behavior -- it skips noop changes and just ignores them. So, for example (and speaking very loosely), if I have 10,000 lines of output, and repeatedly ctrl-l the 9995th line, then it bounces that one line back and forth between being displayed at the 5th from the bottom and being the very bottom, but it doesn't have a noop reaction for the keypress that would have tried to put the 9995th line at the top of the screen.

0xycn:
@joegallo Thanks for the feedback. you're right this keeps the cursor on the same buffer line in all edge cases.

Regarding the behavior difference from Emacs (skipping noop vs. cycling through unreachable positions): Emacs cycles through all three positions even when one is unreachable because it can add blank space above/below the buffer to position the cursor anywhere on screen. I don't think that's possible in tmux, so when a position is unreachable (e.g. not enough scrollback to move the cursor to the top line), your solution of skipping it looks good.

It's not exactly the same as Emacs, but we have to restrict cursor movement if the scroll doesn't succeed. We also need to adjust the cycle logic there's a flaw where if the cursor is already on the first visible line, C-l won't recenter the line first and instead jumps straight to bottom.

0xycn:
Please see the latest commit: 0f1538f

joegallo:
Oh, yeah, I like this a version lot. It is very slightly different than emacs in behavior (as discussed, and I think those limitations make sense in context), but I think the behavior is quite excellent and makes sense here.

🚀

nicm:
Looks good, applied to OpenBSD now with some trivial style fixes, will be in GitHub later. Thanks to both of you!

joegallo:
This is on `master` here now as 962e56559162c27ca39249a4c254dacc2e948266. 🚀

PR #5070, opened 2026-05-07
copy-mode: add option to disable one-past-EOL cursor

maxonvim:
## Summary

- adds a new window option, `copy-mode-virtual-edit`, with values `off` and `onemore`
- defaults to `onemore` to preserve existing behavior
- when set to `off`, `cursor-right` in copy mode no longer moves into a virtual cell past end of line

## Motivation

In vi-style copy mode, many users expect horizontal movement to stay on real text columns. Today, tmux always allows one-past-EOL movement and there is no built-in way to disable it.

This change keeps current behavior by default, but allows strict motion for users who want it.

## Details

- adds option definition in `options-table.c`
- threads an `onemore` flag through `grid_reader_cursor_right()`
- applies `copy-mode-virtual-edit` consistently in copy-mode cursor placement paths in `window-copy.c`
  (including end-of-line, history bottom, incremental-search reset, and relevant helper motions)
- preserves rectangle-selection behavior

## Compatibility

No behavior changes unless `copy-mode-virtual-edit` is explicitly set to `off`.

## Notes

- `CHANGES` includes a short entry for the new option.
- build verified locally in nix shell:
  - `nix-shell -p autoconf automake pkg-config libevent ncurses bison --run './autogen.sh && ./configure && make -j4'`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,7 @@
 CHANGES FROM 3.6 TO 3.6a
 
-* Add a copy-mode-virtual-edit window option (off or onemore) to control
-  whether copy mode cursor-right can move one cell past end of line.
+* Do not allow copy mode cursor-right to move one cell past end of line with
+  vi keys.
 
 * Fix a buffer overread and an infinite loop in format processing (reported by
   Giorgi Kobakhia, issue 4735).
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,8 @@
 CHANGES FROM 3.6 TO 3.6a
 
+* Add a copy-mode-virtual-edit window option (off or onemore) to control
+  whether copy mode cursor-right can move one cell past end of line.
+
 * Fix a buffer overread and an infinite loop in format processing (reported by
   Giorgi Kobakhia, issue 4735).
 
--- a/grid-reader.c
+++ b/grid-reader.c
@@ -45,15 +45,21 @@ grid_reader_line_length(struct grid_reader *gr)
 
 /* Move cursor forward one position. */
 void
-grid_reader_cursor_right(struct grid_reader *gr, int wrap, int all)
+grid_reader_cursor_right(struct grid_reader *gr, int wrap, int all,
+    int onemore)
 {
 	u_int			px;
 	struct grid_cell	gc;
 
 	if (all)
 		px = gr->gd->sx;
-	else
+	else if (onemore)
 		px = grid_reader_line_length(gr);
+	else {
+		px = grid_reader_line_length(gr);
+		if (px != 0)
+			px--;
+	}
 
 	if (wrap && gr->cx >= px && gr->cy < gr->gd->hsize + gr->gd->sy - 1) {
 		grid_reader_cursor_start_of_line(gr, 0);
--- a/options-table.c
+++ b/options-table.c
@@ -111,9 +111,6 @@ static const char *options_table_allow_passthrough_list[] = {
 static const char *options_table_copy_mode_line_numbers_list[] = {
 	"off", "default", "absolute", "relative", "hybrid", NULL
 };
-static const char *options_table_copy_mode_virtual_edit_list[] = {
-	"off", "onemore", NULL
-};
 
 /* Status line format. */
 #define OPTIONS_TABLE_STATUS_FORMAT1 \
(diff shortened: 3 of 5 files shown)


nicm:
Does this need to be an option? Why not make it always work like this if `vi` keys are turned on?

nicm:
It's actually not really how emacs works either, but the difference in emacs is that if the line is exactly the same as or longer than visible width, it is wrapped and a `\` is inserted as the last character. tmux obviously cannot do this so I guess this is why we have the invisible last position.

maxonvim:
> Does this need to be an option? Why not make it always work like this if `vi` keys are turned on?

That could be more intuitive for people who already signed up for the Vim experience. Want me to change it to that?

nicm:
Yes I think that would be better.

maxonvim:
Okay now it depends on the vi keys option

nicm:
Looks fine to me, applied to OpenBSD now, will be in GitHub later. Thanks!

mckellygit:
Hi,

Just a comment.
If we select a line in copy-mode-vi (V) the selection highlight still goes one past the end of line, which is different than if we start selection and then extend it to the end of line.

thx,
-m

nicm:
@maxonvim  `regress/copy-mode-test-vi.sh` now fails, I think due to this change - any chance you could update it?

maxonvim:
Updated the test with #5172

mckellygit:
Also, in rectangular mode I can no longer extend the selection past the length of the current line, but I might want to, because lines above included in my selection are longer and I want all those chars out to my selection right side edge.   In vi I can do this, so its no longer as functional as it was or the same as vi.

nicm:
How does it work in vi?

nicm:
We should match vi here, if the old way was closer we should revert this.

mckellygit:
you can extend the rectangular selection past the end of the line, if lines above (or below) in your current selection are longer than the current line - so you can still move right into the open space.
This only makes sense in rectangular mode.

mckellygit:
Let me try without this commit, I don't want to blame this commit until I prove it, but I agree we should be able to match vi as close as we can easily.

mckellygit:
I actually think its a vi setting that can be changed and I probably have this set one way and default vi is another.  I'll report back more details soon,

mckellygit:
Yes, I have in my vim:
set virtualedit=block

This makes the spaces following the end of the line accessible to cursor movements, so you can extend the selection as much as you need. Yanking that text will have whitespace padding inserted to make a rectangular block

So I would say the current method is good, but I sure wish I had a way to get what I have for vi (and what I swear it used to be a while ago)

thx

mckellygit:
I think it would be great if we did have an additional option for this, at least for rectangular selection mode.  It would default to off but I could turn it on to be able to select columns like I could before.  I can create a PR if you like.
Imho, selecting right-most columns is a strong feature of tmux and I really can't do that anymore.

mckellygit:
This might be the smallest change I can think of to get back the ability to extend the rectangular selection but leave everything else as it was:
```
diff --git window-copy.c window-copy.c
index f06b179b..ce33c040 100644
--- window-copy.c
+++ window-copy.c
@@ -5403,8 +5403,8 @@ window_copy_update_cursor(struct window_mode_entry *wme, u_int cx, u_int cy)
    u_int                maxx;
    int              allow_onemore;

-    allow_onemore = (data->screen.sel != NULL && data->rectflag);
-    if (cy < screen_size_y(s)) {
+    if ( (!data->rectflag) && (cy < screen_size_y(s)) ) {
+       allow_onemore = (data->screen.sel != NULL && data->rectflag);
        py = screen_hsize(data->backing) + cy - data->oy;
        maxx = window_copy_cursor_limit(wme, py, allow_onemore);
        if (cx > maxx)
```
We could wrap this in an option that defaults to off if you like, just let me know.
@maxonvim what do you think ?

nicm:
I don't really want a load of options for obscure vi behaviour that barely anyone will use, but other than that I don't care how vi mode works so long as there is some consensus and if in doubt we defer to vi. So if this is the way it works, that's fine. Can you please attach a file with the diff or open a PR.

mckellygit:
Thx yes will do.

davidchambers:
I just upgraded tmux and was delighted by this change! :heart:

Thank you @maxonvim and everyone else who contributed to this quality-of-life improvement. :clap:

PR #5071, opened 2026-05-09
Avoid overshooting Sixel height in sixel_parse_write

bptato:
Prior to this patch, sixel_parse_write would unconditionally round up the image height to a multiple of 6.  Then, sixel_size_in_cells would round up the result to the cell height, so when that isn't a multiple of 6 you could end up with placeholders that are larger than the original image in cells.

(sixel_set_pixel does the same thing as the replaced code, but without expanding the height to a multiple of 6.)


nicm:
Applied, thanks!

Issue #5024, opened 2026-04-23
tmux server livelock in tty_draw_line under control-mode load (separate from #4969)


tarekrached:
## Summary

tmux server enters a livelock (100% user-mode spin, never returns to `poll()`) inside the screen redraw path. All subsequent commands against the server (`tmux ls`, `tmux has-session`, `tmux info`, any new `attach-session`) queue on the command socket and never complete. Existing clients remain attached and visible but unresponsive. Recovery requires `SIGKILL` on the server PID — `tmux kill-server` and `SIGTERM` are both ignored (details in Workaround below).

## Related prior art

- **#4969** ("Bug: `tty_draw_line` enters long running loop under specific circumstances", closed 2026-04-04): same function, same loop, fixed by commit `cd60de44` which clamps `nx` at the top of `tty_draw_line`. **That fix is present in this build** (confirmed at lines 149–150 of `tty-draw.c@dd62c2f`). This report is a different path into the same loop.
- **#4932** ("Infinite CPU loop in `screen_redraw_draw_scrollbar` due to unsigned integer underflow", open): different helper but same class — `screen_redraw_screen` → helper that can iterate billions of times.

## Environment

- macOS 26.4.1 (arm64, build 25E253)
- tmux build: **HEAD-dd62c2f** (reports as `next-3.7`), installed via Homebrew HEAD tap
  - `/opt/homebrew/bin/tmux -> ../Cellar/tmux/HEAD-dd62c2f/bin/tmux`
- Workload: ~5 concurrent control-mode clients (`tmux -C attach-session`) driven by `agent-deck` v1.7.50 (Go TUI that polls each tracked session every ~2s via `capture-pane`, `display-message`, `list-clients`, `set-option`). Sessions display Unicode-heavy content: braille spinners (`⠐⠁⠃⠇⠧⠷⠿`), box-drawing, emoji in `status-left`.

## Process state

Both incidents: process state in `ps`: `Rs` (runnable + session leader), not `S` or `D` — the event loop is *not* in a syscall. CPU consistent with a single-threaded spin at 100%. First incident server uptime at detection ~17h; second incident ~6h14m.

## Evidence: two independent incidents, five `lldb` samples

All samples: `lldb --batch -p <server_pid> -o "thread backtrace all" -o "detach" -o "quit"`, on the main thread (`com.apple.main-thread`), same binary (`HEAD-dd62c2f`), same machine, same workload.

- **Incident A** — ~17h server uptime, two samples ~60s apart
[...]

PR #5084, opened 2026-05-13
Avoid infinite loop in tty_draw_line on width-0 PADDING cell

leonasdev:
## The bug

`tty_draw_line`'s main loop advances the cell index with `i += gcp->data.width`. A PADDING cell always has `data.width == 0`, so if an orphan PADDING cell appears mid-line, `i` never advances and the loop spins forever. The server pegs at 100% CPU with all IPC blocked (`tmux ls` times out).

The pre-pass at the top of `tty_draw_line` already handles this for **leading** orphan PADDING (comment: *"If there is padding at the start, we must have truncated a wide character. Clear it."*). The same guard is missing in the main loop for the mid-line case.

## Reproducer

Generate a test file (requires U+2003 EM SPACE between emoji, not plain ASCII spaces):

```sh
python3 -c "print(('\u26a0\ufe0f\u2003' * 500 + '\n') * 1000)" > /tmp/bug.txt

Start a clean outer tmux session and open a popup with a nested tmux inside:

tmux -Ltest -f/dev/null new-session
tmux -Ltest -f/dev/null display-popup -E "tmux -Ltest2 -f/dev/null new-session"

Inside the popup, open nvim with no config and set a statuscolumn containing
a click region followed by %l and a trailing space:

nvim --clean /tmp/bug.txt

:set number
:lua vim.wo.statuscolumn = "%@Fake@  %T%l "

Press Ctrl-D / Ctrl-U rapidly. The outer tmux server hangs within a few keystrokes;
timeout 3 tmux -Ltest -f/dev/null ls exits 124.
```

## Fix

Guard the loop increment so `i` always advances by at least 1, mirroring the existing handling of orphan PADDING at the line start.

### DIFF
--- a/tty-draw.c
+++ b/tty-draw.c
@@ -328,8 +328,10 @@ tty_draw_line(struct tty *tty, struct screen *s, u_int px, u_int py, u_int nx,
 		memcpy(&last, gcp, sizeof last);
 		if (empty != 0)
 			i += empty;
-		else
+		else if (gcp->data.width > 0)
 			i += gcp->data.width;
+		else
+			i += 1;	/* Orphan PADDING; advance to avoid infinite loop. */
 	}
 
 out:


nicm:
Try this one instead please https://github.com/tmux/tmux/issues/5024#issuecomment-4420312414.

leonasdev:
> Try this one instead please [#5024 (comment)](https://github.com/tmux/tmux/issues/5024#issuecomment-4420312414).

Tested with the patch applied and the hang is gone.

nicm:
I have applied my patch to OpenBSD now, it should be in GitHub later. Let me know if you see any other problems. Thanks!

PR #5088, opened 2026-05-14
Handle duplicate terminal replies and split paste-end keys

Empty-Jing:
Problem
tmux could leak terminal control sequences into pane input in two cases:
- duplicate DA/DA2 replies during terminal feature detection
- fragmented bracketed-paste end sequences such as `ESC [ 201 ~`

Root cause
After tmux had already recorded terminal features, later DA/DA2 replies were no longer consumed by the dedicated parsers and could fall through into normal pane input handling.
In addition, bracketed-paste parsing did not keep enough state to reliably recognize a paste-end sequence when it arrived split across multiple reads.

Fix
This patch continues to consume valid DA/DA2 replies so they do not leak into pane input, even when they are duplicated.
It also tracks bracketed-paste state in tty key parsing and keeps waiting for a partial paste-end prefix, so split `ESC [ 201 ~` sequences are recognized correctly instead of being forwarded to the pane.

Environment
I reproduced this in a remote tmux session accessed from Windows PowerShell over SSH.
Client:
- Windows PowerShell
- OpenSSH client
Server:
- Ubuntu 24.04
- tmux running inside the SSH session

Reproducer
A reliable reproducer was:
1. Connect from Windows PowerShell to the remote host over SSH.
2. Start or attach to tmux.
3. Paste any content with `Win+V`.
4. Immediately click another window.
In this setup, the bracketed-paste end sequence could arrive split across reads and leak into pane input.
See the attached screenshot for an example.
image


nicm:
Can you make this MR just the paste bits and remove the DA bits? I don't think we want to do that part.

Empty-Jing:
> Can you make this MR just the paste bits and remove the DA bits? I don't think we want to do that part.

Updated. I dropped the DA/DA2 handling changes and kept only the bracketed paste-end fix for the split paste-end sequence. The MR should now be paste-only.

nicm:
OK I see what you are doing.

I would probably get rid of `tty_keys_paste_end_partial` and just always bump it to 500 if pasting is active. Would that hurt anything?

Empty-Jing:
> OK I see what you are doing.
> 
> I would probably get rid of `tty_keys_paste_end_partial` and just always bump it to 500 if pasting is active. Would that hurt anything?

I looked at that path more closely. I think always bumping it to 500 while bracketed paste is active would probably work, but it widens the scope more than we need.
Right now `tty_keys_paste_end_partial()` only extends the delay when we are already inside bracketed paste and the buffered input is specifically a prefix of `\033[201~`, so it is only protecting a split paste-end sequence. If we apply the 500ms delay to all partial keys during paste, then any partial escape-like input inside the pasted data would also take the longer path.
So I kept the narrower check because it seems to fix the broken split `paste end` case without changing behavior for the rest of the pasted stream. If you prefer, I can simplify the helper or inline it, but I think the narrower condition is the safer behavior.

nicm:
Thanks, I have applied it to OpenBSD now, with the function but with a couple of trivial changes. Should be in GitHub later. Let me know if you see any problems. Thanks!

PR #5090, opened 2026-05-15
Unescape #: in format modifier arguments

### LINKED ISSUE #4073: Cannot substitute colon?
### Issue description

I can't seem to figure out a syntax to replace a colon (`:`) in a `#{s/.../.../:@v}` expression.
First, I can't use `#{s/:/_/:@v}` because the `:` terminates the substitution prefix. Okay, that makes sense, I guess it needs to be escaped somehow.
However, escaping the colon (`#{s/#:/_/:@v}`) doesn't seem to work either because while the `s/` expression seems to be recognized correctly now, the `#:` is never replaced with `:` afterwards:
```bash
$ tmux set @v 'foo:bar'
$ tmux display-message -v '#{s/#:/_/:@v}'
# expanding format: #{s/#:/_/:@v}
# found #{}: s/#:/_/:@v
#  expanding format: #:
#  result is: #:
#  expanding format: _
#  result is: _
# modifier 0 is s
# modifier 0 argument 0: #:
# modifier 0 argument 1: _
# format '@v' found: foo:bar
#  expanding format: #:
#  result is: #:
#  expanding format: _
#  result is: _
# substitute '#:' to '_': foo:bar
# replaced 's/#:/_/:@v' with 'foo:bar'
# result is: foo:bar
```
Note that it's trying to replace `#:` instead of `:`.
For completeness, I was expecting the result to be "foo_bar".

Similarly, it seems to be impossible to substitute characters *with* colons:
```bash
$ tmux display-message -v '#{s/o/#:/:@v}'
# expanding format: #{s/o/#:/:@v}
# found #{}: s/o/#:/:@v
#  expanding format: o
#  result is: o
#  expanding format: #:
#  result is: #:
# modifier 0 is s
# modifier 0 argument 0: o
# modifier 0 argument 1: #:
# format '@v' found: foo:bar
#  expanding format: o
#  result is: o
#  expanding format: #:
#  result is: #:
# substitute 'o' to '#:': f#:#::bar
# replaced 's/o/#:/:@v' with 'f#:#::bar'
# result is: f#:#::bar
```
Again, the `#:` is never converted to `:`, so `#`s show up in the result. This one can kind-of be worked around though, by using `'#{s/o/#:/;s/##//:@v}'` to get rid of the `#`s in a separate substitution, but that would produce incorrect results if `@v` already contains `#`s that should be preserved.

Am I missing something or is this a bug?

### Required information

```bash
$ tmux -V
tmux 3.4
$ uname -sp
Linux x86_64
$ echo $TERM
tmux-256color
```
[tmux-client-1666311.log](https://github.com/user-attachments/files/16580905/tmux-client-1666311.log)
[tmux-server-1666313.log](https://github.com/user-attachments/files/16580912/tmux-server-1666313.log)

AungMyoKyaw:
Fixes #4073.

#: is used to escape : in format modifiers like #{s/#:/_/:@v},
but after format_skip1() handles the escape during parsing, the raw
#: string was passed to format_expand1() which left it unchanged.

Run modifier arguments through format_unescape() before expansion,
so #: correctly becomes : in substitution patterns and replacements.

Also add test cases for escaped colons in regress/format-strings.sh.

### DIFF
--- a/format.c
+++ b/format.c
@@ -4327,7 +4327,7 @@ format_build_modifiers(struct format_expand_state *es, const char **s,
 {
 	const char		*cp = *s, *end;
 	struct format_modifier	*list = NULL;
-	char			 c, last[] = "X;:", **argv, *value;
+	char			 c, last[] = "X;:", **argv, *value, *unescaped;
 	int			 argc;
 
 	/*
@@ -4394,8 +4394,10 @@ format_build_modifiers(struct format_expand_state *es, const char **s,
 
 			argv = xcalloc(1, sizeof *argv);
 			value = xstrndup(cp + 1, end - (cp + 1));
-			argv[0] = format_expand1(es, value);
+			unescaped = format_unescape(es, value);
 			free(value);
+			argv[0] = format_expand1(es, unescaped);
+			free(unescaped);
 			argc = 1;
 
 			format_add_modifier(&list, count, &c, 1, argv, argc);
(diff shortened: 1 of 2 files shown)


nicm:
I came up with pretty much the same patch, going to close this but thanks.

PR #5091, opened 2026-05-15
Fix selection-mode line and other-end in copy mode

### LINKED ISSUE #5085: `selection-mode line` behaves strangely
### Issue description

I'm sorry that the bulk of this report is the following asciinema recording:
https://asciinema.org/a/7ZS4qcJPhSb1frdV

1. I start an unconfigured tmux server and enter copy mode (vi variant, `C-b [`)
2. I start selection (`Space`) and move around as normal
3. I enable `selection-mode line` (`C-b : send -X selection-mode line`) and start moving around. Moving past the starting point of the selection causes the selection to split in two, moving up further causes only a single line to be selected.
4. I move down, which works as expected
5. I use `o` (`other-end`), the start of the selection inexplicably jumps to the very start of scrollback
6. After using `o`, `selection-mode line` is disabled.

I was trying to see if I could reasonably replicate vim's visual modes (regular, line, block), but besides the fact that the current `selection-mode` / `rectangle-mode` can't be queried via `if -F`, I also found out that line mode works rather strangely.

### Required information

* tmux version (`tmux -V`). 3.6a
* Terminal in use (xterm, rxvt, etc). foot
* $TERM *inside* tmux (`echo $TERM`). tmux-256color
* $TERM *outside* tmux (`echo $TERM`). foot

AungMyoKyaw:
Fixes #5085.

This fixes four problems with selection-mode line in copy mode:

1. selection-mode line did not initialize lineflag, selrx, selry, endselrx, endselry when activated on an existing selection, causing the selection to behave strangely (split in two, jump to scrollback).

2. selection-mode line before begin-selection did not work because begin-selection unconditionally reset selflag to SEL_CHAR.

3. other-end unconditionally reset selflag to SEL_CHAR, disabling line mode.

4. selection-mode word had the same problems as line mode.

Changes:
- window_copy_cmd_selection_mode now properly initializes lineflag and the selection reset positions for both SEL_LINE and SEL_WORD.
- window_copy_cmd_begin_selection now preserves SEL_LINE instead of resetting to SEL_CHAR, and sets up the reset positions.
- window_copy_cmd_other_end no longer clobbers selflag.

### DIFF
--- a/regress/copy-mode-test-emacs.sh
+++ b/regress/copy-mode-test-emacs.sh
@@ -112,5 +112,39 @@ $TMUX send-keys -X next-space-end
 $TMUX send-keys -X copy-selection
 [ "$($TMUX show-buffer)" = "500xyz" ] || exit 1
 
+# Test that selection-mode line works after begin-selection.
+$TMUX send-keys -X history-top
+$TMUX send-keys -X start-of-line
+$TMUX send-keys -X cursor-down
+$TMUX send-keys -X cursor-down
+$TMUX send-keys -X begin-selection
+$TMUX send-keys -X selection-mode line
+$TMUX send-keys -X cursor-down
+$TMUX send-keys -X copy-selection
+[ "$($TMUX show-buffer)" = "$(printf "Another line...\n... @nd then \$ym_bols[]{}")" ] || exit 1
+
+# Test that other-end preserves selection-mode line.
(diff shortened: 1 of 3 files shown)


nicm:
The original submitter has gone away so unless you can reproduce let's close this.

PR #5092, opened 2026-05-15
Gate tty_write on MODE_SYNC to fix #4983

### LINKED ISSUE #4983: DEC 2026 synchronized updates leak structural commands when the BSU/ESU pair spans multiple pane reads
### Issue description

When an application wraps a redraw in `\033[?2026h … \033[?2026l` and the bytes
span more than one `input_parse_buffer` call, the outer terminal commits a
partial intermediate frame instead of an atomic transition. The same sequence
run in the bare outer terminal renders smoothly.

https://github.com/user-attachments/assets/a58bd67d-c8a1-45a0-a657-ef5028f21a23

### Reproducer

```sh
#!/bin/sh
# Inside tmux master: visible blank flash between iterations.
# Outside tmux on a terminal honouring mode 2026: stable screen of x's.

trap 'printf "\033[?2026l\033[2J\033[H"; exit' INT TERM

while :; do
    awk 'BEGIN {
        printf "\033[?2026h\033[H\033[2J"
        for (i = 1; i <= 200; i++) row = row "x"
        for (y = 1; y <= 2000; y++) print row
        printf "\033[?2026l"
    }'
    sleep 0.5
done
```

### Diagnosis

I asked and discussed with Claude Code the next section. I still spent time
on reproduction and verification.

`MODE_SYNC` already gates cell emission in two places:

- `screen_write_cell` at `screen-write.c:2174` skips per-cell emit.
- `screen_write_collect_flush` at `screen-write.c:1823` drops the queued write
  list.

Structural commands are not gated. The following all call
`tty_write(tty_cmd_*, ...)` unconditionally inside `screen-write.c`:
`screen_write_clearscreen`, `_clearendofscreen`, `_clearstartofscreen`,
`_insertline`, `_deleteline`, `_insertcharacter`, `_deletecharacter`,
`_clearcharacter`, `_scrolldown`, `_reverseindex`, `_alignmenttest`, and
`_sixelimage`. The `tty_write` dispatch at `tty.c:1574` has no `MODE_SYNC`
check, so they all reach `c->tty.out` regardless of sync state.

(The `_clearline` family, `_scrollup`, and `_linefeed` are not on this list:
they go through the collect write list / `ctx->scrolled` instead of calling
`tty_write` directly, so the existing line-1823 drop in `collect_flush`
handles them correctly.)

For the reproducer:

1. The first parse call contains BSU + `Clear::All` + early cells.
   `tty_cmd_clearscreen` reaches `c->tty.out` (not gated). The cells after it
   are dropped by `collect_flush` (gated). End of iteration: `tty_sync_end`
   commits an atomic frame on the outer terminal containing only the clear.
2. Subsequent parse calls drop their cells; no structural commands fire, so no
   outer-tty sync wrap occurs. The outer terminal keeps showing the cleared
   frame.
3. The parse call containing ESU sets `PANE_REDRAW`. The next server iteration
   runs `screen_redraw_pane`, which repaints from the grid.

The user sees the cleared frame for the duration between (1) and (3). With a
~400 KB frame on macOS this is hundreds of milliseconds.

<details>
<summary>Server log excerpt — one re-render cycle (full log attached)</summary>

```
1775630162.587503 input_parse_buffer: %0 ground, 1024 bytes  BSU + Clear::All + cells...
1775630162.587765 /dev/ttys011 sync start
1775630162.588312 /dev/ttys011 sync end
1775630162.588397 input_parse_buffer: %0 ground, 1024 bytes
1775630162.591273 /dev/ttys011 sync start
1775630162.592879 /dev/ttys011 sync end
1775630162.592975 input_parse_buffer: %0 ground, 1024 bytes
... (about 400 more 1024-byte parse calls of pure cell data, all dropped by collect_flush) ...
1775630162.744529 input_parse_buffer: %0 ground, 1024 bytes
1775630162.745014 input_parse_buffer: %0 ground, 194 bytes  cells + ESU
1775630162.745141 /dev/ttys011 sync start
1775630162.771497 /dev/ttys011 sync end
1775630163.255763 input_parse_buffer: %0 ground, 823 bytes  BSU + Clear::All + cells...
```

Reading top to bottom: the first 1024-byte parse call contains
`BSU + Clear::All + early cells`. The first sync start/end pair commits a
frame to the outer terminal that contains only the clear (cells dropped by
`collect_flush`). About 400 more 1024-byte parse calls follow, each
dropping their cells; no `tty_cmd_*` fires so there are no further
sync start/end pairs. The last 194-byte parse call contains the ESU,
which sets `PANE_REDRAW`. The next iteration runs `screen_redraw_pane`,
which produces the second `sync start` → `sync end` pair (26 ms long,
the actual repaint). Between the cleared frame at `162.588` and the
filled frame at `162.771` the outer terminal sits on the cleared screen
for ~183 ms — the visible flash.

</details>

### One possible direction (for discussion)

The most localised thing I came up with is gating `tty_write` itself on
`MODE_SYNC`, with a small bypass list for side-channel commands. The post-ESU
`PANE_REDRAW` already repaints from the grid, so suppressing structural
emissions during the sync window shouldn't lose anything visible. Sketch of
what that could look like in `tty.c:1574`:

```c
    if (tty_client_ready(ctx, c)) {
        state = ctx->set_client_cb(ctx, c);
        if (state == -1)
            break;
        if (state == 0)
            continue;
        if (ctx->s != NULL && (ctx->s->mode & MODE_SYNC) &&
            cmdfn != tty_cmd_syncstart &&
            cmdfn != tty_cmd_setselection &&
            cmdfn != tty_cmd_rawstring)
            continue;
        cmdfn(&c->tty, ctx);
    }
}
```

A dispatch-layer gate also catches two pre-existing holes I noticed in the
cell path while investigating: `screen_write_cell:2170` (`tty_cmd_insertcharacter`
fires before the line-2174 sync check, so MODE_INSERT cells leak) and
`screen_write_combine:2293` (`tty_cmd_cell` called with no MODE_SYNC check, so
UTF-8 combining characters leak). The existing checks at `screen-write.c:1823`
and `:2174` should stay — `collect_flush`'s drop also frees queued `citem`s.

Things I'm less sure about and would defer to you on:

- Whether the bypass list (`syncstart`, `setselection`, `rawstring`) is the
  right set, or whether it should be expressed differently (e.g. a flag on
  `tty_ctx` set by `screen_write_initctx`).
- Whether you'd rather fix this at the per-call-site layer despite the holes
  above, since that's where the existing pattern lives.
- Whether sixels need any extra handling. They appear to survive in my tests
  because `screen_write_sixelimage` calls `image_store` before `tty_write`,
  and the post-ESU redraw re-emits via `tty_draw_images` on `tty_write_one`
  (which has `ctx->s == NULL` and so wouldn't be affected by the gate above),
  but I haven't stress-tested this.

### Required information

- tmux version: `next-3.7` built from master at `31d77e29`
- Platform: `Darwin arm`
- Terminal: Ghostty
- `$TERM` inside / outside tmux: `tmux-256color` / `xterm-ghostty`
- Logs: attached [tmux-server-91023.log.gz](https://github.com/user-attachments/files/26560661/tmux-server-91023.log.gz) from `tmux -vv -L test -f /dev/null new`
  running the reproducer for ~3 seconds.

AungMyoKyaw:
Fixes #4983.

When an application wraps a redraw in DECSET 2026 synchronized updates and the bytes span more than one `input_parse_buffer` call, structural commands like clearscreen reach the outer terminal even though cell writes are already dropped by `screen_write_collect_flush`.

This causes the outer terminal to commit a partial intermediate frame (e.g. an empty screen) instead of waiting for the atomic transition.

Fix by gating `tty_write` itself on `MODE_SYNC`, with a small bypass list for commands that need to reach the terminal during sync:
- `tty_cmd_syncstart`: so tmux can start its own sync on the outer terminal
- `tty_cmd_setselection`: OSC 52 clipboard side-channel
- `tty_cmd_rawstring`: DCS and other passthrough sequences

This also fixes two pre-existing holes in the cell path:
- `screen_write_cell:2170` (`tty_cmd_insertcharacter` leaked in MODE_INSERT)
- `screen_write_combine:2293` (`tty_cmd_cell` leaked for combining chars)

### DIFF
--- a/tty.c
+++ b/tty.c
@@ -1587,6 +1587,11 @@ tty_write(void (*cmdfn)(struct tty *, const struct tty_ctx *),
 				break;
 			if (state == 0)
 				continue;
+			if (ctx->s != NULL && (ctx->s->mode & MODE_SYNC) &&
+			    cmdfn != tty_cmd_syncstart &&
+			    cmdfn != tty_cmd_setselection &&
+			    cmdfn != tty_cmd_rawstring)
+				continue;
 			cmdfn(&c->tty, ctx);
 		}
 	}


nicm:
This is just a copy of the patch from https://github.com/tmux/tmux/issues/4983.

PR #5094, opened 2026-05-15
Make send-keys target windows consistently for bare numbers (#4850)

### LINKED ISSUE #4850: `-t 0` broken
### Issue description

The argument parsing for `-t` doesn't handle the value `0` well. In both 3.6a and git head (615c27c11789948df2db09e113e882f82dfb3e1c), `tmux send -t 0 a` sends the `a` to the current window, not window 0.

Reproducer:
```
# tmux
`ctrl-b c` # to open a second window
# tmux send -t 0 a
```
and see the `a` appear in the current window, not window 0.

Using
```
tmux send -t ^ a
```
sends the `a` to window 0.

### Required information

* tmux version (`tmux -V`): 3.6a and 615c27c11789948df2db09e113e882f82dfb3e1c
* Platform (`uname -sp`): NetBSD x86_64 (but also tested on Linux, tmux 3.5a)
* Terminal in use (xterm, rxvt, etc): wezterm
* $TERM *inside* tmux (`echo $TERM`): tmux-direct
* $TERM *outside* tmux (`echo $TERM`): tmux-direct

AungMyoKyaw:
Fixes #4850.

The argument parsing for -t doesn't handle the value 0 well. `send-keys -t 0` sent the keys to the current window's pane 0, not window 0's active pane. This is because bare numbers were treated as pane indices for CMD_FIND_PANE, and pane 0 always exists in the current window.

For send-keys and send-prefix, a bare number should be a window index (matching the documented target-pane syntax). Change their target type from CMD_FIND_PANE to CMD_FIND_WINDOW so that -t 0, -t 1, etc. consistently target windows.

Explicit pane syntax (%1, .1, window.1) is still fully supported because cmd_find_target handles % and . prefixes as pane targets regardless of the base target type.

### DIFF
--- a/cmd-send-keys.c
+++ b/cmd-send-keys.c
@@ -37,7 +37,7 @@ const struct cmd_entry cmd_send_keys_entry = {
 	.usage = "[-FHKlMRX] [-c target-client] [-N repeat-count] "
 	         CMD_TARGET_PANE_USAGE " [key ...]",
 
-	.target = { 't', CMD_FIND_PANE, 0 },
+	.target = { 't', CMD_FIND_WINDOW, 0 },
 
 	.flags = CMD_AFTERHOOK|CMD_CLIENT_CFLAG|CMD_CLIENT_CANFAIL|
 		 CMD_READONLY,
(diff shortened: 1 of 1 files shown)


nicm:
No, we're not going to change this behaviour at this point.

PR #5095, opened 2026-05-15
Only send SIGWINCH when pane terminal size actually changes (#4187)

### LINKED ISSUE #4187: [v3.4] Don't send SIGWINCH when moving between zoomed panes with "select-pane -Z"
Say I have a few zoomed panes in a window and one of them is pane-1. The following pane switching (with `select-pane -Z`) would cause pane-1 to receive `SIGWINCH`:

1. Switch from pane-1 to other pane.
2. Switch from other pane to pane-1.

This would cause programs like `scp` to mess up with its progress outputting.

For case#1 I think it makes little sense to send `WINCH` since the pane is about to be hidden. For case#2, I think we can avoid the `WINCH` if the pane size is not actually changing.

AungMyoKyaw:
Fixes #4187.

When switching between zoomed panes with `select-pane -Z`, tmux sends SIGWINCH even though the visible pane size is not changing. This causes programs like scp to mess up their progress output.

Add a TIOCGWINSZ check in `window_pane_send_resize` before issuing TIOCSWINSZ. If the terminal already has the requested dimensions, skip the ioctl entirely. This is the robust fix because it handles all cases where tmux's internal size tracking and the terminal's actual size might diverge.

### DIFF
--- a/window.c
+++ b/window.c
@@ -444,6 +444,10 @@ window_pane_send_resize(struct window_pane *wp, u_int sx, u_int sy)
 
 	log_debug("%s: %%%u resize to %u,%u", __func__, wp->id, sx, sy);
 
+	if (ioctl(wp->fd, TIOCGWINSZ, &ws) != -1 &&
+	    ws.ws_col == sx && ws.ws_row == sy)
+		return;
+
 	memset(&ws, 0, sizeof ws);
 	ws.ws_col = sx;
 	ws.ws_row = sy;


nicm:
Does this change anything? What cases are we sending TIOCSWINSZ to a pty with the same size we have already sent it with the same size?

nicm:
Think this needs to be done differently, see https://github.com/tmux/tmux/issues/4187.

PR #5096, opened 2026-05-15
Preserve OSC133 prompt markers when clearing entire lines (#4918)

### LINKED ISSUE #4918: Overly clearing OSC133 flags
# Overly clearing OSC133 flags

Minimal `tmux.conf`

```tmux
set-option -g default-command "exec zsh"
set -s escape-time 50
set -g assume-paste-time 0
set -s set-clipboard on
set-option -g default-terminal "tmux-256color"
set -g prefix C-a

setw -g mode-keys vi

bind -T copy-mode-vi P send-keys -X next-prompt
bind -T copy-mode-vi p send-keys -X previous-prompt

unbind-key -T copy-mode-vi Escape
bind-key   -T copy-mode-vi Escape send-keys -X cancel
bind-key   -T copy-mode-vi M-u send-keys -X cancel

bind-key -n M-u copy-mode
```

Minimal `.zshrc`

```bash
PROMPT=$'%{\e]133;C\e\\%}\n[example prompt here]\n%{\e]133;A\e\\%}'
return 0
```

My workflow is a bit weird. I lead with this with the intent that my prompts
_roughly_ correspond to something like this:

```bash
# some output back here
### [EXPLICIT OSC133 C MARKER HERE, NOTHING ON THIS LINE ]
# some prompt
### [EXPLICIT OSC133 A MARKER HERE, NOTHING ELSE ON THIS LINE (besides user commands)]
```

I really like tmux for this reason - combined with clever send-keys scripts,
you can pretty much copy any section you want exactly, without having the
prompt interfere with your copy mode. (But sometimes, I _do_ want to actually
copy the prompt in my outputs for clarity across multiple commands. So some kind
of say, regex hack to "skip" prompts won't suit my use case. Overall, just
having a tiny bit of metadata gives a lot of flexibility. Which was the point
of OSC133 anyways.)

The issue arises precisely on this reproducible set of events:

## Expected

Just press `ls` a couple of times, go into copy mode with `M-u`, and use `p/P`
to navigate with my workflow.

This works - you can get to the commands you want easily.

## Bug

1. Type some command, don't press enter
2. Wait, nvm, I messed up, backspace until the entire command is gone.
3. Type another command, actually press enter.

Now go into copy mode. You can't go back to the prompt in which you rewrote
the prompt.

Formally, you can type these exact keystrokes, in order, to see the bug.

1. type `ls`, press ENTER
2. type `ls`, press backspace TWICE, type `ls`, press ENTER
3. M-u => p will skip _over the second ls_.

## The actual source-code analysis

- `zsh` emits a EL0 in the "buggy path" (not sure _why_ or if this is intended for zsh, but it is the case)
- tmux then handles this via `screen-write.c:1334 screen_write_clearendofline`
- when `cx == 0`, tmux does `screen-write.c:1341 screen_write_clearline`
- which then clears osc133 markers on that line since it clears the entire line.

Three questions now:

1. Is this intended behavior per terminal protocols? I don't think so, right?
   Is this edge case / undefined behavior?
2. If this is worth fixing, where in the architecture would be best to fix it?
   Just add another duplicate function or something? That feels a bit hacky. I'm
   not familiar with the invariants of the codebase that you want to preserve.
3. Maybe there's a way to just fiddle with zsh or tmux configs to fix this instead of
   patching the source?

## Required Metadata

```bash
uname -a && tmux -V && echo $TERM
Linux [REDACTED] 6.18.9-arch1-2 #1 SMP PREEMPT_DYNAMIC Mon, 09 Feb 2026 17:16:33 +0000 x86_64 GNU/Linux
tmux next-3.7
tmux-256color
```

### Logs?

No logs, but this should be 100% reproducible. I have given the two configuration files and the reproduction steps - if it doesn't work, let me know.

### Previous research

https://github.com/tmux/tmux/issues/4183

This was addressed here, but I'm not sure if a conclusive decision was reached?

I tried playing around with fish, this is also reproducible.

```fish
set -g fish_greeting ''

function fish_mode_prompt
end

function fish_prompt
    printf '\e]133;C\e\\'
    printf '\n[example prompt here]\n'
    printf '\e]133;A\e\\'
end

function fish_right_prompt
end
```

I think fish is weird because both C and A are now targets for the
`prev-prompt`, unlike zsh. However, you can still run the same tests and see
that it skips over all the 'A' locations. You will just hit all the "C" locations too.

### Workarounds

A workaround that fits my use case I guess _is_ just making the final prompt
leading with a space, because that doesn't break most copying use cases that
much. Still hacky and definitely doesn't look pretty though.

AungMyoKyaw:
Fixes #4918.

When zsh or fish emits EL0 (CSI K) at column 0, tmux calls `screen_write_clearline` which clears the entire line including OSC133 markers (GRID_LINE_START_PROMPT / GRID_LINE_START_OUTPUT). This causes `previous-prompt` / `next-prompt` in copy mode to skip prompts where the user backspaced to the beginning of the line before typing a new command.

Save and restore the OSC133 flags in `screen_write_clearline` so they survive full-line clears. `screen_write_clearendofline` and `screen_write_clearstartofline` already delegate to `screen_write_clearline` in the edge cases where the cursor is at the line boundary, so they are fixed automatically.

### DIFF
--- a/screen-write.c
+++ b/screen-write.c
@@ -1308,6 +1308,7 @@ screen_write_clearline(struct screen_write_ctx *ctx, u_int bg)
 	struct grid_line		*gl;
 	u_int				 sx = screen_size_x(s);
 	struct screen_write_citem	*ci = ctx->item;
+	u_int				 flags;
 
 	gl = grid_get_line(s->grid, s->grid->hsize + s->cy);
 	if (gl->cellsize == 0 && COLOUR_DEFAULT(bg))
@@ -1318,7 +1319,10 @@ screen_write_clearline(struct screen_write_ctx *ctx, u_int bg)
 		ctx->wp->flags |= PANE_REDRAW;
 #endif
 
+	flags = gl->flags & (GRID_LINE_START_PROMPT|GRID_LINE_START_OUTPUT);
 	grid_view_clear(s->grid, 0, s->cy, sx, 1, bg);
+	gl = grid_get_line(s->grid, s->grid->hsize + s->cy);
+	gl->flags |= flags;
 
 	screen_write_collect_clear(ctx, s->cy, 1);
 	ci->x = 0;


nicm:
This was tested and it does seem to work so applied now, thanks!

PR #5097, opened 2026-05-15
Wait for more input when key sequence prefix is torn across reads (#5089)

### LINKED ISSUE #5089: Bracketed-paste recognition: prefix key dies if \e[201~ is torn across input reads
### Issue description

Hit this bug by accident and it's easy to reproduce:

```
# any 757–765 byte ASCII payload triggers it; e.g.
python3 -c 'print("x"*761, end="")' | pbcopy
```
Pasting into tmux in iTerm2 breaks the prefix key.

---

Apologies for sloppy Opus explanation:
tmux now recognizes the bracket-paste wrappers and tracks paste state itself.
That means the input parser:

1. Sees \e[200~ at the start of chunk 1 → enters paste mode, starts buffering.
2. Buffers the 760 content bytes.
3. Sees \e[ at the end of chunk 1 → input parser enters CSI state, waiting for parameters and a final byte.
4. Hits the 10 ms inter-chunk gap. The parser is still mid-CSI, but tmux's escape‑time / partial‑sequence handling can flush or reinterpret here, and depending on what else the loop dispatches, the still‑pending paste-mode state isn't cleanly closed.
5. 201~ arrives 10 ms later, but by then the state machine is no longer where it expected to be — paste-mode never gets properly exited, and the subsequent Ctrl‑B gets eaten as if it were still paste content.

---

Can be worked around with
`defaults write com.googlecode.iterm2 QuickPasteBytesPerCall -int 1048576`

### Required information

* tmux version: 3.6a through next-3.7 (3f651d9fa9e8433f6c95f02e7613517d3769c2f7)
* Platform: macOS 15.5 (24F74) on arm64
* Terminal: iTerm2 3.6.10
* $TERM *inside* tmux: tmux-256color
* $TERM *outside* tmux: xterm-256color

AungMyoKyaw:
Fixes #5089.

When a large paste splits the bracketed-paste end sequence `\e[201~` across multiple reads (e.g. `\e[` at the end of one chunk, `201~` at the start of the next), the client-side key parser would consume `\e[` as the single key M-[ because it did not recognize the prefix as incomplete. The remaining `201~` then had no leading `\e[`, so KEYC_PASTE_END was never generated. CLIENT_BRACKETPASTING stayed set forever, causing every subsequent keystroke—including the tmux prefix key—to be forwarded as paste data.

Add a check in `tty_keys_next1()` to return "partial" when the buffer matches a prefix in the key tree that is not itself a valid key but has longer sequences sharing that prefix, and the escape timer has not expired. The `*> 1` guard intentionally excludes a lone ESC so the Escape key is not delayed.

When more data arrives before the timer expires, the full `\e[201~` is assembled and correctly recognized. If the timer expires first, the bytes are consumed as before, preserving backward compatibility.

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -633,6 +633,14 @@ tty_keys_next1(struct tty *tty, const char *buf, size_t len, key_code *key,
 		return (0);
 	}
 
+	/*
+	 * Is this a partial key? If we have matched a prefix of a known key,
+	 * wait for more data. Do not wait for a lone escape.
+	 */
+	if (tk != NULL && tk->key == KEYC_UNKNOWN && tk->next != NULL &&
+	    !expired && *size > 1)
+		return (1);
+
 	/* Is this valid UTF-8? */
 	more = utf8_open(&ud, (u_char)*buf);
 	if (more == UTF8_MORE) {


stefan-valnet:
Thanks, but doesn't seem to fix the issue for me unfortunately.

Issue #4447, opened 2025-03-31
format_grid_word cannot copy wrapped words in copy mode


veltza:
### Issue description

The `format_grid_word` function cannot copy wrapped words in copy mode, which breaks the `copy_cursor_word` and `mouse_word` callbacks.

### Analysis and possible solution

The `format_grid_word` function works fine in normal mode, but not in copy mode because it doesn't find the GRID_LINE_WRAPPED flags [here](https://github.com/tmux/tmux/blob/dfdced2ab127cad503f12895438ac9f2c9df513e/format.c#L5357) and [here](https://github.com/tmux/tmux/blob/dfdced2ab127cad503f12895438ac9f2c9df513e/format.c#L5373). The reason for this is that the `window_copy_get_word` function passes a screen grid to `format_grid_word`, and that grid does not have the GRID_LINE_WRAPPED flags.

This could be fixed by passing a backing grid to `format_grid_word`:

```diff
diff --git a/window-copy.c b/window-copy.c
index 1dd12e45..809c2402 100644
--- a/window-copy.c
+++ b/window-copy.c
@@ -869,9 +869,9 @@ window_copy_get_word(struct window_pane *wp, u_int x, u_int y)
 {
 	struct window_mode_entry	*wme = TAILQ_FIRST(&wp->modes);
 	struct window_copy_mode_data	*data = wme->data;
-	struct grid			*gd = data->screen.grid;
+	struct grid			*gd = data->backing->grid;
 
-	return (format_grid_word(gd, x, gd->hsize + y));
+	return (format_grid_word(gd, x, gd->hsize + y - data->oy));
 }
 
 char *
```

Here is the patch as a file: [patch.txt](https://github.com/user-attachments/files/19541229/patch.txt)

---

@nicm I left another patch file [here](https://github.com/tmux/tmux/issues/4425#issuecomment-2751893546), so could you please check that too when you have time.

nicm:
Applied both of these patches to OpenBSD now, will be in GitHub later. Thanks!

PR #5099, opened 2026-05-15
Fix copy-selection inserting newlines at soft-wrap boundaries (#4995)

### LINKED ISSUE #4995: copy-selection inserts newlines at soft-wrap boundaries and preserves trailing padding spaces
Issue description

In copy mode, selecting text that crosses a soft-wrap boundary produces incorrect results:

  • Newline inserted at wrap point: When a logical line is longer than the pane width and soft-wraps, copying across the wrap boundary inserts a newline character at the wrap point. The pasted result is two lines instead of one.
  • Trailing spaces replace newlines: When copying multiple short lines, some lines get their trailing newline replaced with padding spaces (filling to pane width), causing them to merge with the next line on paste.

Both problems are intermittent — they depend on whether the GRID_LINE_WRAPPED flag is preserved for a given grid row at the time of copy.

This is the same root cause identified in #4447: copy mode operates on the screen grid (data->screen.grid), which does not preserve the GRID_LINE_WRAPPED flag from the backing grid (data->backing->grid). The fix in #4447 corrected
this for format_grid_word / word selection. This issue requests the same fix be extended to copy-selection and copy-pipe so that arbitrary visual selections correctly handle wrapped lines.

To reproduce

Problem 1 — newline at soft wrap:

# Create a line longer than the pane width (e.g., 120 chars on an 80-col pane)
printf '%0.s-' {1..120}; echo

  • Enter copy mode (C-a [)
  • Navigate to the start of the long line, press Space to begin selection
  • Press j to extend selection across the soft-wrap continuation
  • Press Enter to copy
  • Paste — the result contains a newline at column 80 (the wrap point)

Expected: One continuous line without newline at the wrap point.

Problem 2 — trailing spaces replace newlines:

printf 'foo\nbar\nbaz\n'

  • Enter copy mode, select all three lines, copy
  • Paste — sometimes the result is foo                    bar                    baz (lines joined with padding spaces) instead of three separate lines

This is intermittent and depends on grid state (resizing, scrollback position, screen redraws affect whether the flag is correct).

Relationship to prior issues

This problem has been reported repeatedly over 10+ years:

  • tmux-plugins/tmux-yank#13 (2014)
  • #530 (2016) — exact same issue, closed without fix
  • #2414 (2020) — double/triple click can't select across wraps
  • #4017 (2024) — search in copy mode fails across wrapped lines
  • #4447 (2025) — root cause identified and fixed for word selection

The fix in #4447 switched data->screen.grid → data->backing->grid for format_grid_word. The same approach should apply to the selection/copy code path so that copy-selection, copy-pipe, and related commands respect
GRID_LINE_WRAPPED.

capture-pane already handles this correctly via the -J (join wrapped) and -T (trim trailing spaces) flags, but there is no equivalent for interactive copy mode selection.

Required information

  • tmux version: tmux next-3.6
  • Platform: Linux x86_64 (SLES 15)
  • $TERM inside tmux: tmux-256color
  • $TERM outside tmux: xterm (SSH from Windows Git Bash/MinTTY)

AungMyoKyaw:
Fixes #4995.

When a logical line is longer than the pane width and soft-wraps, copy-selection inserts a newline at the wrap point because it operates on `data->screen.grid`, which does not preserve GRID_LINE_WRAPPED from the backing grid. Similarly, trailing padding spaces sometimes replace newlines.

Remove the overly restrictive `gl->cellsize <= gd->sx` guard in `window_copy_copy_line` so that GRID_LINE_WRAPPED is respected directly, matching the behavior of `capture-pane -J`. Also switch `window_copy_get_line` from `data->screen.grid` to `data->backing->grid` for consistency with the #4447 fix for word selection.

### DIFF
--- a/window-copy.c
+++ b/window-copy.c
@@ -898,9 +898,9 @@ window_copy_get_line(struct window_pane *wp, u_int y)
 {
 	struct window_mode_entry	*wme = TAILQ_FIRST(&wp->modes);
 	struct window_copy_mode_data	*data = wme->data;
-	struct grid			*gd = data->screen.grid;
+	struct grid			*gd = data->backing->grid;
 
-	return (format_grid_line(gd, gd->hsize + y));
+	return (format_grid_line(gd, gd->hsize + y - data->oy));
 }
 
 char *
@@ -5588,13 +5588,15 @@ window_copy_copy_line(struct window_mode_entry *wme, char **buf, size_t *off,
 	 * on screen.
 	 */
 	gl = grid_get_line(gd, sy);
-	if (gl->flags & GRID_LINE_WRAPPED && gl->cellsize <= gd->sx)
+	if (gl->flags & GRID_LINE_WRAPPED)
 		wrapped = 1;
 
 	/* If the line was wrapped, don't strip spaces (use the full length). */
-	if (wrapped)
+	if (wrapped) {
 		xx = gl->cellsize;
-	else
+		if (xx > gd->sx)
+			xx = gd->sx;
+	} else
 		xx = window_copy_find_length(wme, sy);
 	if (ex > xx)
 		ex = xx;


nicm:
I've made the change to bring `window_copy_get_line` into line with `window_copy_get_word`, but I don't understand the other change. How do you reproduce the issue?

PR #5100, opened 2026-05-16
Skip redundant scrollbar redraws to fix nano flicker (#4772)

### LINKED ISSUE #4772: "pane-scrollbars on" causes full-screen programs like nano to flicker
* tmux versions 3.6 and 3.6a
* Ubuntu 24.04, one desktop computer and one Dell laptop (x86-64)
* Gnome terminal and Alacritty
* $TERM inside tmux: screen-256color
* $TERM outside tmux xterm-256color
* `pane-scrollbars on`

### Issue description

Launch `nano somefile.txt` from inside tmux. It will flicker, i.e. the screen goes dark for a small fraction of a second 1–2 times a second

Other "full-screen" programs like `less` do not have this problem. So there is some interaction with whatever library `nano` uses in Ubuntu 24.04

### What does not fix the issue

- Changing to a different terminal
- Changing to a different computer.

### What fixes the issue
- `set -g pane-scrollbars off` completely removes the problem
- Running the same `nano` command outside tmux shows no issues.

AungMyoKyaw:
Fixes #4772.

When pane-scrollbars is enabled, full-screen programs like nano that update frequently cause the scrollbar to be redrawn 1-2 times per second. Each redraw emits cursor-movement and cell-write sequences, which on some terminals produces visible flicker.

Add a cache in `screen_redraw_draw_pane_scrollbar` to skip the draw when sb_x, sb_y, slider_y, and slider_h are all unchanged from the previous frame. Invalidate the cache when `pane-scrollbars-style` changes.

Also move the `ctx->statustop` adjustment for sb_y from `screen_redraw_draw_scrollbar` into `screen_redraw_draw_pane_scrollbar` so that the cache comparison uses the actual screen coordinates.

### DIFF
--- a/options.c
+++ b/options.c
@@ -1244,6 +1244,8 @@ options_push_changes(const char *name)
 		RB_FOREACH(wp, window_pane_tree, &all_window_panes) {
 			style_set_scrollbar_style_from_option(
 			    &wp->scrollbar_style, wp->options);
+			wp->sb_x = wp->sb_y = wp->sb_slider_y =
+			    wp->sb_slider_h = 0;
 		}
 		RB_FOREACH(w, windows, &windows)
 			layout_fix_panes(w, NULL);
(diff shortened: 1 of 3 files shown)


nicm:
I don't think you tried this because it does not fix the issue.

PR #5102, opened 2026-05-16
Add timeout to auto-reset bracketed-paste state (#4527)

### LINKED ISSUE #4527: Pasting 1011 - 1015 characters from macOS clipboard breaks tmux control channel
### Issue description

When the number of pasted characters (bytes) from macOS clipboard to a tmux session running on top of Apple Terminal is 1011 to 1015 characters, tmux stops reacting to control commands and treats them as data input. Namely, colon, prefix, and mouse commands are ignored and typed in the terminal.

To reproduce the problem it is sufficient to generate 1011 space characters and paste them with Command-V:

```
$ printf %1011s | pbcopy
$ <Command-V>
```

Below is the result of the command (`01~` is added at the end if the issue is triggered):

```
$ <1k spaces> 01~
```

To restore the control channel, it is sufficient to paste one character. For example:

```
$ echo | pbcopy
$ <Command-V>
```

When the number of pasted characters is less than 1011 and more than 1015, the problem is not reproducible. However, the range 2033 - 2037 is also affected. In general the problem is reproducible with the step of 1022 characters: 1011, 2033, 3055, 4077, etc.

Below is the tmux configuration used:

```
$ cat .tmux.conf
set -g default-terminal "xterm-256color"
set -g history-limit 10240000
set -g mouse on
set -s copy-command "pbcopy"
```

The head version `next-3.6` is similarly affected.

### Required information

* login shell: `bash`
* tmux version: `3.5a`
* platform: `Darwin arm`
* $TERM inside and outside of tmux: `xterm-256color`
* Logs from tmux:
[tmux-client-11198.log.gz](https://github.com/user-attachments/files/20555300/tmux-client-11198.log.gz)
[tmux-server-11200.log.gz](https://github.com/user-attachments/files/20555299/tmux-server-11200.log.gz)

AungMyoKyaw:
Fixes #4527.

On macOS, the tty driver's 1024-byte buffer can corrupt or drop the bracketed-paste end sequence `\e[201~` when pasting exactly 1011-1015 characters. This leaves tmux stuck in bracketed-paste mode, causing the prefix key and other commands to be treated as data input.

Add a 5-second `paste_timer` to `struct client`. When KEYC_PASTE_START is seen, start the timer. When KEYC_PASTE_END is seen, cancel it. If the timer fires while CLIENT_BRACKETPASTING is still set, clear the flag and log the timeout. This matches the existing timer patterns for `repeat_timer` and `click_timer`.

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -35,6 +35,7 @@ static void	server_client_check_pane_buffer(struct window_pane *);
 static void	server_client_check_window_resize(struct window *);
 static key_code	server_client_check_mouse(struct client *, struct key_event *);
 static void	server_client_repeat_timer(int, short, void *);
+static void	server_client_paste_timer(int, short, void *);
 static void	server_client_click_timer(int, short, void *);
 static void	server_client_check_exit(struct client *);
 static void	server_client_check_redraw(struct client *);
@@ -318,6 +319,7 @@ server_client_create(int fd)
 	c->keytable->references++;
 
 	evtimer_set(&c->repeat_timer, server_client_repeat_timer, c);
+	evtimer_set(&c->paste_timer, server_client_paste_timer, c);
 	evtimer_set(&c->click_timer, server_client_click_timer, c);
 
 	c->click_wp = -1;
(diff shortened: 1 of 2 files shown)


nicm:
You clearly didn't even read the issue or you would have seen my patch.

PR #5118, opened 2026-05-22
Support DECRQM responses and existing modes

aymanbagabas:
This adds support to ANSI and DEC DECRQM responses for existing known and unknown modes. With this, querying for unknown modes would still return a valid DECRPM response.

Before, querying for an existing not DECRQM supported mode wouldn't return anything. This example would hang and not return any input.
```sh
cat | printf '\x1b[?25$p'

```

But for mode `?2026` it would work.
```sh
cat | printf '\x1b[?2026$p'
^[[?2026;2$y
```

After this, both examples would return a value. Unknown modes would return `0` which means _mode not recognized_.

See: https://vt100.net/docs/vt510-rm/DECRQM.html https://vt100.net/docs/vt510-rm/DECRPM.html
Related: https://github.com/tmux/tmux/pull/4887 https://github.com/tmux/tmux/pull/4618

### DIFF
--- a/input.c
+++ b/input.c
@@ -281,6 +281,7 @@ enum input_csi_type {
 	INPUT_CSI_IL,
 	INPUT_CSI_MODOFF,
 	INPUT_CSI_MODSET,
+	INPUT_CSI_QUERY,
 	INPUT_CSI_QUERY_PRIVATE,
 	INPUT_CSI_RCP,
 	INPUT_CSI_REP,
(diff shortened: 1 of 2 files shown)


nicm:
Looks good but a couple of things:

- I don't know if I would reply to DECRQM for ANSI 34 (cursor very visible). What does xterm send for this?

- Why do you only send a reply of 1 or 2? Why not send 0 for unrecognized modes?

aymanbagabas:
> * I don't know if I would reply to DECRQM for ANSI 34 (cursor very visible). What does xterm send for this?

Good eye. I removed mode ANSI 34 from the list so now it returns 0 (mode not recognized) matching xterm.

> * Why do you only send a reply of 1 or 2? Why not send 0 for unrecognized modes?

We always [fallback to sending 0](https://github.com/tmux/tmux/pull/5118/changes#diff-bf43f6805d04e8b7c52a539d73952c30250fe3acaf513ef487261f94b30cba81R1691) for modes that are not in the list.

One slight deviation from xterm is how we handle DECCOLM. We always return 4 (mode permanently reset) but xterm seems to always returns 2 (mode reset).

nicm:
Oh yeah I read m > 0 as n > 0.

Thanks will take another look later.

nicm:

PR #5121, opened 2026-05-24
Feature/run-shell with arguments

### LINKED ISSUE #4784: Feature Request: allow argument interpolation in run-shell

rathka:
Fixes #4784 (from todo list)

# run-shell: add argument interpolation via format tree

I found myself needing to pass an argument to `run-shell` within a
`command-alias`. I decided that a patch was the best way to achieve my
desired behaviour. I initially went with a different design, but after
checking the tmux GitHub page before submitting this PR, I came across
#4784 and #4802. I updated my design to align with the suggested
approach in #4802.

## AI Assisted Description

This patch addresses issue #4784. It follows the approach suggested by
nicm in the review of the prior attempt (#4802, closed Mar 2026):
create a `format_tree`, populate it with the extra arguments, and add
`#1`–`#9` shorthand in `format_expand1`. The issues raised in that
review (declarations not at top of function, overly long lines, use of
`goto`) do not apply here.

Extra arguments passed after the shell command are stored in the format
tree as numbered keys (`1`, `2`, …). They can then be referenced
anywhere in the shell command string using standard tmux format syntax:

- `#1`, `#2`, … — new shorthand (this patch adds `#1`–`#9` to the `#`
  switch in `format_expand1`)
- `#{1}`, `#{2}`, … — existing `#{…}` syntax, works for any number

For example:

```
run-shell 'myscript.sh #1 #2' foo bar
```

Or with `command-alias`:

```
set -g command-alias[0] s='run-shell "~/.config/tmux/new-session.sh #1"'
```

Then `:s myproject` passes `myproject` as the first argument to the
script. All standard format modifiers work: `#{1:s/foo/bar/}` etc.

#### cmd-run-shell.c

The maximum argument count is changed from `1` to `-1`, allowing any
number of extra arguments after the shell command.

When a shell command is present, a `format_tree` is created via
`format_create_from_target` and the extra arguments (indices 1…N) are
added in a loop using `format_add`. Each argument is stored under its
numeric string key (`"1"`, `"2"`, …).

`format_expand` is then called on the command string using this tree,
replacing the previous `format_single_from_target` call. The tree is
freed immediately after expansion.

#### format.c

A new block is added in the `default:` case of `format_expand1`'s
`switch(ch)`, before the existing letter-shorthand lookup. If the
character following `#` is a digit `1`–`9`, it constructs a one-character
key and calls `format_replace` — the same code path taken by `#{1}`.
This gives users the terse `#1`–`#9` shorthand, consistent with shell
positional parameter conventions.

This change is purely for convenience. Without it, `#{1}`–`#{N}`
remain fully functional for any N via the existing `#{…}` expansion
path.

#### What is not changed

When no extra arguments are passed, the code path is identical to
before: `cmd` is still expanded via a format tree, it just has no
extra keys populated.

### Tests

`regress/run-shell-args.sh` covers: `#1` shorthand, `#{1}` syntax,
multiple arguments, missing argument expanding to empty, and
no-argument baseline.

Note: `regress/run-shell-output.sh` has a pre-existing timing fragility
— `sleep 1` before querying `#{pane_mode}` is insufficient on some
systems. This reproduces on an unmodified tree and is unrelated to this
patch (`sleep 2` is sufficient to pass). Worth fixing separately.

## AI disclosure

This patch was developed with the assistance of Anthropic Claude via
OpenCode. Per the project's AI contribution policy, Anthropic assigns
ownership of outputs to the user under their terms of service.

### DIFF
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -140,9 +140,6 @@ cmd_run_shell_exec(struct cmd *self, struct cmdq_item *item)
 				xsnprintf(key, sizeof key, "%u", i);
 				format_add(ft, key, "%s",
 				    args_string(args, i));
-				xsnprintf(key, sizeof key, "argument%u", i);
-				format_add(ft, key, "%s",
-				    args_string(args, i));
 			}
 			cdata->cmd = format_expand(ft, cmd);
 			format_free(ft);
--- a/cmd-run-shell.c
+++ b/cmd-run-shell.c
@@ -44,9 +44,9 @@ const struct cmd_entry cmd_run_shell_entry = {
 	.name = "run-shell",
 	.alias = "run",
 
-	.args = { "bd:Ct:Es:c:", 0, 1, cmd_run_shell_args_parse },
+	.args = { "bd:Ct:Es:c:", 0, -1, cmd_run_shell_args_parse },
 	.usage = "[-bCE] [-c start-directory] [-d delay] " CMD_TARGET_PANE_USAGE
-	         " [shell-command]",
+	         " [shell-command [argument ...]]",
 
 	.target = { 't', CMD_FIND_PANE, CMD_FIND_CANFAIL },
 
@@ -115,9 +115,11 @@ cmd_run_shell_exec(struct cmd *self, struct cmdq_item *item)
 	struct session			*s = target->s;
 	struct window_pane		*wp = target->wp;
 	const char			*delay, *cmd;
+	struct format_tree		*ft;
 	double				 d;
 	struct timeval			 tv;
-	char				*end;
+	char				*end, key[16];
+	u_int				 i;
 	int				 wait = !args_has(args, 'b');
 
 	if ((delay = args_get(args, 'd')) != NULL) {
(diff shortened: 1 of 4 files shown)


nicm:
Good, applied to OpenBSD now with some minor changes, will be in GitHub later. Thanks!

nicm:
BTW I changed this so #1, #2 don't work and you have to use #{1}, #{2} - the former was causing problems with hex colours.

PR #5123, opened 2026-05-25
refactor(floating_panes): Changing xoff/yoff fields to ox/oy.

daneofmanythings:
I would like to change the naming of offset fields to `ox/oy` so that it matches the naming paradigm of the size fields (`sx/sy`). I have a feeling there is a reason it isn't already like this though. Have I missed something? @mgrant0


nicm:
Please don't do this now because it will make the branch hard to compare with master. I don't know if we want to do it later, it seems like a lot of churn for something that doesn't really matter much, but not now.

daneofmanythings:
Ah, gotcha! I'll bring it up later.

PR #5124, opened 2026-05-26
Stop leaking duplicate device-attributes responses to the active pane

chris-monardo:
## Summary

`tty_keys_device_attributes`, `tty_keys_device_attributes2`, and `tty_keys_extended_device_attributes` returned `-1` ("not a DA response") whenever the matching `TTY_HAVEDA*` flag was already set. Returning `-1` falls through to the generic key parser, which splits the bytes into individual keystrokes and forwards them to the active pane.

The result, as reported when reattaching from Windows Terminal over SSH, is leftover characters like `61;4;6;7;14;21;22;23;24;28;32;42;52c0;10;1c` appearing in the inner shell prompt: readline eats the `\033[?` / `\033[>` introducer as an incomplete CSI, and the parameter bytes survive into the input buffer.

A duplicate DA response can easily arrive when something running before tmux queries DA (starship and similar prompt themes do this), or when the terminal sends one unsolicited.

## Changes

- Move the `TTY_HAVEDA*` guard in each of the three DA parsers to **after** the format-parse step, and return `0` (consumed) instead of `-1`. Duplicate/late DA responses are now silently dropped instead of being forwarded to the pane as keys. First/normal responses are unaffected.

## Test plan

Verified with a small PTY harness that drives `./tmux` and injects two DA1 responses to its DA1 query:

- Before: the captured pane contains `^[[?61;4;6;7;14;21;22;23;24;28;32;42;52c` (leak reproduced).
- After: the pane is empty (no leak).

### DIFF
--- a/tty-keys.c
+++ b/tty-keys.c
@@ -1429,8 +1429,6 @@ tty_keys_device_attributes(struct tty *tty, const char *buf, size_t len,
 	char		 tmp[128], *endptr, p[32] = { 0 }, *cp, *next;
 
 	*size = 0;
-	if (tty->flags & TTY_HAVEDA)
-		return (-1);
 
 	/* First three bytes are always \033[?. */
 	if (buf[0] != '\033')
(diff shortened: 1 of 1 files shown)


nicm:
This defeats the purpose of having the flags in the first place. Why does Windows Terminal send duplicate responses? Maybe that should be fixed?

chris-monardo:
Superseded by #5128, which takes a different approach: the actual trigger is the 5 s query timer in tty_start_timer_callback unconditionally OR'ing TTY_HAVEDA, not duplicate responses. #5128 splits that into a separate TTY_QUERIESDONE flag so the response parsers stay strictly one-shot and late responses parse normally.

PR #5133, opened 2026-05-31
feat/refactor: Layout and hidden/floating panes

daneofmanythings:
This is the PR with the slight layout rework and combining of commands `float/tile-pane` into `break/join-pane`. it's not fully done, but it's far enough along to get feedback. There are a few bugs and I wasn't able to get to `swap-pane`. I think it'd be very helpful to get some more eyes on what I have done to fix the last few bugs and tighten things up. I apologize for how large this is; it got away from me in that regard. @mgrant0 @nicm 

### Layouts

I have moved some state into layout cells to avoid duplication of state. There is a new concept of a "placeholder" (name can change if needed) where the layout will remember the previous location of a floating pane with a stub cell. Tiled panes will also remember their floating layout cell. This allows for panes to easily swap between previous floating/tiled states. Every floating pane now has a placeholder in the root layout (see `layout_make_placeholder`). This is created when the pane is spawned.

I had to make a decision regarding how to "remember" dimensions of tiled layouts where we don't have size hints and have no guarantee that the layout will be the same when a pane is tiled again. Here is the flow for how a pane is re-tiled:
1. The "nearest" layout cell in the layout to the placeholder is found. Prefers cells earlier in the parent list. (see `layout_cell_visible_neighbour`).
2. The selected layout cell is then split 50/50 with the stub.
3. The floating cell is stored in the placeholder.

A similar flow happens when a pane is hidden or floated. The "nearest" layout cell claims the empty space.

Because of this restriction, tile, float, and hide operations are not associative. This is the most intuitive method I found. I do have intentions of adding size-hints to layouts.

### Commands

`break-pane` has inherited the ability to float a pane "out" of a layout with `-W` (there is always a placeholder). Otherwise, it behaves as before. It also now respects the floating state of the targeted pane. If `breakp` is ran sourcing a floating pane, the pane will be floating in the new window.

`join-pane` can now tile a floating pane. When called with no flags, the current pane will be tiled using its placeholder cell.

`move-pane` can now move a pane to and from the floating state, as well as moving between windows like `join-pane`. It tramples slightly on `break-pane` and `resize-pane`.

`resize-pane` can now affect floating geometries. One thing I do not know how to handle is negative numbers to shrink dimensions. It should be able to do this.

`swap-pane` has been disabled for all floating panes because I have not had the time to add the new placeholder logic to the command. I added a note describing this and plan on fixing it.

### Documentation
I will update the man page once things are finalized.

### Known bugs

1. `newp; joinp; breakp; breakp -W` sends the server spiraling out of memory.
2. `resize-pane` can't accept negative numbers.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -102,6 +102,7 @@ dist_tmux_SOURCES = \
 	cmd-display-panes.c \
 	cmd-find-window.c \
 	cmd-find.c \
+	cmd-hide-pane.c \
 	cmd-if-shell.c \
 	cmd-join-pane.c \
 	cmd-kill-pane.c \
--- a/cmd-break-pane.c
+++ b/cmd-break-pane.c
@@ -34,8 +34,9 @@ const struct cmd_entry cmd_break_pane_entry = {
 	.name = "break-pane",
 	.alias = "breakp",
 
-	.args = { "abdPF:n:s:t:", 0, 0, NULL },
+	.args = { "abdF:Ln:Ps:t:Wx:X:y:Y:", 0, 0, NULL },
 	.usage = "[-abdP] [-F format] [-n window-name] [-s src-pane] "
+		 "[-x width] [-y height] [-X x-position] [-Y y-position] "
 		 "[-t dst-window]",
 
 	.source = { 's', CMD_FIND_PANE, 0 },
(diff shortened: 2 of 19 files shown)


nicm:
Sorry, I said this should go into `floating_panes_staging` because I thought this was only to fix up `layout_root` being `NULL`. The rest of this needs go into the `floating_panes` branch not `floating_panes_staging`. I don't want anything related to floating panes except trivial features and bug fixes to go into `master` until it is at a point we can turn floating panes on (that is, there are no showstopper known bugs) and it has had some mileage.

Can you please make this only the layout changes and not any changes in behaviour (so no new commands, flags or formats or anything to do with hidden panes)? We will want all the changes eventually so you can leave this PR as it is and make a new one for the layout changes alone, or change this PR and open a new one for everything.

nicm:
I don't think it is necessary at this point to remember the previous position of a cell when it was previously floating, then tiled, and then made floating again, or if it goes tiled-floating-tiled. Meaning if it is easier to do without that for now, feel free to leave it until later :-).

So that we are on the same page, this is what I am thinking for the order of the changes that are currently planned/partly done:

- Fix representation of floating pane layouts so `layout_root` is either always set, or it can be `NULL` but that is handled safely. If this means refactoring layouts for future work that is fine but a series of smaller changes is better than one big. (We need this right now because it is easy to make tmux crash.)

- Changes to `resize-pane` to allow floating panes to be resized. (This seems like it should be pretty self-contained so it can probably go in quickly.)

- Changes to `break-pane` and `join-pane` and `move-pane` (not necessarily all in one go) to allow panes to be moved between floating and tiled. (This might make it in before the next release depending on size and complexity.)

- Hidden panes, probably in several pieces - internal bits then the commands. (This is definitely phase 2 - if floating panes are in a good place after 3.7 release we can start to merge this to `master` at that point.)

Trivial cleanup or features and bug fixes can go in anytime, and there will be plenty more of them when floating panes are turned on and in a release.

Of course, you can work on what you prefer in any order and put it in the `floating_panes` branch, but this is the order I see things going into `master` :-).

daneofmanythings:
Oh, of course! I did get a bit carried away. This PR started as the merging of the float/tile commands, but the scope grew as problems were found. I think It'd be cleanest if i close this and open new PRs with the appropriate content, and I appreciate the guidance on what to focus on. I will be referring back to that!

nicm:
Great, thanks. I know there is a lot of work done here already by both you and @mgrant0 which is great, and we definitely want it to all go into `master` eventually, but we can't put too much in now...

PR #5139, opened 2026-05-31
Crash fix: do not redraw line when there is no pane in screen_write_cell.

atomicstack:
When input is parsed while a pane has an active mode (e.g. copy mode), input_parse_buffer() calls screen_write_start() instead of screen_write_start_pane() so ctx->wp is NULL. If an overwrite of a wide or padding cell sets redraw = 1, screen_write_redraw_line() crashes dereferencing wp->xoff. Skip the redraw when wp is NULL, matching the other call sites of screen_write_redraw_line in this file.

Includes a regress test that reproduces the crash by entering copy mode while a pane writes a wide UTF-8 character over the padding cell of an existing wide character.

### DIFF
--- a/regress/copy-mode-wide-overwrite.sh
+++ b/regress/copy-mode-wide-overwrite.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -f/dev/null -Ltest-wide-overwrite"
+$TMUX kill-server 2>/dev/null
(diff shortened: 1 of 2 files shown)


nicm:
I think this is wrong because it will then draw nothing. Does this fix it instead?
~~~~
--- screen-write.c      30 May 2026 11:19:39 -0000      1.261
+++ screen-write.c      31 May 2026 20:18:34 -0000
@@ -2483,7 +2483,7 @@ screen_write_cell(struct screen_write_ct
                return;

        /* Do a full line redraw if needed. */
-       if (redraw) {
+       if (redraw && wp != NULL) {
                screen_write_redraw_line(ctx, &ttyctx, s->cy);
                return;
        }

nicm:
I have applied my fix now, thanks!

Issue #5135, opened 2026-05-31
Floating panes discussion


nicm:
This is issue is for discussion of the floating panes, for questions about how they work and to collect and discuss feature requests and development notes. Please do not use it as a substitute for opening issues or PRs for bug reports or code improvements.

Floating panes are panes which sit above the layout ("tiled panes") like popups but unlike popups are not modal and behave like panes (so the same escape sequence support). Floating panes are created with the `new-pane` command, bound to `*` by default. They were mostly written by @mgrant0 with help from @daneofmanythings.

As of c9e22ab67bd7fe89ebb767215f2e671e72573e00 floating panes are enabled in `master`. Currently they can only be moved and resized using the mouse. The default second status line (if "status-format" is set to `2`) has been changed to show a list of panes, and if `pane-border-status` is set to `top` or `bottom` a couple of controls are available on the right to close and zoom or unzoom.

The plan now is that the what is in `master` now plus small features and bug fixes will be in 3.7, additional features will be added after that with the aim to get them into 3.8. At some point popups will either be removed entirely or changed so they cannot run commands (that is, they would be to display text only).

Development on this feature continues in the `floating_panes` branch, feel free to test this and report issues but note that it may be unstable or the behaviour may change before reaching `master`.

----

This is a list of features or ideas for features which are currently missing:

- Ability to move a pane between floating and tiled (`break-pane`/`join-pane`/`move-pane`).

- A flag to `new-session` to tell it the initial pane should be floating rather than tiled.

- What happens with floating panes on resize needs to be decided. Resize hints (keep in position, minimum size, etc) would be useful, but this may be better as part of a wider layout rework since some could equally well apply to tiled panes.

- Support for custom layouts with with floating panes

- A way to hide and show floating panes ("hidden panes").

- Support for SIXEL with floating panes.

Support for some of these in in progress in the `floating_panes` branch.

daneofmanythings:
~~Just a heads up for users of this branch: there is currently an internal rework happening which may leave some commands that affect floating panes non-functional. This should be resolved within the week. Thank you for your understanding, and thank you for testing the branch with us!~~

Morantron:
I tried to make use of this feature in [tmux-fingers](https://github.com/morantron/tmux-fingers), but I'd need borderless floating panes to make it work.

My current approach for rendering highlighted hints on top of the current pane is:

- capture the target pane
- create an empty pane (fingers pane) and print to it
- swap target pane and fingers pane

[...]

PR #5142, opened 2026-06-01
fix(floating_panes_staging): Added plumbing for 'placeholder' layout cells to fix `NULL` root layouts.

daneofmanythings:
This is a fix to the [first known issue](https://github.com/tmux/tmux/issues/4800#issuecomment-4543971922) in #5135. The implementation is also forward looking, laying the tracks for future layout operations like hiding panes and toggling floating state.

Some state has been moved away from `struct window_pane` and into `struct layout_cell`s' new `flags` field when that state concerns aspects of the layout.

Layout cells can now have the state `LAYOUT_CELL_CONCEALED` (previously `PANE_HIDDEN`) which is used for the "placeholder" cells. Whenever a pane is created as a floater, an additional cell is created and inserted into the root layout as the placeholder for the float. This ensures that there will always be a root layout when a window has a pane. I am using the "conceal/reveal" semantics for layout cells over "hide/show" because a layout cell can be concealed, but not be related to a hidden pane. This divide in semantics is meant to keep related but distinct states clear.

With the movement of state to layout cells, some helper functions have been added to `window.c` which retrieve the state of a window pane; mainly convenience functions to avoid having to write out long dereference chains. This abstraction of pane state will give room for future changes in layout implementation.

I believe I have torn out everything related to hiding panes and commands. Let me know what y'all think!

@mgrant0 @nicm

**EDIT1**: I have a feeling that this is too untested of a change to add to staging. Would a PR against `floating_panes` be more appropriate?

### DIFF
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -159,7 +159,7 @@ cmd_resize_pane_mouse_update(__unused struct cmd *self, struct cmdq_item *item)
 	if (wp == NULL || c == NULL || c->session != s)
 		return (CMD_RETURN_NORMAL);
 
-	if (~wp->flags & PANE_FLOATING) {
+	if (window_pane_is_floating(wp)) {
 		c->tty.mouse_drag_update = cmd_resize_pane_mouse_update_tiled;
 		cmd_resize_pane_mouse_update_tiled(c, &event->m);
 		return (CMD_RETURN_NORMAL);
--- a/cmd-select-pane.c
+++ b/cmd-select-pane.c
@@ -160,7 +160,7 @@ cmd_select_pane_exec(struct cmd *self, struct cmdq_item *item)
 			server_redraw_window_borders(markedwp->window);
 			server_status_window(markedwp->window);
 		}
-		if (wp->flags & PANE_FLOATING) {
+		if (window_pane_is_floating(wp)) {
 			window_redraw_active_switch(w, wp);
 			window_set_active_pane(w, wp, 1);
 		}
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -69,63 +69,6 @@ const struct cmd_entry cmd_split_window_entry = {
 	.exec = cmd_split_window_exec
 };
 
-static struct layout_cell *
-cmd_split_window_get_floating_cell(struct cmdq_item *item, struct args *args,
-    struct window *w, struct window_pane *wp)
-{
-	struct layout_cell	*lc = NULL;
-	char			*cause = NULL;
-	u_int			 x, y, sx, sy;
-
-	if (window_pane_floating_geometry(w, wp, &x, &y, &sx, &sy, item, args,
-	    &cause) != 0) {
-		cmdq_error(item, "invalid floating pane geometry %s", cause);
-		free(cause);
-		return (NULL);
-	}
-
-	/*
-	 * Floating panes sit in layout cells which are not in the layout_root
-	 * tree so we call it with parent == NULL.
-	 */
-	lc = layout_create_cell(NULL);
-	lc->xoff = x;
-	lc->yoff = y;
-	lc->sx = sx;
-	lc->sy = sy;
-
-	return (lc);
-}
-
-static struct layout_cell *
-cmd_split_window_get_tiled_cell(struct cmdq_item *item, struct args *args,
-    struct window *w, struct window_pane *wp, int flags)
-{
-	enum layout_type	 type;
-	struct layout_cell	*lc = NULL;
-	char			*cause = NULL;
-	int			 size;
-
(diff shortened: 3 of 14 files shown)


nicm:
OK a few things:

- TBH I think it is confusing to have different terms `concealed`/`revealed` and `hidden`/`shown` especially since they are clearly the same thing, since there is no other "pane is hidden" flag? But...

- ...can we do without the  concealed/hidden stuff for now? I don't want to have to worry about hidden panes yet.

- I don't think there is any need for `ZOOMED` to be moved. It looks like the only places you use it you have the pane?

- I'd prefer if the other flags weren't moved either, but I'm not certain you always have the pane when you need them. But if you get rid of `CONCEALED`, then all you have is `FLOATING`?

- I think `window_pane_has_lcflag` is unnecessary and should probably be disposed of entirely and just check the flag directly. Or it should be a function in the `layout_*` namespace eg `layout_cell_has_flag` and take a `struct layout_cell`.

- Do you need to move `cmd_split_window_get_floating_cell` and `cmd_split_window_get_tiled_cell`? I'm not sure that these should be in the layout code. Anyway, unless this has to be done now this seems like unnecessary churn and should be done separately.

nicm:
Maybe I am making too much work by trying to not include concealed cells, but this is a big diff with lots of moving code around.

If you can either remove the `cmd_split_window_get_floating_cell` changes for now or do the first it would make it smaller.

Then we could add `window_pane_is_floating` right now separately which would make the main diff a bit smaller as well.

And could we add the layout flags and remove `LAYOUT_FLOATING` separately too?

Then we would have a much smaller diff which really only added placeholder cells.

daneofmanythings:
1. They are slightly different. I am using the concealed flag to stop panes from being drawn and used in the `layout_fix_*` functions. Placeholders are concealed as well as hidden panes. I can change the semantics though. It's not a big deal.
2. The concealed flag is being used for the placeholders.
3. Hm, yeah `ZOOMED` didn't need to be moved. I do think it should be part of the layout logic though. Doesn't need to be part of this.
4. See 2 above.
5. I can get rid of `window_pane_has_lcflag`. It is just an internal helper to reduce repeated boilerplate. I'll check flags directly.
6. I tried using the `*_geometry` functions, and didn't like the api at all. I wrote them to minimally abstract out the geometry logic, and what I found when working on this is that what i actually wanted was to get back the layout cell (or possibly update parameters as for the floating version). I can revert it if you want. I do want the new functions to be the api though. They slot in very well to the other commands I was working on.

I also would like to know if you would like floating panes to be proportionally resized when the layout is resized. I didn't implement that, but it wouldn't be hard to add because placeholders link to their panes. It occurred to me after that this is probably the correct behavior.

daneofmanythings:
Just saw your second message come through. I'd be happy to chop this up further. How would you like me to do it? Shall I change this PR to a sequence of commits, or open a sequence of PRs? Would you still like all of the against staging?

nicm:
OK keeping concealed for layout vs hidden for panes is fine.

It probably doesn't matter where the zoomed flag lives, it is really a property of the window. But let's leave it alone for now. If you want to clean it up so it belongs to the layouts later, sure.

I think moving the code around in (6) is fine but can you do it separately (before or after is fine)?

I don't mind if you make separate PRs or separate commits but if you make separate commits and there are changes then it might be harder to manage?

I will review each change and apply to OpenBSD and to `master` in GitHub, then merge it into the floating panes branches. This means there are smaller changes to review, and if anyone sees a bug it is easier to track down what caused it, and to sort out any conflicts.

So I would do at least these if possible:

1 Add `window_pane_is_floating` and the other accessors.

2 Move `LAYOUT_FLOATING` to a flag on the cell (you can add the other flags in the header at the same time but they won't be used yet).

3 Add the placeholders stuff (this will be the main diff, the previous are just prep to make this easier to review). If we look at it at this point and see other obvious ways to split it up at this point we can do that then...

4 Move `cmd_split_window_get_floating_cell` etc around (you could do this before (3) if you like).

daneofmanythings:
Excellent. I'll start on these tomorrow! It's 1am for me at the moment. I'm thinking ill prepare all of them, and then rebase and submit as they get merged. Our timezone difference can add an annoying delay, so I'll have them prepped.

It's not that important, but `LAYOUT_FLOATING` wasn't being used for floating panes. Michael said it must have been a vestige of an earlier implementation. It was only being used for custom layouts and some printing, but layout cells for floating panes are `LAYOUT_WINDOWPANE`. The custom layout support for floats can be added later.

nicm:
OK great, there is no rush. I mostly work with the US so unfortunately it does tend to mean I am most available in the early morning before they are awake...

If `LAYOUT_FLOATING` is not needed and you just want to make a diff to remove it first, that would be good.

Issue #5145, opened 2026-06-01
crash: NULL window_pane deref in screen_write_redraw_line when status line has a wide/utf-8 cell


undont:
current master (74b91d6), macos 26.5 / arm64. tmux segfaults the moment a client
attaches if the status line contains a wide (utf-8) cell. this is not the
copy-mode grid crash from #4962, it's in the status redraw path.

minimal repro, no plugins, no config of mine:

    cat > /tmp/min.conf <<'EOF'
    set -g status on
    set -g status-interval 1
    set -g status-right '宽 %H:%M:%S '
    EOF
    tmux -L x -f /tmp/min.conf new-session -d
    tmux -L x attach

dies within a second of attaching, once the clock ticks and the line redraws.

backtrace (stock 74b91d6 built -g -O0):

    frame #0  screen_write_redraw_line   screen-write.c:1132   int xoff = wp->xoff
    frame #1  screen_write_cell          screen-write.c:2578
    frame #2  screen_write_putc(ch=' ')  screen-write.c:398
    frame #3  status_redraw              status.c:437
    frame #4  screen_redraw_update       screen-redraw.c:744
    frame #5  screen_redraw_screen       screen-redraw.c:811
    frame #6  server_client_check_redraw server-client.c:2117

wp is NULL, EXC_BAD_ACCESS at 0x30 (offset of xoff in struct window_pane).

i had a poke at the source, might be wrong on the details:

status_redraw looks like it builds its write ctx with
screen_write_start(&ctx, &sl->screen), so ctx->wp would be NULL there (the status
line isn't a pane). from what i can tell each redraw clears the line with spaces
and then draws the formatted content over it (status.c ~437), and if the previous
redraw left a wide char the line is GRID_LINE_EXTENDED, so the space-clear seems
to hit screen_write_overwrite() and set redraw=1 (screen-write.c:2491). that
looks like it then takes the `if (redraw) screen_write_redraw_line(...)` branch
added in ecfe6e4e, where `int xoff = wp->xoff` runs unconditionally -> NULL deref.

that would also fit why it only shows up when the status line has an extended cell
and the content changes between redraws (a static or plain-ascii bar seems to get
skipped), but i'm guessing, you'll know better.

@nicm this is the status-line segfault i mentioned over in #4962, looks unrelated
to that one and not from your test patch (that only touches grid.c/tmux.h).

nicm:
I think this is the same as https://github.com/tmux/tmux/pull/5139. Does this fix it?
[...]

PR #5146, opened 2026-06-01
Fix status bar wide char crash

adithyabsk:
#### Description

Segfault in `screen_write_redraw_line` when the status bar contains a wide character (emoji) and the terminal is resized. The server crashes, killing all active sessions. I've included a regression test.

### Required information

**Version**: mainline commit (e778f7f8)
**OS**: Linux (kernel 5.19.0-45-generic, x86_64)
**Trigger**: Resize terminal with an emoji in `status-right`.

**Backtrace (from core dump)**:

```gdb
#0  screen_write_redraw_line (ctx=..., yy=0) at screen-write.c:1136
        wp = <optimized out>
#1  screen_write_cell (ctx=..., gc=...) at screen-write.c:2578
        wp = 0x0
        now_gc = {data = "🤖", width = 2}
        tmp_gc = {data = " ", width = 1}
        sx = 40
#2  screen_write_putc (ch=32 ' ') at screen-write.c:398
#3  status_redraw (c=...) at status.c:437
        ctx = {wp = 0x0, s = 0x155555518bc0}
        n = 38, width = 40
#4  screen_redraw_update () at screen-redraw.c:744
#5  screen_redraw_screen () at screen-redraw.c:811
#6  server_client_check_redraw () at server-client.c:2117
#7  server_client_loop () at server-client.c:1528
```

`ctx->wp` seems to be unconditionally dereferenced

### DIFF
--- a/regress/status-redraw-wide-char-crash.sh
+++ b/regress/status-redraw-wide-char-crash.sh
@@ -1,0 +1,69 @@
+#!/bin/sh
+
+# Regression test: crash in screen_write_redraw_line when status bar
+# contains a wide character (emoji).
+#
+# When the status line is redrawn, it is first filled with spaces
+# (width 1). If the previous content had a wide character (width 2),
+# writing a space over it triggers screen_write_overwrite() which sets
+# redraw = 1. Then screen_write_redraw_line() dereferences ctx->wp,
+# but wp is NULL for status-bar screens, causing a SIGSEGV.
+#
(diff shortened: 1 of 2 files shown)


adithyabsk:
oh, I see this was raised here: https://github.com/tmux/tmux/issues/5145

and fixed here: https://github.com/tmux/tmux/commit/4d91ff8d8d57105001af91ee384527dd92e6a83d

hence my conflict

PR #5147, opened 2026-06-01
fix(floating_panes_staging): Removed LAYOUT_FLOATING.

daneofmanythings:
LAYOUT_FLOATING was old code that is no longer being used. The custom layout code it was a part of needs to be redone anyway.


nicm:
I've applied this but I also removed `floating_lc` from `layout_construct` since it is not used. Thanks!

nicm:
I'm going to add "Make custom layouts work with floating panes" to the FP todo list in the issue but it is really not urgent for them to work :-).

PR #5149, opened 2026-06-02
refactor(floating_panes_staging): Moved args based layout cell creation logic into layout.c

daneofmanythings:
Also, getting a floating cell can also now update the values of layout cell parameter. This was annoying to do with the old api.


nicm:
I've applied these now, thanks!

Issue #18, opened 2015-06-07
SourceForge mentions around the codebase


alistra:
https://github.com/tmux/tmux/search?p=7&q=Sourceforge&type=Code&utf8=✓ Search for SourceForge still has some mentions in files. Most Notably the `CHANGES` file and copyright notices have peoples' names with @users.sourceforge emails. 

Should this be cleaned up?

nicm:
No these are ok

Issue #16, opened 2015-06-07
tmux start-server without forking


LaraCraft93:
I want to be able to start the tmux server without forking the process, but I found no way on last tmux release.

I found this commit on cvs-archived repository:

https://github.com/ThomasAdam/tmux-cvs-archived/commit/4ae54388081f3802700eae2ebcd876efcb2714f5

why it's not in the main tree?

nicm:
I don't like it :-).

zxcvbn97:
Without forking the process can be useful for debugging purpose. Why not?

nicm:
It doesn't make it appreciably easier and tmux doesn't log anything to stdout.

LaraCraft93:
I need it because I want to start a default tmux session on my system and restart the session if it don't exist anymore. For do this with a forking process I need a infinite loop,  and it is ugly.

It's a functional implementation, not only for debugging.

ntnn:
@LaraCraft93 You can define the socket-path, tmux doesn't start a second server on the same socket.

ThomasAdam:
It was useful once or twice when I was debugging tmux server code for something, but it's generally not a problem now to just either use abort() (in rare cases), or shove some log_debug() calls in, etc., etc.

devurandom:
I would also like to see this feature in order to make tmux interact more nicely with a service supervisor like systemd or s6.

mmonaco:
+1 to this. I think it would be useful for learning and debugging if there was an option to `start-server`to not fork which also implies or could be used in conjunction with an option to log to `stdout` / `stderr`, and also to imply `set-option -s exit-empty off`.

This isn't only about systemd, I'd be interested in this for running manually too. However, for those interested my current user-unit workaround is:

```
[Service]
Type=forking
GuessMainPID=true
ExecStart=/usr/bin/tmux set-option -g exit-empty off \; start-server
ExecStop=-/usr/bin/tmux kill-server
```

but this doesn't solve the logging problem.

I also like to keep the server running in order to keep startup latency to a minimum.

PR #5150, opened 2026-06-02
fix(floating_panes_staging): Floating panes are now always a part of the root layout.

daneofmanythings:
~~A 'placeholder' layout cell is a stubbed `struct layou_cell` which holds the position of a window pane in the root layout. More generally, it will "remember" the previous cell when switching between a tiled or floating state.~~

~~It's first use, this PR, is to anchor floating cells into the root layout so that as long as a window has a pane, there will be a non-null root layout. The placeholder is created and inserted into the root layout as floating panes are spawned.~~

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -104,7 +104,7 @@ cmd_split_window_exec(struct cmd *self, struct cmdq_item *item)
 		flags |= SPAWN_EMPTY;
 
 	if (is_floating)
-		lc = layout_get_floating_cell(item, args, w, wp, lc, &cause);
+		lc = layout_get_floating_cell(item, args, w, wp, &cause);
 	else
 		lc = layout_get_tiled_cell(item, args, w, wp, flags, &cause);
 	if (cause != NULL) {
--- a/layout.c
+++ b/layout.c
@@ -46,8 +46,6 @@ static int	layout_set_size_check(struct window *, struct layout_cell *,
 		    enum layout_type, int);
 static void	layout_resize_child_cells(struct window *,
 		    struct layout_cell *);
-void		layout_redistribute_cells(struct window *, struct layout_cell *,
-		    enum layout_type);
 
 /* Create a new layout cell. */
 struct layout_cell *
@@ -252,6 +250,8 @@ layout_fix_offsets1(struct layout_cell *lc)
 	if (lc->type == LAYOUT_LEFTRIGHT) {
 		xoff = lc->xoff;
 		TAILQ_FOREACH(lcchild, &lc->cells, entry) {
+			if (lcchild->flags & LAYOUT_CELL_FLOATING)
+				continue;
 			lcchild->xoff = xoff;
 			lcchild->yoff = lc->yoff;
 			if (lcchild->type != LAYOUT_WINDOWPANE)
@@ -261,6 +261,8 @@ layout_fix_offsets1(struct layout_cell *lc)
 	} else {
 		yoff = lc->yoff;
 		TAILQ_FOREACH(lcchild, &lc->cells, entry) {
+			if (lcchild->flags & LAYOUT_CELL_FLOATING)
+				continue;
 			lcchild->xoff = lc->xoff;
 			lcchild->yoff = yoff;
 			if (lcchild->type != LAYOUT_WINDOWPANE)
@@ -275,6 +277,10 @@ void
 layout_fix_offsets(struct window *w)
 {
 	struct layout_cell	*lc = w->layout_root;
+
+	/* Root consists of a single floating cell */
+	if (lc->flags & LAYOUT_CELL_FLOATING)
+		return;
 
 	lc->xoff = 0;
 	lc->yoff = 0;
@@ -484,8 +490,11 @@ layout_resize_adjust(struct window *w, struct layout_cell *lc,
 
 	/* Child cell runs in a different direction. */
 	if (lc->type != type) {
-		TAILQ_FOREACH(lcchild, &lc->cells, entry)
+		TAILQ_FOREACH(lcchild, &lc->cells, entry) {
+			if (lcchild->flags & LAYOUT_CELL_FLOATING)
+				continue;
 			layout_resize_adjust(w, lcchild, type, change);
+		}
 		return;
 	}
 
@@ -497,6 +506,8 @@ layout_resize_adjust(struct window *w, struct layout_cell *lc,
 		TAILQ_FOREACH(lcchild, &lc->cells, entry) {
 			if (change == 0)
 				break;
+			if (lcchild->flags & LAYOUT_CELL_FLOATING)
+				continue;
 			if (change > 0) {
 				layout_resize_adjust(w, lcchild, type, 1);
 				change--;
@@ -515,7 +526,7 @@ void
 layout_destroy_cell(struct window *w, struct layout_cell *lc,
     struct layout_cell **lcroot)
 {
-	struct layout_cell     *lcother, *lcparent;
+	struct layout_cell     *lcother = NULL, *lcparent;
 
 	/*
 	 * If no parent, this is either a floating pane or the last
@@ -530,11 +541,13 @@ layout_destroy_cell(struct window *w, struct layout_cell *lc,
 		return;
 	}
 
-	/* Merge the space into the previous or next cell. */
-	if (lc == TAILQ_FIRST(&lcparent->cells))
-		lcother = TAILQ_NEXT(lc, entry);
-	else
-		lcother = TAILQ_PREV(lc, layout_cells, entry);
+	if (~lc->flags & LAYOUT_CELL_FLOATING) {
+		/* Merge the space into the previous or next cell. */
+		if (lc == TAILQ_FIRST(&lcparent->cells))
+			lcother = TAILQ_NEXT(lc, entry);
+		else
+			lcother = TAILQ_PREV(lc, layout_cells, entry);
+	}
 	if (lcother != NULL && lcparent->type == LAYOUT_LEFTRIGHT)
 		layout_resize_adjust(w, lcother, lcparent->type, lc->sx + 1);
 	else if (lcother != NULL)
@@ -603,6 +616,8 @@ layout_resize(struct window *w, u_int sx, u_int sy)
 	 * out proportionately - this should leave the layout fitting the new
 	 * window size.
 	 */
+	if (lc->type == LAYOUT_WINDOWPANE && (lc->flags & LAYOUT_CELL_FLOATING))
+		return;
 	xchange = sx - lc->sx;
 	xlimit = layout_resize_check(w, lc, LAYOUT_LEFTRIGHT);
 	if (xchange < 0 && xchange < -xlimit)
@@ -915,6 +930,8 @@ layout_resize_child_cells(struct window *w, struct layout_cell *lc)
 	count = 0;
 	previous = 0;
 	TAILQ_FOREACH(lcchild, &lc->cells, entry) {
+		if (lcchild->flags & LAYOUT_CELL_FLOATING)
+			continue;
 		count++;
 		if (lc->type == LAYOUT_LEFTRIGHT)
 			previous += lcchild->sx;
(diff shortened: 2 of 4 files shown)


daneofmanythings:
I should mention, we may want floating panes to resize as windows resize. It is possible for a floating pane to "disappear" if the window is resized small enough that the offsets are outside of the bounds, or for it to spill off the side boundaries. I have a patch to fix this, but I'd like to know what the desired behavior is. Right now, I scale the position and size based on the change in window size.

nicm:
I'm not sure, we probably don't want to squash everyone's floating panes if they temporarily resize so it might be best to leave them even if they are outside the window. But then it would be nice to have a key to bring them all back on to the window?

Maybe when hidden panes are done we can just be automatically hidden.

daneofmanythings:
I'm not sure what the most expected behavior is. My passing thought was to have an option which will keep floating panes proportional to the window size, or let them keep their absolute values.

nicm:
I don't think we need an option, just pick a behaviour that seems reasonable. I would either:

- leave them outside the window and provide a way to "gather" them;
- move them all so they are visible (this seems least desirable but does match the behaviour for tiled panes);
- ignore the whole issue for now and when hidden panes are done, they can be automatically hidden and when shown can be moved on to the visible window.

daneofmanythings:
I chose ignore and wait!

daneofmanythings:
After some great feedback from @mgrant0, we have changed and simplified the way things are handled. 'Placeholders' are gone, and as a result the concept of a "concealed" cell is gone too. State is much more easily managed. All layout cells are a part of the root layout. How the cell is drawn depends on its flags, currently only `LAYOUT_CELL_FLOATING`, but `LAYOUT_CELL_HIDDEN` is coming. 

There were a couple other downstream effects which I also believe simplified the code. `layout_get_floating_cell` is much more focused in what it does, and there is an accompanying `layout_floating_pane` which enables an analogous logical flow as `layout_get_tiled_cell` with `layout_split_pane`. Also, easier floating pane creation outside of `layout_get_floating_cell`.

nicm:
Why is `spit-window` operating on the active pane? It probably shouldn't so that, commands should use the target pane.

And then why is this needed?
~~~~
                if (wpp == NULL) /* fallback */
                        wpp = TAILQ_LAST(&w->panes, window_panes);
~~~~
If there is a window where `w->active` is `NULL` but `w->panes` is not empty, it is a bug and we should figure out how that is happening and fix it. If we want to support having no panes at all (I am not convinced we should), then both `w->active` will be `NULL` and `w->panes` will be empty, so this will have no effect.

nicm:
Another few things:

- `layout_get_floating_cell` uses `wp` but it is still marked `__unused`.

- I'm not sure I follow what this is doing. Why does the new `layout_cell` for a floating pane have as its parent the `layout_cell` of whatever pane happened to be active? That seems confusing, it seems like they should always go together at the top level or in some special position?

- Why does `layout_fix_offsets` need to change a neighbouring cell for floating cells? Surely it doesn't need to change anything for them at all, since their offsets will not change and they are ignored in `layout_fix_offsets1`?

daneofmanythings:
Thanks for the feedback! I removed the logic that guarded against creating a floating pane in a pane-less window and cleaned up `split-window`.

- I believe I did remove it, and that is what it see in the diff; regardless, it is corrected now.
- Creating a new floating pane works similarly to how splitting a pane works. The caller provides a layout cell and the new cell is put after it in the tree; The logic is a simplified version of what is in `layout_split_pane`. There is nothing special about a floating pane in the layout root. It has a flag indicating it is floating and that flag is used to exclude it from the tiling/resize/offset calculations. Could you point me to where the code clarity needs to be improved? I'll make it more explicit and add comments.
- This was an error on my part. The condition just needs to return if the root is a single floating cell.

Also found and fixed a couple more bugs.

nicm:
I don't understand what `layout_get_floating_cell` needs the active pane for, you are not splitting the active pane. Shouldn't it take `w->layout_root` from the `struct window *`? It is confusing that it is operating on the cell of some random pane when it is creating an entirely new floating pane which should not be related to any existing pane.

daneofmanythings:
That has been removed. It was logic i hadn't thought all the way through when i refactored.

nicm:
It is still taking a pane - why does it need to know about any other pane? Can't it take just the window or the root cell?

daneofmanythings:
It could. The thought was to give the caller the ability to put the new cell in a specific spot in the tree. Should i remove this?

nicm:
Is there a need to be able to put a floating cell in a specific spot? The position of the cell in the tree doesn't matter for floating cells, right? I think it would be better if they were always at the top level and never inside any other cell with tiled cells (aside from the root cell).

Another thing - are the changes around `layout_destroy_cell` needed? I assume you had broken `layout_cell_get_neighbour` out because you wanted to reuse it, but it is only called from `layout_destroy_cell` now - that code doesn't not apply to floating panes, they are handled earlier, so could the code go back to pretty much how it was?

Thanks!

daneofmanythings:
Hmm, good question. The impetus of this logic comes from wanting to "toggle" between floating and tiling state for a pane. If we keep a pane "anchored" in its "tiled" spot then that operation becomes nearly trivial. Also, consider what we need to do when we add hidden panes. My thought for those was to add a flag that keeps them from being drawn in a similar way to how the floating logic is skipping cells in the calculations, so the scattered if checks are going to be in there anyway assuming you like that implementation for them.

Yes, it is used all over the work for the floating/tiling/hiding logic. the `direction` version will be broken out of static for `resize-pane` when I implement middle split resizing. Shall i revert it and add it in later?

nicm:
Yes please keep this change minimum possible and do the destroy changes later. I want to keep the risk of a regression in layouts as low as practical for the first phase of floating panes. (You can change layouts much as you like in the branch later once we get them turned on by default.)

My concern with having floating cells scattered all through the tree is that you then need to cope with eg parent cells that only have floating cells. Maybe that kind of thing just falls out nicely but it seems unnecessary and confusing unless it is absolutely necessary.

Anyway at the moment it is not needed so would it be better just to always make them in the layout root? If they need to be moved later they can be.

daneofmanythings:
Okay, I'll see what I can do. I now think I understand why a previous implementation had `LAYOUT_FLOATING` as a container type. My gut is telling me that there is going to be an equal amount of conditional logic to keep things tidy with a floating container pinned to the root, but I'll have to see how the implementation shakes out.

I have the logic to handle parents with all undrawn cells. I didn't find it that confusing, but we can wait on it.

Thanks for the feedback!

nicm:
Won't it just be the same? Adding them always to the root now is just forcing a layout that *could* occur even if they could be placed anywhere, so it can be handled the same way. We are just making the create path a little simpler for now and perhaps sidestepping any issues.

daneofmanythings:
If we don't containerize the floating panes into their own cell type, I foresee all the same skipping logic in the resize/offset/drawing logic remaining. Since we have no guarantee about how the layout progresses over time, I'm thinking the least intrusive way to do this is to insert the layout cell before the "first" `LAYOUT_WINDOWPANE` in the tree in the same parent, or make a new `LAYOUT_TOPBOTTOM` for them both. I believe the state of the root container would then have a cells field consisting of every floating cell followed by the start of the tiled layout, be it another parent container or a single `LAYOUT_WINDOWPANE`. I would prefer this over making a new container type, but I'm not certain it will make things less confusing. It might! I need to work out the implementation to see the edge cases. Is this what you're suggesting?

The mental model I have been using for layouts is that each `LAYOUT_WINDOWPANE` leaf corresponds to a pane that has a meaningful position in the layout tree. The flags like `LAYOUT_CELL_FLOATING` and `LAYOUT_CELL_HIDDEN` then change the way in which they are drawn on screen, but they are still part of the root in a way that matters. Not saying this is correct, just explaining my thought process.

nicm:
OK, I don't really mind how they are represented all that much.

The idea of keeping floating cells where they would go if they were tiled seems like it does have advantages, so I'm happy if you want to go that way. But if it is like that, you will have to deal with, for example, a parent cell that contains one tiled and N floating panes. I don't know what is involved in dealing with that. 

But what I'm asking for right now is - could the floating cells just be put all in the same place, for example the root cell. This would be a valid layout with the final design (for example if you had one tiled cell and you then made two floating cells, that's what you would end up with, right?) but it would mean we don't have to worry about them being anywhere else. I think it would be fine to disable having *no* tiled cells for the moment to avoid there being any case where floating cells are not sitting in the top cell with at least one either pane or parent cell.

The idea is to avoid making too many changes that aren't actually need yet.

daneofmanythings:
Yeah I can do that. I wanted to make sure that I understood what you wanted! I'll stuff them all at the top, and adjust/remove from `layout.c` accordingly (also reverting unneeded changes.) Restricting to at least 1 tiled cell solves some annoying edge cases we would hit at this point. I'll incorporate that too.

mgrant0:
Up to now, floating panes have not been in the tree rooted at layout_root.  They had their own layout cell but they were dangling.  Dane and I discussed putting the floating panes into the layout tree and it seemed like the logical place to put a floating pane was next to whatever the active pane was when the floating pane was created.  The advantage here being that if you tile a pane in the future, it had a default place to go.  

Stacking them (or rather appending them) to the root node in the layout tree is fine in my opinion too.  In any case, when a pane is marked as floating, it is skipped when drawing the tiled panes but it has a place to live in the tree and that's what we need at the moment.

daneofmanythings:
Floating panes are now confined to the end of `w->layout_root->cells` field!

**EDIT1**: I should mention that I did not add code to enforce there always being a tiled pane. That can be added later if needed.

nicm:
This looks good to me, but there is a bug. Do this, then once the tiled pane has exited, run `exit` in the floating pane:
~~~~
./tmux -vvLx new 'sleep 0.1' \; new-pane
~~~~
tmux will crash:
~~~~
#0  thrkill () at /tmp/-:2
#1  0xdb203c2d96a06852 in ?? ()
#2  0x0000042de057e11b in _libc_abort () at /usr/src/lib/libc/stdlib/abort.c:51
#3  0x0000042de05a3614 in wrterror (d=Variable "d" is not available.
) at /usr/src/lib/libc/stdlib/malloc.c:378
#4  0x0000042de05a9581 in find_chunknum (d=Variable "d" is not available.
) at include/machine/tcb.h:43
#5  0x0000042de05a4bb0 in ofree (argpool=0x7bac77807a40, p=0x42df6aa60a0, clear=Variable "clear" is not available.
)
    at /usr/src/lib/libc/stdlib/malloc.c:1659
#6  0x0000042de05a4833 in _libc_free (ptr=0x42df6aa60a0) at /usr/src/lib/libc/stdlib/malloc.c:1729
#7  0x0000042b99639833 in layout_free_cell (lc=0x42df6aa60a0) at layout.c:100
#8  0x0000042b996b246f in window_destroy (w=0x42de5ce0c00) at window.c:352
#9  0x0000042b996b1a94 in window_remove_ref (w=0x42de5ce0c00, from=0x42b995a71d2 "notify_callback")
    at window.c:402
#10 0x0000042b996446ad in notify_callback (item=0x42dc1644200, data=0x42de5cf8460) at notify.c:166
#11 0x0000042b995eeea2 in cmdq_fire_callback (item=0x42dc1644200) at cmd-queue.c:726
#12 0x0000042b995ee8a1 in cmdq_next (c=0x0) at cmd-queue.c:785
#13 0x0000042b9966db12 in server_loop () at server.c:266
#14 0x0000042b9965081e in proc_loop (tp=0x42e1471aa00, loopcb=0x42b9966dae0 <server_loop>) at proc.c:210
#15 0x0000042b9966d714 in server_start (client=0x42de5cdaa00, flags=402718720, base=0x42e1472a400,
    lockfd=6, lockfile=0x42de5cd9200 '�' <repeats 200 times>...) at server.c:248
#16 0x0000042b995d8bd2 in client_connect (base=0x42e1472a400, path=0x42de5d0c700 "/tmp/tmux-1000/x",
    flags=402718720) at client.c:165
#17 0x0000042b995d8089 in client_main (base=0x42e1472a400, argc=4, argv=0x7bac77807f48, flags=402718720,
    feat=0) at client.c:285
#18 0x0000042b9967f614 in main (argc=4, argv=0x7bac77807f48) at tmux.c:567
Current language:  auto; currently asm
~~~~
Maybe `w->layout_root` is being freed twice?

ASAN or valgrind might be more specific.

nicm:
If this is not easy to fix and you want to just enforce there always being at least one tiled pane, I am happy with that.

daneofmanythings:
I believe I fixed it. The layout root wasn't being NULL'd out because of old logic I overlooked.

nicm:
Great, I am out today but I'll try it when I get back.

nicm:
Looks good, applied now, thanks!

If you want to keep working on this now in `floating_panes` branch (or make a new branch if you like), that's great.

daneofmanythings:
Certainly. I'm thinking I'll try to atomize the command work I did and implement it in floating panes!

Issue #4381, opened 2025-02-23
Add support for edge case resizing in control mode


brads55:
### Issue description

When you have a layout such as the following:

```
HSplit(
    VSplit(
        HSplit(Pane, Pane),
        HSplit(Pane, Pane)
    ),
    VSplit(
        HSplit(Pane, Pane),
        HSplit(Pane, Pane)
    )
)
```

(i.e, two 2x2 grids contained in a horizontal split)

From what I can gather there is no way to resize the center horizontal split outside of using the mouse, since resize-pane looks like it only resizes the splits in the containing cell and nothing further up the tree.

It would be great if there was a way to do this without using the mouse so that control mode clients can cope with this kind of resize.

(Even if it was just exposing layout_resize_layout, and having a way to do multiple resizes at once before the layout is reconstructed).

nicm:
I'll add it to the todo list but I wouldn't hold your breath TBH. Thanks.

PR #5151, opened 2026-06-03
feat(floating_panes_staging): `resize-pane` can now affect floating panes.

daneofmanythings:
I added some additional logic for `resize-pane` to handle negative numbers as that is a common use case for resizing a floating pane, and floating panes have all four borders that `U/D/L/R` handle. I have chosen 'n' as a prefix to the number argument to signify that it should be negative. I am not thrilled about this solution, but am unsure about what would be better. Negative numbers currently also affect tiled targets, so `-D 10` is the same as `-U n10`. I have a feeling this behavior isn't weird enough to implement restriction to the flags.

### DIFF
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -60,10 +60,10 @@ cmd_resize_pane_exec(struct cmd *self, struct cmdq_item *item)
 	struct window		*w = wl->window;
 	struct layout_cell	*lc = wp->layout_cell;
 	enum layout_type	 type;
-	const char		*errstr;
+	const char		*errstr, *argval;
 	char			*cause;
 	u_int			 adjust, shift = 0;
-	int			 x, y, status;
+	int			 x, y, status, neg = 0;
 	struct grid		*gd = wp->base.grid;
 
 	if (args_has(args, 'T')) {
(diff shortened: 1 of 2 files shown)


nicm:
Why can't it be `-D-10`?

daneofmanythings:
Mmm, yeah okay. I'll convert `U/D/L/R` to take arguments. Otherwise it parses out to `-10` as a flag that is unrecognized.

daneofmanythings:
I have made the adjustment. One tiny sticking point, because `U/D/L/R` now accept arguments to handle `-` as a literal and not a flag marker, doing something like `resize-pane -U` is no longer valid. I had to change 4 key bindings that used that. I'm not sure how many configs that will break. A bonus is that now multiple directions can be specified in one call of the command.

nicm:
You can make them optional arguments with `::` in the getopt string.

daneofmanythings:
Oh cool, didn't know that. Fixed!

daneofmanythings:
while I am adding this, should I also knock out #4381?

nicm:
Up to you, if you like :-).

daneofmanythings:
I'll throw up a PR against master once some more stuff gets merged. The work I did with layouts will let me solve it. This PR will stand on its own for now.

PR #5156, opened 2026-06-04
cmd-list-keys: escape special characters in list-keys output

### LINKED ISSUE #5153: `list-keys` is not escaping special characters
### Issue description

`list-keys` is not escaping special characters like `{`, `}`, `#`, `;`, etc. The output of the command cannot be used against `source-file` command.

For example the output contains
```
bind-key    -T copy-mode    M-{                       send-keys -X previous-paragraph
bind-key    -T copy-mode    M-}                       send-keys -X next-paragraph
```
instead of `M-\{`, `M-\}`.

### Required information

* tmux version (`tmux -V`).
`tmux next-3.7`, latest git master cfeade60557b5b4c0f3bf0cf4cb78fcc787fcac8

* Platform (`uname -sp`).
`Linux unknown`, Arch Linux

* Terminal in use (xterm, rxvt, etc).
foot

* $TERM *inside* tmux (`echo $TERM`).
`tmux-256color`

* $TERM *outside* tmux (`echo $TERM`).
`foot`

[tmux-client-2851252.log](https://github.com/user-attachments/files/28555502/tmux-client-2851252.log)
[tmux-out-2851254.log](https://github.com/user-attachments/files/28555505/tmux-out-2851254.log)
[tmux-server-2851254.log](https://github.com/user-attachments/files/28555549/tmux-server-2851254.log)

pengfeixx:
## Summary
- Wrap key_string_lookup_key() output with args_escape() in cmd_list_keys_format_add_key_binding()
- This escapes special characters ({, }, #, ;, etc.) so list-keys output can be used directly with source-file

## Changes
| File | Change |
|------|--------|
| cmd-list-keys.c | Added args_escape() wrapper around key_string_lookup_key() |

## Root Cause
The list-keys command outputs key bindings without escaping special characters. This means the output cannot be sourced back into tmux, since characters like { and } are interpreted as tmux syntax.

## Test Plan
- Run tmux, set some bindings with special characters
- Run list-keys and verify output can be piped to source-file
- Example: bind-key M-\{ ... should appear as M-\{ in list-keys output

Fixes tmux/tmux#5153

### DIFF
--- a/cmd-list-keys.c
+++ b/cmd-list-keys.c
@@ -155,7 +155,7 @@ cmd_list_keys_format_add_key_binding(struct format_tree *ft,
 	format_add(ft, "key_prefix", "%s", prefix);
 	format_add(ft, "key_table", "%s", bd->tablename);
 
-	format_add(ft, "key_string", "%s", key_string_lookup_key(bd->key, 0));
+	format_add(ft, "key_string", "%s", args_escape(key_string_lookup_key(bd->key, 0)));
 
 	s = cmd_list_print(bd->cmdlist, CMD_LIST_PRINT_ESCAPED|
 	    CMD_LIST_PRINT_NO_GROUPS);


nicm:
Thanks but there is really no point in making a PR from someone else's change right after they propose it.

pengfeixx:
Understood, thank you for letting me know! I'll look for other issues to contribute to next time.

Issue #2346, opened 2020-08-12
[Feature Request] Command to kill session groups


serban:
## Request

Please provide a command to kill all sessions in a session group.

## Background

I use session groups to show different windows of the same session on multiple displays simultaneously.

I create and destroy session groups frequently, but tmux does not provide a convenient way to kill all the sessions in a session group at once. I work around this with a shell function similar to this:

```fish
function kg --argument-names group
  tmux list-sessions -F '#{session_group} #{session_name}' \
      | grep "^$group " | cut --delimiter=' ' --fields=2 \
      | while read --line session
    tmux kill-session -t =$session
  end
end
```

It would be very nice to be able to do this from within tmux. After all, tmux provides `kill-server`, `kill-session`, `kill-window`, and `kill-pane`. `kill-group` would be a natural addition, since session groups are a native tmux abstraction.

With `kill-group`, users could destroy a session group in just a couple of keystrokes using a binding like this:

```
bind-key K confirm-before kill-group
```

## Thanks

I use tmux all day, every day. This is excellent software. Thanks!

nicm:
I think this should be a flag on kill-session.

nicm:
You could just put this in a shell script and run it with `confirm-before` and `run-shell`.

serban:
Thanks for the suggestion. I'm happy to use the `run-shell` workaround -- it works great. Still, this would be a useful command to have as a means of manipulating a native tmux abstraction.

If you agree that such a feature would be a good idea, I could send you a patch for it.

nicm:
Sure if you send a patch I will look at it, I suggest adding a flag to `kill-session` to do this.

Although I'm not a big fan of grouped sessions and would rather like to get rid of them, as soon as we have an alternative way to do independent current windows (probably an extension of the `active-pane` flag). But that might not be for a while so by all means go for it :-).

nicm:
Added to todo list.

PR #5157, opened 2026-06-04
Add -g flag to kill-session to kill all sessions in a group

jiku0730:
This implements the request in #2346 (also on the wiki todo list): a `-g` flag for `kill-session` to kill all sessions in the target session's group.

If the target session is not in a group, only that session is killed as before. The man page is updated.


nicm:

PR #5158, opened 2026-06-04
Add animation framework with window-switch slide and pane layout animations

jonaburg:
Adds a per-client animation system layered on tmux's existing overlay API. A frame timer paces redraws; each tick advances the animation state and calls server_redraw_client, which invokes a draw callback that paints the interpolated frame on top of the normal redraw output.

Implemented animation kinds:

- Window-switch slide: triggered from session_set_current. Composes two windows' pane contents at a horizontal translation that smoothdamps to zero. Active-window status highlight slides to its new tab position by copying tgt-styled cells from the status grid. In-flight retargeting is supported via a new overlay_key return code 2 (passthrough).

- Pane layout: resize/split/kill all lerp pane bounds. Dying panes are rendered from a screen snapshot taken before tmux frees the grid. Hooks live in cmd-resize-pane, cmd-split-window, cmd-kill-pane, and server_destroy_pane (for Ctrl-D / process exits).

New session options:
  animation-enable          (default on)
  animation-window-switch   slide | off (default slide)
  animation-pane-layout     (default on)
  animation-duration        ms  (default 120)
  animation-pane-duration   ms  (default 80)
  animation-easing          smoothdamp | linear | ease-in-out
  animation-tau             ms  (default 40)
  animation-frame-interval  ms  (default 8)
  animation-status-highlight (default on)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -84,6 +84,7 @@ endif
 # List of sources.
 dist_tmux_SOURCES = \
 	alerts.c \
+	animation.c \
 	arguments.c \
 	attributes.c \
 	cfg.c \
--- a/animation.c
+++ b/animation.c
@@ -1,0 +1,1099 @@
+/* $OpenBSD$ */
+
+/*
+ * Animated transitions for tmux. Frames are paced by an animation_timer
+ * libevent; each tick advances state and forces a client redraw, which
+ * invokes overlay_draw to paint the frame.
+ */
+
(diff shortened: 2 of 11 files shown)


jonaburg:
closing due to monoliths not conforming to PR guidelines. if you have any interest I'd be happy to work on it and split it up appropriately, but as this is more of an aesthetic choice probably better to keep as a fork or set of patches anyway

PR #5162, opened 2026-06-05
Add option to pass keypresses through while status message is displayed

Steampunkery:
I like to use status messages as a notification system. I find it annoying that status messages always eat a keypress or totally ignore keypresses with no option to simply pass keypresses through to the active pane as if the status message wasn't even there. I implemented this functionality in `display-message` with the option `-P`, which implies `-NC`. I updated the man page as well.

I read the style guide and tried to stick to it. Please let me know if there is anything that I need to change, add, or remove.

### DIFF
--- a/alerts.c
+++ b/alerts.c
@@ -315,10 +315,10 @@ alerts_set_message(struct winlink *wl, const char *type, const char *option)
 		if (visual == VISUAL_OFF)
 			continue;
 		if (c->session->curw == wl) {
-			status_message_set(c, -1, 1, 0, 0,
+			status_message_set(c, -1, 1, 0, 0, 0,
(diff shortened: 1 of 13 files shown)


nicm:
Looks OK, thanks, but you spelt `passthu_keys` wrong. I don't really like the abbrevation `passthru` anyway, so maybe just change it to `pass`?

Steampunkery:
Hi, sorry for the delay. I pushed a commit to my fork that fixes what you asked for, and I rebased onto master. I don't think I have permissions to reopen this PR though.

nicm:
It won't let me reopen it either since you force pushed, can you make a new one please.

Issue #5017, opened 2026-04-20
do not freeze panes when scrolling in copy-mode


barrettruth:
### Issue description

Scroll in a pane (in copy mode) with stuff going on

the panes freeze (intentionally so)

### Required information

* tmux version (`tmux -V`). 3.6a
* Platform (`uname -sp`). nixos
* Terminal in use (xterm, rxvt, etc). ghostty
* $TERM *inside* tmux (`echo $TERM`). tmux-256color
* $TERM *outside* tmux (`echo $TERM`). ghostty-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`). N/A

I'm open to work on this especially given the pr i submitted related to pane freezing with display-command. [in the past]().

Could be a cool extension to scrolling.

I am most interested if:

1. you see any obvious reasons this is hard (i.e. makes it different than just not freezing)
2. *How* to specify this config option.

Happy (and would prefer) to work on this feature.

nicm:
You mean this one:

- A toggle in copy mode to automatically refresh, either when new output is added - with some sort of time limit, say at most once a second - or if that is tricky, just once a second.

We added an `r` key (`send-keys -X refresh-from-pane`) to refresh from the pane.

I think here the idea was just when the pane updates we automatically refresh, but not too often (a fixed limit say once a second would be fine I'd say). So it would need to have a way to know if the pane has changed and a timer to check and do a refresh if it has, plus obviously it would need to be as not-annoying as possible so preserve scroll position, selection etc.

I would probably add a copy mode command to turn this off and on so it would default to on or off (I don't really mind which) and you would do `send -X toggle-refresh` or `refresh-on` or `refresh-off` to turn on or off if you wanted it to start or stop.

barrettruth:
ah. i see this is a quite an undertaking, at least my desired feature of a near 60-hz refresh.

i will close this out of selfishness (a simple one-second-off refresh to me is not my use case)

thanks for the response - i may chef something up in the next few days 🤔

PR #5164, opened 2026-06-06
Use correct X position for visible range checks in screen_write_fast_copy.

barrettruth:
## context

found while testing a copy-mode live-refresh experiment for #5017.

reproduces on current `master` without any refresh changes; does not reproduce on 3.6a.

## problem

1. with a horizontally split window, put the right pane into copy-mode with line
2. move the cursor
3. the position indicator in the top-right of the right pane is:

```text
[
```

not:

```text
[X/Y]
```

## solution

this only happens when the pane has a non-zero `xoff`. the same pane looks fine
when it is the only pane/full-width.

use `xoff + s->cx` for the visible range lookup/checks, matching the coordinate
space used by the rest of the screen-write visibility paths.

## repro

the following script makes a new window in the current session to reproduce the issue:

```sh
left='i=0; while :; do printf "left %06d abcdefghijklmnopqrstuvwxyz abcdefghijklmnopqrstuvwxyz abcdefghijklmnopqrstuvwxyz abcdefghijklmnopqrstuvwxyz\n" "$i"; i=$((i + 1)); sleep 0.02; done'
right='i=0; while :; do printf "right %06d ABCDEFGHIJKLMNOPQRSTUVWXYZ ABCDEFGHIJKLMNOPQRSTUVWXYZ ABCDEFGHIJKLMNOPQRSTUVWXYZ ABCDEFGHIJKLMNOPQRSTUVWXYZ\n" "$i"; i=$((i + 1)); sleep 0.02; done'

set -- $(tmux new-window -d -P -F '#{window_id} #{pane_id}' -n repro-5164 "$left")
win=$1
right_pane=$(tmux split-window -d -P -F '#{pane_id}' -h -t "$win" "$right")

tmux set-option -w -t "$win" mode-keys vi
tmux set-option -w -t "$win" copy-mode-line-numbers off
tmux select-layout -t "$win" even-horizontal
tmux resize-pane -t "$right_pane" -x 62

sleep 2

tmux copy-mode -t "$right_pane"
tmux send-keys -t "$right_pane" -X history-bottom
tmux select-pane -t "$right_pane"
tmux select-window -t "$win"

tmux display-message 'press k in the right pane and watch the top-right copy-mode indicator'
```

### DIFF
--- a/screen-write.c
+++ b/screen-write.c
@@ -624,21 +624,23 @@ screen_write_fast_copy(struct screen_write_ctx *ctx, struct screen *src,
 	struct grid_line	*gl, *sgl;
 	struct grid_cell	 gc;
 	u_int			 xx, yy, cx = s->cx, cy = s->cy;
-	int			 yoff = 0;
+	int			 xoff = 0, yoff = 0;
 	struct visible_ranges	*r;
 
 	if (nx == 0 || ny == 0)
 		return;
-	if (wp != NULL)
+	if (wp != NULL) {
+		xoff = wp->xoff;
 		yoff = wp->yoff;
+	}
 
 	for (yy = py; yy < py + ny; yy++) {
 		if (yy >= gd->hsize + gd->sy)
 			break;
 		s->cx = cx;
 		screen_write_initctx(ctx, &ttyctx, 0, 0);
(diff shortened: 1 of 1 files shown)


nicm:

Issue #5160, opened 2026-06-05
do not freeze panes when scrolling in copy-mode (#5017, revisited)


barrettruth:
Context: #5107 (closed because of inactivity).

Hi Nicholas, I know you've got a lot on your plate with the release of 3.6b - great stuff!

I just revisited this issue. You proposed a `send -X toggle-{refresh-{on,off}}`. I propose a window-local `copy-mode-refresh-interval` in milliseconds (0 disables) which implicitly allows for both dynamic refresh rates *and* the ability to toggle on/off (given it'd require a line of scripting).

I am implicitly biased to this config shape rather than an unconfigurable rate because I find it strange to have a 0.5s or 1s refresh rate, _even_ if I set the option. I'm personally more likely, in 2026, to conflate this with the terminal glitching/being slow rather than an intentional feature. The proposed shape is very much "all or nothing."

This comes to the meat of the issue. If we choose this config shape, people will crank it up (likely to 60hz) to get a truly "live" copy-mode pane. The core question is:

> Do we care about supporting this? If so, we'll need to optimize the copy-mode pane rendering. This comes at the cost of complexity.

As you know, right now the copy-mode backing grid is the entire history (i.e. viewport + scrollback). This is already quite inefficient... ***but it is simple and doesn't matter given the existing use case***.

Cutting to the chase: I've already explored the before/after changes and written a port that moves refreshing from `O(total history + visible pane)` to `O(new output + visible pane)` (i.e. independent of history size). Big discontinuities like resize/reflow/clear-history still fall back to full rebuild, but normal scrolling output no longer scales with total scrollback. The diff is +430/-20, requiring the addition of logic and structs to track viewable lines.

I don't want to dive into anything too deep but get your take on this, specifically:

- How has your opinion on the config option change?
- Do you think a "live" scrollback buffer is a valuable sell? Of course, nearly everyone I've talked to is extremely excited at the idea of this feature - but I know this has no bearing on whether it should actually exist.

Thanks for your time!

nicm:
I don't think we need the ability to configure the refresh period. An off/on toggle is enough. I don't mind if it then live updates or does it periodically, whichever you prefer.

But I think it does need to have a way to turn it off and copy from a snapshot. Starting to copy and then having the thing you are copying change or scroll away would be frustrating to say the least.

barrettruth:
Ah. That makes a lot of sense. Two more comments:

1. We can disable re-rendering of the pane on entering a selection...
[...]

PR #5165, opened 2026-06-06
Automatically refresh copy mode from the pane

barrettruth:
This makes copy mode track the live pane while you are in it. It combines the auto-refresh commands here with the incremental backing-grid sync from #5178 as suggested.

## changes

- New pane output is pulled into copy mode on a 50ms timer.
- At the bottom the view follows output; scrolled up, the scroll position is held.
- Refresh pauses while a selection is in progress, so text is copied from a stable snapshot.
- A view of another pane (`copy-mode -s`) is never auto-refreshed.
- OFF! by default. Controlled from copy mode with `refresh-on`, `refresh-off` and `refresh-toggle`; `refresh-toggle` takes the `r` key from `refresh-from-pane`, which is removed (redundant once copy mode is live).

## how it works

Each refresh reconciles the backing grid with the pane in place instead of recloning the whole scrollback. It copies only the lines that scrolled in or were collected since the last refresh, driven by the grid scroll counters (`scroll_added`, `scroll_collected`, `scroll_generation`), and falls back to a full clone when the size or scroll generation does not line up.

The incremental result was compared against a fresh clone with a differential oracle across append / drop / noop / large-change / repeated cases: byte-identical, 12/12, and 12/12 again under ASan with no reports.

### refresh interval 50ms justification

Per-tick total at the 50ms interval:

| terminal | log tail p99 | duty | firehose p99 | duty |
|---|---|---|---|---|
| 80x24  | 2.0 ms | 4%  | 4.3 ms | 9%  |
| 120x40 | 4.7 ms | 9%  | 7.2 ms | 14% |
| 200x50 | 7.8 ms | 16% | 9.7 ms | 19% |

(duty = per-tick cost / 50ms.)

we can see two things:

1. new solution is (desirably so) *flat* in history depth
2. Even under the firehose, refresh stays under ~20% of wall time

50ms also matches the existing `WINDOW_COPY_DRAG_REPEAT_TIME` (appeal to precedence, lol)

## Open questions

1. On by default? **I enabled it.**
2. Removing `refresh-from-pane`? I thought it redundant once refresh is live. **I removed it and made `r` toggle it instead.**
3. Ability to disable/enable option? With it on by default and no option, the only opt-out is `refresh-off` or rebinding `r`. Didn't seem like you cared about it so far
4. The refresh rate of 50ms, but I found after performance analysis this was just fine.

***This is very open to the test and use of others; I'm currently daily-driving with no problems but would prefer to not break all of copy-mode in the next release :sweat_smile: ***


nicm:
OK is this instead of your changes to update live (ie not at intervals) or just until that is ready?

I did wonder if maybe we could have it on by default since starting a selection will pause, but one of the things copy mode is useful for is to pause output, so I do think we need to make sure that is still possible.

barrettruth:
Honestly, this is best communicated with a video. Here's the play-by-play of the attached:


https://github.com/user-attachments/assets/82e1e9c8-3640-4bd9-8c3f-9c316048bde0

There's a continuous stream of logs in the buffer.

1. I enter copy mode (it's disabled by default). Everything freezes. Expected.
2. I press a keybinding (_not included in this pr, ofc_) and enable live-refresh. The cursor remains at the bottom, _live-tailing the output_. Every one second, you can see the buffer contents update. The copy-mode annotation in the top-right reflects this.
3. I press k to go up. You can see the bottom logs keep getting outputted, but the screen viewport stays fixed. Note that when I press k, the cursor was initially on a new, empty line. Then, that line is naturally populated by the logs since it remains in the viewport. In other words, the viewport is the live, _actual_ contents of that buffer in the overall view (i.e. viewport + history).
4. I exit copy mode.

Lastly, to be extra-explicit:

> If you are in live-refresh copy-mode and your cursor is on the bottom, it will live-tail. If you press V to enter copy-mode, the screen always _automatically freezes_. Or, if you press k, the screen always freezes. Selections are unimpacted in this way. preserving this utility of copy-mode.

So, I hope this informs your perception. This feature will probably evolve if merged as performance is improved. Questions of:

1. enable by default
2. a copy-mode keybinding to toggle refresh view
3. whether to persist the live-refresh state across different copy-mode entrances (e.g. enter copy-mode, turn on live refresh, exit, re-enter -> copy-mode currently is off (should it stay on?))

are still up for you to decide in this pr!

nicm:
No, I meant: this code refreshes every second but my impression was you didn't like that and wanted it to refresh as soon as there is a change?

I am fine with refreshing every second if that's what you want but if you prefer it to refresh immediately then let's do that (assuming that doesn't mean just setting the timer to one millisecond, we can't do that).

barrettruth:
Yes - I'm not a fan of the 1 second refresh long-term. I want it higher.

However, I'm uncomfortable given the current inefficient performance of copy-mode viewing (see [this](https://github.com/tmux/tmux/issues/5160#issue-4598498874)) to turn it up any higher than one second first.

Why not do it now? The proper performance patch I've made is 3-400 lines and enough complexity on its own. I think it'd be best shipped separately.

My thought was:

1. ship live refresh
2. ship performance improvements
3. turn up live refresh interval

I'm happy to ship the performance improvements first.

nicm:
I don't think we should add a feature and then change how it works on everyone a few weeks later.

So I would do the performance changes first.

barrettruth:
The pr description has been updated according to the latest changes.

nicm:
OK this looks good but it should be off until you press `r`, not on by default.

barrettruth:
done, and re-phrased the documentation accordingly. thanks for the reviews and let me know if you want any revisions to the phrasing.

nicm:
Looks good. I changed the text in the man page slightly. Applied to OpenBSD now, will be in GitHub later. Thanks!

barrettruth:
hi @nicm, have there been any issues with this feature? :thinking: 

while i spent a long time designing it, i'm surprised that i haven't seen anything specifically horrible go down.

feel free to @ me as always, and thanks for maintaining the project.

nicm:
I have not heard of any issues but probably not many people use this feature or don't use it often, I expect most people just copy quickly and are then done, the don't need to worry about refresh.

grothesque:
Thanks, I find this new feature very useful!  I just compiled tmux from git to try it out.

I enable automatic refresh whenever copy mode is entered with
```tmux
set-hook -g after-copy-mode 'send-keys -X refresh-on'
```
Is this the intended way to make automatic refresh the default? Would it make sense to provide a simpler way, perhaps a global option?

I agree that automatic refresh should remain disabled by default, but I am not sure that a per-copy-mode toggle bound to `r` is the most useful interface. The new feature feels more like a user preference: I expect many users will generally want it either always enabled or always disabled.

The new feature works for me as expected, but when using it with the [Pi coding agent](https://pi.dev/), I see heavy flickering. Pi itself does not flicker inside tmux outside of copy-mode.  The following is an analysis of the problem by an LLM agent.  The agent had access to tmux and Pi sources.

----------------

Pi is a differential-rendering TUI. While streaming, it can render as often as every 16 ms. Each render is wrapped in DEC synchronized output:

```text
CSI ? 2026 h
... cursor movement, line clearing, and updated content ...
CSI ? 2026 l
```

Normally tmux respects this and does not expose intermediate states.

The new copy-mode refresh path appears capable of interrupting that synchronization:

1. `input_parse_buffer()` sets `PANE_UNSEENCHANGES` whenever input arrives while the pane is in a mode.
2. `window_copy_refresh_timer()` runs every 50 ms. When it sees unseen changes, it refreshes the backing buffer, redraws the entire copy-mode screen, sets `PANE_REDRAW`, and clears `PANE_UNSEENCHANGES`.
3. The refresh timer does not check whether `wp->base.mode` contains `MODE_SYNC`.
4. When the forced pane redraw occurs, `screen-redraw.c` explicitly calls `screen_write_stop_sync(wp)` if synchronized output is active.

Therefore, if Pi’s synchronized update is split across PTY reads and the refresh timer fires between them, tmux may terminate the synchronized update early and redraw an intermediate state – for example, after lines have been cleared but before they have been rewritten.

Pi generates larger and more frequent cursor-oriented updates than a simple append-only command, which may explain why the problem is readily visible with Pi but not with ordinary command output.

Would it make sense for `window_copy_refresh_timer()` to defer refreshing while:

```c
wp->base.mode & MODE_SYNC
```

The `PANE_UNSEENCHANGES` flag could remain set, allowing the next timer invocation after the synchronized update completes to refresh from the completed frame.

nicm:
I don't think we should have an option but I am OK with adding a flag to `copy-mode`, that is how we usually do this. Then people can just change their key bindings (although they can equally well bind `copy-mode \; send -X refresh-on`, or do what you have done).

nicm:
I've added an item to the copy mode menu to turn this on and off (as well as line numbers):

image

Although this menu is getting a bit big...

I am seeing a lot of flickering as well.

barrettruth:
The problem with pi is inevitable and expected. keystrokes are being sent from keyboard -> terminal -> pi -> tmux -> underlying program...

Anyways Nicholas, let me know if there's an "action item" here I can work on. We did decide against providing a "tunable" rate earlier, I believe...

nicm:
Couldn't it defer the refresh until the sync finishes? Wouldn't that fix the flickering? Meaning if the timer fires and a sync is in progress, do not refresh, then you can either just wait for the next timer or have a callback fired into copy mode to do the refresh when the sync is canceled.

grothesque:
> The problem with pi is inevitable and expected. keystrokes are being sent from keyboard -> terminal -> pi -> tmux -> underlying program...

What is the relation between Pi flickering in tmux auto-refresh copy-mode and the keyboard?  I see the flickering without touching the keyboard.  It's enough if Pi is outputting a long model response.

barrettruth:
unfortunately i don't use tmux anymore and secondly i can't actually see any glitching ever :sweat_smile: 

@grothesque unironically a good opportunity to leverage pi to solve this...

nicm:
You stopped using tmux in the four hours since asking if there were any problems with this? :-)

Lots of programs use mode 2026 now so it would probably be possible to recreate this with others, but yes the best bet would probably be to point Claude or Codex at it and ask it to make you a test script.

barrettruth:
> You stopped using tmux in the four hours since asking if there were any problems with this? :-)

not even trolling, yes...

point is I can and may put up a pr with it but if it doesn't sort of "immediately" resolve the issue I have little incentive to work on it, and the filer can take over. I'll start on it now.

nicm:
OK well it is up to you, it you don't and someone comes up with a test case that does not require installing random software then I can look at it, best bet is to open a new issue though.

grothesque:
I attach a Python script that recreates the Pi flickering.  In order to see the flickering, I simply run the script under tmux (without any custom config), enter copy-mode, and press `r`.

To create the script, I captured the raw PTY output of Pi (with the help of tmux), and then let an LLM agent analyze it and write a script that reproduces the possibly relevant aspects of it.  I then iterated with the LLM to simplify the script.  (I wouldn’t have the time to do such work without machine assistance.)

Perhaps Pi’s way of handling output is suboptimal, but it is the way it is, and in a real terminal Pi does not flicker.  I believe that tmux should gracefully handle even such difficult cases.

[flicker.py](https://github.com/user-attachments/files/30010578/flicker.py)

grothesque:
@barrettruth [wrote](https://github.com/tmux/tmux/pull/5392#issuecomment-4971018150):
> it sounds like you can take this over.

So you yourself are not interested in the correctness of a feature that you submitted just one month ago?

barrettruth:
So far, you've been able to:

- Natively reproduce the issue (and thus more expediently verify the correctness of the fix)
- Create an intelligent, concise reproduction
- Offer specific, legitimate code-level fixes to code path AND to my suggested fix

If there's any value I can provide, @ me in a PR.

grothesque:
@nicm I’m not sure whether the basic idea behind this PR is good.  For sure some automatic way to refresh while in copy mode is desirable (that’s what made me discover this PR), but I’m not sure whether the concrete approach that this PR uses (polling every 50 ms) is the way to go.

It’s not just the flickering.  Suppose that (with the current PR) a user is in copy mode with auto-refresh enabled, and a program keeps producing output.  The copy mode view is being refreshed, but once the user starts a selection this stops.  This is probably the reasonable thing to do, but it shows a conceptual conflict between copy mode and auto-refresh.  Without continuous refreshes, this inconsistency would not exist.

Before I discovered this PR, my impression was that the old `r` (`refresh-from-pane`) *almost* did the right thing.  (I can explain what I believe should have been adjusted.)  I wanted to be able to configure tmux such that once I scroll down with the mouse wheel in copy mode (I have `set-option -g mouse on`), instead of quitting copy mode and jumping to the most recent output, the copy mode gets refreshed.

The idea was to trigger a refresh only when new content would possible appear (e.g. when the cursor is moved beyond the current last line, or when `M->` is pressed).  This would preserve the old behavior of copy mode, but at the same time allow to seamlessly see new output.  It would *not* allow following new output continuously from copy mode, but I believe that this is not the purpose of copy mode.  Regular mode can be used instead.

For me, such a simpler auto refresh seems sufficient.  It would come without the 50 ms polling, the flickering, etc.  It wouldn’t even have to change the meaning of the `r` binding.

nicm:
If you want to change it to update on demand rather than periodically that's fine. If it is fundamentally broken as it is it should either be fixed or changed to something else, I don't mind which...

grothesque:
tmux 3.7b can be *configured* (see below) to refresh copy mode on demand in a way that, at least for my needs, works better than this PR.

This configuration-only solution relies on 5ff385f8a9a43647956f467bafbac09120ee60a2. Before that commit, `refresh-from-pane` preserved the copy-mode view’s distance from the bottom. Consequently, output produced since copy mode was entered caused the view and the text under the cursor to jump forward. Since that commit, the view remains anchored to the same position in retained history.

This PR, which has been merged into master, removes `refresh-from-pane` and therefore breaks this configuration-only solution. I consider that a regression: a useful, stateless primitive has been removed and replaced by a more specialized, stateful polling mechanism.

I suggest retaining `refresh-from-pane`. I also suggest considering reverting this PR for the following reasons.

If the goal is to make newly produced output available in copy mode, I believe an on-demand solution similar to the one below is preferable. It preserves copy mode’s stable-snapshot behavior and refreshes only when the user tries to navigate into output that may not yet be present in the snapshot. I’m not claiming that my exact configuration-only solution is perfect; it is merely a proof of concept.

If the goal is to follow a command’s output continuously in real time, I do not see what remaining in copy mode adds. The regular pane view already provides that behavior, without polling or making copy mode’s snapshot change in the background.

The synchronized-output flickering reported above is one concrete failure mode introduced by periodic refresh. Even if that particular problem is fixed separately, the polling design still adds complexity around redraws, selections, and deciding when the snapshot is allowed to change.

One known limitation of the proof of concept is that tmux 3.7b does not clear `pane_unseen_changes` after `refresh-from-pane`. Once set, it remains set for the rest of that copy-mode session, so some later near-bottom refreshes may be redundant. A native on-demand implementation could handle this cleanly.

----------------

Here is a tmux 3.7b configuration snippet that automatically invokes `refresh-from-pane` when downward navigation approaches the end of copy mode’s current snapshot. It also retains `r` for an explicit refresh and refreshes before `M->` jumps to the end.

<details>
<summary>Proof-of-concept tmux configuration snippet</summary>

```tmux
################ Refresh when reaching the bottom in copy mode ################
# With tmux 3.7b, an explicit redraw is needed after refresh-from-pane.
bind-key -T copy-mode r \
  send-keys -X refresh-from-pane \; \
  refresh-client

# Enter copy mode without exit-on-bottom.
bind-key -T root WheelUpPane \
  if-shell -F '#{||:#{alternate_on},#{pane_in_mode},#{mouse_any_flag}}' \
    'send-keys -M' \
    'copy-mode'

# Near the bottom, refresh before scrolling. Exit if scrolling reaches the
# refreshed end; during selection or synchronized output, only scroll.
bind-key -T copy-mode WheelDownPane \
  select-pane \; \
  if-shell -F \
    '#{&&:#{!:#{selection_present}},#{!:#{synchronized_output_flag}},#{e|<=:#{scroll_position},5}}' \
    'send-keys -X refresh-from-pane ; send-keys -N5 -X scroll-down-and-cancel ; refresh-client' \
    'send-keys -N5 -X scroll-down'

# Refresh as above, but never exit, to avoid sending key presses to the running program.
bind-key -T copy-mode Down \
  if-shell -F \
    '#{&&:#{pane_unseen_changes},#{!:#{selection_present}},#{!:#{synchronized_output_flag}},#{==:#{scroll_position},0},#{e|==:#{copy_cursor_y},#{e|-:#{pane_height},1}}}' \
    'send-keys -X refresh-from-pane ; send-keys -X cursor-down ; refresh-client' \
    'send-keys -X cursor-down'

# Bind C-n as well.
bind-key -T copy-mode C-n send-keys Down

# Same for other movement key bindings...
bind-key -T copy-mode M-Down \
  if-shell -F \
    '#{&&:#{pane_unseen_changes},#{!:#{selection_present}},#{!:#{synchronized_output_flag}},#{e|<=:#{scroll_position},#{e|/:#{pane_height},2}}}' \
    'send-keys -X refresh-from-pane ; send-keys -X halfpage-down ; refresh-client' \
    'send-keys -X halfpage-down'

bind-key -T copy-mode C-Down \
  if-shell -F \
    '#{&&:#{pane_unseen_changes},#{!:#{selection_present}},#{!:#{synchronized_output_flag}},#{e|<=:#{scroll_position},1}}' \
    'send-keys -X refresh-from-pane ; send-keys -X scroll-down ; refresh-client' \
    'send-keys -X scroll-down'

bind-key -T copy-mode NPage \
  if-shell -F \
    '#{&&:#{pane_unseen_changes},#{!:#{selection_present}},#{!:#{synchronized_output_flag}},#{e|<=:#{scroll_position},#{e|-:#{pane_height},2}}}' \
    'send-keys -X refresh-from-pane ; send-keys -X page-down ; refresh-client' \
    'send-keys -X page-down'

bind-key -T copy-mode C-v send-keys NPage

bind-key -T copy-mode Space send-keys NPage

# Refresh when jumping to the end.
bind-key -T copy-mode M-> \
  send-keys -X refresh-from-pane \; \
  send-keys -X history-bottom \; \
  refresh-client
```

</details>

nicm:
I don't really mind what it does so long as it is off by default.

Now we have automatic refresh we should probably fix it, and it seems like there must be use cases or @barrettruth would never have bothered adding it in the first place. But I'm OK having a "refresh once" command as well as on/off.

I'll take a look at your test case at some point.

barrettruth:
The point of this PR is that anywhere in the scrollback history, things may be updated at any time. This PR enables your view of the scrollback to actually always represent the most up-to-date version of what those lines actually ought to show. Not "near the bottom of the pane," not when you press a key - _always_.

I apologize for the back and forth and lack of robustness in this feature. Please leave me out of the conversation henceforth.

> The synchronized-output flickering reported above is one concrete failure mode introduced by periodic refresh. Even if that particular problem is fixed separately, the polling design still adds complexity around redraws, selections, and deciding when the snapshot is allowed to change.

This paragraph is entirely AI-generated.

grothesque:
@nicm wrote:

> I don't really mind what it does so long as it is off by default.

My point is that the previously existing functionality, `refresh-from-pane`, is useful (IMHO more useful than the replacement) and should be preserved, including the key binding.  In my eyes, the new `r` is an inferior replacement for it.

> Now we have automatic refresh we should probably fix it, and it seems like there must be use cases or @barrettruth would never have bothered adding it in the first place. But I'm OK having a "refresh once" command as well as on/off.

While I was [initially enthusiastic](https://github.com/tmux/tmux/pull/5165#issuecomment-4957489105) about this PR, I now have serious doubts whether bolting auto refresh onto tmux’ existing copy mode is technically the right choice.  You can judge this much better than me.

I also do not see the need for it.  As @barrettruth says, auto refresh is relevant for programs that continuously update text anywhere in the scrollback history.  (And this without user interaction, since in copy mode user input goes to tmux itself.)  Personally, I am not aware of any such programs, and if they exist, they are niche.

grothesque:
@barrettruth wrote:

> This paragraph is entirely AI-generated.

The language of this particular paragraph sounds indeed robotic, but the technical point is entirely my own: bolting-on auto refresh on top of tmux copy mode is difficult to get right, and I am not sure whether this complexity (in terms of user interface and implementation) is justified.

I hand-wrote a reply, but then asked an agent to review it in light of a long discussion about the technical details.  Since it was almost midnight, I did not bother to reformulate.

Overall, I spent at least a full working day of my human time on this PR.  The result so far are the `flicker.py` script and the proof-of-concept on-demand refresh configuration.

nicm:
OK, I don't use this feature myself and @barrettruth no longer uses it, so the question is really what it should look like to bring most value to people who will use it.

He was not the first person to request automatic refresh, just the first to implement it, so there is an audience out there for it, and we should probably keep it.

But I think we should fix the issues with mode 2026 if possible and add a "refresh one time" key to map back to `refresh-from-pane`.

PR #5168, opened 2026-06-06
fix(master): Possible fix to mouse events not being correctly identified for pane border buttons.

daneofmanythings:
There was a regression with the mouse event detection for pane borders. The change in `server-client.c` i believe should happen regardless. 

I am less sure about the change in `window.c`. I believe that the bottom border of a pane should be identified as the top border of the lower pane for pane border buttons to be correctly detected. I am unsure if this change will have unintended side effects. Is there a better way to approach this?

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -695,12 +695,8 @@ server_client_check_mouse_in_pane(struct window_pane *wp, int px, int py,
 				bdr_bottom = fwp->yoff + fwp->sy;
 				if (py == bdr_bottom)
 					break;
-				if (window_pane_is_floating(wp)) {
-					/* Floating pane, check top border. */
-					bdr_top = fwp->yoff - 1;
-					if (py == bdr_top)
-						break;
-				}
+				if (py == bdr_top)
+					break;
 			}
 		}
 		if (fwp != NULL)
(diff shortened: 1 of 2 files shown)


nicm:
How do you reproduce this?

daneofmanythings:
Try to click the default border buttons (zoom or close) on a tiled pane. Nothing happens. It will work on a floating pane, but if the pane is zoomed, the button press to close or unzoom it will not work.

The change in `server-client.c` allows the top border to be detected on mouse events in tiled panes.

This is not enough though, when there is a pane "underneath" a pane spacially in the layout, the top border of the mouse events target pane is being detected as the bottom border of the pane "above" it from the call of `window_get_active_at` at `server-client.c:911`. All I did was bump up the detection logic by one, though that may have unintended side effects. If this is not the correct solution to this, have you got a direction I should look to do it properly?

nicm:
Yeah, this works for `pane-border-status top` but it break `pane-border-status off` - you can't drag the border anymore.

daneofmanythings:
You are right. I will need to look into how to do this better. If you want a quick triage to get the behavior mostly there, apply the change in `server-client.c`.

nicm:
Take a look at this: [pbs.diff.txt](https://github.com/user-attachments/files/28727651/pbs.diff.txt)

It makes sense to me and seems to fix it.

daneofmanythings:
Makes sense to me too. I have modified the fix!

nicm:
OK I have applied this now, thanks!

PR #5170, opened 2026-06-07
mode-tree: guard mode_tree_key against an empty tree

thorr18:
mode_tree_key() dereferences mtd->line_list[mtd->current] without checking
line_size, so a key, wheel, or right-click delivered while the tree is empty
(line_list == NULL, line_size == 0) crashes the server.

line_size == 0 only happens when the underlying collection is genuinely empty:
mode_tree_build() rebuilds without the filter when a filter matches nothing
(no_matches), so an empty tree is never a filtered view to recover, and
mode_tree_get_current() returns NULL. So just exit the mode. *key is cleared
first because the callers run their key switch before checking the return value
and some cases dereference the (now NULL) current item.

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -1224,6 +1224,11 @@ mode_tree_key(struct mode_tree_data *mtd, struct client *c, key_code *key,
 	u_int			 i, x, y;
 	int			 choice, preview;
 
+	if (mtd->line_size == 0) {
+		*key = KEYC_NONE;
+		return (1);
+	}
+
 	if (KEYC_IS_MOUSE(*key) && m != NULL) {
 		if (cmd_mouse_at(mtd->wp, m, &x, &y, 0) != 0) {
 			*key = KEYC_NONE;


nicm:
I think in `mode_tree_key` you could just return 1 because the mode should not exist without any content? Doesn't it just immediately exit the mode anyway?

thorr18:
Good call - Done; it's a single guard now.

Should just exit: `line_size == 0` only ever means an empty collection. `mode_tree_build()` rebuilds without the filter when a filter matches nothing (the `no_matches` path), so an empty tree is never a filtered view to recover, and `mode_tree_get_current()` returns NULL; there's nothing to act on.

One thing I had to add: `*key = KEYC_NONE` before `return (1)`. The callers run their key `switch` before they check the returned "finished" value (eg `window_tree_key`: the `switch` is at ~1260, the `if (finished)` only at ~1344), and a few cases dereference the current item - `window_tree_pull_item()` does `item->session` unconditionally, reached from `m` / `x` / Enter - which is NULL on an empty tree. So a bare `return 1` leaves `*key` set and the switch still runs it, just relocating the crash into the caller. Clearing the key makes the switch fall through.

I also moved the guard to the top of `mode_tree_key` (before the mouse-block) so it covers the right-click path too - `mode_tree_display_menu()` has the same `line_list[line]` deref - which let me drop the second guard the first version had. So it's now just:

```c
	if (mtd->line_size == 0) {
		*key = KEYC_NONE;
		return (1);
	}
```

nicm:

Issue #5169, opened 2026-06-06
Consider adding XDG session variables to default update-environment


shbernal:
### Issue description

PR #4965 added `WAYLAND_DISPLAY` to the default `update-environment` list. I think there may be a related remaining Linux/Wayland case for XDG session identity variables such as:

- `XDG_SESSION_TYPE`
- `XDG_CURRENT_DESKTOP`
- possibly `XDG_SESSION_DESKTOP`

On Linux Wayland desktops, some programs and wrappers use these variables to decide which graphical/session backend they are running under. When a tmux server/session is used alternately from SSH and from a local graphical session, tmux may refresh display-related variables but leave the XDG session identity stale or missing.

Example from my machine:

```text
systemd --user environment:
  DISPLAY=:0
  WAYLAND_DISPLAY=wayland-1
  XDG_CURRENT_DESKTOP=Hyprland
  XDG_SESSION_TYPE=wayland

tmux session environment:
  DISPLAY=:0
  missing WAYLAND_DISPLAY in tmux 3.6b package default
  missing XDG_CURRENT_DESKTOP
  stale/missing XDG_SESSION_TYPE
```

This can leave programs launched inside tmux in a mixed state: they see a graphical display (`DISPLAY=:0`) but do not see the Wayland/XDG session context. In practice this can make graphical helpers or media programs choose the wrong backend or fail. One concrete example was a pinentry wrapper selecting a GUI backend from `DISPLAY`, while the rest of the session environment did not match the local Wayland desktop.

### Current behavior

With a fresh tmux server on my installed tmux:

```sh
tmux -Ltest -f/dev/null new-session -d
tmux -Ltest show-options -gqv update-environment
tmux -Ltest kill-server
```

Output:

```text
DISPLAY
KRB5CCNAME
MSYSTEM
SSH_ASKPASS
SSH_AUTH_SOCK
SSH_AGENT_PID
SSH_CONNECTION
WINDOWID
XAUTHORITY
```

I also checked current master and saw `WAYLAND_DISPLAY` is now present after #4965, but I did not see the XDG session variables above.

### Expected behavior / request

Would it make sense to include one or more XDG session identity variables in the default `update-environment` list on platforms where they are relevant?

For example:

```text
XDG_SESSION_TYPE
XDG_CURRENT_DESKTOP
XDG_SESSION_DESKTOP
```

Or, if these are considered too desktop/Linux-specific for the default, would a note in the manual/FAQ for Wayland users be acceptable?

[...]

PR #5171, opened 2026-06-07
Add XDG session variables to update-environment

shbernal:
Adds XDG_CURRENT_DESKTOP, XDG_SESSION_DESKTOP, and XDG_SESSION_TYPE to the default update-environment list so tmux refreshes Wayland/XDG session identity along with display variables.\n\nCloses #5169.\n\nTested with:\n- make\n- git diff --check\n- throwaway tmux server with controlled XDG values\n- control-mode attach refresh from stale XDG values to fresh values


nicm:

PR #5172, opened 2026-06-07
Update copy mode vi test result

maxonvim:
## Summary

- update the vi copy-mode regression for the new `next-word-end` end-of-line behavior
- add an explicit `next-word` step so later stateful checks keep their intended starting position

## Testing

- `TEST_TMUX=../tmux sh copy-mode-test-vi.sh` against a locally built tmux in nix shell


nicm:
Great stuff, applied, thanks!

PR #5175, opened 2026-06-08
Do not silently change invalid names given to commands.

### LINKED ISSUE #5173: tmux converts dots to underscores
### Issue description

tmux automatically converts dots (.) in session names to underscores (_) when creating a new session. However, this normalization is not applied consistently when attempting to attach to the same session, resulting in an unexpected error.

Actual behavior:
tmux silently converts dots in the session name to underscores when creating the session.
When attempting to attach using the original dotted name, tmux interprets the dots as pane/window selectors rather than part of the session name.
This results in an error (can't find pane: 1.9) instead of attaching to the session.

`tmux new -s application_3.1.9`
`tmux ls` message: `application_3_1_9: 1 windows (created Thu June 4 16:56:11 2026)`
`tmux a -t application_3.1.9` message: `can't find pane: 1.9`

Expected behavior:
tmux should either:
- Accept the dotted session name and internally map it to the normalized underscore version, or
- Reject dotted session names at creation time with a clear error message, or
- Apply consistent normalization rules across both creation and attachment.

### Required information

* tmux version: tmux next-3.7
* Platform: Linux x86_64 (Ubuntu 24.04)
* Terminal in use: xterm
* $TERM *inside* tmux (`echo $TERM`).
* $TERM *outside* tmux (`echo $TERM`).

barrettruth:
closes #5173

use nicholas's solution: reject invalid session, window names instead of rewrite: add `check_name` and use it _everywhere_ a command takes a name (session, window, pane title, buffer).

there's no way to reject escape sequences and automatic renames so those still modify silently,

2 open questions here:

1. buffer names keep `forbid=""` (from 746dd91e), so `{#,:,.}` are still allowed in them and only control chars get rejected. in other words, buffers end up reject-consistent but not character-level consistent with sessions/windows. **lmk if you want to rather forbid `#` in buffers too**
2. i normalized error messages as follows - this alright with you?
```
invalid session name: %s
invalid session group name: %s
invalid window name: %s
invalid pane title: %s
invalid buffer name: %s
```

### DIFF
--- a/cmd-break-pane.c
+++ b/cmd-break-pane.c
@@ -62,6 +62,14 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 	int			 idx = target->idx, before;
 	const char		*template;
 
+	if (args_has(args, 'n')) {
+		if (!check_name(args_get(args, 'n'), ":.")) {
+			cmdq_error(item, "invalid window name: %s",
+			    args_get(args, 'n'));
+			return (CMD_RETURN_ERROR);
+		}
+	}
+
 	before = args_has(args, 'b');
 	if (args_has(args, 'a') || before) {
 		if (target->wl != NULL)
@@ -81,7 +89,7 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 			return (CMD_RETURN_ERROR);
 		}
 		if (args_has(args, 'n')) {
-			window_set_name(w, args_get(args, 'n'));
+			window_set_name(w, args_get(args, 'n'), ":.");
 			options_set_number(w->options, "automatic-rename", 0);
 		}
 		server_unlink_window(src_s, wl);
@@ -111,10 +119,10 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 
 	if (!args_has(args, 'n')) {
 		name = default_window_name(w);
-		window_set_name(w, name);
+		window_set_name(w, name, ":.");
 		free(name);
 	} else {
-		window_set_name(w, args_get(args, 'n'));
+		window_set_name(w, args_get(args, 'n'), ":.");
 		options_set_number(w->options, "automatic-rename", 0);
 	}
 
--- a/cmd-new-session.c
+++ b/cmd-new-session.c
@@ -99,15 +99,26 @@ cmd_new_session_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	}
 
-	tmp = args_get(args, 's');
+	tmp = args_get(args, 'n');
 	if (tmp != NULL) {
 		name = format_single(item, tmp, c, NULL, NULL, NULL);
-		newname = clean_name(name, "#:.");
-		if (newname == NULL) {
-			cmdq_error(item, "invalid session: %s", name);
+		if (!check_name(name, ":.")) {
+			cmdq_error(item, "invalid window name: %s", name);
 			free(name);
 			return (CMD_RETURN_ERROR);
 		}
+		free(name);
+	}
+
+	tmp = args_get(args, 's');
+	if (tmp != NULL) {
+		name = format_single(item, tmp, c, NULL, NULL, NULL);
+		if (!check_name(name, ":.")) {
+			cmdq_error(item, "invalid session name: %s", name);
+			free(name);
+			return (CMD_RETURN_ERROR);
+		}
+		newname = clean_name(name, ":.");
 		free(name);
 	}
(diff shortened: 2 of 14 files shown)


nicm:
Thanks. I think buffer names may as well not be allowed to contain `#` as well. `.` and `:` are OK.

barrettruth:
Thanks, just fixed.

nicm:
OK the idea is good but I think we need to change `check_name`. I don't think it should reject names that `utf8_stravis` will sort out, it is fine to silently let that happen.

So I think it would be better to change `check_name` not to call `clean_name` and instead just check for forbidden characters itself?

barrettruth:
fixed. lmk if you want the empty + utf8_isvalid rejection moved out of `check_name`

nicm:
Hmm. Maybe we are going too far with this.

The reason for forbidding `:` and `.` is because tmux uses them itself for targets, so I think forbidding them in session and window names is probably fine. IIRC we have not allowed them in session names for years.

`#` is a bit trickier... it just seems a bit unexpected to suddenly start forbidding it.

The idea of removing it is that something in a pane could use an escape sequence to rename a window or change a pane title to contain `#(foo)` and if that was then evaluated in the status line, `foo` could be executed.

But that only applies to escape sequences. There is no security boundary with commands - anything that could run `tmux renamew '#(foo)'` could equally well just run `foo` itself directly. So there is no requirement to remove `#` from a name set with `rename-window`.

So perhaps we should allow `#` unless the name or title came from an escape sequence?

That is easy enough for session and paste buffer names but all window names go into `window_set_name` and pane titles into `screen_set_title` so they would be a little more complicated.

Or maybe the whole thing is silly and aside from escape sequences we just allow commands to set anything and clean away only invalid UTF-8? If someone creates a session called `:.0` then they won't be able to target it, but does that matter?

What do you think?

barrettruth:
Agreed. I reverted it to only reject `:` and `.` in session and window names and leave `#`, pane titles, and buffers as before.

nicm:
This isn't enough because `window_set_name` and `screen_set_title` still replace `#` so even though you can set them with `rename-window` they will be removed.

They should be allowed from `rename-window`/`select-pane`/`automatic-rename` but forbidden from `input.c`.

And the `check`/`clean` calls should match for each.

barrettruth:
sorry for the back and forth. fixed it at the root (window_set_name and screen_set_title now take a forbid set). # is only stripped from escape sequences and kept from rename-window/select-pane/automatic-rename. this is what you were looking for, correct?

nicm:
A couple of them look off. I asked ChatGPT to summarise them against what the rules should be. I took a quick look and it seems not to be lying:

Class/source | Calling function | Desired set | PR does | Result
-- | -- | -- | -- | --
Command session name | cmd_new_session_exec, cmd_rename_session_exec | :. | :. | OK
Command session group name | cmd_new_session_exec | :. | :. | OK
Command window name | cmd_new_session_exec, cmd_new_window_exec, cmd_rename_window_exec, cmd_break_pane_exec | :. | :. | OK
Command pane title | cmd_select_pane_exec | empty | empty | OK
Command buffer name | paste_set, paste_rename via cmd_set_buffer_exec | empty | empty | OK
Escape-seq pane title | input_exit_osc, input_exit_apc → screen_set_title | # | # | OK
Escape-seq window name | input_exit_rename → window_set_name | :.# | # | Wrong
Automatic window rename | check_window_name → window_set_name | :. | empty | Wrong
Default window name from argv/shell | default_window_name → parse_window_name | :. | empty | Wrong
Spawned window final storage | spawn_window | :. | :. | OK

Here are the rules:

Class | Command | Escape sequence
-- | -- | --
Session names | :. | :.#
Window names | :. | :.#
Buffer names | empty | #
Pane titles | empty | #

nicm:
Are you still working on this? Think we need it sorted out one way or another for 3.7.

barrettruth:
sorry for the churn. went through your table row by row this time. escape-seq window name forbids `:.#` now (before as you pointed out was just just #), and `automatic-rename` + the default window name forbid `:.` (both were empty).

lastly, i also caught one that wasn't on your list: `break-pane`'s default-name branch was passing an empty set while every other window name uses `:.`, so
that's `:.` now too.

please check again :sweat_smile: is this what you were after?

nicm:
Thanks, this whole area is a bit of a mess so thanks for sticking with it.

I have applied this now with a few fairly minor changes. I added defines for the "forbid" values to stop copying `":."` everywhere, pulled the format expansion out of `spawn.c` so we are not doing it twice for window names and moved some code around a bit to try and make it a bit tidier.

PR #5177, opened 2026-06-08
Add new -h arg to choose-tree and choose-client to hide current pane in preview

### LINKED ISSUE #5056: Feature request: allow opening choose-tree in a pop-up
### Issue description

This has been previously proposed in https://github.com/tmux/tmux/issues/2836, but the benefits and impact on UX were not explained.

An example use case for this capability is quickly and unobtrusively switching sessions with a custom key binding (e.g., `tmux bind S choose-tree -sNZO time`, which is very similar to the default `s` key binding).

The problem is that `choose-tree` displays the list of sessions / windows / panes in the top-left corner, which doesn't look great on large or wide displays:

Image

Foregoing the `-Z` flag does not provide a consistent experience, because it might lead to the same top-left corner anchoring, depending on the currently-focused pane:

Image

Comparatively, opening the selection menu in a centered popup is consistent regardless of the focused pane, is easy to focus on visually, and reduces screen-wide transitions / flickering:

Image

In my particular case, the current behavior causes mild eye strain.

I would be deeply grateful if you would reconsider supporting this feature.

### Required information

* tmux version (`tmux -V`): `tmux 3.5a`
* Platform (`uname -sp`): `Linux x86_64`
* Terminal in use: Alacritty
* $TERM *inside* tmux (`echo $TERM`): `tmux-256color`
* $TERM *outside* tmux (`echo $TERM`): `alacritty`
* Logs from tmux (`tmux kill-server; tmux -vv new`): N/A

mgrant0:
In `choose-tree` and `choose-client`, add new -h arg to omit the current pane from being in this list.  This is useful if you bind it to something like this:

```
tmux bind S 'new-pane  -x50% -y50% -E; choose-tree -h'
```
or
```
tmux bind S 'new-pane  -x50% -y50% -E; choose-client -h'
```

### DIFF
--- a/cmd-choose-tree.c
+++ b/cmd-choose-tree.c
@@ -33,8 +33,8 @@ const struct cmd_entry cmd_choose_tree_entry = {
 	.name = "choose-tree",
 	.alias = NULL,
 
-	.args = { "F:f:GK:NO:rst:wyZ", 0, 1, cmd_choose_tree_args_parse },
-	.usage = "[-GNrswZ] [-F format] [-f filter] [-K key-format] "
+	.args = { "F:f:GhK:NO:rst:wyZ", 0, 1, cmd_choose_tree_args_parse },
+	.usage = "[-GhNrswZ] [-F format] [-f filter] [-K key-format] "
 		 "[-O sort-order] " CMD_TARGET_PANE_USAGE " [template]",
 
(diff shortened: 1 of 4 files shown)


nicm:
Applied now, I made a few changes to the man page bits. Thanks!

PR #5178, opened 2026-06-08
Refresh copy mode from the pane incrementally instead of recloning the whole history.

barrettruth:
Superseded by #5165, which folds this incremental backing-grid sync together with the copy-mode auto-refresh feature.

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -292,7 +292,7 @@ grid_free_line(struct grid *gd, u_int py)
 }
 
 /* Free several lines. */
-static void
+void
 grid_free_lines(struct grid *gd, u_int py, u_int ny)
 {
 	u_int	yy;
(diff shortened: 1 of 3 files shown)


nicm:
OK I thought this would be a monster diff but this is pretty simple. Makes sense to me, although I didn't test it yet, so just nits really:

- `grid_free_line` can stay `static`? Only `grid_free_lines` is used.

- The `_total` suffix on `appended_total` seems a bit unnecessary. The term we use in the code for "drop" is "collect" so I might rename `dropped` to `collected`. Also `appended` sounds odd for this, so I'd probably just go with `added`. It feels like some prefix on these would be nice, like maybe `scroll_added`, `scroll_collected`, `scroll_generation`? I dunno, not that fussy about this TBH.

- Also bikeshedding but returning 0/1 from `window_copy_sync_backing` would be better than -1/0 and would match other code in the same file.

barrettruth:
> OK I thought this would be a monster diff but this is pretty simple.

i found a way simpler solution, and after profiling, kept the 20% of the changes with the 80% performance implications xd

changes made - pr is ready.

nicm:
This seems to work for me.

Given these are quite small and self-contained, I'm not sure there is any point in doing them separately. Do you want to roll the `refresh-on`/`refresh-off`/`refresh-toggle` into this?

Do we want to remove `refresh-from-pane`?

barrettruth:
moving discussion to #5165 ;)

PR #5181, opened 2026-06-08
Add -B to new-pane to block and return a return code from the process.

mgrant0:
`new-pane -B` blocks the issuing client's command queue until the pane's command exits, then returns that command's exit
  status as the client's exit code — so a script can do:
```
status=$(tmux newp -B whiptail ...; echo $?)
```
 The server and other clients are never blocked.


nicm:

PR #5183, opened 2026-06-09
refactor(floating_panes): Layout cell position in root layout.

daneofmanythings:
~~(WIP)~~

Currently, layout cells are appended to the container at the top of the layout tree. This PR aims to meaningfully place them inside the tree to enable float/tile operations.


daneofmanythings:
When a floating pane is created, the layout cell is put after the provided window pane in the root layout. The window pane can be specified in `new-pane` with the target.

**Note**: the functionality is purposely simple at this stage. `-b` can be added later to enable `SPAWN_BEFORE`.

PR #5187, opened 2026-06-09
fix(master): Removed stale variable declarations.

daneofmanythings:
A recent change to `kill-pane` has left stale variables in `cmd_kill_pane_exec`. These are generating compiler warnings. The logic that utilized these variables has been moved into a helper function.


nicm:
Thanks, I already did this, not sure why GitHub is still showing this as different.

PR #5188, opened 2026-06-10
refactor(floating_panes): updating tile-pane and float-pane to use new layout mechanics.

daneofmanythings:
~~(WIP)~~

I am updating `tile-pane` and `float-pane` to use the new layout logic system. Once this is solid, I will be moving the functionality into the other existing commands.

I will detail here how I plan on accomplishing this ~~(bugs are still being squashed)~~. Note that function names can be changed; they were made hastily to get coding.

Given that we would like to be able to restore floating layout geometry, some new state needs to be added to `struct layout_cell` to track this. It was previously added to `struct window_pane`. I am not certain adding four new fields is the correct approach, but it is the simplest for now.

When a tiled pane is "floated":
1. Any previous geometry is fetched from the layout cell.
2. The geometry it occupies in the tiled layout needs to be reclaimed by either a sibling, or in the case that it is the last tiled pane in the parent, a sibling of the parent reclaims the geometry (example: `splitw; splitw -h; floatp -t2; floatp -t1`). This is currently being done in `layout_cell_remove_tile`.
3. The floating geometry needs to be determined using a combination of provided arguments, stored dimensions, and defaults in a priority of that order. I broke out some logic to reuse the argument parsing. This is happening in `layout_cell_floating_args`.
4. The expected layout and pane touch-ups happen.

When a floating pane is "tiled":
1. The floating dimensions are saved in the layout cell.
2. A neighbour is selected from `layout_cell_get_neighbour` (a neighbour is a sibling that exists in the tiled layout). Since floating panes now have meaningful layout positions, this will restore the pane to its previous position.
3. We perform a "split" (two functions have been broken out of `layout_split_pane` to reuse logic) in place and recursively resize siblings/parents as necessary to make room. This is happening in `layout_cell_insert_tile`. None of the old tiled dimensions are stored nor used for this calculation. There is no guarantee of consistent layout state between floating and tiling, so the simplest thing to do is a 50/50 split of the parents type. Logic can be added in the future to track IF the state hasn't changed, the old tiling dimensions can be used.
4. The expected layout and pane touch-ups happen.

A lot of the PR so far is pulling out existing logic into their own functions for reuse with two new workhorse functions `layout_cell_remove_tile` and `layout_cell_insert_tile`. I am anticipating using these in the logic to hide/show panes, which is why they aren't local to the tile/float commands.

Let me know if ya'll have any thoughts or questions on what I have here. ~~The code is still being debugged, so is less useful to look at, but don't hesitate to provide feedback anyway.~~  Thanks!

@nicm @mgrant0

### DIFF
--- a/cmd-tile-float-pane.c
+++ b/cmd-tile-float-pane.c
@@ -88,7 +88,7 @@ cmd_float_pane_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	}
 
-	layout_cell_remove_tile(w, lc);
+	layout_remove_tile(w, lc);
 	layout_cell_floating_args(item, args, w, &sx, &sy, &ox, &oy, &cause);
 	if (cause != NULL) {
 		cmdq_error(item, "failed to float pane: %s", cause);
@@ -134,7 +134,7 @@ cmd_tile_pane_exec(struct cmd *self, struct cmdq_item *item)
 	}
 
 	layout_save_size(lc);
-	if (layout_cell_insert_tile(w, lc) == 0)
+	if (layout_insert_tile(w, lc) == 0)
 		return (CMD_RETURN_ERROR);
 
 
--- a/cmd-tile-float-pane.c
+++ b/cmd-tile-float-pane.c
@@ -38,10 +38,10 @@ static enum cmd_retval	cmd_tile_pane_exec(struct cmd *, struct cmdq_item *);
 
 const struct cmd_entry cmd_float_pane_entry = {
 	.name = "float-pane",
-	.alias = NULL,
+	.alias = "floatp",
 
-	.args = { "t:x:y:w:h:", 0, 0, NULL },
-	.usage = "[-h height] [-w width] [-x x] [-y y] "
+	.args = { "dt:x:X:y:Y:", 0, 0, NULL },
+	.usage = "[-x height] [-y width] [-X x-position] [-Y y-position] "
 		 CMD_TARGET_PANE_USAGE,
 
 	.target = { 't', CMD_FIND_PANE, 0 },
@@ -52,9 +52,9 @@ const struct cmd_entry cmd_float_pane_entry = {
 
 const struct cmd_entry cmd_tile_pane_entry = {
 	.name = "tile-pane",
-	.alias = NULL,
+	.alias = "tilep",
 
-	.args = { "t:", 0, 0, NULL },
+	.args = { "dt:", 0, 0, NULL },
 	.usage = CMD_TARGET_PANE_USAGE,
 
 	.target = { 't', CMD_FIND_PANE, 0 },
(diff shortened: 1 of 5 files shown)


daneofmanythings:
Okay, I believe I have gotten it to a testable point. Things I am unsure about:
- It seems nice to be able to update the dimensions of floating geometry whilst not having to specify everything. Previously stored geometry is used where no args specify otherwise. The solution I came up with is to reuse the logic from `layout_get_floating_cell`. I'm not sure I like how it ended up. I can't seem to lock down `layout_get_floating_cell`; it has been on a journey.
- The naming of some functions, primarily `layout_insert_tile` and `layout_remove_tile`. They handle giving up, and taking back tiling dimensions on float/tile operations, and will be used to do a similar thing with hide/show.

nicm:
@mgrant0 should look at this preferably :-).

daneofmanythings:
I'm going to wait for a review to merge this. I feel it is warranted.

mgrant0:
Some comments by just reading through the code changes:
1. Remember to stack all your declarations at the top of functions.  (It's a BSD coding style requirement).
2. in layout_create_cell, I see 'is_node' on lines 314 and 316 which means it's a container.  Would you mind naming this 'is_container' instead?  You do use the term 'container' in a comment below on line 565 by the way.
3. layout_resize_set how about naming it layout_resize_set_size.
4. in layout_split_sizes, you pass in the flags but only use SPAWN_BEFORE.  I would pass in the SPAWN_BEFORE flag itself so it is more obvious what the function is doing when it is called.
5. layout_cell_floating_args, should this be named layout_cell_floating_parse_args perhaps?
6. similarly, should layout_get_floating_cell be named layout_get_new_floating_cell?
7. something does not look right in layout_remove_tile, you remove a tile but only add 1 column or line to the other adjacent cell then call it recursively?  you can walk me through this on a call if you like.
8. layout_insert_tile, i'm a little unclear here on this, let's talk about it too
9. in the man page, you have "If the pane has previously been floating, the geometry is restored for".  I think you mean "If the pane had..." but in the next line you have "dimensions not specified by the provided arguments:", do you really mean "not"?  Probably need to split this into more than one sentence.
10. "The pane's takes half the relevant dimension from the nearest tiled pane in the", the "pane's take" is wrong, I think you mean "The pane takes half of the relevant space" rather than 'dimension'.
11. just a question, in window_pane_is_hidden you return 0 if lc==NULL, what does it mean if the pane has no layout_cell in this case if it isn't hidden?
I am going to try to build this pr and try it, we can discuss if you want.

mgrant0:
@nicm I've looked at this now with Dane and it looks good.  Have a look and make sure all the names of things are the way you like them.

daneofmanythings:
There are a few changes that I need to implement from our discussion that the next commit will contain.

daneofmanythings:
@nicm I am going to merge this so I can keep working. Let me know if you find anything I should change!

PR #5190, opened 2026-06-10
Rename new-pane -B to -W.

mgrant0:


nicm:

PR #5192, opened 2026-06-10
feat(master): resize floating panes

daneofmanythings:
`resize-pane` can not affect floating panes. The boundaries are specified with `U/D/L/R` and negative values may be given.

### DIFF
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -41,8 +41,8 @@ const struct cmd_entry cmd_resize_pane_entry = {
 	.alias = "resizep",
 
 	.args = { "D::L::MR::Tt:U::x:y:Z", 0, 1, NULL },
-	.usage = "[-DLMRTUZ] [-x width] [-y height] " CMD_TARGET_PANE_USAGE " "
-		 "[adjustment]",
+	.usage = "[-MTZ] [-U up] [-D down] [-L left] [-R right] "
+		 "[-x width] [-y height] " CMD_TARGET_PANE_USAGE,
 
 	.target = { 't', CMD_FIND_PANE, 0 },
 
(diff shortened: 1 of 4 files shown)


nicm:
Looks good but a couple of minor comments:

- `-U up` is a bit odd, what is an `up`? I would make it `-U size` or `-U lines` and `-L size` or `-L columns`.

- Since `strtonum` can now return negative numbers, `adjust` needs to be `int` not `u_int`

- I think there needs to be some range checks. If `lc->xoff` is 10 and I do `-L -2147483647` (with `int adjust`) it will wrap which may be UB (depending on compiler flags). Range checking the size as well as the offsets is probably wise too although you won't have UB from unsigned integer overflow.

- `cause` and `flag` and `flags` can all be declared on one line.

daneofmanythings:
On it. Is there a use case for comically large numbers? I struggle to think of one. I am thinking of bounding the adjustments to a factor of window size, like 2x/4x. Do you think the command should fail if a boundary is exceeded, or clamp and succeed silently?

nicm:
In `layout_get_floating_cell` we report an error.

I don't know if there is a use case but we have to set the limit somewhere. We enforce `WINDOW_MAXIMUM` in a few places but not consistently, and 10000 is not really much less silly than `INT_MAX`. I guess the important thing is not to overflow `int`s if possible and not letting the resulting sizes or offsets be bigger than +/- 10000 would achieve that :-).

daneofmanythings:
Okay, I have added error handling to changing the dimensions on floating panes using a familiar api to what already exists. The two new functions are `layout_resize_floating_pane` (see `layout_resize_pane`) and `layout_resize_floating_pane_to` (see `layout_resize_pane_to`). Isolating the error handling seemed like the right thing to do. I also added a little helper define `PANE_MAXIMUM`, and set it to 10000.

A small annoyance: If a user wants to do a resize and specify multiple dimensions at once, they are able to. However, if a partial application of the resizing happens in the wrong order resulting in a temporary invalid dimension (for example `lc->sx` is 20, and `-R30 -L-30` may result in `lc->sx = -10`) the command will fail. The use case for this is using `resize-pane` to move the pane around, so it shouldn't be an issue once that functionality is implemented into an appropriate command.

nicm:
Looks good, applied to OpenBSD with some minor changes, will be in GitHub soon. Thanks!

PR #5193, opened 2026-06-10
fix(floating_panes): `swap-pane` can now affect floating panes.

daneofmanythings:
It seems like the new layout logic works with floating panes in `swap-pane`. I couldn't get it to break.

**EDIT**: as for the second commit message, I noticed that the guard had an `&&` previously. Is there a reason to restrict swapping two floating panes?

### DIFF
--- a/cmd-swap-pane.c
+++ b/cmd-swap-pane.c
@@ -79,12 +79,6 @@ cmd_swap_pane_exec(struct cmd *self, struct cmdq_item *item)
 	if (src_wp == dst_wp)
 		goto out;
 
-	if (window_pane_is_floating(src_wp) &&
-	    window_pane_is_floating(dst_wp)) {
-		cmdq_error(item, "cannot swap floating panes");
-		return (CMD_RETURN_ERROR);
-	}
-
 	server_client_remove_pane(src_wp);
 	server_client_remove_pane(dst_wp);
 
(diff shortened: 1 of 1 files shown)


nicm:
I wonder if the xor bit is why it's broken in master.

daneofmanythings:
Could be! That is old logic that got through.

nicm:
I'll try deleting it and see if it fixes it tomorrow and if it does we can just do this in master.

Does swap -U/-D work? How does it determine the up and down pane for floating panes?

daneofmanythings:
It does, and it determines them the same way as for tiled. It looks for the previous/next pane by number and swaps them. Should there be logic that restricts `U/D` to only target tiled if floating and floating if tiled?

nicm:
Not sure it makes sense to swap by number for floating panes. I guess it works for tiled because the pane indexes are always close in position but floating panes could be anywhere? May be better doing something than turning it off though although yes probably should only swap with panes of the same type.

If you have time could you check if swapping with the marked pane works (it's on the right click menu but either seems to be always disabled, may be a problem with the menu format, or doesn't work).

daneofmanythings:
There is a bit of reason to swapping between floating and tiled. Since floating panes exist meaningfully in the layout (in floating_panes branch), their tiled positions carry through the swap. It make no sense on master at the moment.

Yep, I'll look at marked panes!

nicm:
OK let's just leave `swap-pane` disabled in `master` for the moment if it needs the layout changes to really work.

daneofmanythings:
Okay sounds good! Merging this to `floating_panes` then.

PR #5194, opened 2026-06-11
fix(master): fixed format for 'Swap Marked' in the pane menu.

daneofmanythings:
Requested from this [conversation](https://github.com/tmux/tmux/pull/5193#issuecomment-4675267227).

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -65,7 +65,7 @@
 	" ''" \
 	" '#{?#{&&:#{!:#{pane_floating_flag}},#{>:#{window_panes},1}},Swap Up,}' 'u' {swap-pane -U}" \
 	" '#{?#{&&:#{!:#{pane_floating_flag}},#{>:#{window_panes},1}},Swap Down,}' 'd' {swap-pane -D}" \
-	" '#{?pane_marked_set,,-}Swap Marked' 's' {swap-pane}" \
+	" '#{?#{pane_marked_set},,-}Swap Marked' 's' {swap-pane}" \
 	" ''" \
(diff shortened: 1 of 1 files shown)


nicm:
Are you sure that changes anything? Both should be the same.

daneofmanythings:
I thought so too, but it wasn't expanding for some reason. Explicitly adding the octo and curlies made it be found. I used `display-message -v` to diagnose it. My power is out currently, but I can paste some outputs when it comes back on.

daneofmanythings:
Specifically the output said that the key wasn't found, so it assumed false. Thats why the menu was always off.

nicm:
OK I'll take a look because it should work.

nicm:
I tried this and actually it seem to be working fine, so I don't know what happened, I must have got confused when I thought it was wrong before...

But since `swap-pane` is disabled entirely for floating panes in `master`, I'm going to turn this menu option off as well.

nicm:
This is the same in both floating and tiled panes:
~~~~
$ tmux selectp -m
$ tmux display -p '#{?pane_marked_set,Y,N}'
Y
$ tmux selectp -M
$ tmux display -p '#{?pane_marked_set,Y,N}'
N
~~~~
And the menu option does seem to be fine... I dunno. Sorry for wasting time here.

nicm:
If you like, you can turn the menu options back on in `floating_panes` if `swap-pane` is working so we don't forget to do it later :-).

daneofmanythings:
This is so strange. I just did the test today which failed last night, and now its working fine. Glad it's working now.

PR #5199, opened 2026-06-12
CONTRIBUTING: fix shell code block language

iztanos:
Fix a Markdown code fence language tag from `base` to `bash` for a shell command example.


nicm:
Fixed, thanks.

PR #5201, opened 2026-06-12
fix(master): remove old floating logic from `swap-pane`.

daneofmanythings:
Even though swapping is disabled with floating panes, leaving this old code in there will only add noise when the time comes to enable it.


nicm:
Done, thanks!

PR #5203, opened 2026-06-13
regress: add command-alias test

iztanos:
Add a regression test for command-alias expansion.

The test defines a zoom alias for resize-pane -Z, invokes it, and checks #{window_zoomed_flag} to verify that the alias had the expected effect.


nicm:
Added, thanks!

Issue #5116, opened 2026-05-21
Add option to make cursor fixed relative to text and not to screen


yousef8192:
In tmux copy mode if you set keybinds to scroll using "scroll-up" or "scroll-down" what happens when you scroll is that the cursor position is fixed relative to the screen. However in vim the behavior is that the cursor position is fixed relative to the text itself. I hope the demonstration below makes it more clear.

Tmux 
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6b2cf231-af69-45dc-850d-bfb079761f69" />

Vim
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c8fd4279-6a96-4208-8269-c7e2f0b55810" />

Furthermore, having scroll behavior as described in this issue implies an important yet simple option to be added too to make the experience similar to vim. This option is called "scrolloff" and it determines how many lines the cursor should be away from the top/bottom end of the screen during scrolling. For example if scrolloff is set to 5 then there must be at least 5 lines between the cursor and the top/bottom end when scrolling as follows:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/adf0bbc4-2c08-4b85-b0c9-6e563c43883d" />

Tmux is one of my favorite tools and I use it daily. This option makes the experience be consistent for people like me who use vim or in general are used to scrolling with the behavior described above so without this option it is becomes very uncomfortable to have different scrolling mechanisms to get used to simultaneously. I opened an issue similar to this earlier with an assumption that vim motions are familiar but apparently more clarification was needed so sorry for the inconvenience. If there are any further clarifications needed please do let me know. Thanks a lot for your time and effort!

nicm:
I'm not going to do this but maybe someone else will (a lot of people have sent changes to be more compatible with vi). But it should either just be the default for vi mode, or a new command, not an option.

nicm:
Added to todo list [here](https://github.com/tmux/tmux/wiki/Contributing#small-things).

PR #5216, opened 2026-06-13
Add copy-mode-scrolloff option for vi mode, issue 5116.

catterer:
As per https://github.com/tmux/tmux/issues/5116 , this code changes the behavior of scroll-up and scroll-down in vi mode. When scrolling with Ctrl+e and Ctrl+y, the cursor tries to stick to the current line, unless it goes over the scrollup threshold. 

Added a new `copy-mode-scrollup` option to configure said threshold.

AI disclaimer: AI was used to update the manpage. The code is human-written.

### DIFF
--- a/tmux.1
+++ b/tmux.1
@@ -2322,6 +2322,11 @@ that line.
 (emacs: C\-Down)
 .Xc
 Scroll down.
+If
+.Ic mode\-keys
+is
+.Ic vi ,
+the cursor is fixed relative to the text.
 .It Xo
 .Ic scroll\-down\-and\-cancel
 .Xc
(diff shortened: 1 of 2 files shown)


nicm:
Thanks. I don't think this should be an option. Pick a value you like and use it. What does vi do? 0?

catterer:
vim has a scrolloff setting: https://vimdoc.sourceforge.net/htmldoc/options.html#'scrolloff', default is 0

nicm:
OK what is the default?

nicm:
Oh sorry, you said. Just make it 0 then in tmux.

catterer:
done, no option anymore

nicm:

yousef8192:
Thanks a lot this works as intended, however though there is a small issue
1. when we are at the very top (i.e. can't scroll up any more), ctrl+y moves the cursor down
2. vice versa when we are at the very bottom (i.e. can't scroll down any more), ctrl+e moves the cursor up

What should happen is that the cursor should stay fixed and doesn't move.

PR #5224, opened 2026-06-15
configure.ac: Prefer wide ncurses libs

### LINKED ISSUE #4844: tmun-3.6a error - missing or unsuitable terminal: xterm with ncurses-6.6_p20260131
### Issue description

After I upgraded to the latest ncurses-6.6_p20260131 - tmux-3.6a can't starting in my lxc with gentoo. It shows an error - missing or unsuitable terminal: xterm.

Downgrading ncurses to 6.6-p20260124 resolv an error. 

Maybe this is an ncurses issue only.

* tmux version (`tmux -V`) - tmux 3.6a
* Platform (`uname -sp`). - Linux Intel(R) Core(TM) i7 CPU 860 @ 2.80GHz
* Terminal in use (xterm, rxvt, etc) - xterm
* $TERM *inside* tmux (`echo $TERM`) - tmux-256color
* $TERM *outside* tmux (`echo $TERM`) - xterm
* Logs from tmux (`tmux kill-server; tmux -vv new`) - missing or unsuitable terminal: xterm

Polynomial-C:
otherwise startup of tmux may fail with very recent ncurses:

  `missing or unsuitable terminal: xterm-256color`

Fixes: #4844

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -297,24 +297,24 @@ fi
 # Look for ncurses or curses. Try pkg-config first then directly for the
 # library.
 PKG_CHECK_MODULES(
-	LIBTINFO,
-	tinfo,
+	LIBTINFOW,
+	tinfow,
 	[
-		AM_CPPFLAGS="$LIBTINFO_CFLAGS $AM_CPPFLAGS"
-		CPPFLAGS="$LIBTINFO_CFLAGS $SAVED_CPPFLAGS"
-		LIBS="$LIBTINFO_LIBS $LIBS"
+		AM_CPPFLAGS="$LIBTINFOW_CFLAGS $AM_CPPFLAGS"
+		CPPFLAGS="$LIBTINFOW_CFLAGS $SAVED_CPPFLAGS"
+		LIBS="$LIBTINFOW_LIBS $LIBS"
 		found_ncurses=yes
 	],
 	found_ncurses=no
 )
 if test "x$found_ncurses" = xno; then
 	PKG_CHECK_MODULES(
-		LIBNCURSES,
-		ncurses,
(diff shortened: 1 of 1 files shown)


nicm:
Might it. Why?

Polynomial-C:
> Might it. Why?

Dunno. I got the hint from https://bugs.gentoo.org/972804#c2

nicm:
Hmm I wonder why. I don't know there should be any differences in how the terminfo is read with w and not-w.

ThomasDickey:
It depends (Debian read the notes, Gentoo didn't).

 I answered that in https://bugs.gentoo.org/910430 (Gentoo packagers didn't read my answer and respond accordingl).

nicm:
OK the issue seems like it may be that Gentoo is building a `libtinfo` that is genuinely `--disable-widec`:
~~~~
if [[ ${target} == ncurses*w ]] ; then
		conf+=( --enable-widec )
	else
		conf+=( --disable-widec )
	fi
~~~~
That means it cannot read the `xterm` and `xterm-256color` files generated by a `--enable-widec` build. In fact with `--disable-widec` they are not built at all:
~~~~
"terminfo.tmp", line 5820, terminal 'xterm': entry is larger than 4096 bytes
~~~~
The limit is 4096 without `--enable-ext-colors` and that is only supported with `--enable-widec`.

TBH tmux doesn't care if ncurses has wide characters or not, so we could probably just do what you are suggesting and prefer the `libtinfow` library. Although I expect Gentoo should really just be making them the same library which I assume is what @ThomasDickey is suggesting.

nicm:
I have applied this change now, thanks!

PR #5236, opened 2026-06-16
refactor(master): Added the 'neighbour' logic and `layout_remove_tile`.

daneofmanythings:
This also includes the fixes from #5235, so that it works on its own.

### DIFF
--- a/layout.c
+++ b/layout.c
@@ -588,12 +588,57 @@ layout_resize_adjust(struct window *w, struct layout_cell *lc,
 	}
 }
 
+/* Find and return the nearest neighbour to a cell in a specific direction. */
+static struct layout_cell *
+layout_cell_get_neighbour_direction(struct layout_cell *lc, int direction)
+{
+	struct layout_cell	*lcneighbour = lc;
+
+	while (1) {
+		if (direction)
+			lcneighbour = TAILQ_NEXT(lcneighbour, entry);
+		else
+			lcneighbour = TAILQ_PREV(lcneighbour, layout_cells,
+			    entry);
+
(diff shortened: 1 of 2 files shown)


nicm:
Looks good, and good comments. Some minor style stuff:

- In `layout_cell_get_neighbour_direction` if you called it `lcneigh` or `lcn` you wouldn't need to wrap `TAILQ_PREV` across two lines. If you *do* wrap, it use braces - only leave the braces out if there is actually one line, not one statement.

- `return lcother;` needs `()`s. Maybe we should get rid of this convention... I can't see anyone wanting to do `#define return(foo) ...` these days...

- `cells space` -> `cell's space`.

daneofmanythings:
Fixed it up.

I didn't fully understand braces/no braces. Thanks! I do catch myself missing `()`s around return statements sometimes; this one slipped through the cracks. If we get rid of the convention, it's probably a good idea to get rid of every `()` return statement instead of having things be mixed. I'm fine with the `()`s.

nicm:
Applied now, thanks!

Although now I look at it maybe `layout_remove_tile` should be `layout_remove_tiled_cell`? Not worth changing now but maybe as part of a bigger diff if you think it is better.

daneofmanythings:
I am totally on board changing the name of `layout_remove_tile` and its partner `layout_insert_tile`. I think adding `cell` to it is slightly unclear as the layout cell is still in the layout tree, it is just the tiled space that is being given up. Possibly `layout_remove_tiled_space` is better? Or maybe the more evocative verbage of `layout_collapse_tile`?

PR #5244, opened 2026-06-18
control: avoid NULL deref in session-window-changed control notification

### LINKED ISSUE #5243: control mode: NULL dereference in control_notify_session_window_changed crashes the server
### Summary

`control_notify_session_window_changed()` (control-notify.c) dereferences `s->curw->window->id` with no check that `s->curw` is non-NULL. The `session-window-changed` control-mode notification is **deferred** through the command queue and holds a reference to the session, so it can fire *after* the session has been destroyed: `session_destroy()` sets `s->curw = NULL` but the notification's held reference keeps the `struct session` alive. When the deferred notification then runs, `s->curw` is NULL and `s->curw->window` is a NULL dereference, crashing the **entire server** (all sessions and clients), not just the control client.

Reachable from ordinary commands (no malformed input) whenever a control-mode client (`tmux -CC` / `-C`) is attached.

### Version

Reproduced on `master` (`next-3.7`), commit `55b2e74`. The code path is long-standing.

### The code (control-notify.c)

```c
void
control_notify_session_window_changed(struct session *s)
{
	struct client	*c;

		if (!CONTROL_SHOULD_NOTIFY_CLIENT(c))
			continue;

		control_write(c, "%%session-window-changed $%u @%u", s->id,
		    s->curw->window->id);   /* s->curw may be NULL */
	}
}
```

Sibling functions in this file already guard their window access (`control_notify_window_pane_changed` checks `w->active == NULL`; `control_notify_window_layout_changed` checks `TAILQ_FIRST(&w->winlinks)`); this one has no guard.

### Crash

```
#0  control_notify_session_window_changed (s=0x...) at control-notify.c:232
#1  notify_callback (item=..., data=...) at notify.c:152
#2  cmdq_fire_callback (item=...) at cmd-queue.c:725
#3  cmdq_next (c=0x0) at cmd-queue.c:784        <- global queue drain (c == NULL)
...
Program terminated with signal SIGSEGV.
```

### Root cause

1. `session_set_current()` (e.g. from `select-window`) sets `s->curw` and calls `notify_session("session-window-changed", s)`.
2. `notify_add()` takes a reference (`session_add_ref(s)`) and **defers** the work via `cmdq_append(NULL, cmdq_get_callback(notify_callback, ne))` — it does not run synchronously.
3. Later in the same command-queue drain, `kill-session` runs `session_destroy()`, which sets `s->curw = NULL` and drops one reference; the notification's reference keeps the session allocated.
4. The deferred `notify_callback` fires and dispatches `control_notify_session_window_changed(s)` with `s->curw == NULL` → SIGSEGV.

### Reproduction

Deterministic. With a control client attached, run one batched command list that selects a different window and then kills that session in a single command-queue drain:

```sh
SOCK=/tmp/bug.sock
tmux -S $SOCK new-session -d -s s -n win0
tmux -S $SOCK new-window  -t s: -n win1

# attach a control-mode client and keep its stdin open so it stays attached
mkfifo /tmp/bug.fifo; sleep 1000 > /tmp/bug.fifo &
tmux -S $SOCK -C attach-session -t s < /tmp/bug.fifo > /tmp/bug.out 2>&1 &
sleep 1   # let the control client finish attaching

OTHER=$(tmux -S $SOCK list-windows -t s -F '#{window_id}' | grep -v "$(tmux -S $SOCK display-message -p -t s '#{window_id}')" | head -1)

# crash:
tmux -S $SOCK select-window -t "$OTHER" \; kill-session -t s
```

Two details matter for reliability: the control client's stdin must stay open (else it detaches early), and `select-window` must target a window other than the current one (otherwise `session_set_current` isn't called and no notify is queued).

### Fix

Skip the notification when there is no current window. A PR is linked below.

bmaurer:
Fixes #5243.

`control_notify_session_window_changed()` dereferences `s->curw->window->id` without checking that `s->curw` is non-NULL. The `session-window-changed` notification is deferred through the command queue (`notify_add()` → `cmdq_append()`) and holds a reference to the session, so it can run after `session_destroy()` has set `s->curw = NULL` — the held reference keeps the `struct session` alive. The deferred notification then dereferences NULL and crashes the whole server.

It is reachable with a control client (`tmux -CC`) attached by, for example, selecting a different window and then killing that session in a single command list, so the deferred notify fires after `curw` was cleared.

This skips the notification when there is no current window, matching how the other window-touching notify functions in this file guard their access (e.g. `control_notify_window_pane_changed` checks `w->active`).

Full analysis, backtrace, and a deterministic reproduction are in #5243.

### DIFF
--- a/control-notify.c
+++ b/control-notify.c
@@ -224,6 +224,15 @@ control_notify_session_window_changed(struct session *s)
 {
 	struct client	*c;
 
+	/*
+	 * A deferred session-window-changed notification can fire after the
+	 * session has been destroyed (which sets curw to NULL) but is kept
+	 * alive by the notification's reference. There is no current window to
+	 * report, so skip the notification.
+	 */
+	if (s->curw == NULL)
+		return;
+
 	TAILQ_FOREACH(c, &clients, entry) {
 		if (!CONTROL_SHOULD_NOTIFY_CLIENT(c))
 			continue;


nicm:

BTW if you are opening a PR you do not need to open an issue as well, just one will do :-).

PR #5250, opened 2026-06-20
Add scrollbar auto-hide feature.

mgrant0:
This may only be a holding PR.  Ultimately you may want to pull the screen_redraw branch first into master then pull this into master afterwards.


mgrant0:
The useful part is that the scene is already a final visibility map. For auto-hide, I could let
  the pane mark its normal content first, then when the scrollbar is revealed, mark only the
  scrollbar width columns over the pane edge. When it hides, rebuilding the scene without those
  scrollbar spans naturally restores pane spans underneath. That fits the scene architecture
  cleanly.

  What I might change or add after doing this:

  - Add a shared scrollbar geometry helper. Right now layout, redraw, mouse hit-testing, and
    resize/border code all need related-but-slightly-different answers: reserve area, overlay draw
    area, hover interaction area, scrollbar width, transparent pad. A helper returning those
    rectangles would reduce duplication and make left/right/pad behavior harder to drift.

  - Give auto-hide/overlay scrollbars a more explicit scene concept. I added
    REDRAW_SCROLLBAR_OVERLAY, but the scene still treats scrollbar spans mostly like normal
    scrollbar spans. That is fine, but a small redraw_scrollbar_geom or clearer span fields could
    make “drawn width” versus “hit-test width including transparent pad” more obvious.

  - Consider a narrower redraw path for hide. Currently hide sets PANE_REDRAW, which is correct
    because pane content underneath must be repainted. The scene could theoretically support
    redrawing just the overlaid columns, but that would add complexity. I would keep the current
    full-pane redraw unless performance proves it matters.

  The only awkward part is not really the scene; it is that mouse hit testing lives outside the
  scene, so it needs to know about the same overlay geometry. That is where I’d improve the
  structure first.

mgrant0:
scrollbar became green block, like status line colour.
image

fixed this now.

Issue #5251, opened 2026-06-20
floating panes cannot have 100% width/height


Morantron:
### Issue description

Floating panes cannot have its width/height set to `100%`. Popups have the ability to span to 100% with or without borders, should floating panes too?

- Open tmux with empty config
- Try `tmux new-pane -x 100%`
- tmux errors with `position too large`

### Required information

Please provide the following information. These are **required**. Note that bug reports without logs may be ignored or closed without comment.

* tmux version: `tmux next-3.7` (commit https://github.com/tmux/tmux/commit/cc922de7587bb74f93809bc51bb36228856bb287)
* Platform: `Linux unknown` (alpine linux with docker image`crystallang/crystal:1.19-alpine`)
* Terminal in use: `iTerm2`
* $TERM *inside* tmux: `tmux-256color`
* $TERM *outside* tmux: `xterm`
* Logs from tmux 

[tmux-server-2795.log](https://github.com/user-attachments/files/29159739/tmux-server-2795.log)
[tmux-out-2795.log](https://github.com/user-attachments/files/29159738/tmux-out-2795.log)
[tmux-client-2792.log](https://github.com/user-attachments/files/29159737/tmux-client-2792.log)

nicm:
They definitely can exist so this is just that the 100% is the content so there then isn't room for borders - but we should probably make it work. Probably a % should be total width including borders.

Does it work if you do `-Bnone`?

Morantron:
> Does it work if you do -Bnone?

The same error happens with `-Bnone`

daneofmanythings:
Thank you for the report @Morantron! Can you try out the fix in #5254 and let me know if that is the behavior you expect?

nicm:
Please try this again with `master` and let us know if there are still problems. Thanks!

Morantron:
works fine! thank you! 🙏

PR #5247, opened 2026-06-19
refactor(master): floating layout cells are placed meaningfully in the layout.

daneofmanythings:
Up until this point, floating layout cells were stuffed at the end of the top node of the layout tree. The changes prior to this have enabled this PR to function. Layout cells are now created after the target window panes cell of the `new-pane` command. This has no real affect as of now. However, this will allow for floating and tiling operations to maintain the cell's layout position.

PR #5254, opened 2026-06-20
fix(master): floating panes now accomodate for border size.

daneofmanythings:
Possible fix for #5251.

A few notes:
- This feels kind of hacky. I'm adjusting the size and positions based on border state.
- There was a maximum of window size for pane sizes which didn't make much sense to me. We already dont enforce a floating pane's spawn offsets to be in the window, so why enforce the size? I set it to PANE_MAXIMUM which feels more appropriate to me.
- This doesn't account for the pane border option that is already set. This PR intersects slightly with #5247 and would like that one to be finished before fully fixing up this one.

### DIFF
--- a/arguments.c
+++ b/arguments.c
@@ -998,7 +998,7 @@ args_string_percentage(const char *value, long long minval, long long maxval,
 		copy = xstrdup(value);
 		copy[valuelen - 1] = '\0';
 
-		ll = strtonum(copy, 0, 100, &errstr);
+		ll = strtonum(copy, 0, 1000, &errstr);
 		free(copy);
 		if (errstr != NULL) {
 			*cause = xstrdup(errstr);
@@ -1066,7 +1066,7 @@ args_string_percentage_and_expand(const char *value, long long minval,
 		copy[valuelen - 1] = '\0';
 
 		f = format_single_from_target(item, copy);
-		ll = strtonum(f, 0, 100, &errstr);
+		ll = strtonum(f, 0, 1000, &errstr);
 		free(f);
 		free(copy);
 		if (errstr != NULL) {
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -94,7 +94,7 @@ cmd_resize_pane_exec(struct cmd *self, struct cmdq_item *item)
 	server_unzoom_window(w);
 
 	if (args_has(args, 'x')) {
-		x = args_percentage(args, 'x', 0, INT_MAX, w->sx, &cause);
+		x = args_percentage(args, 'x', 0, PANE_MAXIMUM, w->sx, &cause);
 		if (cause != NULL) {
 			cmdq_error(item, "width %s", cause);
 			free(cause);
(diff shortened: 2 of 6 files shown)


nicm:
I haven't tried this but if I do 100% the pane should take up the entire width with both borders still visible (that is, in first and last column) except with -Bnone.

The pane should not have any part outside the window or have any uncovered columns unless I do 101% or 99%.

This might mean making 100% a special case.

nicm:
I don't mind if the size (% or absolute) is with borders or without borders, so long as we are consistent. But we should always cover exactly the whole area with 100%.

nicm:
OK this works well for me but:

- `tmux newp -x10000 -X0` works, but I can't do `tmux newp -x101% -X0`?

- It definitely needs to look at the window option if there is no `-B`. `resize-pane` does too - it is not intuitive that `tmux newp -x100% -X0 -Bnone` works but these don't:
~~~~
tmux set -g pane-border-lines none
tmux newp -x100% -X0
tmux resizep -x100%

daneofmanythings:
I had a thought yesterday too; How should i handle a change in the global border option? I assuming those changes should propagate to all panes regardless of if it were a float spawned with `-Bnone`. I should also add logic that updates the cell size when the pane border option is changed to and from `none`.

nicm:
You should not need to do anything if `pane-border-lines` is changed to `none` because it can only affect floating panes and they will just lose their border.

daneofmanythings:
Options are now considered! I bumped up the max when decoding a percentage from a string to 1000, that is what was preventing `101%`. I am fairly certain this won't have unintended side effects as all the other call sites seem to reference flat values. It's only the commands that take percentages.

nicm:
This is not quite right - you are calling `window_pane_get_pane_lines` on the `wp` that the function is given, but if that is floating it may have its own option value which is not the same as the window. I think you may need to add `window_get_pane_lines`.

Also I think it would be nicer to pass in `enum pane_lines` than to pass in a string and compare it with "none" because that bypasses the option choice list. Maybe parse `-B` into `lines` before calling `layout_get_floating_cell`?

All the `return (CMD_RETURN_ERROR);` after the `lc` assignment in `cmd_split_window_exec` leak `lc` and `sc.environ` and `sc.argv` so you could fix that while you are there if you like (there need to be a `goto fail`).

I think `resize-pane` needs to change as well to handle `PANE_LINES_NONE`? (It *can* call `window_pane_get_pane_lines`).

daneofmanythings:
I have fixed the line detection in `layout_get_floating_cell`. I chose to keep the border detection in the function instead of adding a parameter for borders because that code would be required at every call site. I think that is what you were suggesting? I believe it now handles things like all other options handling is done.

Good catch on `cmd-split-window.c`. Added the `fail` tag.

I don't think `resize-pane` needs extra border handling because the function it calls to resize the cell, `layout_resize_floating_pane_to` already handles that logic at `layout.c:814`.

nicm:
Looks good, I have applied this with some minor changes so it looks up `lines` in `cmd-split-window.c` rather than repeating the lookup twice. Thanks!

PR #5256, opened 2026-06-21
add flag to dismiss popup when click outside

jake-stewart:
This PR adds `-O` to `display-popup` to opt into dismissing when the user clicks outside of the popup.

I would find this useful because I sometimes spawn popups by clicking on the statusline and closing them requires either clicking the popup or using the keyboard. For some popups there is not much room to allow closing on click so I have to use the keyboard which is a mode switch.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -54,8 +54,8 @@ const struct cmd_entry cmd_display_popup_entry = {
 	.name = "display-popup",
 	.alias = "popup",
 
-	.args = { "Bb:Cc:d:e:Eh:kNs:S:t:T:w:x:y:", 0, -1, NULL },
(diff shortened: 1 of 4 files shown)


nicm:
I don't really want to add anything more to popups because I want to get rid of them.

At the moment I'm leaning towards just changing `display-popup` to open a floating pane. Maybe we need to add a "modal" flag for panes to handle cases like this (close when click outside).

The big difficulty with that is that floating panes do not sit on top of a zoomed pane.

nicm:
Thanks, but I think we are not going to do this. The path forward is that we will add modal floating panes, then put popups in the bin. This can be supported as part of that.

See this issue: https://github.com/tmux/tmux/issues/5258. I'll add a note about this feature to it as well. There is plenty of stuff to do for floating panes if you wanted to work on them :-).

PR #5260, opened 2026-06-22
fix(master): `split-window` can be used to create a new tiled pane when there is no tiled layout.

### LINKED ISSUE #5213: If you end up with no tiled panes, it is impossible to create any
There needs to be a way.

Maybe `split-window` should create one if there are no tiled panes? Not sure. Seems a bit inconsistent.

This might not be a problem when we have float-tile because you can just make a floating pane and then tile it.

daneofmanythings:
This fixes #5213. The logic to create and swap in a new parent node has been pulled into its own function. This logic is used in many places.

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -70,6 +70,38 @@ const struct cmd_entry cmd_split_window_entry = {
 	.exec = cmd_split_window_exec
 };
 
+static struct layout_cell *
+cmd_split_window_new_tiled(struct cmdq_item *item, struct args *args,
+    struct window *w, struct window_pane *wp, int flags, int new_pane,
+    char **cause)
+{
+	struct layout_cell	*lcnew, *lc = wp->layout_cell;
+	struct layout_cell	*lcroot = w->layout_root;
+
+	/*
+	 * When no panes are tiled and `new-pane` was used, a new cell is
+	 * created and inserted at the top of the root node. A new root node is
+	 * created if necessary.
+	 */
+	if (new_pane && !layout_cell_has_tiled_child(lcroot)) {
+		lcnew = layout_create_cell(NULL);
+		layout_set_size(lcnew, w->sx, w->sy, 0, 0);
+		if (lcroot->type == LAYOUT_WINDOWPANE) {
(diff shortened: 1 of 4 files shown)


daneofmanythings:
Fixed two bugs in the implementation.

nicm:
This doesn't apply, does it need to be updated?

daneofmanythings:
Hmm, strange. I rebased against master with no conflicts. Hopefully it applies now.

nicm:
OK this works but I don't think it is intuitive that `split-window` is an error until suddenly it is not.

Maybe `split-window` should create a new tiled pane always? If you are in a floating pane it should split the last used tiled pane? Or is that weird?

daneofmanythings:
I'm not sure of the use case for being in an active floating pane and then running a `split-window` with no target to purposefully split the last active tiled pane. But more to the point, it feels weird having the command `split-window` doing something that isn't "splitting a window." I think I am going to reorganize things so that if there is no tiled pane, the user needs to call `new-pane -L` instead so the semantics line up better.

daneofmanythings:
Okay, I am much happier with this. I put the logic for making a new tiled pane and put it into `cmd-split-window.c` and restricted that behavior to `new-pane` invocations. I also touched up the documentation a lot to accurately reflect the state of the commands. It had drifted a bit; I think I had forgotten to update it a few times. The behavior of making a new tiled pane when there is no tiled layout is also now documented.

nicm:
I think we should not do this at all, I don't like `split-window` and `new-pane` behaving more and more differently. They should have *one* case where they are different and that is whether they *always* make a floating pane (`new-pane`) or not (`split-window`). A few flags that only one of them support is OK too but that just means we don't include it in the usage, the code is the same.

I think if people end up with no tiled panes they should just make a floating pane and then tile it. That will be a lot easier anyway because making a pane and tiling it will be on key bindings and menus, instead of having to go dig the fact you need to run `new-pane -L` out of the manual.

daneofmanythings:
I'm closing this PR for now based on our off platform discussion. I'll make new PRs more inline with the design goals.

PR #5261, opened 2026-06-22
fix(master): memory leak in `layout.c:layout_get_tiled_cell`

daneofmanythings:
If there is a failure in parsing the args, the original value of `cause` was being leaked.


nicm:
Done, thanks!

Issue #5229, opened 2026-06-16
Incrementally transferring functionality from `floating_panes` to `master`


daneofmanythings:
This issue will track the work being done to slowly move functionality over to the master branch.

- ~~Add tiling predicates: #5228~~
- ~~Add 'neighbour' detection with insert: #5236~~
- ~~All layout cells exist meaningfully within the root layout: #5247~~
- ~~Tiled layout cells can be floated: #5262~~
- ~~Removing tiles from a layout and pulling out functionality from `layout_split_pane`: #5285~~ 
- ~~Floating layout cells can be auto tiled: #5293~~
- ~~Add pane menu buttons, pane border buttons, and a keybind for float/tile operations: #5321~~

PR #5262, opened 2026-06-22
feat(master): `break-pane` now has `W/x/y/X/Y` to float a tiled pane.

daneofmanythings:
This PR is the next step in #5229. Window panes can now be floated out of their tiled positions. There is also some cleanup with error handling and the float args api. I have left in the new `saved` fields of `struct layout_cell`. They will be used in the coming PRs to restore geometry.

### DIFF
--- a/cmd-break-pane.c
+++ b/cmd-break-pane.c
@@ -34,9 +34,10 @@ const struct cmd_entry cmd_break_pane_entry = {
 	.name = "break-pane",
 	.alias = "breakp",
 
-	.args = { "abdPF:n:s:t:", 0, 0, NULL },
-	.usage = "[-abdP] [-F format] [-n window-name] [-s src-pane] "
-		 "[-t dst-window]",
+	.args = { "abdPF:n:s:t:Wx:X:y:Y:", 0, 0, NULL },
+	.usage = "[-abdPW] [-F format] [-n window-name] [-s src-pane] "
+		 "[-t dst-window] [-x width] [-y height] [-X x-position] "
+		 "[-Y y-position]",
 
 	.source = { 's', CMD_FIND_PANE, 0 },
 	.target = { 't', CMD_FIND_WINDOW, CMD_FIND_WINDOW_INDEX },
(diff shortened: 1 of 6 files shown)


daneofmanythings:
I have rebased this to fit with the floating pane geometry fix. I should probably mention that I have reverted some of the api surface to how things were previously with `layout_floating_args_parse`. The previous implementation is what I was using when I had the misguided "placeholder" plan, and so I am moving back to what makes the most sense. There may come a day where tiled parsing needs the same treatment, but i can't see it happening soon.

nicm:
Thanks:

I think `layout_get_floating_cell` should stay in `layout.c` with `layout_get_tiled_cell`.

Using `cause` as an error marker is unintuitive, it is better to explicitly return success or failure then the caller can use that to decide whether to check the error. Can you make `layout_floating_args_parse` return 0/1 (false/true ie failure/success) or -1/0 (failure/success) instead? I'd change `layout_resize_floating_pane_to` and `layout_resize_floating_pane` as well while you are there.

Also you reverted this:
~~~~
-               ox = args_percentage_and_expand(args, 'X', -sx, w->sx,
+               ox = args_percentage_and_expand(args, 'X', -(int)w->sx, w->sx,
~~~~
Was it intentional?

daneofmanythings:
I have update the error signaling of those functions with failure/success semantics, and restored the old lower bound when parsing floating args. I had changed the lower bound because i can see a user being confused why a negative offset for differing sized panes sometimes fails. Having it be consistent for all sized panes seemed less confusing. But, following that logic leads me to want to make bounds `-PANE_MAXIMUM` and `PANE_MAXIMUM`. Regardless, I have changed it back!

nicm:
I don't really mind too much about the bounds, but the idea is that putting a pane at -99999 is pretty useless so we should make sure at least some of it (even just 1 row or column of border) is visible.

nicm:
Applied, thanks!

When we can do the reverse, this should definitely go on the pane menu.

PR #5266, opened 2026-06-23
Fix 5056 floating choose tree

mgrant0:
Fix for when choose-tree is run in a floating pane like a popup to hide the empty pane.  Also fixes the example in the manpage -l option example to use new-pane instead of splitw.


mgrant0:
fix_5056_floating_choose_tree and feature_add_h_hide_pane_in_choose are functionally different:

  - fix_5056_floating_choose_tree contains 4d369480, which changes choose-tree startup selection when
    the mode is opened from a floating pane. Instead of initially tagging/selecting the floating pane,
    it looks through window->last_panes for the first non-floating pane and uses that for the preview;
    if none exists, it falls back to the window. This is automatic behavior and only affects choose-tree
    pane-mode selection.

  - feature_add_h_hide_pane_in_choose adds an explicit -h flag to choose-tree and choose-client. With
    -h, the pane containing the mode is hidden from preview/list handling, intended for commands like
    new-pane ...; choose-tree -h. This is opt-in behavior.

  For master: commit 4d3694807df13a6b3a77537bce03b8fefb4854ed itself is not integrated into master. git
  branch --contains only shows it on fix_5056_floating_choose_tree, and git cherry -v master
  fix_5056_floating_choose_tree still reports it as not in master.

  master does have the related -h solution via upstream commit a0f4038d:

  a0f4038d Add a -h flag to choose-tree and choose-client to hide the pane containing the mode, intended
  for use with floating panes.

  So the answer is: 4d369480 was not integrated in a different way. master integrated a different
  solution to the same general problem. The missing behavior from master is the automatic fallback to
  the last non-floating pane when running plain choose-tree from a floating pane. In master, you need
  choose-tree -h / choose-client -h to avoid showing the mode pane’s empty preview.

PR #5271, opened 2026-06-24
fix(master): Added border offset calculation to `move-pane`

daneofmanythings:
`move-pane` was overlooked when the floating geometry made to accommodate differing border styles. I have also corrected a strange lower bound when calculating the positions.


nicm:
Applied, thanks!

PR #5272, opened 2026-06-24
Scrollbar autohide

### LINKED ISSUE #4725: Scrollbar should be hidden if entering copy-mode from alternate screen
### Issue description

To reproduce:
1. empty tmux config, run `set -g pane-scrollbars on`
2. run `lazygit` (the screen shows lazygit with no scrollbar from tmux)
3. enter copy mode
4. the scrollbar appears, pushing lines like this

Image

Alternate screen doesn't scroll like normal screen, copy mode can be used for copying text but scrollbar won't be helpful, and displaying it breaks text layout

See original discussion in https://github.com/tmux/tmux/issues/4699#issuecomment-3603520338

### Required information

* tmux version (`tmux -V`). 3.6
* Platform (`uname -sp`). Linux x86_64
* Terminal in use (xterm, rxvt, etc). wezterm
* $TERM *inside* tmux (`echo $TERM`). tmux-256color
* $TERM *outside* tmux (`echo $TERM`). xterm-256color
* Logs from tmux (`tmux kill-server; tmux -vv new`). probably irrelevant

cc: @mgrant0

### LINKED ISSUE #5132: Scrollbar overlay mode for width=0
I started using it but the reflowing that can not be prevented because of a width of at least one column is layouted messes up rich full width terminal history pretty badly. I briefly looked at it and decided to not touch that code at this time (or fire up an LLM agent).. but it looked like a three minute job, kind of. Sorry for not providing a patch.. got a deadline fast approaching 😅

mgrant0:
This branch was cherry picked from the screen_redraw_scrollbar_autohide branch.

This branch removes the pane-scrollbars-autohide on/off option and moves this into the pane-scrollbars option:

      - off: no scrollbars
      - on: scrollbars always shown, reserved width
      - modal: scrollbars only in modes, overlay + auto-hide
      - auto-hide: scrollbars generally available, overlay + auto-hide

 Remove modal scrollbar reflow
      - Change modal scrollbars to stop reserving layout width.
      - Ensure pane-scrollbars=modal no longer shrinks/reflows copy-mode content when scrollbars appear.

The man page is also updated.


nicm:
Applied now, thanks!

PR #5274, opened 2026-06-25
kitty images: render via unicode placeholders (grid-resident)

### LINKED ISSUE #4902: Support kitty image protocol
Leaving this here to collect feedback on whether this change in `-portable` is going to be worth it or not -- as it does in some ways "compete" with sixel.

Kitty's image protocol allows for a kitty-aware terminal to render images directly to it.  Terminals such as `kitty`, `ghostty`, and probably others all support this, but such images cannot be rendered inside tmux as tmux isn't aware of this.

I've created a proof-of-concept for this on the `ta/kitty-img` branch in the main `tmux.git` repo.

There's still things to do, but it seems mostly functional from what I can tell.  For example, with the image below:

Image

meisbokai:
## kitty images: render via unicode placeholders (grid-resident)

Ref #4902. Replaces the broken terminal-overlay approach for kitty graphics with the **Unicode placeholder protocol**, making images grid-resident text the same way sixel is handled. This is what makes images actually work through tmux: clip to their pane, scroll with the text, and coexist — none of which the overlay approach could do.

### How it works

For each captured kitty APC, tmux:

1. Transmits the image data **once** to each kitty client with the transmit-only action **`a=t`** (no placement), followed by **`a=p,U=1`** to create an invisible *virtual* placement that defines the render rectangle. `a=t` (not `a=T`) is essential: `a=T` also creates a real placement at the cursor that ghosts when the placeholder cells later move on redraw. Per the graphics-protocol spec, the real on-screen image is driven solely by the placeholder cells.
2. Writes **`U+10EEEE` placeholder cells** into the pane grid spanning `cols × rows`, each carrying the image id in its foreground colour and the row in a combining diacritic. Because these are ordinary cells, tmux's existing redraw/scroll/clip machinery handles everything.

### Root-cause bugs found and fixed along the way

- **Transmit action.** `a=T` left a ghost placement; switched to `a=t` + `a=p,U=1`.
- **Locale-dependent encoding.** `U+10EEEE` and the diacritics were built via `wctomb`/`utf8_fromwc`, which silently fail outside a UTF-8 locale and produce invisible cells. Now encoded as literal UTF-8 bytes.
- **PNG IHDR recovery.** `b64_pton` was asked to decode the *entire* payload into a 24-byte buffer and returned `-1`, so PNGs sent without `s=`/`v=` (the common case) fell back to a 10×10 rectangle. Now decodes only the 32-byte prefix.
- **SGR image-id collision.** A plain palette index for the id is emitted via terminfo `setaf`, which for ids **38 and 48** (the SGR extended-colour introducers) produces a broken escape — so every ~38th image silently failed to render. Now uses `COLOUR_FLAG_256` so tmux always emits the explicit `\e[38;5;<id>m` form.
- **The ghost on pane split.** Image rows were terminated with `screen_write_linefeed(.., 1, ..)`, marking every line `GRID_LINE_WRAPPED`. On pane reflow, `grid_reflow` then joined the following line onto each image line, dropping the cells and leaving stale images on screen. Fixed to use a real newline.
- **Aspect/letterbox.** Clipping `cols` and `rows` to the pane independently broke the aspect ratio, so kitty letterboxed images with large borders. Now scales both axes by the same factor.
- **Memory.** Each image was heap-allocated but never freed. Now handed to the per-screen image list (`image_store`, as sixel does), giving it a proper lifetime and bounding it via the global image cap (also fixed: the cap used `==` so it only ever freed one image).

### Verification

Built and tested with an automated harness that drives a real Ghostty (detached tmux session → `ghostty -e tmux attach` → GNOME portal screenshot → pixel analysis). All of the following pass:

- image renders through tmux
- aspect preserved across wide/tall/square sources and **real PNGs**, both with explicit dimensions and via IHDR recovery
- a pane split produces **no duplicated/ghost images** (single image, two-image, and PNG cases)
- `clear` removes the image cleanly
- memory is bounded

### Known limitation

The image data is transmitted to clients attached at the time the image is captured. A client that attaches later (detach/reattach) will not yet see existing images — this needs per-client transmit tracking on redraw and is left as future work.

### Build

```
./configure --enable-kitty-images && make
```

### DIFF
--- a/image-kitty.c
+++ b/image-kitty.c
@@ -18,6 +18,7 @@
 
 #include <sys/types.h>
 
+#include <resolv.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -45,6 +46,11 @@ struct kitty_image {
 	int		 z_index;     /* z=: z-index */
 	char		 compression; /* o=: 'z'=zlib, 0=none */
 	char		 delete_what; /* d=: delete target (used with a=d) */
+
+	/* Set after this image has been transmitted to a kitty terminal so
+	 * we transmit the payload only once and let the terminal retain the
+	 * placement, rather than re-transmitting on every redraw. */
+	int			transmitted;
 
 	/* Cell size at the time of parsing (from the owning window). */
 	u_int		 xpixel;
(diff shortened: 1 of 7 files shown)


nicm:
@ThomasAdam

ThomasAdam:
I've had a quick glance at this.

Seems alright.

@meisbokai How much of this was AI-generated/assisted? Certainly many of the things listed in the description weren't even a problem with my original implementation, so I'm curious how some of these "conclusions" have come about.

meisbokai:
Hi @ThomasAdam ,

Thank you very much for looking at this PR.

The switch to using grid-resident placeholder cells (the sixel approach) is mine, with AI assistance on parts of the implementation. However, after the implementation, I used an AI agent to drive an automated screen-capture test harness against a real Ghostty to stress-test it.

Some of the findings:
 - a=T vs a=t: the transmit-with-placement action left a real placement that ghosted when the placeholder cells moved on redraw — switched to transmit-only.
 - UTF-8 encoding: U+10EEEE and the row/column diacritics encoded via wctomb produced empty (invisible) cells in non-UTF-8 locales — switched to literal UTF-8 bytes.
 - PNG IHDR recovery: applications that omit s=/v= (the common f=100 case) carried no dimensions — recover them from the PNG IHDR.
 - image id 38/48 collision: a plain palette id emitted via terminfo setaf broke for ids 38 and 48 (the SGR extended-colour introducers), so every ~38th image silently failed — switched to COLOUR_FLAG_256 to force the explicit escape.
 - GRID_LINE_WRAPPED reflow: marking image lines as wrapped made grid_reflow join the next line onto them during a pane split, dropping cells and leaving ghosts — terminate each row with a real newline instead.
 - image cap off-by-one: the retention cap used ==, so it only ever freed one image — fixed.

My apologies regarding the description, the issues were not from your original implementation, but rather from the automated tests. Sorry for the trouble. 

Should I reword the PR description?

(P.S. I'm still a junior developer, so I'd genuinely value any feedback on the approach itself, not just the code. Happy to rework anything that's off.)

nicm:
@mgrant0 has some ideas about this I think so something will probably happen with it at some point :-).

mgrant0:
Yup this is ok my radar due next few weeks.

ThomasAdam:
Thank you, @mgrant0 -- I don't have the time at the moment to pick this up, but will happily review the final patch set as I am still interested in this.

mgrant0:
Have a look at this branch: https://github.com/tmux/tmux/tree/4902-image-support

It's my attempt to generalise the image support and should handle both kitty and sixel images everywhere, in multiple attached terminals as well.

It uses kitty's idea of breaking the image into grid size rectangles the size of characters.  It's work in progress.

nicm:
I'm going to close this in favour of @mgrant0's work to do the same for SIXEL as well. @meisbokai thanks for opening this, if you want to test/contribute to the other PRs that would be great.

PR #5276, opened 2026-06-25
fix(master): fixed `resize-pane` command parse regression.

daneofmanythings:
Some invocation styles of `resize-pane` were not compatible with the new argument parsing. A check has been added which looks at arg of a direction flag, if it does not start with a digit (likely a flag), then we look for the arg string. I have also fixed a bug that didnt't correctly track the 'opposite' state used for offsetting floating pane movement.

### DIFF
--- a/arguments.c
+++ b/arguments.c
@@ -186,6 +186,12 @@ args_parse_flag_argument(struct args_value *values, u_int count, char **cause,
 out:
 	s = args_value_as_string(new);
 	log_debug("%s: -%c = %s", __func__, flag, s);
+	if (optional_argument && *s == '-') {
+		args_free_value(new);
+		free(new);
+		new = NULL;
+	}
+
 	args_set(args, flag, new, 0);
 	return (0);
 }
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -18,6 +18,7 @@
 
 #include <sys/types.h>
 
+#include <ctype.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -149,6 +150,8 @@ cmd_resize_pane_exec(struct cmd *self, struct cmdq_item *item)
 		argval = args_get(args, flag);
 		if (argval == NULL)
 			argval = "1";
+		else if (!isdigit(*argval) && args_count(args) > 0)
+			argval = args_string(args, 0);
 
 		adjust = strtonum(argval, INT_MIN, INT_MAX, &errstr);
 		if (errstr != NULL) {
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -148,10 +148,12 @@ cmd_resize_pane_exec(struct cmd *self, struct cmdq_item *item)
 			continue;
 
 		argval = args_get(args, flag);
-		if (argval == NULL)
-			argval = "1";
-		else if (!isdigit(*argval) && args_count(args) > 0)
-			argval = args_string(args, 0);
+		if (argval == NULL) {
+			if (args_count(args) == 0)
+				argval = "1";
+			else
+				argval = args_string(args, 0);
+		}
 
 		adjust = strtonum(argval, INT_MIN, INT_MAX, &errstr);
 		if (errstr != NULL) {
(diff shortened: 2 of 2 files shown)


nicm:
Hmm actually I don't like this. I thought `-L -t` would work but you're right it parses `-t` as an argument to `-L`. I think this will need to be changed in `arguments.c` and that would break eg `-123`...

daneofmanythings:
Yeah, this felt kinda hacky and fragile. I'm not familiar with the argument or parsing logic, but I'll take a look!

daneofmanythings:
I am curious though, can you see a way to defeat the logic with the current flag options for resize-pane?

nicm:
No, we need to make it so `L::` treats the next argument as belonging to `L` only if it doesn't start with `-`. We'd need to check what commands with `::` would be affected. It may be enough for it only to work for [A-Za-z], so not numbers.

daneofmanythings:
Okay. I'm thinking the logic would end up as something like this.
```c
if (argval == NULL) {
    if (args_count(args) == 0)
        argval = "1"
    else
        argval = args_string(args, 0)
}
```
There is one other tiny thing. If a user were to specify multiple directional flags and rely on string args for the values, things might get wonky. This is not supported behavior and it doesn't clash with backwards compatibility, so it is probably not worrying about at the moment.

daneofmanythings:
The only two commands affected are `resize-pane` and `move-pane` and both are using "::" on the same flags for the same reason. This feels much better. I am a little concerned that there could be some malformed state which is handled and defaulted to "1" without error, but I can't think of a concrete example.

feiyehua:
Hello, it seems the parser still use `-t` as `-L`'s argument, and didn't reach `cmd_resize_pane_exec` at all. Here the log:
```
1782466329.155986 control_read_callback: /dev/pts/6: resize-pane -L -t "%0" 11; list-windows -F "#{window_id} #{window_layout} #{window_flags} #{window_visible_layout}"
1782466329.156168 yylex_token: end at WS
1782466329.156232 yylex_token: resize-pane
1782466329.156280 yylex_token: end at WS
1782466329.156319 yylex_token: -L
1782466329.156355 yylex_token: end at WS
1782466329.156389 yylex_token: -t
1782466329.156424 yylex_token: end at WS
1782466329.156488 yylex_token: %0
1782466329.156524 yylex_token: end at ;
1782466329.156558 yylex_token: 11
1782466329.156596 yylex_token: end at WS
1782466329.156630 yylex_token: list-windows
1782466329.156664 yylex_token: end at WS
1782466329.156697 yylex_token: -F
1782466329.156738 yylex_token: end at EOF
1782466329.156776 yylex_token: #{window_id} #{window_layout} #{window_flags} #{window_visible_layout}
1782466329.156816 cmd_parse_build_commands 0:0: resize-pane
1782466329.156851 cmd_parse_build_commands 0:1: -L
1782466329.156886 cmd_parse_build_commands 0:2: -t
1782466329.156920 cmd_parse_build_commands 0:3: %0
1782466329.156954 cmd_parse_build_commands 0:4: 11
1782466329.156988 cmd_parse_build_commands 1:0: list-windows
1782466329.157022 cmd_parse_build_commands 1:1: -F
1782466329.157058 cmd_parse_build_commands 1:2: #{window_id} #{window_layout} #{window_flags} #{window_visible_layout}
1782466329.157105 args_parse_flags: next -L
1782466329.157140 args_parse_flag_argument: -L = -t
1782466329.157174 args_parse_flags: next %0
1782466329.157208 args_parse: flags end at 3 of 5
1782466329.157243 args_parse: 3 = %0 (type STRING)
1782466329.157276 args_parse: 4 = 11 (type STRING)
1782466329.157322 cmdq_append </dev/pts/6>: [control_error/0x6047b07351e0]
1782466329.157360 cmdq_next <global>: empty
1782466329.157395 cmdq_next </dev/pts/6>: enter
1782466329.157430 cmdq_next </dev/pts/6>: [control_error/0x6047b07351e0] (1), flags 0
1782466329.157477 control_vwrite: /dev/pts/6: writing line: %begin 1782466329 389 1
1782466329.157519 control_vwrite: /dev/pts/6: writing line: parse error: command resize-pane: too many arguments (need at most 1)
```

PR #5279, opened 2026-06-26
Enlarge style_tostring buffer to prevent fatal overflow on large styles

### LINKED ISSUE #5278: style_tostring(): server aborts ("fatal: xsnprintf: overflow") for styles over 256 bytes
### Issue description

Rendering a status line whose `#[...]` style serialises to more than 256 bytes aborts the whole server (all sessions lost). `style_tostring()` (style.c) formats into a fixed `static char s[256]` with `xsnprintf()`, which is **fatal on overflow** (`fatalx("xsnprintf: overflow")`) rather than truncating. It runs on every render — it's an argument to `log_debug()` in `format_draw()`, so it's evaluated regardless of log level.

### To reproduce

```sh
tmux -L bug new-session -d 'sleep 60'
tmux -L bug set -g status-right '#[fg=colour1,bg=colour2,us=colour3,acs,bright,dim,underscore,blink,reverse,hidden,italics,strikethrough,double-underscore,curly-underscore,dotted-underscore,dashed-underscore,overline,range=user|aaaaaaaaaaaaaaaa,align=absolute-centre,list=on,fill=colour200,width=4294967295,pad=4294967295]X'
tmux -L bug attach     # status renders -> "lost server"
```

These are all valid attributes; the style serialises to ~290 bytes. With `-v` the server log ends with `fatal: xsnprintf: overflow`.

### Expected

A long-but-valid style should render (truncated if needed), not abort the server.

### Environment

- tmux 3.6a and master (`next-3.7`)
- macOS (darwin); not platform-specific — the path is in `style.c`.

PR to follow.

angerman:
Fixes #5278.

`style_tostring()` formats into a fixed 256-byte buffer with `xsnprintf()`, which is fatal on overflow, and it runs on every status redraw (it is an argument to `log_debug()` in `format_draw()`). A style that sets enough attributes serialises to more than 256 bytes and aborts the whole server.

Enlarge the buffer to 1024 — large enough for the largest possible style. Every serialised component is bounded: the attribute list is ~155 bytes, `range=user|…` by `range_string[16]`, the colours by `colour_tostring`, and the numeric fields by `%u`, summing to ~550.

The reproducer from the issue no longer crashes; built warning-clean.

### DIFF
--- a/style.c
+++ b/style.c
@@ -285,7 +285,13 @@ style_tostring(struct style *sy)
 	struct grid_cell	*gc = &sy->gc;
 	int			 off = 0;
 	const char		*comma = "", *tmp = "";
-	static char		 s[256];
+	/*
+	 * A style with many attributes set serialises to well over 256 bytes
+	 * (the attribute list alone is ~155, plus colours, range, width, pad
+	 * and so on). xsnprintf() below is fatal on overflow, so size this for
+	 * the largest possible style.
+	 */
+	static char		 s[1024];
 	char			 b[21];
 
 	*s = '\0';


nicm:

PR #5280, opened 2026-06-26
Add a link= style attribute for OSC 8 hyperlinks in formats

angerman:
Builds on #5279 (its commit is the first one here; this PR shows both commits until #5279 merges, then reduces to just the `link=` commit).

Related to #911 / #2563, which added OSC 8 hyperlinks for pane content. This reuses that infrastructure — the per-screen hyperlink set, the grid cell link ID, and the `hyperlinks` terminal feature — to emit links from formats/styles, which those issues did not cover.

Styled formats had no way to emit a clickable OSC 8 hyperlink: a raw escape sequence in a format is mangled by format expansion, and `#[...]` had no link attribute. This adds `#[link=URI]`, ended by `#[nolink]`, an empty `#[link=]`, or `#[default]`. It applies to any format drawn with styles, so it works throughout the status line (`status-left`, `status-right`, the window list, `status-format`) and in other styled surfaces such as `pane-border-format`.

`format_draw()` resolves the URI into the target screen's hyperlink set and stores the resulting ID on the drawn cells. `tty_draw_line()` already emits the OSC 8 sequence for any cell carrying a link, gated by the existing `hyperlinks` terminal feature, so the output path is unchanged. The URI doubles as the internal ID, so repeated links share one entry and the ID stays stable across redraws.

Emitted only with `set -ga terminal-features '*:hyperlinks'` (same gate as pane hyperlinks). Application-set window/pane titles are `clean_name`'d (`#` → `_`), so a program in a pane cannot inject `#[link=]` via `#{pane_title}` / `#W` — the same protection that already covers `#[fg=]`.

Documented in `tmux.1` and `CHANGES`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,10 @@
 CHANGES FROM 3.6b TO 3.7
 
+* Add a link= style attribute (with nolink to end it) which makes styled text
+  an OSC 8 hyperlink, for example in the status line:
+  status-right "#[link=https://example.com]link#[nolink]". Emitted only to
+  terminals with the hyperlinks feature.
+
 * Add floating panes. These are panes which sit above the layout ("tiled
   panes") like popups but unlike popups are not modal and behave like panes (so
   the same escape sequence support). Floating panes are created with the
--- a/format-draw.c
+++ b/format-draw.c
@@ -723,7 +723,7 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 	struct grid_cell	 gc, current_default, base_default;
 	struct style		 sy, saved_sy;
 	struct utf8_data	*ud = &sy.gc.data;
-	const char		*cp, *end;
+	const char		*cp, *end, *link_uri;
 	enum utf8_state		 more;
 	char			*tmp;
 	struct format_range	*fr = NULL, *fr1;
@@ -837,6 +837,21 @@ format_draw(struct screen_write_ctx *octx, const struct grid_cell *base,
 			sy.gc.bg = base->bg;
 			sy.gc.fg = base->fg;
 		}
+
+		/*
+		 * Resolve the style's #[link=...] URI into a hyperlink ID in
+		 * the target screen's set and store it on the cell, so the
+		 * cells drawn below carry an OSC 8 link (emitted later by
+		 * tty_draw_line). The URI doubles as the internal ID so
+		 * repeated links share one entry and the ID stays stable across
+		 * redraws (no spurious status redraws from grid_compare).
+		 */
+		link_uri = style_link(&sy);
+		if (link_uri != NULL && os->hyperlinks != NULL)
+			sy.gc.link = hyperlinks_put(os->hyperlinks, link_uri,
+			    link_uri);
+		else
+			sy.gc.link = 0;
 
 		/* If this style has a fill colour, store it for later. */
 		if (sy.fill != 8)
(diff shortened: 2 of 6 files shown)


nicm:
Idea is OK, thanks, but I don't want to add 256 bytes to `struct style`...

nicm:
I wonder if it could hold a `link_id`. Maybe we need to have a global `struct hyperlinks` somewhere to hold the ones that belong to a style?

angerman:
Thanks! Reworked to avoid growing `struct style`: it now stores a `u_int` ID into a global `style_hyperlinks` set (in style.c) instead of the URI itself. `style_parse()` puts the URI there and keeps the ID; `format_draw()` resolves it back and puts it into the target screen's set as before; a small `style_link()` helper does the lookup (and `style_tostring()` uses it). `struct style` grows by 4 bytes instead of 256.

The set shares the existing global hyperlink cap, and the URI is its own internal ID so identical links dedup to one entry.

Now that #5279 has landed I've rebased onto master, so this is a single commit.

nicm:
Looks good, thanks, but `uri` in `hyperlinks_uri` has no length limit, so a long URI can trigger the bug you just fixed in `style_tostring`. I think we need to either set a limit on the allowed length of a URI (I would make it 1024 bytes and increase the buffer in `style_tostring` to 2048) or change `style_tostring` to allocate the string. I don't really mind which.

How are you using this feature?

angerman:
Went with the limit + bigger buffer: `style_tostring()` now caps the serialised URI at 1024 (`%.1024s`) and uses a 2048-byte buffer, so it can't overflow regardless of URI length — without relying on the parser's token limit.

I'm using it to make an item in `status-right` a clickable link.

nicm:
No, I don't like that, I meant cap the URI length in `hyperlinks.c`, so they are not added at all, not truncate them in `style_tostring`. If we do that then a style can't be converted back from a string to a style.

angerman:
Done — `hyperlinks_put()` now rejects URIs longer than 1024 bytes (`MAX_HYPERLINK_URI`) so they are never stored, and reverted the `style_tostring()` truncation (it just uses a 2048-byte buffer). No truncation, so a style still round-trips through a string.

nicm:
I have applied the `hyperlinks.c` part of this with some minor changes.

The rest looks good but `nolink` doesn't seem to work for me. I do this:
~~~~
tmux set -g status-right '#[link=http://example.com]example#[nolink] foo'
~~~~
But the terminal is highlighting the entire `status-right` as a link, and if I look at what tmux is emitting:
~~~~
1782752360.210393 /dev/ttyp3: \033]8;id=tmux2;http://example.com\033\\
1782752360.210407 /dev/ttyp3: \033[59C
1782752360.210419 /dev/ttyp3: example foo
1782752360.210438 /dev/ttyp3: \033]8;;\033\\
~~~~
The link is not being reset until after the `foo` rather than before.

I had a quick look but I can't see what is causing this. Do you see this as well?

nicm:
If I do `#[link=]` instead of `#[nolink]` it works - I think `nolink` should just be inside the `else if (end > 2 && strncasecmp(tmp, "no", 2) == 0)` case.

nicm:
Yes, that was it. I moved the `nolink` check up a bit and it is fine.

nicm:
I've applied this to OpenBSD now with the fix above and a few other trivial changes, will be in GitHub later. Thanks!

PR #5283, opened 2026-06-26
fix(master): invalidated scene after changing z_indexes

daneofmanythings:
Most noticeable when using `select-pane` targeting a floating pane that is obscured by one higher in the z_index.


nicm:
Applied, thanks!

PR #5284, opened 2026-06-26
refactor(master): Pulled out logic to swap a node into the place of a leaf.

daneofmanythings:
This functionality will be used all over in the future. Pulling it out now. I'm not precious about the naming; can be changed to whatever is best. Also tiny bugfix.


nicm:
Applied, thanks!

PR #5285, opened 2026-06-27
refactor(master): Pulled logic out of `layout_split_pane` for reuse, and added `layout_insert_tile`

daneofmanythings:
`layout_insert_tile` is not being used in this PR, but it is there to show why the logic was pulled out of `layout_split_pane`.

`layout_insert_tile` is the inverse function to `layout_remove_tile` and is used to place cells back into tiled layouts.


nicm:
Applied, thanks!

Issue #5282, opened 2026-06-26
split-window and floating panes


nicm:
The fact that `C-b %` and `C-b "` just show an error when used in a floating pane is useless and annoying.

I think they should mean "make me a new floating pane in a usable position near my current one", so `split-window` in a floating pane should make a new floating pane:

- `-h` make a pane on the right, or left with `-b` as well;
- `-v` make a pane below, or right with `-b` as well;

As far as size goes I would have it use some simple heuristics to pick the best size, maybe some thing like:

- Start with a pane the same size as current one.
- If it won't fit in one direction or the other, make it smaller in that direction.
- Unless that size is getting some value of silly (maybe 25% of the current pane size?), in which case make it at least that size and and move the position so it  remains in the window, allowing the old and new pane to overlap.

We could *maybe* have `-f` make it full width or full height.

This way when I press `%` or `"` to get a new pane - for which I have years of muscle memory - it gives me something that will be immediately usable at least some of the time. Of course if I don't like the size or position or I want it tiled I'll have to change it, but if I just want to for example run a quick command and copy something then if it is immediately usable in even 50% of cases that will save a lot of time.

Of course `new-pane` could accept these flags too, but the default bindings (and the ones people have modified) for are for `split-window` and it is better to make them work usefully than make everyone change them.

daneofmanythings:
#5286 is a quick fix to get better behavior when using keybinds.

I am thinking that the default behavior for spawning floating panes should be the heuristic approach with the cascading as a fallback for when there is no more space. The next PR will be an implementation of that. Then fine tuning can be done with how it interacts with flags for better control.

daneofmanythings:
Finished as of #5338.

PR #5286, opened 2026-06-27
feat(master): Better `split-window` behavior

daneofmanythings:
Part 1 of #5282.

Also includes updates to documentation (99% of the PR).


nicm:
Applied, thanks!

PR #5288, opened 2026-06-27
clean_name: guard against NULL forbid

### LINKED ISSUE #5287: SIGSEGV in clean_name via cmd-break-pane -n (NULL forbid deref)
### Issue description

`break-pane -n "name"` on a 2+ pane window crashes the tmux server with
SIGSEGV in `clean_name()` at `tmux.c:293` (the `strchr(forbid, *cp)`
call) because `cmd-break-pane.c` passes `0` (NULL) as the `forbid`
argument to `window_set_name()` on the `-n` branch.

### Required information

* `tmux version` (`tmux -V`): `next-3.7` (built from master at
  `91e30f4f Merge branch 'obsd-master'`)
* Platform (`uname -sp`): `Darwin arm` (macOS 26.5.1, Apple Silicon)
* Terminal in use: `kitty` 0.41.0 (also reproduced in Alacritty, Ghostty, iTerm2)
* `$TERM` inside tmux: `tmux-256color`
* `$TERM` outside tmux: `xterm-kitty`
* The same crash is reachable on stock `tmux 3.7` once the unrelated
  commit `30e133d3` ("Fix default window check in break-pane", from
  Dane Jensen) is applied.

### Steps to reproduce

1. Build tmux from master:
   ```
   git clone https://github.com/tmux/tmux
   cd tmux
   ./autogen.sh && ./configure --enable-debug && make
   ```
2. Start a fresh server and a 2-pane window:
   ```
   ./tmux -Ltest -f/dev/null new-session -d -s test 'sleep 999'
   ./tmux -Ltest split-window -v
   ```
3. Run `break-pane -n "anything"`:
   ```
   ./tmux -Ltest break-pane -n "anything"
   server exited unexpectedly
   ```

### Crash trace (macOS arm64, master @ 91e30f4f)

```
Exception: EXC_BAD_ACCESS (SIGSEGV)
Subtype:   KERN_INVALID_ADDRESS at 0x0000000000000000

Thread 0 Crashed:
0  _platform_strchr     +12
1  clean_name           +96    tmux.c:293
2  window_set_name      +28    window.c:426
3  cmd_break_pane_exec  +1080  cmd-break-pane.c:168
4  cmdq_fire_command    +32    cmd-queue.c:649
5  cmdq_next            +1436  cmd-queue.c:774
6  server_loop          +80
7  proc_loop            +68
8  server_start         +568
9  client_main          +1040
10 main                 +2032
```

The faulting source line is `tmux.c:293`:

```c
for (cp = copy; *cp != '\0'; cp++) {
    if (strchr(forbid, *cp) != NULL)   // forbid is NULL -> SIGSEGV
        *cp = '_';
}
```

and the caller path is `cmd-break-pane.c:168` (`window_set_name(w, name, 0)`,
where `0` is NULL).

### Server log (-vv)

The server log ends right before the crash with the incoming command
(`message: client-XXXX command: break-pane -n myname`) but no further
lines — the SIGSEGV happens before any further log output. Excerpt of
the last lines before the crash:

```
1782541523.366914 cmd_parse_build_commands: break-pane -n myname
1782541523.366922 cmdq_get_command: [break-pane/0x...] group 2935
1782541523.366945 cmdq_next <client-XXXX>: [break-pane/0x...] (0), flags 0
1782541523.366950 message: client-XXXX command: break-pane -n myname
1782541523.366955 cmdq_fire_command <client-XXXX>: (2935) break-pane -n myname
[server terminated by SIGSEGV here]
```

### Root cause

`clean_name()` was introduced in commit `d339ab51` ("Sanitize pane titles
and window and session names more consistently and strictly", 2026-04-22).
Its loop dereferences `forbid` unconditionally. Every caller added in
that commit (`cmd-new-session.c`, `cmd-new-window.c`,
`cmd-rename-session.c`, `cmd-rename-window.c`, `screen.c`, `paste.c`,
`names.c`, `spawn.c`, `input.c`) passes a non-NULL forbid string
(`"#:."`, `"#"`, `WINDOW_NAME_FORBID`, `WINDOW_NAME_FORBID_EXT`,
`""`, etc.).

Then in commit `eb653314` / `64e83caf` ("Be more strict about what
names and titles we allow and reject them immediately...",
2026-06-15), `cmd-break-pane.c` was rewritten and the only caller that
passes a NULL `forbid` was introduced:

```c
window_set_name(w, name, 0);   /* cmd-break-pane.c:167 */
```

The `0` is almost certainly meant as "no transformation" (matching the
pre-3.7 2-arg `window_set_name(w, name)` behavior), but the new
3-arg API interprets `0` as a NULL pointer and `clean_name` doesn't
handle NULL.

### Affected versions

* `master` @ `91e30f4f` (current tip) — crash reachable via
  `break-pane -n "x"` on a 2+ pane window.
* `next-3.7` / `3.7-rc` — same as master.
* `tmux 3.7` (released) — bug 1 in `cmd-break-pane.c` (`if (name != NULL)`
  inversion) masks the NULL path for `break-pane -n` (the `-n` value is
  silently ignored). Once bug 1 is fixed (already in `30e133d3`), bug 2
  fires immediately.

### Suggested fix

Add a NULL guard inside `clean_name()` so any future caller can pass
NULL to mean "no rewrites":

```diff
--- a/tmux.c
+++ b/tmux.c
@@ -289,9 +289,11 @@ clean_name(const char *name, const char* forbid)
        if (*name == '\0' || !utf8_isvalid(name))
                return (NULL);
        copy = xstrdup(name);
-       for (cp = copy; *cp != '\0'; cp++) {
-               if (strchr(forbid, *cp) != NULL)
-                       *cp = '_';
+       if (forbid != NULL) {
+               for (cp = copy; *cp != '\0'; cp++) {
+                       if (strchr(forbid, *cp) != NULL)
+                               *cp = '_';
+               }
        }
        utf8_stravis(&new_name, copy, VIS_OCTAL|VIS_CSTYLE|VIS_TAB|VIS_NL);
        free(copy);
```

This treats NULL `forbid` the same as the empty-string `forbid` (which
`paste.c` and `screen.c` already pass — see `paste.c:222,290` and
`screen.c:255,272`). Existing callers that pass non-NULL forbid
strings are unaffected.

A PR with this patch is at https://github.com/tmux/tmux/pull/5288.

gilbertwong96:
## Summary

`clean_name()` dereferences `forbid` without a NULL check. A NULL
`forbid` is reachable today via `cmd-break-pane.c`, which passes `0`
(NULL) to `window_set_name()` on the `-n` branch.

Triggers: `break-pane -n "anything"` on a 2+ pane window crashes the
server.

Fixes #5287

## Repro

```
$ ./tmux -Ltest -f/dev/null new-session -d -s test 'sleep 999'
$ ./tmux -Ltest split-window -v
$ ./tmux -Ltest break-pane -n "anything"
server exited unexpectedly
```

## Crash trace (macOS arm64, master @ 91e30f4f)

```
0  _platform_strchr     +12
1  clean_name           +96    tmux.c:293
2  window_set_name      +28    window.c:426
3  cmd_break_pane_exec  +1080  cmd-break-pane.c:168
4  cmdq_fire_command    +32
5  cmdq_next            +1436
```

`tmux.c:293` is `if (strchr(forbid, *cp) != NULL)` inside the
unconditional `for` loop.

## Patch

```diff
--- a/tmux.c
+++ b/tmux.c
@@ -289,9 +289,11 @@ clean_name(const char *name, const char* forbid)
        if (*name == '\0' || !utf8_isvalid(name))
                return (NULL);
        copy = xstrdup(name);
-       for (cp = copy; *cp != '\0'; cp++) {
-               if (strchr(forbid, *cp) != NULL)
-                       *cp = '_';
+       if (forbid != NULL) {
+               for (cp = copy; *cp != '\0'; cp++) {
+                       if (strchr(forbid, *cp) != NULL)
+                               *cp = '_';
+               }
        }
        utf8_stravis(&new_name, copy, VIS_OCTAL|VIS_CSTYLE|VIS_TAB|VIS_NL);
        free(copy);
```

The empty-string `forbid` already means "no rewrites" (used by
`paste.c` and `screen.c`); this just extends the same convention to
NULL, which is what `cmd-break-pane.c:167` appears to have intended
with `window_set_name(w, name, 0)`.

## Background

`clean_name()` was added in `d339ab51` (2026-04-22). Every caller in
that commit passes a non-NULL forbid, so the function never crashed.
The `cmd-break-pane.c` rewrite in `eb653314` / `64e83caf`
(2026-06-15) introduced the NULL caller via
`window_set_name(w, name, 0)`, which has been latent since.

On the released `tmux 3.7` the inverted condition in
`cmd-break-pane.c` (`if (name != NULL)` should be `if (name == NULL)`;
fixed on master in `30e133d3`) routes `break-pane -n "x"` through the
wrong branch, so the NULL forbid path isn't hit there. As soon as
`30e133d3` is in, this crash fires.

## Notes

* Behaviour-preserving for existing callers — only callers passing
  `forbid == NULL` change behavior, and there are no other such
  callers in the tree.
* The 5-line diff is intentionally the smallest defensible fix; an
  alternative is to change `cmd-break-pane.c:167` to pass
  `WINDOW_NAME_FORBID` instead of `0`, but that papers over the
  unsafe API.

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -289,9 +289,11 @@ clean_name(const char *name, const char* forbid)
 	if (*name == '\0' || !utf8_isvalid(name))
 		return (NULL);
 	copy = xstrdup(name);
-	for (cp = copy; *cp != '\0'; cp++) {
-		if (strchr(forbid, *cp) != NULL)
-			*cp = '_';
+	if (forbid != NULL) {
+		for (cp = copy; *cp != '\0'; cp++) {
+			if (strchr(forbid, *cp) != NULL)
+				*cp = '_';
+		}
 	}
 	utf8_stravis(&new_name, copy, VIS_OCTAL|VIS_CSTYLE|VIS_TAB|VIS_NL);
 	free(copy);


nicm:
You explained the right problem and then gave the wrong fix, checking for `NULL` in `check_name` is just hiding the bug. This is already fixed in `master` so this fix needs to go into 3.7. 

I'll add it to the list for 3.7b should we release one.

nicm:
I have made `release_3.7a` branch with this fixed, please let me know if you use it and still see problems.

I'll release 3.7a but I'll to see if there is anything else first. Thanks!

gilbertwong96:
Thank you very much., I just learned these two commit: 30e133d36aa158850f1b3961345c1515d4abf0fa and d0bf3bf06d4d007f62348671e6aab00bb15cfb7e. Will appreciate if you make a fix on 3.7.

PR #5293, opened 2026-06-28
feat(master): Floating panes can be automatically tiled with `join-pane`

daneofmanythings:
When `join-pane` is invoked on a floating `src-pane` and no `dst-pane` is specified, the pane is put back into the layout in its old position.

This minimally gets the feature working and it is not comprehensive. If there is desire, `join-pane` can be more thoroughly reworked so that floating dimensions are saved across ALL tiling operations, not just auto-tiling. This will be a larger diff and is not necessary at this time. It's on my radar for if an issue is raised in the future.

I also edited the error reporting semantics in remove/insert tile. I don't know why I had them as "true/false" instead of "success/failure".

### DIFF
--- a/cmd-join-pane.c
+++ b/cmd-join-pane.c
@@ -364,6 +364,44 @@ cmd_join_pane_zindex(struct cmdq_item *item, struct winlink *wl,
 }
 
 static enum cmd_retval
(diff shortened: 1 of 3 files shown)


nicm:
Applied now, thanks!

PR #5304, opened 2026-06-30
Redraw pane after disabling sync updates

japinli:
Commit 65a032b2 reordered the 2026 case but failed to carry over the PANE_REDRAW flag due to copy-paste.

For example:

``` bash
tmux -Ltest -f/dev/null new
```

Then in the tmux session, use `nvim --clean cfg.c`, I got the following output:

image

### DIFF
--- a/screen-write.c
+++ b/screen-write.c
@@ -1011,6 +1011,8 @@ screen_write_stop_sync(struct window_pane *wp)
 		evtimer_del(&wp->sync_timer);
 	wp->base.mode &= ~MODE_SYNC;
 
+	wp->flags |= PANE_REDRAW;
+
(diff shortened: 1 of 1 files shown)


nicm:
Whoops I committed this accidentally. I think this should be in `screen_write_stop_sync` anyway, do you want to move it there?

japinli:
> Whoops I committed this accidentally. I think this should be in `screen_write_stop_sync` anyway, do you want to move it there?

I've updated this as you suggested. Please take a look.

sluongng:
I ran into this exact issue this morning. Would love to see the fix merged

nicm:

PR #5309, opened 2026-07-01
Update tty-features.c

XhstormR:
ghostty also support overline features which tmux already implemented. https://github.com/ghostty-org/ghostty/issues/2314


nicm:

PR #5313, opened 2026-07-01
Vertical Status Column

Jon-Schneider:


nicm:
Thanks - not that I am ungrateful, but can you put some text in your PR (preferably in your own words) explaining what you are doing, what the design is, tradeoffs, any new options or commands etc, rather than just dumping a 3000 line diff. This is too big for everything to just be obvious.

Jon-Schneider:
I'm sorry, this was an accident. It's a personal fork I wasn't intending to push up, at least not without reviewing the entire thing in detail (which I won't have time to do for a few weeks) because while it's seemingly functional it's primarily agent-coded.

I love the project, and again I'm very sorry!

PR #5316, opened 2026-07-02
cmd-split-window: fix layout cell double free on spawn failure

Uzaaft:
This was tricky for me to to fully understand as I do not fully grasp the tmux internals, but I try to explain the finding as good as possible inside the git msg. 

Stumbled upon this while wiring in libghostty-vt into tmux and replace the vt mechanism the repo has. I'm planning to maintain it as a fork but if you'd be interested in any efforts to upstream it I'd be happy to do it!

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -167,6 +167,12 @@ cmd_split_window_exec(struct cmd *self, struct cmdq_item *item)
 	if ((new_wp = spawn_pane(&sc, &cause)) == NULL) {
 		cmdq_error(item, "create pane failed: %s", cause);
 		free(cause);
+		/*
+		 * On failure spawn_pane has already destroyed the layout cell
+		 * (via layout_close_pane in its fork-failure path); clear lc so
+		 * the fail path below does not destroy it a second time.
+		 */
+		lc = NULL;
 		goto fail;
 	}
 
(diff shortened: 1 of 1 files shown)


Uzaaft:
FYI I'm trying to narrow this down to a minimal repro that can consistently trigger this.

nicm:
Makes sense, but there is an omission - I think the `fail` path should be destroying `new_wp` as well.

It might be nicer if all the style etc validation was hoisted up to before the cell is created, where `-B` is checked.

I don't have time to look at this more now, but I will look again later.

If you are making changes and you find bugs do report them, but we are not going to change the parser.

Uzaaft:
> If you are making changes and you find bugs do report them, but we are not going to change the parser.

I expected this, but it was worth checking in with you guys 🫡

Uzaaft:
> Makes sense, but there is an omission - I think the `fail` path should be destroying `new_wp` as well.

Addressed this point about your feedback. 

Re style validation: Looking at it, clueless as to how it *should* look atp.

nicm:
Looks good, I have applied this to OpenBSD now, will be in GitHub later. Thanks!

I took a quick look at moving the various argument checks higher up but it is not as simple as I thought so I just left it for now.

PR #5318, opened 2026-07-02
feat(master): `swap-pane` enabled for floating panes.

daneofmanythings:
I have removed the guard to prevent floats from being the destination or source of `swap-pane`. This was unblocked by #5247. Additionally, the `-U` and `-D` flags ignore panes that are not tiled.

### DIFF
--- a/cmd-swap-pane.c
+++ b/cmd-swap-pane.c
@@ -80,6 +80,10 @@ cmd_swap_pane_exec(struct cmd *self, struct cmdq_item *item)
 		server_redraw_window(dst_w);
 
 	if (args_has(args, 'D')) {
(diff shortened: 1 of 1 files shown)


nicm:
I think `-U` and `-D` working in floating panes is confusing, maybe they should be an error or ignored? (And continue to be hidden on the menu.)

daneofmanythings:
Oh right. It should definitely error out, I forgot to add that.

daneofmanythings:
Okay, the check is added and the previous pane menu format excludes floats from swapping up/down already.

nicm:
Applied now, thanks!

PR #5321, opened 2026-07-03
feat(master): Added pane menu and border buttons for tile/float operations.

daneofmanythings:


nicm:
Applied, thanks. I also turned "Swap Marked" back on.

PR #5322, opened 2026-07-03
Dirty whole screen on clear in synchronized update

twilligon:
Since 565db46, ending a synchronized update redraws only the lines that were marked dirty while it was active. screen_write_clearscreen clears the entire view but gives screen_write_should_draw_lines the same range as clearendofscreen (s->cy, sy - s->cy), so after the update ends, only lines from the cursor down are marked, and any lines *above* the cursor keep their old content on the terminal. This is invisible with a common `\e[H\e[2J` because the cursor is already at the top left, but lines stay stale with the cursor elsewhere. Dirty the screen's full range instead.


nicm:

PR #5323, opened 2026-07-04
fix(master): logic bug causing restored geometries to slowly drift.

daneofmanythings:
Restoring the floating geometry of a pane will slightly drift because of a missing check in `layout_floating_args_parse`. The checks have been added.


nicm:
Applied, thanks.

PR #5324, opened 2026-07-04
refactor(master): Added `struct layout_geometry` to hold layout values.

daneofmanythings:
I have been wanting to group up the layout size and positioning values for a while. I have implemented what I was thinking to see how it feels. We need to track at least 2 geometries for layout cells (current geometry with saved floating) and grouping them up feels better to me. It adds a little namespacing that was previously handled by prefixing fields with things like `saved_*`. I prefer this approach because it also simplifies function signatures and allows us to return all geometry information from functions. I am not dynamically allocating this anywhere as that doesn't make sense, and in one function i return the struct as a value, treating it like returning four numbers. I have not fully incorporated this method yet. I'd like to get some thoughts on it first before the diff gets incomprehensible. When fully integrated, `layout_set_size` would become obsolete, as we would be able to assign the geometry directly.

### DIFF
--- a/cmd-break-pane.c
+++ b/cmd-break-pane.c
@@ -51,8 +51,6 @@ cmd_break_pane_float(struct cmdq_item *item, struct args *args,
     struct window *w, struct window_pane *wp)
 {
 	struct layout_cell	*lc = wp->layout_cell;
-	u_int			 sx = lc->saved_sx, sy = lc->saved_sy;
-	int			 ox = lc->saved_xoff, oy = lc->saved_yoff;
 	char			*cause = NULL;
 	enum pane_lines		 lines = window_get_pane_lines(w);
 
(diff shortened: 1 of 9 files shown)


nicm:
Can you fix the conflicts? I did it but I think I broke something and don't have time to work out what...

Also can you please:

- Make  `layout_geometry_init` take a `struct layout_geometry *` and assign the members, instead of returning a struct like that which is a bit ugly.

- Avoid struct assignment, which is something of a C language misfeature, in favour of `memcpy` or explicit assignment.

- Similarly, avoid pass-by-value for structs and use pointers.

daneofmanythings:
Oh of course! I'll make the changes and fix the conflicts. I didn't notice that conflicts needed resolution.

daneofmanythings:
I have implemented changes from your feedback. I also fixed a few miscellaneous bugs.

nicm:
Applied, thanks!

PR #5329, opened 2026-07-05
Feature-PR: "Team mode" that allows multiple cursors and independent windows/panes

gehaxelt:
Hi there, 

first of all a disclaimer: I have a CS background, but I'm not a proficient C/C++ programmer, and without a deep understanding of tmux's codebase, I simply asked Claude Opus to implement a feature in tmux that my colleagues and I were missing for a long time. Thus, I don't expect this PR to be merged as-is, but hope that it can serve as inspiration and a PoC for a "team mode"  feature. 

The idea of "team mode" is that multiple people can join the same tmux session (`tmux new -s teammode` and `tmux a -t teammode`) and have separate cursors and independent navigation between the panes and windows of the session. 

As you can see in the screenshot below, there are 3 different "users" (terminals), all connected to the same session. In the top two terminals both users are in different panes can work independently in the shared session. Bottom right is the third user that is in the second window of the session, independently of the other users first window.  

Screenshot_2026-07-05_13-38-18

To some extent, this mirrors the behavior and functionality offered by screen, where users can work independently in a shared screen session. 

I'd appreciate if someone with better understanding of tmux could have a look and help bring such a feature into upstream tmux :) 

All the best and thanks in advance,
@gehaxelt

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -1,3 +1,23 @@
+# tmux with teammode
+
+This is a fork of tmux that adds **teammode**: a multi-user mode (enabled by
+default) in which several users attached to the same session each work
+independently. Each user gets their own active pane and their own current
+window, so they can type in different panes and navigate to different windows
+at the same time, and each user sees the other users' cursors drawn as coloured
+markers. Set `set -g teammode off` for classic shared behaviour.
+
+See [TEAMMODE.md](../TEAMMODE.md) for details and usage.
+
+> **Disclaimer:** the teammode changes were fully vibe-coded and should be used
+> with care. They work in manual testing, but have not been exhaustively
+> reviewed or tested — do not rely on them in critical setups without your own
+> verification.
+
+The original tmux README follows.
+
+---
+
 # Welcome to tmux!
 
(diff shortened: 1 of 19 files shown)


nicm:
Thanks.

It is confusing to give this a name "team mode" when it looks like it is  just an extension of `-factive-pane` to windows - modes in tmux are already something different.

I think this needs more thought. We already have session groups which allow an independent current window. What happens to them? Do we have two mechanisms for the same thing? I don't like the sound of that very much.

Another thing - and the reason I didn't add something like this when I added `active-pane` - is that I don't like how that works very much. I have more than once been tempted to remove it. And this seems inconsistent because "teammode" is a session property but the window is per-client - `active-pane` is a client property.

So I would like to see a more comprehensive design here. Is there a better way we can do this internally? What is going to happen with grouped sessions (people use them and we probably need something *like* them, although it would be nice if it was better integrated)? Could this (and `active-pane`) be easier to use? How does it interact with targets? In short, what would this look like in an ideal world, taking into account the existing features that cover this.

Issue #5327, opened 2026-07-05
new-pane on a zoomed window crashes the server in 3.7b


junegunn:
### Full disclosure

I noticed the crash while testing floating panes (new-pane from a zoomed pane). This report was fully generated by Claude Code, which refined the reproduction, ran the AddressSanitizer builds, and collected the logs. To be honest, I don't have a full understanding of the details, so please take the analysis with a grain of salt; the reproductions and logs should stand on their own.

### Issue description

Creating a floating pane while the window is zoomed crashes the 3.7b server, taking all sessions with it. No client needs to be attached:

```sh
tmux -f /dev/null -L test new-session -d -x 100 -y 30 -s t
tmux -L test split-window -t t
tmux -L test resize-pane -Z -t t
tmux -L test new-pane -t t -x 40 -y 10 -X 5 -Y 5 'sleep 60'   # server exits
```

Interestingly, running the server with `-v` or `-vv` makes the crash disappear, but the new pane is then created with a corrupt size of 0x0 (visible in the attached server log and in `list-panes`). This and the ASan report below suggest a use-after-free whose visible outcome depends on heap state.

Master (5356c62e) does not visibly crash on this either, but it does not seem to be fixed there; an AddressSanitizer build of master reports a heap-use-after-free on the same reproduction, so it looks latent rather than resolved:

```
ERROR: AddressSanitizer: heap-use-after-free
READ of size 4 at 0x60800000f6b8 thread T0
    #0 layout_fix_panes layout.c:455
    #1 window_set_active_pane window.c:564
    #2 spawn_pane spawn.c
    #3 cmd_split_window_exec cmd-split-window.c:167

freed by thread T0 here:
    #1 layout_free_cell layout.c:106
    #2 window_set_active_pane window.c:564
    #3 spawn_pane spawn.c
    #4 cmd_split_window_exec cmd-split-window.c:167

previously allocated by thread T0 here:
    #1 xmalloc xmalloc.c:33
    #2 layout_get_floating_cell layout.c:1666
    #3 cmd_split_window_exec cmd-split-window.c:133
```

The full report is attached. The ASan build was:

```sh
./configure --enable-utf8proc CFLAGS="-fsanitize=address -g -O1" LDFLAGS="-fsanitize=address"
```

run with `ASAN_OPTIONS=log_path=/tmp/asan-tmux` in the environment of the first client, since the daemonized server's stderr is not visible; the report is then written to `/tmp/asan-tmux.<pid>`.

[...]

PR #5331, opened 2026-07-05
fix(master): Push the zoom when creating a floating pane.

daneofmanythings:
Triage for #5327

### DIFF
--- a/layout.c
+++ b/layout.c
@@ -1663,6 +1663,7 @@ layout_get_floating_cell(struct cmdq_item *item, struct args *args,
 	    cause) != 0)
 		return (NULL);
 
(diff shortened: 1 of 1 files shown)


nicm:
I think it should just unzoom always.

daneofmanythings:
It seemed right to mirror what was in ~~`layout_split_pane`~~ `layout_get_tiled_cell` and localize the change so as not to touch other code paths because this is a temporary fix. Should I make a more global change?

nicm:
No it's ok so long as it works:-).

daneofmanythings:
I believe it does! This is what is done in `spawn.c:spawn_editor:669` as well.

nicm:
Applied, thanks!

PR #5332, opened 2026-07-06
fix(layout): break infinite loop in layout_resize_adjust when no progress

Uzaaft:
The while (change != 0) loop in layout_resize_adjust spins forever when shrinking (change < 0) but all tiled children are at minimum size:
layout_resize_check returns 0 for every child so change never reaches
zero. Track a progress flag across each pass and break when a full iteration makes no progress.

this was found while fuzzing my libghostty-vt fork with bombadil.


nicm:
Applied to OpenBSD now, although I changed `progress` to `changed` which I think is clearer. Will be in GitHub later. Thanks!

PR #5333, opened 2026-07-06
fix(layout): prevent zero-height pane from border subtraction

Uzaaft:
layout_fix_panes subtracts 1 from sy for the pane-status border. When a pane cell is 1 row tall with pane-border-status top, sy becomes 0. A zero-height pane triggers pathological behaviour in mode draw routines (e.g. mode_tree_draw iterating with an underflowed row bound). Only subtract the border when sy > 1.


nicm:

PR #5334, opened 2026-07-06
fix(mode-tree): guard mode_tree_draw against zero height

Uzaaft:
The draw loop break condition (i > mtd->offset + h - 1) underflows to UINT_MAX when h is 0, causing iteration over all line items on a zero-row grid and a 100% CPU spin. Bail out early when height is zero.

I believe this is the last one for today(but no promises 😂)

### DIFF
--- a/mode-tree.c
+++ b/mode-tree.c
@@ -835,6 +835,12 @@ mode_tree_draw(struct mode_tree_data *mtd)
 	if (mtd->line_size == 0)
 		return;
 
+	w = mtd->width;
+	h = mtd->height;
+
+	if (h == 0)
+		return;
+
 	memcpy(&gc0, &grid_default_cell, sizeof gc0);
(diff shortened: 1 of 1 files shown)


nicm:
How do you get `h` to be zero?

Uzaaft:
@nicm Due to PR: #5333,  sy can never reach 0, so mode_tree_draw never sees h == 0. This is purely defensive, to guard against any future path that might produce a 0-height pane through a different mechanism than what was fixed in #5333

Uzaaft:
So admittedly, this can be dropped, but I had already made the change locally and thought why not, in case something in the future adds back the h == 0 case

nicm:
I think there is no harm in this, will look later.

nicm:
Applied to OpenBSD now, although I made it check `w` as well. Will be in GitHub later. Thanks!

PR #5337, opened 2026-07-06
Do not disable and reenable the mouse on every redraw

### LINKED ISSUE #5303: Mouse mode re-asserted (off→on) on every redraw even when unchanged, with re-enable outside DEC 2026 sync brackets
### Environment

```
Darwin arm
tmux 3.7
xterm-256color   # (inside tmux: tmux-256color)
```

- Platform: macOS (Apple Silicon)
- Outer terminal where the flicker is visible: Ghostty (advertises DEC 2026 `Sync`)
- `tmux show -g mouse` → `mouse on`
- Reproduces on **tmux 3.7**, not on **3.6**.

### Description

When a pane emits frames wrapped in DEC 2026 synchronized-output markers (`\e[?2026h … \e[?2026l`), tmux 3.7 honors the boundaries and redraws the host once per frame. On **every** redraw tmux emits a full mouse-mode **disable→enable cycle even though the net mouse mode never changes** (stays on), and the **re-enable lands outside** the synchronized-update brackets.

Root cause, confirmed from source — it's a structural side effect, not a deliberate mouse re-assert:

1. **Disable (inside sync).** `screen_redraw_screen()`:
   - `tty_sync_start(tty)` → `\e[?2026h` (`screen-redraw.c:1667`)
   - `tty_update_mode(tty, 0, NULL)` (`screen-redraw.c:1668`) — forces the mode word to **0**. The intent is cursor-hide; zeroing the *whole* mode word also clears the mouse bits, so this emits `\e[?1006l\e[?1000l\e[?1002l\e[?1003l` as **collateral**.
   - draw, then `tty_sync_end(tty)` → `\e[?2026l` (`screen-redraw.c:1702`)
2. **Re-enable (outside sync).** `server_client_reset_state()` runs immediately after the redraw (`server-client.c:1737`), *after* sync was already closed:
   - re-merges session `mouse on` (`server-client.c:2084-2098`)
   - `tty_update_mode(tty, mode, s)` (`server-client.c:2105`) now sees `changed & ALL_MOUSE_MODES` true (because step 1 zeroed it) and re-emits `\e[?1006l\e[?1000l\e[?1002l\e[?1003l` + `\e[?1006h\e[?1000h\e[?1002h`.

`tty_update_mode` is itself already diff-based (`tty.c:880`, `:888`) and would emit **zero** mouse bytes in steady state — the churn only happens because the redraw path zeroes the mode first and the restore happens in a separate, later `tty_update_mode` outside the brackets.

### Why 3.7-only

Controlled A/B with one emitter (6 frames, 200 ms apart), counting host redraws (`?2026h`) and mouse re-enables (`?1000h`):

| Emitter | host redraws (`?2026h`) | mouse re-enables (`?1000h`) |
|---|---|---|
| frames wrapped in DEC 2026 | 9 | 11 |
| frames not wrapped | 3 | 5 |

3.7 added `DECRQM ?2026` handling, so apps that probe for sync support now wrap each frame, which makes tmux do ~3× as many host redraws — one mouse off→on cycle each. 3.6 didn't answer the probe, apps didn't wrap, redraws coalesced, and the churn was too rare to notice.

### Minimal reproduction (tmux + `printf` only, no application involved)

```sh
script -q /tmp/wire.txt sh -c '
  tmux -Ltest kill-server 2>/dev/null
  tmux -Ltest -f/dev/null new "tmux set -g mouse on; \
    for i in 1 2 3 4 5 6; do printf \"\033[?2026h\033[H\033[2Jframe %d\033[?2026l\" \$i; sleep 0.2; done; \
    sleep 0.3"
'

# Inspect what tmux sent to the host terminal:
cat -v /tmp/wire.txt | grep -oE '\^\[\[\?(2026[hl]|100[0-3][hl]|1006[hl])' | sort | uniq -c
```

Observed (per-frame, verified on `tmux 3.7`): a mouse **disable** cycle `?1006l ?1000l ?1002l ?1003l` followed by **re-enable** `?1006h ?1000h ?1002h`, repeating once per frame even though `mouse` stays `on` throughout. On a `Sync`-capable outer terminal (e.g. Ghostty) the re-enable falls *after* the closing `?2026l`, so a sync-honoring host still observes a real off→on transition.

### Expected

- A redraw should not clear mouse reporting: zero only the bits it means to (cursor), e.g. preserve mouse mode at `screen-redraw.c:1668`.
- Any mouse-mode change that is necessary should stay **inside** the `?2026h…?2026l` brackets so a conforming host never sees the intermediate off state.

### Actual

Every redraw emits a mouse off→on cycle (off inside sync, on outside sync) regardless of whether the mode changed.

### Impact

Terminals that re-evaluate pointer shape on mouse-tracking-mode changes (Ghostty) flicker the pointer arrow↔I-beam on every frame/keystroke. Apple Terminal and iTerm2 get the identical churn without reshaping the pointer, so the visible symptom is terminal-specific — but the redundant per-frame mouse churn and the out-of-bracket restore are tmux behaviors.

### Possible fixes

1. `screen-redraw.c:1668` — preserve input bits, e.g. `tty_update_mode(tty, tty->mode & ALL_MOUSE_MODES, NULL)`, so a *draw* never touches *mouse reporting*; or
2. perform the `reset_state` re-enable inside the redraw's sync brackets so the off→on never escapes to the host.

aymanbagabas:
With the mouse on, tmux sends a full mouse disable then enable to the
client on every redraw, even though the mode never changes. On frequent
redraws (a status line clock, or apps wrapping frames in DEC 2026) it
flickers. Fixes #5303.

redraw_draw() calls tty_update_mode(tty, 0, NULL) to hide the cursor, but
passing 0 clears the whole mode word, so the mouse is turned off too;
reset_state() then turns it back on. The fix keeps every mode except the
cursor:

    tty_update_mode(tty, tty->mode & ~CURSOR_MODES, NULL);

Not 2026 specific: any redraw hits it, 2026 just makes them frequent.
Adds regress/screen-redraw-mouse.sh, which fails without the fix.

### DIFF
--- a/regress/screen-redraw-mouse.sh
+++ b/regress/screen-redraw-mouse.sh
@@ -1,0 +1,51 @@
+#!/bin/sh
+
+# Test that a redraw does not churn the mouse mode.
+#
+# When the mouse is on, redraw_draw() must preserve the input modes and only
+# drop the (visual) cursor while drawing a frame. A buggy redraw turned every
+# mode off (including mouse) and let reset_state turn it back on, so each
+# refresh emitted a full disable/enable of the mouse (ESC[?1000l ... ESC[?1000h)
+# to the client. That flicker is what this test guards against.
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -Ltest -f/dev/null"
+TMUX2="$TEST_TMUX -Ltest2 -f/dev/null"
+$TMUX kill-server 2>/dev/null
+$TMUX2 kill-server 2>/dev/null
+
+TMP=$(mktemp)
+trap "rm -f $TMP; $TMUX kill-server 2>/dev/null; $TMUX2 kill-server 2>/dev/null" \
+	0 1 15
+
+# Inner session with the mouse enabled, attached inside an outer pane so the
+# outer pane captures whatever the inner tmux writes to its client terminal.
+$TMUX2 new -d -x80 -y24 || exit 1
+$TMUX2 set -g mouse on || exit 1
+$TMUX new -d -x80 -y24 "$TMUX2 attach" || exit 1
+sleep 1
+
+# Capture the inner tmux's raw client output, then force several redraws. A
+# buggy redraw disables the mouse every time, so a few attempts make the test
+# reliable without depending on the timing of a single refresh.
+$TMUX pipe-pane -O "cat >$TMP" || exit 1
+sleep 1
+i=0
+while [ $i -lt 5 ]; do
+	$TMUX2 refresh-client || exit 1
+	sleep 0.5
+	i=$((i + 1))
+done
+$TMUX pipe-pane || exit 1
+
+# After startup, a redraw must not disable the mouse (ESC[?1000l).
+if grep -q '\[?1000l' "$TMP"; then
+	echo "[FAIL] redraw churned the mouse mode"
+	exit 1
+fi
+
+[ -n "$VERBOSE" ] && echo "[PASS] redraw preserved the mouse mode"
+exit 0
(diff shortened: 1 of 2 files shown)


nicm:
Nice one, thanks. Applied to OpenBSD now, will be in GitHub later. Thanks!

PR #5338, opened 2026-07-06
feat(master): modified `split-window` behavior on floating panes.

daneofmanythings:
This PR implements differing behavior when using `splitw-window` on a floating pane. The new pane is created adjacent to the target based upon the provided `h/v/b` flags. Both panes are resized to keep the new pane fully visible on screen. This PR will be touched up once #5324 is finished. Let me know if the behavior is acceptable!

### DIFF
--- a/cmd-join-pane.c
+++ b/cmd-join-pane.c
@@ -458,7 +458,14 @@ cmd_join_pane_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	}
 
-	lc = layout_get_tiled_cell(item, args, dst_w, dst_wp, &flags, &cause);
+	if (args_has(args, 'h'))
+		flags |= SPAWN_HORIZONTAL;
+	if (args_has(args, 'b'))
+		flags |= SPAWN_BEFORE;
+	if (args_has(args, 'f'))
+		flags |= SPAWN_FULLSIZE;
+
+	lc = layout_get_tiled_cell(item, args, dst_w, dst_wp, flags, &cause);
 	if (cause != NULL) {
 		cmdq_error(item, "size or position %s", cause);
 		free(cause);
--- a/cmd-join-pane.c
+++ b/cmd-join-pane.c
@@ -458,7 +458,7 @@ cmd_join_pane_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	}
 
-	lc = layout_get_tiled_cell(item, args, dst_w, dst_wp, flags, &cause);
+	lc = layout_get_tiled_cell(item, args, dst_w, dst_wp, &flags, &cause);
 	if (cause != NULL) {
 		cmdq_error(item, "size or position %s", cause);
 		free(cause);
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -84,7 +84,7 @@ cmd_split_window_exec(struct cmd *self, struct cmdq_item *item)
 	struct window_pane	*wp = target->wp, *new_wp = NULL;
 	struct layout_cell	*lc = NULL;
 	struct cmd_find_state	 fs;
-	int			 input, empty, is_floating, newp, flags = 0;
+	int			 input, empty, is_floating, npe, flags;
 	const char		*template, *style, *value;
 	char			*cause = NULL, *cp, *title;
 	const struct options_table_entry *oe;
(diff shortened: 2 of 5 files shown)


nicm:
It looks OK to me, there are a couple of minor things:

- If I do `C-b %` and the panes do not fit exactly they are rounded down, it would be better if the new pane was made slightly bigger so instead of:

avail = 11
old = 5
new = 5

We end up with:

avail = 11
old = 5
new = 6

- If there is exactly the right amount of space for a pane below it puts it one cell off screen so the border is hidden. I don't think it should do that, all borders should be kept visible. For example I pressed `C-b "` in the active pane:

image

There are a few behaviours that seem a little silly maybe but they are probably more complicated than it is worth to change. You can just use `C-b g` to move the pane where you want it anyway.

I really nice feature would be to remember the last geometry and have an "undo" key... but it might be tricky to do with mouse resize since you would want the geometry *before* the resize started, not an incremental.

nicm:
An undo key would be great because then if people are like  "wah C-b % resized my pane" they can just press C-b g R or whatever and itll be back. Even if it doesnt work for mouse...

mgrant0:
Or maybe a key sequence with c-Z since that's perhaps more memorable?

nicm:
`C-z` is already used.

nicm:
Maybe `C-b g C-z` or `C-b g C-\`, I'd like to keep all the "pane move" stuff under `C-b g` now.

mgrant0:
C-b g C-z sounds good to me.

daneofmanythings:
Interesting, All borders were shown when I was testing, must be my window size. I'll look into it.

I don't really like that the operation can change the dimensions of the original pane, but as you mentioned, maybe it's more trouble than its worth at the moment?

It is fully possible to have an undo queue now that `struct layout_geometry` exists. Each cell can remember it's previous geometry states. The user can go forward and backward as they like through the history. When the state is updated, the tail of the queue is freed and the new geometry is pushed onto it. This probably only makes sense for floating geometry because tiled geometry interferes with the whole tiled layout. Only thing I am immediately unsure about is ignoring intermediate geometries when a mouse resize is happening.

**EDIT1**: I have an idea of how to ignore when dragging; doesn't seem too bad.

daneofmanythings:
@nicm What are the silly behaviors? I'll have a think if i can remedy them easily.

nicm:
The main things are:

- If you do `C-b %` on a pane that is say in the centre of the window and there is no room on the right, it will resize the pane smaller, even though there is tons of space on the left. It would be more sensible to either make the new pane on the left instead, or shift the pane left to make more space on the right.

- If you do `C-b %` or `C-b "` and then go back to the previous pane and do the same again, it will make a second new pane that exactly covers the last new one. 

Both of these would require looking at where other panes are rather than just the sizes, so they are going to be more complicated.

I don't think we need a geometry queue necessarily, just one previous geometry would be enough. A queue is probably holding more state than most people will every use.

daneofmanythings:
First point should be pretty easy to resolve. It makes sense to me to move the target pane, and then do the resize as a last resort. Trying to spawn the new pane in different locations doesn't immediately guarantee that there is enough room to avoid a resize.

The second one is harder, but I have an idea of how to solve it.

One previous geometry it is!

daneofmanythings:
I didn't end up liking the solutions to fix those two silly points. Moving the original pane around felt bad to me. I have fixed the bugs and refactored things to make them more clear. I also added the `-f` flag so that the new pane takes up all the space on the side it was spawned. This is how I will mostly be using this. Let me know how you like it!

It seems to me that if a user wants panes to intelligently take up space as they split them, they should be using a tiled layout. I'm comfortable shipping this for now with the silly behavior intact. I will adjust things from user feedback when given.

*The "previous floating cell geometry" feature will be added in a separate PR. I also cleaned up a tiny bit of `split-window` and added `SPAWN_HORIZONTAL` for the floating splits to use.*

nicm:
I think too much logic that belongs to `cmd_split_window_exec` is now in `layout.c`. `layout.c` should not be responsible for setting `SPAWN_*` flags. If you want to do this much, move the helper functions into `cmd-split-window.c` but expose them so others can use them like `cmd_attach_session`.

daneofmanythings:
I have moved all the `SPAWN_*` state back into `split-window` and `join-pane`. The only args being used outside of command files are related to the geometry of panes, `l/p/x/X/y/Y`.

One thing to note that is starting to smell off: `SPAWN_BEFORE` is part of `struct spawn_context`'s `flag` field and it is not used in any of the places where `spawn_context->flags` are used. However is it related and I'm not sure there is a better spot to put it. I have added two more flags, `SPAWN_HORIZONTAL` and `SPAWN_SPLIT` which now also exist as part of the listed flags for `struct spawn_context`. That makes three flags that are hitching a ride. Probably not a big deal as they are all conceptually related to the same thing, but it is something to keep an eye on.

nicm:
Applied now, thanks!

PR #5343, opened 2026-07-07
Report the terminal's own theme to panes rather than guessing from the background

### LINKED ISSUE #5342: Theme reporting (2031/DSR 996) ignores the terminal's reported theme and always guesses from background colour
When a pane inside tmux queries the colour scheme (`CSI ? 996 n`) or receives an unsolicited mode 2031 notification, tmux answers with a value computed from the background colour luminance, even when the outer terminal has itself reported its theme via `CSI ? 997 ; Ps n`. The terminal's authoritative value is discarded, so tmux can report the opposite of what the terminal reports.

### Environment

- tmux master (next-3.8), also present in 3.7.
- Outer terminal: any terminal that supports mode 2031 / DSR 996 (e.g. Ghostty), i.e. one that answers `CSI ? 996 n` with its own light/dark preference.

### What happens

tmux does query the outer terminal:

- `tty.c:377` sends `\e[?2031h\e[?996n` when the terminal supports it.
- The reply `\e[?997;1n` / `\e[?997;2n` is parsed in `tty-keys.c:219-220` into `KEYC_REPORT_DARK_THEME` / `KEYC_REPORT_LIGHT_THEME`, and `server_client_report_theme()` stores it in `c->theme`.

So the client theme correctly reflects the terminal. But the value a pane receives comes from `window_pane_get_theme()` (`window.c:2420`), which prioritises the background colour:

```
/* Derive theme from pane background color, if it's not the default colour. */
theme = colour_totheme(window_pane_get_bg(wp));
if (theme != THEME_UNKNOWN)
    return (theme);
/* ... only now fall back to a client that has a theme (c->theme) ... */
```

For a default pane, `window_pane_get_bg()` (`window.c:2324`) does not return "default". It falls through to `window_get_bg_client()` (`window.c:2342`), which returns the attached client's `tty.bg`, i.e. the terminal background queried over OSC 11. `colour_totheme()` (`colour.c:347`) then classifies it purely by luminance (`r+g+b > 382 => light`).

Because the terminal always answers OSC 11, `tty.bg` is essentially always set, so the `c->theme` branch (the value the terminal actually reported via 997) is never reached. The same background-first fallback is also used for the client theme itself in `server_client_update_theme_colours()` (`server-client.c:1197`), but there it is only used when `c->theme` is `THEME_UNKNOWN`, which is the correct precedence. `window_pane_get_theme()` has the precedence reversed.

### Result

- Query the terminal directly: `printf "\e[?996n"` returns the terminal's OS/config scheme.
- Query tmux running in the same terminal: tmux returns the luminance of the background colour.

These disagree whenever the background's luminance does not match the declared scheme, for example a light scheme with a slightly dark-tinted background, or any custom `background` set in the terminal that crosses the threshold the other way. Apps that use 2031/996 (delta, vim, etc.) then see the wrong theme inside tmux.

### Expected

The terminal is the authority. When the outer terminal has reported a theme via `997` (mode 2031 / DSR 996), tmux should report that value. The background-colour luminance guess should only be used as a fallback for terminals that do not support the sequence, which is what the original change intended (see #4353: "guess the theme from the background colour on terminals which do not themselves support the escape sequence").

### Suggested fix

In `window_pane_get_theme()`, prefer a known client theme (`c->theme`, from the `997` report) and only fall back to `colour_totheme(window_pane_get_bg(wp))` when no client reports a theme, mirroring the precedence already used in `server_client_update_theme_colours()`.

aymanbagabas:
When a client's terminal reports its light/dark mode via mode 2031 or DSR 996, that report is authoritative and should be forwarded to panes that query DSR 996 themselves.

`window_pane_get_theme` instead guessed from the pane background colour first and only consulted the client's reported theme when the background was default. So any pane with an explicit background overrode the terminal's own answer, and a client on a light terminal could be told the theme is dark (and vice versa).

This reverses the precedence: check attached clients for a reported theme first, and fall back to the background-colour guess only for terminals that do not report one. This matches the precedence already used in `server_client_update_theme_colours`.

Includes a regress test (`regress/theme-report.sh`) that nests an inner client inside an outer light-background pane while giving the inner server a dark window background: the pane's DSR 996 query must answer `997;2` (light, from the terminal), not `997;1` (dark, from the bg guess). Verified the test fails on the unfixed binary and passes on the fixed one.

Fixes #5342, follow-up to #4353.

### DIFF
--- a/regress/theme-report.sh
+++ b/regress/theme-report.sh
@@ -1,0 +1,57 @@
+#!/bin/sh
+
+# Test that the theme reported to a pane (mode 2031 / DSR 996) follows the
+# terminal's reported theme, not a guess from the background colour.
+#
+# An inner client is attached inside an outer tmux pane. The outer pane has a
+# light background, so the outer server answers the inner client's DSR 996 with
+# light. The inner server is given a dark window background, so a guess from the
+# background colour would say dark. A pane in the inner server then queries DSR
+# 996: the answer must be light (2), following the client's reported theme, not
+# dark (1) from the background.
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+INNER="$TEST_TMUX -LtestI$$ -f/dev/null"
+OUTER="$TEST_TMUX -LtestO$$ -f/dev/null"
+$INNER kill-server 2>/dev/null
+$OUTER kill-server 2>/dev/null
+
+TMP=$(mktemp)
+trap "rm -f $TMP; $INNER kill-server 2>/dev/null; $OUTER kill-server 2>/dev/null" \
+	0 1 15
+
+# Inner server, dark background, keep panes alive across the query respawn.
+$INNER new-session -d -x80 -y24 || exit 1
+$INNER set -g remain-on-exit on
+$INNER set -g window-style 'bg=black'
+
+# Outer server with a light background, running the inner client attached so the
+# inner client has a real terminal that reports a theme.
+$OUTER new-session -d -x80 -y24 || exit 1
+$OUTER set -g window-style 'bg=white'
+$OUTER new-window "$INNER attach" || exit 1
+sleep 2
+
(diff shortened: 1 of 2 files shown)


nicm:

PR #5348, opened 2026-07-08
refactor(master): `layout_set_size` now uses `struct layout_geometry`.

daneofmanythings:
I have also changed the `sx` and `sy` fields of `struct layout_geometry` to signed. Some calculations use signed values to hold temporary sizes and check against going negative. This change allows for inline assignment of field values in these calculations. There is a loss of semantic significance though. I kind of like it, but I can certainly change it back.

This PR is groundwork for a size history in floating panes.


daneofmanythings:
Closing as not planned.

PR #5349, opened 2026-07-08
Do not exit copy mode on scroll-exit if there is an active selection

X1aomu:
Right now, when you enter copy mode with `-e` (which is the default behavior when you scroll up with the mouse), scrolling back to the bottom exits copy mode automatically. This is super handy for a quick peek at the pane history, but it gets a bit annoying when you're trying to select text with the mouse.

For instance, if you want to select text that starts above the current screen: you'd scroll up to enter copy mode, start your selection, and then scroll down to extend it. But if you scroll all the way back to the bottom, copy mode exits right away and cuts off your selection before you can finish.

To make this feel smoother, this change stops copy mode from exiting if there is an active selection. If you don't have anything selected, it will still exit automatically just like before.

This applies to all the usual scroll-exit actions, whether you are scrolling down, paging down, or using the scrollbar.

I have also added a regression test and updated the manual page to reflect this.

### DIFF
--- a/regress/copy-mode-scroll-exit.sh
+++ b/regress/copy-mode-scroll-exit.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -LtestA$$ -f/dev/null"
+$TMUX kill-server 2>/dev/null
+
+cleanup()
+{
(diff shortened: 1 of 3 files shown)


mgrant0:
I've folded these changes into https://github.com/tmux/tmux/pull/5353 which is still work in progress.  Thanks

nicm:
I have applied this now with a tweak from Michael to remove the helper function. It should be in GitHub later on, please let me know if you see any problems. Thanks!

X1aomu:
That's fine. I'll try it when the changes are made to the master branch.

Issue #5336, opened 2026-07-06
display-popup repaints/flickers when a background pane redraws its full region (scroll / alternate-screen) — 3.7


noamsto:
tmux 3.7b

Since 3.7, an open `display-popup` is repainted whenever a pane behind it redraws a region that overlaps the popup. On 3.6a this never happened while a popup was open.

It's easy to see the boundary: a background pane that only *appends* lines below the popup does not flicker — but as soon as its output **scrolls** (passes the pane height, shifting the rows behind the popup), the popup repaints on every scroll. A fullscreen / **alternate-screen** program (vim, btop, or any TUI) redraws its whole pane each frame, so the popup flickers continuously.

This looks like the fix for #4920 ("popup window gets overwritten by background updates") firing on every full-region background update, not only when the popup is genuinely overwritten.

**Repro (3.7b):**
1. `tmux new-session`
2. In the pane: `while :; do date +%s.%N; sleep 0.1; done`
3. `tmux display-popup -w 60% -h 60% -E 'sleep 60'`
4. No flicker while output stays below the popup; **flicker begins once the pane scrolls.** An alt-screen TUI in step 2 (e.g. `btop`) flickers immediately.

**Expected:** open popup stays static (3.6a behavior).
**Actual:** popup repaints on every overlapping background redraw.

Terminal: kitty (`xterm-kitty`), synchronized output (`sync`) enabled — flicker persists with `sync` on.

Could the popup be repainted only over the region actually overwritten (or an option to keep background panes from repainting an open popup)?

ibehnam:
Yes, I have the same issue and burned all my Fable quota, thinking it was my config change that caused it... Downgrading to 3.6 solved it.

nicm:
I think this is fixed now?

PR #5350, opened 2026-07-08
Clip the status line and menu against an open popup overlay

noamsto:
Two related overlay-clipping bugs: content that isn't clipped against the
client overlay can be drawn over an open `display-popup`. Found while looking
into #5336.

### 1. Status line drawn over a popup

`redraw_draw()` draws the status line with `tty_draw_line(tty, sl, 0, i,
UINT_MAX, 0, y + i, NULL)`. `tty_draw_line()` writes straight to the terminal
and does not consult the overlay — callers are expected to pass already-clipped
ranges (as `tty_draw_pane()` does), but the status caller passes the full width.
A status-only redraw also does not set `REDRAW_OVERLAY`, so the overlay is not
repainted afterwards. So whenever the status line is redrawn while a popup is
open — clock tick, activity, an option change — the status line paints over the
popup.

Reproduce:
```sh
tmux -f/dev/null new-session -d -x80 -y24
tmux set -g status-interval 1
tmux set -g status-right '%H:%M:%S'
tmux display-popup -w60 -h6 -yS -E 'sleep 30'
```
Each clock tick draws the time over the bottom of the popup. (Also reproducible
with `tmux refresh-client -S` while the popup is open.)

This commit splits each status line into the ranges left visible by the overlay
(`tty_check_overlay_range`) and draws only those, the same way panes and borders
are drawn. Repainting the overlay on every status redraw would also hide it but
makes an open popup flicker on each tick (see #5336), so clipping is preferred.

### 2. Menu open over a popup uses the wrong ranges

In `popup_check_cb()`, the menu branch computes the ranges left visible by the
menu into `mr`, but when intersecting them with the popup it indexed
`r->ranges[i]` — `r` is `&pd->r`, its own output buffer, still holding the
result of a previous call — instead of `mr->ranges[i]`. With a menu open over a
popup this yields wrong visible ranges. Fixed to use `mr->ranges[i]`.

### Note

The per-caller clip loops (`tty_draw_pane`, `tty_cmd_redrawline`, the status
draw) all repeat the same check_overlay_range → loop → draw idiom, and the
status caller's bug was simply forgetting it. Moving the clip into
`tty_draw_line()` itself would remove the duplication and make the omission
impossible; I kept the change minimal here but am happy to send that version
instead if you prefer.

### DIFF
--- a/popup.c
+++ b/popup.c
@@ -244,7 +244,7 @@ popup_check_cb(struct client* c, void *data, u_int px, u_int py, u_int nx)
 		*/
 		for (i = 0; i < mr->used; i++) {
 			server_client_overlay_range(pd->px, pd->py, pd->sx,
-			    pd->sy, r->ranges[i].px, py, r->ranges[i].nx,
+			    pd->sy, mr->ranges[i].px, py, mr->ranges[i].nx,
 			    &pd->or[i]);
 		}
 
--- a/screen-redraw.c
+++ b/screen-redraw.c
@@ -1588,8 +1588,10 @@ redraw_draw(struct client *c, struct window_pane *wp, int flags)
 	struct screen		*sl;
 	struct redraw_scene	*scene;
 	struct window_pane	*loop;
-	u_int			 width, i, y, lines;
+	u_int			 width, i, y, lines, j;
 	struct redraw_span	*first;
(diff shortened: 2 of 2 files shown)


nicm:

PR #5351, opened 2026-07-08
4772 pane scrollbars on causes alt screen to flicker

### LINKED ISSUE #4772: "pane-scrollbars on" causes full-screen programs like nano to flicker

mgrant0:
Based on the first fix, the second commit has been discarded.


nicm:
Applied, thanks!

PR #5353, opened 2026-07-09
Copy mode selection cleanups

mgrant0:
1. Fixed the selection so that it does not change when the pane scrolls
2. Made it so you can modify the selection by dragging in it.
3. Added regression tests

### DIFF
--- a/regress/copy-mode-selection-scroll.sh
+++ b/regress/copy-mode-selection-scroll.sh
@@ -1,0 +1,113 @@
+#!/bin/sh
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -LtestA$$ -f/dev/null"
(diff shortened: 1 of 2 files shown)


mgrant0:
Some of the changes in here were to fix some flashing issues which were introduced by the selection changes.  I bet this can be cleaned up so those additional changes are not necessary.  e.g. adding window_copy_redraw_lines1 and window_copy_write_line1.

nicm:
There is a lot of stuff in here all mixed together, some bug fixes and some features...

What is the need for `WINDOW_COPY_CMD_NOREDRAW`? Is it just an optimization? `NOTHING` should only redraw the first two lines so is "don't do anything" for a few rare cases really needed?

What is the flashing? Do you mean the flickering when dragging the selection? Which bit is the fix for that?

mgrant0:
I've completely reworked this PR.  This PR now only contains fixes to the selection in copy-mode, allows the selection to be modified by draging it more.  It also adds regression tests now.

nicm:
Applied now, thanks. I added a couple of temporaries to make lines shorter in `window_copy_mouse_in_selection`.

PR #5355, opened 2026-07-09
Fix flickering in copy-mode where the lines were being needlessly redrawn

mgrant0:
1. Fix flickering when at the top in copy-mode and scrolling up with mousewheel
2. Similarly fix flickering when at the bottom and scrolling down with mousewheel
3. Fix flickering when the copy-mode cursor moved or the selectionn changed, it was redrawing the line that the cursor was on over and over.
4. Flickering was caused by excessive clearing and redrawing where only redrawing was necessary to change the selection.


nicm:
Done, thanks.

PR #5361, opened 2026-07-10
fix(master): Two bugs in `layout-set`.

daneofmanythings:
Overlooked setting of the other cell in the `layout_set_main_*` functions. Offsets are ignored as they are fixed further down the function.

**EDIT1**: will test when main compiles again.

### DIFF
--- a/layout-set.c
+++ b/layout-set.c
@@ -136,7 +136,7 @@ layout_set_first_tiled(struct window *w)
 }
 
 static void
-layout_set_link_untiled(struct window *w, struct layout_cell *lcroot)
+layout_set_link_floating(struct window *w, struct layout_cell *lcroot)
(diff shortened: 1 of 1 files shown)


daneofmanythings:
Realized that floating panes are left referencing the old root. Adjusted logic so they are wired into the new layout.

nicm:
What is "untiled"? If something is not tiled, it is floating, right?

daneofmanythings:
Right now yes, but I was trying to future proof the semantics. Floats, hidden, maybe modal floats. Anything that doesn't take up space in the tiled layout. I have changed the semantics for now to floating. Also found a bug and fixed it.

nicm:
Applied, thanks.

PR #5363, opened 2026-07-11
fix(xmalloc): add overflow check to xcalloc

YayoRazo:
xcalloc() wraps calloc() but does not validate that nmemb * size fits in size_t before calling calloc().  Every other allocation wrapper in xmalloc.c already performs this check (xreallocarray delegates to reallocarray which checks internally).

Add nmemb > SIZE_MAX / size before calloc(), matching the pattern used by the compat reallocarray and recallocarray implementations.

### DIFF
--- a/xmalloc.c
+++ b/xmalloc.c
@@ -44,6 +44,8 @@ xcalloc(size_t nmemb, size_t size)
 
 	if (size == 0 || nmemb == 0)
 		fatalx("xcalloc: zero size");
+	if (nmemb > SIZE_MAX / size)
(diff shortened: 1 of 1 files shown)


nicm:
Why did you close this? It looks right?

`xcalloc` was added to ssh in r1.18 in 2006 (https://cvsweb.openbsd.org/checkout/src/usr.bin/ssh/xmalloc.c?rev=1.38)... I guess I must have copied the file before that and then added `xcalloc` independently, forgetting the check.

YayoRazo:
Closed by mistake while fixing the title format — the fix is reopened at #5374. Sorry for the noise.

PR #5365, opened 2026-07-11
fix(compat): guard buffer size multiplication in getline

YayoRazo:
In getdelim(), *bufsiz * 2 can overflow before the value reaches realloc(). Add a SIZE_MAX / 2 guard before the multiplication.

Dropped the fgetln hunk after review: reallocarray() there already guards the multiply.

### DIFF
--- a/compat/getline.c
+++ b/compat/getline.c
@@ -74,8 +74,14 @@ getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
 		}
 		if (ptr + 2 >= eptr) {
 			char *nbuf;
-			size_t nbufsiz = *bufsiz * 2;
+			size_t nbufsiz;
 			ssize_t d = ptr - *buf;
+
+			if (*bufsiz > SIZE_MAX / 2) {
+				errno = EOVERFLOW;
+				return -1;
+			}
+			nbufsiz = *bufsiz * 2;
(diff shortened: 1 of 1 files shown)


nicm:
getline yes, fgetln no - bufsz has not changed so if reallocarray succeeded, the multiply must be safe.

It might be better to sync getline from NetBSD if it has any other fixes.

YayoRazo:
Dropped the fgetln hunk — you're right, reallocarray already guards that multiply. Kept the getline one since getdelim uses plain realloc(*buf, *bufsiz*2) with no overflow check.

YayoRazo:
Checked NetBSD: the file/external version this is copied from is the same code with no overflow guard, so there is nothing newer to sync there, and NetBSD's libc getdelim uses libc internals (__srefill, __sferror, fp->_r) that do not fit a compat shim. I did take its errno choice: the overflow case now sets EOVERFLOW (what POSIX and NetBSD use) instead of ENOMEM. I verified the path by temporarily lowering the SIZE_MAX / 2 threshold and reading an over-long line: getline() returns -1 with errno == EOVERFLOW (75).

nicm:
Thanks. It looks like the only remaining user of `fgetln` is in `prompt-history.c`... it'd be nice if that used `getline` then we could drop the `fgetln` compat bits entirely.

Issue #5126, opened 2026-05-27
Sixel image previews leave trailing pixels in window @0 only; new splits inherit the artifacts


hamzabow:
### Issue description

Sixel image previews leave trailing pixels in window @0 of a session. The newly rendered image is correct, but cells from the previous preview that aren't overwritten by the new one remain visible. Reproducible with `tmux -f /dev/null -L bugrepo` (no user config). The yazi file manager is what's emitting the sixel sequences in my repro; happy to try with a simpler emitter if requested.

**Note on testing against master:** the CONTRIBUTING template asks for the latest git HEAD. I built and tested master (`next-3.7`), but it has a different issue: sixel previews don't render at all (image flashes for a fraction of a second and disappears). I couldn't get a clean repro of the original ghosting bug on master because of this — so this report is against tag `3.6b`. If that's a blocker, please let me know and I can try to bisect or test specific commits.

**Reproduced on two terminal emulators.** Initially observed in Windows Terminal; the same window-@0-only pattern also reproduces in **WezTerm** (which has well-tested first-class sixel support). Two independent terminal implementations producing the same artifact makes a tmux-side issue more likely than a terminal-side one.

**Key observations narrowing the cause (3.6b)**
- Only window @0 is affected. Creating a new window (`prefix + c`) → no ghosting in yazi there.
- Per-*window*, not per-pane: splitting window @0 causes the new split pane to also display leftover sixel pixels. Splitting a working window (@1+) does not.
- Killing and recreating window 0 from an attached client (`kill-window -t :0; new-window -t :0`) fixes it. The same commands run from a `session-created` hook do NOT fix it.
- `allow-passthrough`, `terminal-features ',*:sixel'`, $TERM, and pane dimensions are identical between the affected pane in @0 and an unaffected pane in @1 (verified with `tmux display -p`).
- Redraws don't clear the trailing pixels: tested `refresh-client -S`, `Ctrl+L`, hard reset (`printf '\033c'`), and pane resize.
- In tmux 3.4 the same symptom occurred in *every* window. After upgrading to 3.6b built with `--enable-sixel` it became restricted to window @0 — so the upgrade was a partial fix.

**Steps to reproduce**
1. `tmux kill-server`
2. `tmux -f /dev/null -L bugrepo new`
3. In window 0, open yazi, hover several images with different aspect ratios in sequence.

### Required information

* tmux version: `tmux 3.6b` (built from tag `3.6b` with `--enable-sixel`)
[...]

PR #5366, opened 2026-07-11
Redraw only dirty lines when synchronized update ends

Joost1982:
### Problem

On the `release_3.7c` branch (and released 3.7b), ending a synchronized update (`DECRST 2026`) sets `PANE_REDRAW`, repainting the whole pane (added in e802909d as the fix for #5304). The full repaint erases any passthrough content drawn over the pane: with `allow-passthrough on`, a sixel image sent by yazi appears for a fraction of a second and is then wiped by the repaint. This is the "image flashes and disappears" behaviour mentioned as a side note in #5126.

Bisect on my machine confirms 3.7a works and 3.7b does not, and the regressing change is exactly the `PANE_REDRAW` added in the 2026 case of `input_csi_dispatch_rm_private()`.

Simply reverting those two lines is not enough: that brings back the #5304 corruption (verified with `nvim --clean` + scrolling - output stays stale because tty output is suppressed while `MODE_SYNC` is set).

### Fix

Port the approach from master (565db46a and d33d5b7d) to this branch: while a pane is in sync mode, record which lines are actually written at the three places that suppress tty output during `MODE_SYNC` (`screen_write_redraw_line`, `screen_write_collect_flush`, `screen_write_cell`), plus the whole scroll region if there was any scrolling. When sync ends, redraw only those dirty lines instead of the entire pane. Lines never touched during the sync block - such as those covered by a passthrough image - are left alone.

### Testing

Built with `--enable-sixel`, tested on foot 1.27 / Arch Linux:

* yazi image preview inside tmux: image now persists (before: flashed once and disappeared)
* `nvim --clean cfg.c` + `C-d` scrolling (the #5304 reproduction): renders correctly, no stale content
* smoke: scrollback, horizontal split, alternate screen enter/exit all behave as before

Refs #5126, #5304.

### DIFF
--- a/input.c
+++ b/input.c
@@ -1955,8 +1955,6 @@ input_csi_dispatch_rm_private(struct input_ctx *ictx)
 			break;
 		case 2026:
 			screen_write_stop_sync(ictx->wp);
-			if (ictx->wp != NULL)
-				ictx->wp->flags |= PANE_REDRAW;
 			break;
 		case 2031:
 			screen_write_mode_clear(sctx, MODE_THEME_UPDATES);
--- a/screen-redraw.c
+++ b/screen-redraw.c
@@ -1324,6 +1324,7 @@ screen_redraw_draw_pane(struct screen_redraw_ctx *ctx, struct window_pane *wp)
 
 	if (wp->base.mode & MODE_SYNC)
 		screen_write_stop_sync(wp);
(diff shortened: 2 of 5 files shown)


nicm:
No, sorry, this change is too big for a patch release, this will need to stay as it is until 3.8.

PR #5357, opened 2026-07-10
control: stop queueing output and notifications to exiting clients

bmaurer:
## TL;DR

Once a control (`-CC`) client is marked `CLIENT_EXIT` — evicted as `"too far behind"` by `control_check_age()`, or told to go by `detach-client` — the server still queues **new** work for it:

- `CONTROL_SHOULD_NOTIFY_CLIENT()` checks only `CLIENT_CONTROL` and `control_state`, so every notification keeps being formatted and queued for a client that is already being evicted for not consuming data;
- `control_write_output()` keeps creating `%output` blocks for it until the exit handshake completes.

None of it is ever meaningfully consumed: the exit handshake only waits for blocks that were *already* queued (`control_all_done()`), and after `CLIENT_EXITED` is set, `server_client_check_exit()` returns early so even `control_discard()` never runs again.

Queueing that data is not just pointless work — it is what the server later `writev()`s to the dying client's tty. In the incident behind #5356, that tty's file description had been switched to blocking under the server's feet, and the write of exactly these queued notification bytes parked the single-threaded server inside the kernel (`n_tty_write`) forever, hanging every session and every future tmux command.

This PR stops queueing for exiting clients: pane output takes the existing `ignore` path (consuming pane offsets the way `no-output` clients do), and the notify macro excludes `CLIENT_EXIT`. Two lines. Either this or #5356 alone prevents the wedge — both were verified independently with the reproducer below — but each stands on its own.

## Why a client can linger long after CLIENT_EXIT

The exit handshake is asynchronous and the window can be arbitrarily long:

- `server_client_check_exit()` sends `MSG_EXIT` only after already-queued output drains to the tty (`control_all_done()`) — on a slow terminal that alone takes a while;
- after `MSG_EXIT`, a `wait-exit` client deliberately lingers (reading stdin until EOF) with its socket open, so the server keeps the `struct client` and its tty fds;
- if the client's terminal is dead (the scenario that triggers `"too far behind"` eviction in the first place), the client can additionally block forever in its final `%exit` write. In the production incident behind #5356 we found such zombie clients 22 and 30 days old.

During that entire window, master keeps formatting and queueing notifications for the doomed client.

## Evidence

Reproducer (script below), `--mode age`, on the full wedge scenario from #5356. This PR does **not** touch client.c, so the client-side O_NONBLOCK flip on the shared tty file description still happens — the difference is purely whether the server has queued bytes to write afterwards:

| binary | outcome |
|---|---|
[...]

PR #5368, opened 2026-07-11
RFC: resynchronise control mode clients that fall behind, instead of pausing or killing them

bmaurer:
This is an RFC. I'd like a read on the direction before investing more time in it.

## Motivation

The goal is to make the whole EternalTerminal <-> tmux stack behave well end to end: attach with -CC over a remote link, step away while a pane spews output, and come back to a session that is current and responsive. Today every hop handles a slow consumer badly. On the ET side, megabytes queue in kernel buffers so ^C takes a minute or more to bite (MisterTea/EternalTerminal#631, fixed with per-link flow control in MisterTea/EternalTerminal#730). On the tmux hop the only tool is pausing: iTerm2 sets pause-after, tmux emits %pause, and the user gets a "press to unpause" banner. That's bad UX — when you're looking at a buffer you want to see it; you always press the button. The defaults for clients that don't opt in are worse: the pane buffer grows until a 300 second kill, and a sole client that stops reading gets pane reads disabled, at which point the program in the pane blocks in write() with nothing to time it out.

## What this does

End-to-end flow control instead. When tmux can't write to a control client it queues locally, as now. When a client falls further behind than the pane's history keeps, tmux discards the backlog and refreshes the client's state: a repaint of the pane inside a synchronized-update bracket, plus a replay of the missed scrollback out of history. The aim is to be strictly better than pausing — up to history-limit behind, the refresh reconstructs everything the client missed, so resyncing loses nothing; past that, history is already trimming lines the client never saw, so holding the backlog only postpones the same loss. When incremental catch-up is cheaper, tmux just keeps streaming; the refresh is what an unpause would have sent anyway, minus the button.

The refresh is plain terminal control codes carried in the normal %output stream, so the client protocol doesn't change and existing clients render it as-is (unmodified iTerm2 included). By default pause-after clients are resynced too and never see %pause; `control-resync keep-pause` restores the old behaviour and `off` the legacy kill.

Three patches: a small factoring of the control output path, a grid API for tracking the history sequence, then the feature. One new server option; the thresholds derive from history-limit rather than adding knobs. The shape follows TTY_BLOCK, which already does discard-then-redraw for slow regular terminals.

## Disclosure and testing

The implementation is largely AI-generated, directed and reviewed by me. It has been tested heavily with automation: in-tree regress tests including a 26-case repaint-fidelity suite that round-trips every generated repaint through a second tmux instance as an oracle, and a coverage-guided fuzzer over the serialiser. The fuzzing caught two bugs in the new repaint code introduced by this series (stale cells under tab gaps, and a cursor restore that ignored origin mode); both are fixed in what's posted here. What it has not had is real-world end-to-end time. Before taking this further I plan to set up automated testing that runs iTerm2 -> injected proxy -> tmux, so I can reconstruct real flow-control scenarios (slow links, stalls, dead readers) against real client behaviour.

Related: #5356 / #5357 came out of the same investigation.

### DIFF
--- a/control.c
+++ b/control.c
@@ -30,27 +30,44 @@
 #include "tmux.h"
 
 /*
- * Block of data to output. Each client has one "all" queue of blocks and
- * another queue for each pane (in struct client_offset). %output blocks are
- * added to both queues and other output lines (notifications) added only to
- * the client queue.
- *
- * When a client becomes writeable, data from blocks on the pane queue are sent
- * up to the maximum size (CLIENT_BUFFER_HIGH). If a block is entirely written,
- * it is removed from both pane and client queues and if this means non-%output
- * blocks are now at the head of the client queue, they are written.
- *
- * This means a %output block holds up any subsequent non-%output blocks until
- * it is written which enforces ordering even if the client cannot accept the
- * entire block in one go.
+ * A queued block of output. Each client has one "all" queue and a per-pane
+ * queue; a block on both holds up any later block until it is written, which
+ * keeps output ordered against notifications. LINE is a verbatim notification
+ * line; PANE is pane output of size bytes read at cp->offset; REPAINT is a
+ * resync repaint, rendered lazily into its own buffer when the drain loop first
+ * reaches it and escaped from there.
  */
+enum control_block_type {
+	CONTROL_BLOCK_LINE,
+	CONTROL_BLOCK_PANE,
+	CONTROL_BLOCK_REPAINT
+};
 struct control_block {
-	size_t				 size;
-	char				*line;
+	enum control_block_type		 type;
 	uint64_t			 t;
+
+	size_t				 size;		/* PANE */
+	char				*line;		/* LINE */
+	struct evbuffer			*repaint;	/* REPAINT */
+
+	/* History sequence when a PANE block was created (resync watermark). */
+	struct grid_history_state	 history;
 
 	TAILQ_ENTRY(control_block)	 entry;
 	TAILQ_ENTRY(control_block)	 all_entry;
+};
+
+/*
+ * Exclusive pane state. ON streams normally; OFF is disabled and holds the
+ * pane read gate; PAUSED has been paused with %pause; RESYNC has discarded its
+ * backlog and owes a repaint (queued as a REPAINT block) while it keeps running
+ * with its output discarded.
+ */
+enum control_pane_state {
+	CONTROL_PANE_ON,
+	CONTROL_PANE_OFF,
+	CONTROL_PANE_PAUSED,
+	CONTROL_PANE_RESYNC
 };
 
 /* Control client pane. */
@@ -65,9 +82,15 @@ struct control_pane {
 	struct window_pane_offset	 offset;
 	struct window_pane_offset	 queued;
 
-	int				 flags;
-#define CONTROL_PANE_OFF 0x1
-#define CONTROL_PANE_PAUSED 0x2
+	enum control_pane_state		 state;
+
+	/*
+	 * History sequence and pane width when the client fell behind, used to
+	 * bound the RESYNC scrollback replay; a discontinuity or width change
+	 * degrades it to a repaint only.
+	 */
+	struct grid_history_state	 resync_state;
+	u_int				 resync_sx;
 
 	int				 pending_flag;
 	TAILQ_ENTRY(control_pane)	 pending_entry;
@@ -91,6 +114,15 @@ struct control_state {
 	struct bufferevent		*write_event;
 
 	struct monitor_set		*subs;
+
+	/*
+	 * Output watchdog. resync_last_progress is stamped whenever the write
+	 * buffer actually drains; the timer recovers panes that make no progress
+	 * for too long, the case where both the pane-read and socket-write paths
+	 * are stalled and no other callback fires. See control_client_stalled.
+	 */
+	struct event			 output_timer;
+	uint64_t			 resync_last_progress;
 };
 
 /* Low and high watermarks. */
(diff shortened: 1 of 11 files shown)


nicm:
I don't know if we want two mechanisms here, and I don't remember why we chose not to do this and to go with the explicit prompt instead. @gnachman?

In any case this should be chosen by the control client with a flag, not by the user with a `control-resync` option.

gnachman:
> End-to-end flow control instead

True e2e flow control would stop reading from the TTY when the clients are backed up. This is what unix intended and works really well, but we can't always have it.

Ify ou have a non-control-mode tmux client attached as well as a control mode clinet, the non-control-mode client can sink tty input super fast compared to the control-mode client (which is typically over an ssh link). We didn't want to provide backpressure all the way to the tty because that would unnecessarily throttle the non-control-mode tmux.

Pause/unpause is the workaround to allow control mode clients to resynchronize while non-control-mode clients have no speed limit. I privately doubt this happens very often, but I really don't know. As I recall we do provide backpressure to the end-to-end (i.e., we stop reading from the TTY when we get behind) in the case that all clients are control mode, though I could be misremembering that.

> On the tmux hop the only tool is pausing: iTerm2 sets pause-after, tmux emits %pause, and the user gets a "press to unpause" banner. That's bad UX — when you're looking at a buffer you want to see it; you always press the button

I agree that the UX is bad. Auto-unpause (an optional feature in iTerm2) addresses it. Probably it could be improved. 

>  I don't remember why we chose not to do this and to go with the explicit prompt instead

I believe I originally preferred not to enable pausing by default, and @nicm wanted it, but I don't recall the rationale. There's an [absurdly lengthy thread](https://github.com/tmux/tmux/issues/2217) here where all the design occurred. I don't have the time to read all 150+ comments, so I'll share Claude's findings (sorry! I hate when people share LLM output with me but in this case it's my best option)

> The thread's explicit reasoning on this is fairly light, but it's there, and the underlying logic is clear from the surrounding design. Here's what actually drove it.
>
> What the thread records:
> - nicm flagged early that resuming was the hard part (line 540): "How do we get a client out of snapshot mode again? This is complicated I expect." And later (line 2299): "The other question is how you are going to get out of snapshot mode... Or are you just going to pause it until the user says go again?"
> - Your answer (lines 2449-2453): "I was thinking of having the user do it manually. There could be an automatic unpause feature, but it would be off by default. Or added later on. I'll have to see how it feels." — with refresh-client -A '%0:continue' working for either approach.
>
> So the recorded decision was: ship the simple thing (manual, user hits a button), leave auto-unpause as a possible off-by-default feature for later.
>
> Why manual made sense (the logic behind it):
>
> 1. Unpausing isn't free — it's a full resync. Pausing means tmux throws away buffered output for that pane. To come back, iTerm2 can't just resume the stream; it has to atomically re-fetch the pane's state the same way it does on initial attach: capture-pane -P for terminal state, screen contents, history, uservars, etc. (your big semicolon-delimited command at lines 586-589 / 2080), ignoring %output until that completes. That's expensive and can lose history, with a visible hiccup.
> 2. Auto-unpause would thrash. The pane got paused precisely because it's chatty and you're behind. If tmux auto-resumed, you'd very likely blow past the age threshold again almost immediately and re-pause — pause → resync → pause → resync in a loop, which is worse than staying paused. There's no signal that says "the pane has calmed down enough to keep up now," so any auto-resume is a guess.
> 3. The data-loss tradeoff is the user's to make. Pause mode exists so the user can consciously trade reliability (every byte preserved) for responsiveness. Since resuming means accepting lost history for that pane, it's appropriate to make the human explicitly opt back in rather than have tmux silently decide.
> 4. Simplicity / ship-it pragmatism. Both you and nicm wanted to land the core watermark + pause machinery first. Manual unpause is a one-liner (%0:continue); a good automatic policy is genuinely hard (see point 2), so it was deferred rather than blocking the feature.
>
> Net: it wasn't so much a hard design principle as "manual is correct-by-default because resuming is a lossy resync you shouldn't trigger automatically, and a smart auto-resume is hard enough to punt on." Want me to check the iTerm2 side to see whether an auto-unpause option was ever actually added later?

---

> a repaint of the pane inside a synchronized-update bracket, plus a replay of the missed scrollback out of history

It sounds like you'd send a control sequence to erase history and then output a bunch of history and just go from there. That's pretty similar to how unpausing works, except the logic is in the tmux server rather than the client. Since I think the aim is to improve performance over auto-unpause, can you clarify how that would work? 

Also, there's a bunch of state that also needs to be synchronized (e.g., DECSET, keyboard modes, pending input of partially-received control sequences). If tmux is at the wheel of resynchronization it will need to drive everything the client already does for unpausing via control sequences. That is maybe possible but kind of a hassle. For example, terminating a partially received control sequence without weird side effects sounds like a hard problem!

bmaurer:
Sorry for the delay in responding. It sounds like the auto unpause may be the better option here -- I didn't see it when I was using the product!

at the end of the day, this sort of unpausing (both my PR and what exists) is not a great control flow mechanism. if the client is not able to keep up (say cat'ing a huge log file) the net outcome is you get buffer bloat for a while, then catch up, then bloat again.

Ideally what we'd want is a channel that can just downclock to X frames per second if the client can't keep up with full terminal output. if we have a good protocol I bet AI could one shot some very cool stuff!

nicm:
That's pretty much how it works for regular (not control) clients, but you will still end up with buffers growing and then dropping because a full redraw means kilobytes of data being sent. tmux redraws at most every 100 milliseconds when it judges the terminal cannot keep up, and it can be longer if the buffer is not emptying. It also aggressively discards invisible output (such as scrolling) to reduce data volume, which it cannot do with control mode. The issue is that as George mentioned the control mode protocol refresh is two way - the client needs to request what it needs to reconstruct the state. So you would need to switch to a push mechanism where tmux simply says "I've dropped a load of data, you need to refresh - here is everything you need".

PR #5369, opened 2026-07-11
fix(cmd): guard realloc size arithmetic against overflow

YayoRazo:
In cmd_template_replace(), the size passed to xrealloc(), len + strlen(s) * 3 + 1, can overflow before the allocation. Add a SIZE_MAX guard before it.

Dropped the format.c hunks after review: those sites use xreallocarray(), which already guards the multiplication.


nicm:
I think you need this to be `slen >= SIZE_MAX / 3` because if `slen` is exactly `SIZE_MAX / 3`, then `SIZE_MAX - (slen * 3) - 1` will be `(size_t)-1`?

nicm:
I think I accidentally applied your previous change but in any case both changes are in now. Thanks!

PR #5370, opened 2026-07-11
fix(input): return to ground state when input buffer overflows

YayoRazo:
When an OSC, DCS, or CSI sequence exceeds the input buffer size, INPUT_DISCARD is set but the state machine stays in its current non-ground state, stalling all further input processing until the sequence terminator arrives.

Force a return to ground state when INPUT_DISCARD is detected outside ground, so subsequent input can be processed immediately.

### DIFF
--- a/input.c
+++ b/input.c
@@ -1016,6 +1016,13 @@ input_parse(struct input_ctx *ictx, const u_char *buf, size_t len)
 		if (itr->handler != NULL && itr->handler(ictx) != 0)
 			continue;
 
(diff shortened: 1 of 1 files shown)


YayoRazo:
Closing. INPUT_DISCARD already bounds memory and lets the sequence terminate; forcing ground mid-sequence re-parses the remaining bytes as literal text (visible garbage for long CSI/OSC/DCS payloads). No actual bug here.

PR #5371, opened 2026-07-11
fix(hyperlinks): add per-set cap to prevent cross-pane eviction

YayoRazo:
The global MAX_HYPERLINKS cap of 5000 was shared across all panes, so a single pane could flood unique URIs and evict legitimate links from other panes.

Add a per-hyperlinks-set cap (MAX_HYPERLINKS_PER_SET = 1000) so one pane exhausting its own cap only affects its own links.

### DIFF
--- a/hyperlinks.c
+++ b/hyperlinks.c
@@ -42,6 +42,7 @@
 
 #define MAX_HYPERLINKS 5000
 #define MAX_HYPERLINK_URI 1024
+#define MAX_HYPERLINKS_PER_SET 1000
 
 static long long hyperlinks_next_external_id = 1;
 static u_int global_hyperlinks_count;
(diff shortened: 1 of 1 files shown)


YayoRazo:
Closing. The per-set cap makes hyperlinks_put return 0, so gc->link is 0 and the incoming link is dropped instead of the oldest being evicted, inverting the existing global LRU. Verified: emitting 1500 distinct OSC 8 links registers all 1500 on master but only 1000 with this patch, dropping the newest 500. hl->count only decreases on eviction or screen reset, so a pane emitting more than 1000 distinct links silently stops registering new ones until the screen clears. It also doesn't stop cross-pane eviction, which is still driven by the global 5000 cap. Wrong approach.

PR #5372, opened 2026-07-11
fix(server): log waitpid errors instead of killing the server

YayoRazo:
fatal() kills the entire tmux server on an unexpected waitpid() error, taking down all sessions.  Log the error and return instead, matching the graceful handling of ECHILD.


nicm:
This is fine as it is.

PR #5373, opened 2026-07-11
fix(server): use fstat/fchmod to avoid TOCTOU race on server socket

YayoRazo:
server_update_socket used stat() then chmod() on socket_path, creating a TOCTOU race. Use fstat()/fchmod() on the already-open server fd instead. Also resolve the -S flag path's parent directory with realpath() to prevent symlink-based socket hijacking.

### DIFF
--- a/server.c
+++ b/server.c
@@ -348,7 +348,7 @@ server_update_socket(void)
 	if (n != last) {
 		last = n;
 
-		if (stat(socket_path, &sb) != 0)
+		if (server_fd == -1 || fstat(server_fd, &sb) != 0)
 			return;
 		mode = sb.st_mode & ACCESSPERMS;
 		if (n != 0) {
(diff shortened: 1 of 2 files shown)


YayoRazo:
Closing. Verified both problems: the tmux.c -S hunk doesn't build (dirname/basename without <libgen.h>) and is out of scope, and the server.c change swaps chmod(socket_path) for fchmod(server_fd), but fchmod on a bound AF_UNIX socket fd doesn't re-permission the on-disk socket file. On Linux it returns 0 yet leaves the mode unchanged (it acts on the sockfs inode); on BSD it returns EINVAL. So the attach indicator would silently stop updating. The TOCTOU it targets also isn't reachable, since the socket lives in the 0700 per-uid directory make_label enforces.

PR #5374, opened 2026-07-11
fix(xmalloc): add overflow check to xcalloc

YayoRazo:
xcalloc() wraps calloc() but does not validate that nmemb * size fits in size_t before calling calloc().  Every other allocation wrapper in xmalloc.c already performs this check (xreallocarray delegates to reallocarray which checks internally).

Add nmemb > SIZE_MAX / size before calloc(), matching the pattern used by the compat reallocarray and recallocarray implementations.

### DIFF
--- a/xmalloc.c
+++ b/xmalloc.c
@@ -44,6 +44,8 @@ xcalloc(size_t nmemb, size_t size)
 
 	if (size == 0 || nmemb == 0)
 		fatalx("xcalloc: zero size");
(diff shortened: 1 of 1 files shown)


nicm:
I have made this change but I chose to use a slightly different form of the condition to match OpenSSH (where this file originally came from years ago). I also switched some `fatalx` to `fatal` also to match. This change is in OpenBSD now and should be in GitHub as soon as sync is fixed (hopefully this week). Thanks!

PR #5375, opened 2026-07-11
fix(tmux): use O_DIRECTORY and fstat to avoid TOCTOU race in make_label

YayoRazo:
make_label used lstat(base) then later constructed a path using base, creating a window where an attacker could swap the directory.  Open the directory with O_DIRECTORY and validate via fstat() on the fd instead.

### DIFF
--- a/tmux.c
+++ b/tmux.c
@@ -190,6 +190,7 @@ make_label(const char *label, char **cause)
 	u_int		  i, n;
 	struct stat	  sb;
 	uid_t		  uid;
+	int		  dir_fd;
 
 	*cause = NULL;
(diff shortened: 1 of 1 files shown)


YayoRazo:
Closing. lstat doesn't dereference the final component, so a symlinked tmux-<uid> hits the !S_ISDIR check and is rejected; open(base, O_DIRECTORY) without O_NOFOLLOW follows the symlink, so fstat sees the target directory and it passes instead. And the dir fd is closed before the socket is bound by pathname, so the TOCTOU window is unchanged. Net result: it drops the existing symlink guard without closing the race.

PR #5376, opened 2026-07-11
fix(job): use _exit instead of exit in child process after fork

YayoRazo:
In the child process path after fork(), fatal() calls exit() which invokes atexit handlers and flushes inherited stdio buffers.  Use _exit() instead, matching the pattern already used by spawn.c and cmd-pipe-pane.c.


nicm:

PR #5377, opened 2026-07-11
fix(signal): use volatile sig_atomic_t for signal flags

YayoRazo:
Signal-state flags in server.c and client.c were plain int instead of volatile sig_atomic_t.  Also add WUNTRACED to the client waitpid loop to match the server handler.


nicm:
I don't think you read this code.

PR #5378, opened 2026-07-11
fix(file): add O_NOFOLLOW to file operations

YayoRazo:
File operations in client.c and file.c followed symlinks without protection.  Add O_NOFOLLOW to open() calls for lock files, file write, and file read paths.  Resolve path traversal via realpath() in file_get_path.

### DIFF
--- a/client.c
+++ b/client.c
@@ -81,8 +81,10 @@ client_get_lock(char *lockfile)
 
 	log_debug("lock file is %s", lockfile);
 
-	if ((lockfd = open(lockfile, O_WRONLY|O_CREAT, 0600)) == -1) {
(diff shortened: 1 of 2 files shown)


nicm:
Lock files maybe, but what if the user wants to open a symlink? Won't this break that?

YayoRazo:
You're right, it does break that. With O_NOFOLLOW, save-buffer/pipe-pane to a symlinked path fails with ELOOP ("Too many levels of symbolic links") where before it wrote through the link. And the realpath() step only runs on the relative-path branch for files that already exist, so it doesn't achieve the traversal goal either. Closing.

PR #5379, opened 2026-07-11
fix(file): add O_CLOEXEC to file read and write open calls

YayoRazo:
File descriptors opened in file_write_open and file_read_open were missing O_CLOEXEC, so they could leak across exec.  Add O_CLOEXEC to the open() flags.

### DIFF
--- a/file.c
+++ b/file.c
@@ -599,7 +599,7 @@ file_write_open(struct client_files *files, struct tmuxpeer *peer,
 
 	cf->fd = -1;
 	if (msg->fd == -1)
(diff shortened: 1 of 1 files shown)


nicm:
What forks do we have that do not call closefrom?

YayoRazo:
None — every fork+exec path already calls closefrom(STDERR_FILENO+1) (client.c:505, job.c:180, spawn.c:472, cmd-pipe-pane.c:163), and these two opens are in the client, which doesn't exec. So the fds can't leak across exec and O_CLOEXEC adds nothing here. Closing — thanks.

PR #5380, opened 2026-07-11
fix(cmd): add recursion and length guards to command parser

YayoRazo:
Add a maximum nesting depth for %if/%elif blocks in config files, a maximum token length in the lexer, and an underflow guard in cmd_unpack_argv to protect against malformed packed buffers.


YayoRazo:
Closing. On master, deeply nested %if is already handled gracefully: bison's parser stack hits "memory exhausted" and returns a normal parse error without crashing the server (verified with 5000 nested %if). So the recursion cap isn't solving a reachable problem, and the token-length and cmd_unpack_argv guards are unreachable or redundant. Withdrawing.

PR #5381, opened 2026-07-11
fix(control,server-client): add length guards to input handlers

YayoRazo:
Add a buffer length check before evbuffer_readln in control mode to prevent OOM from a single huge line.  Add max length checks on MSG_IDENTIFY_TERM and MSG_IDENTIFY_CWD to bound term name and cwd path lengths.

### DIFF
--- a/control.c
+++ b/control.c
@@ -457,6 +457,11 @@ control_read_callback(__unused struct bufferevent *bufev, void *data)
 	enum cmd_parse_status	 status;
 
 	for (;;) {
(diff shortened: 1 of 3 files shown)


nicm:
What problem is this solving?

YayoRazo:
Closing. The term/cwd guards are redundant with the imsg frame bound (MAX_IMSGSIZE), and the control_wait_exit hunk references locals that don't exist in that function. No real problem being solved — sorry for the noise.

Issue #2354, opened 2020-08-24
[Feature Request] Add style for copy_cursor_word and copy_cursor_line


kevinhwang91:
### Issue description

Similar to highlight under keyword and cursor line in vim, It will improve efficiency in copy mode.

nicm:
I have no idea what this means.

kevinhwang91:
![image](https://user-images.githubusercontent.com/17562139/91058652-4b6bf480-e65b-11ea-874e-79855fc8ce7c.png)

`split` is the `keyword` under the current cursor, `bind s split-window -v` is the cursor line.

nicm:
I don't have time to do this now, I've added it to the todo list, thanks.

PR #5391, opened 2026-07-14
Add copy-mode-current-line-style

leohenon:
Implements cursor-line part of #2354.

Adds `copy-mode-current-line-style` to set the style of the line containing the cursor in copy mode.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -1430,6 +1430,15 @@ const struct options_table_entry options_table[] = {
 	  .text = "Style of current line number in copy mode."
 	},
 
+	{ .name = "copy-mode-current-line-style",
+	  .type = OPTIONS_TABLE_STRING,
+	  .scope = OPTIONS_TABLE_WINDOW,
+	  .default_str = "default",
+	  .flags = OPTIONS_TABLE_IS_STYLE,
+	  .separator = ",",
+	  .text = "Style of the line containing the cursor in copy mode."
+	},
+
 	{ .name = "copy-mode-line-number-style",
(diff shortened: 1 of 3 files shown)


nicm:
Thanks but I think it would be better if we could pass an `is_cursor_line` variable into the existing styles that you could check, instead of adding another entire option.

leohenon:
Sounds good, reworked to check `is_cursor_line` in `window-style` and `window-active-style` instead. Both variants are cached with the pane styles, so there is no cost when the variable is not used.

nicm:
Sorry, I thought this was for one of the existing copy mode styles. It can't work like this, `tty.c` cannot know about copy mode. 

Wouldn't it be enough just to use it change `copy-mode-line-number-style` or something? Does it actually need to change the whole line?

I think you might need to add something like `copy-mode-line-style` after all...

leohenon:
Reverted to the previous copy-mode style option, applied in `window-copy.c` when writing the line without touching `tty.c`.

It does need the whole line so the highlight can track the cursor across the full width, not just the number. 

A style option also means it takes any style (underline, bold, reverse), not just colours, which is why I did it that way originally. 

Named it `copy-mode-current-line-style` to match `copy-mode-current-match-style` and `copy-mode-current-line-number-style`.

leohenon:
With `copy-mode-current-line-style 'bg=#3c3c3c'`:

current-line-style1

nicm:
I think this is OK, got confused about what you were trying to do... will look at it again later.

nicm:

PR #5392, opened 2026-07-14
Do not refresh copy mode during synchronized output

barrettruth:
## Problem

copy-mode auto refresh can redraw in the middle of DEC synchronized output, exposing an incomplete application frame as flicker.

## Solution
defer automatic refresh while synchronized output is active, leaving the pending update to refresh normally after the synchronized frame ends.

related #5165

@grothesque LMK if this helps

### DIFF
--- a/regress/copy-mode-refresh-sync.sh
+++ b/regress/copy-mode-refresh-sync.sh
@@ -1,0 +1,90 @@
+#!/bin/sh
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -LtestA$$ -f/dev/null"
+$TMUX kill-server 2>/dev/null
+
+cleanup()
+{
+	$TMUX kill-server 2>/dev/null
+	rm -rf "$TMP"
+}
+fail()
+{
+	echo "$1"
+	cleanup
+	exit 1
+}
+wait_done()
+{
+	i=0
+	while [ $i -lt 50 ]; do
+		[ -f "$DONE" ] && return
+		i=$((i + 1))
+		sleep 0.1
+	done
+	fail "pane command did not finish"
+}
+wait_sync()
+{
+	i=0
+	while [ $i -lt 50 ]; do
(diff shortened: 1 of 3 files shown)


grothesque:
It still flickers, but perhaps a bit less heavily.

The following patch gets rid of the flickering completely:
```diff
diff --git a/window-copy.c b/window-copy.c
index 654bbd4c..e672508a 100644
--- a/window-copy.c
+++ b/window-copy.c
@@ -3051,8 +3051,7 @@ window_copy_refresh_timer(__unused int fd, __unused short events, void *arg)
                follow = (data->oy == 0 &&
                    data->cy == screen_size_y(&data->screen) - 1);
                window_copy_do_refresh(wme, follow);
-               window_copy_redraw_screen(wme, 1);
-               /* The timer runs outside key handling, so force a repaint. */
+               /* window_copy_size_changed() rebuilt the screen; repaint it. */
                wp->flags |= PANE_REDRAW;
                wp->flags &= ~PANE_UNSEENCHANGES;
        }
```

But I believe that it might be too heavy-handed.

I don’t know the tmux code base, and I unfortunately don’t have the time to dig into it, but perhaps the above gives you a hint.

Here is what the LLM says about why your fix (i.e. this PR) is insufficient:

----------------

The proposed `MODE_SYNC` check reduces the flicker, but only prevents refresh while a synchronized update is actively in progress. Pi emits many short synchronized updates, so the timer frequently runs between completed updates. It then performs an immediate full copy-mode redraw and also sets `PANE_REDRAW`, potentially causing a second repaint in a separate synchronized transaction. Experimentally removing the immediate `window_copy_redraw_screen()` call eliminates the flicker completely. That change needs refinement because the mode screen’s cursor and search highlighting must still be finalized before the scheduled repaint.

grothesque:
The fix proposed above indeed removes the flickering, but it also prevents the copy-mode cursor from being redrawn upon auto-refresh.  Otherwise it feels OK, but I haven’t analyzed it in-depth.

I posted a simple flicker-reproducing script in https://github.com/tmux/tmux/pull/5165#issuecomment-4969964229.

barrettruth:
it sounds like you can take this over.

PR #5395, opened 2026-07-14
Minor doc fixes

mathstuf:


nicm:
I think you missed a line here:
````diff
@@ -4087,6 +4088,8 @@ may be followed by
 to specify a percentage of the available space.
 .Fl p
 is a shorthand option for this.
+.Fl B
+sets the pane border lines for floating panes; see
 The
 .Fl b
 option causes the new pane to be created to the left of or above

nicm:
Good stuff - applied to OpenBSD now, will be in GitHub later. Thanks!

PR #5397, opened 2026-07-15
Do not loop forever in tty_draw_line on zero-width cells not marked as padding (#5024 follow-up)

### LINKED ISSUE #5024: tmux server livelock in tty_draw_line under control-mode load (separate from #4969)

buihongduc132:
## Summary

Commit `281e8ff7` (2026-05-13) closed #5024 by making `tty_draw_line_get_empty()` return `1` for orphan `GRID_FLAG_PADDING` cells, so the draw loop advances instead of spinning on `i += gcp->data.width` (which is `0` for padding).

That fix only covers the padding case. A **non-padding** cell whose `utf8_data.width == 0` still falls through every clause of `tty_draw_line_get_empty()`, returns `empty == 0`, and the loop then executes:

```c
if (empty != 0)
    i += empty;
else
    i += gcp->data.width;   /* width == 0 -> i never advances -> livelock */
```

`i` never moves, the `for (;;)` in `tty_draw_line()` never terminates, one core pegs at 100% (state `Rs`), every socket command hangs, and the only recovery is `kill -9` on the server.

Such cells appear in practice when a **variation selector (U+FE0F), a combining mark, or other zero-width code point** is emitted standalone into a coloured cell and is not folded into the preceding base character — exactly the output pattern of the animated, emoji-bearing status lines produced by several AI CLI TUIs (pi, hermes, codex) and the control-mode pollers described in #5024.

## Fix

Treat a zero-width cell as a one-cell empty advance, mirroring the existing padding clause:

```c
if (gc->data.width > nx)
    empty = nx;
else if (gc->flags & GRID_FLAG_PADDING)
    empty = 1;
else if (gc->data.width == 0)        /* new */
    empty = 1;
else if (gc->flags & GRID_FLAG_SELECTED)
    empty = 0;
...
```

The cell is visually dropped (one column redrawn as space-equivalent) and the next full redraw self-corrects it. This matches how padding is already handled and avoids changing the loop's structure.

## Reproducer

`tools/tty-draw-zero-width-livelock-repro.sh` is a deterministic reproducer. It starts a throwaway server on a private socket (`-L repro-\$\$`, the real server is never touched), injects the exact pathological cell (UTF-8 `EF B8 8F`, `size = 3`, `width = 0`, non-padding, `bg = 2`) via gdb, forces a redraw, and measures. Struct offsets are auto-computed from the source tree so it tracks ABI changes.

## Verification (master HEAD `15746a1b`, 2026-07-15)

Back-to-back, same script, same injected cell:

| binary | `refresh-client` rc | CPU ticks / 3s | `display` rc | `ps` state | verdict |
|--------|---------------------|----------------|--------------|------------|---------|
| master unpatched | **124 (timeout)** | **569** | **124 (timeout)** | **`Rs`** | livelocked |
| with this patch | 0 | 0 | 0 | `Ss` | survives |

Only the private test server was affected (killed on exit); the real server was untouched.

Reproducer requirements: `gdb`, passwordless `sudo` (ptrace attach), the tmux source tree (for the offsetof probe).

## Notes

- This is a residual gap left by #5024's fix, not a regression of it.
- I can drop the `tools/` reproducer commit if you would rather not carry a gdb-based diagnostic in the tree; happy to keep just the one-line fix.
- I have not attempted to address the underlying question of how a zero-width code point ends up as a standalone, non-padding cell in the grid — only the immediate livelock, consistent with how the padding case was handled.

### DIFF
--- a/tools/tty-draw-zero-width-livelock-repro.sh
+++ b/tools/tty-draw-zero-width-livelock-repro.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+# ══════════════════════════════════════════════════════════════════════════════
+# Deterministic reproducer for the tty_draw_line livelock caused by a
+# non-padding, zero-width grid cell.
+#
+# Context
+#   Upstream tmux commit 281e8ff7 (2026-05-13) fixed the #5024 livelock but
+#   only for GRID_FLAG_PADDING (orphan padding) cells: those now return
+#   empty=1 from tty_draw_line_get_empty() and the draw loop advances by 1.
+#
+#   A *non-padding* cell with utf8_data.width == 0 (e.g. a stray variation
+#   selector U+FE0F, a combining mark, or any zero-width code point that the
+#   grid did not fold into the preceding cell) is NOT classified empty: it
+#   falls through every clause in tty_draw_line_get_empty() and the draw loop
+#   then executes `i += gcp->data.width;` with width == 0. i never advances,
+#   the for(;;) never terminates, and the tmux server livelocks on one core
+#   (state Rs). The only recovery is kill -9 on the server PID.
+#
+#   This reproducer plants that exact pathological cell directly into the live
+#   server grid via gdb, then forces a redraw. There is no timing window: an
+#   unpatched server livelocks every run; a patched server survives every run.
+#
+# Requirements
+#   gdb, passwordless sudo (ptrace attach), the tmux source tree (for struct
+#   offsets, auto-computed below), a built ./tmux binary.
+#
+# Safety
+#   Runs on a private throwaway socket (-L repro-$$). The real 'default'
+#   server is never touched. On livelock the server can only be recovered
+#   with kill -9 — we apply that only to the private PID.
+#
+# Usage
+#   tools/tty-draw-zero-width-livelock-repro.sh ./tmux        # RED or GREEN
+#   tools/tty-draw-zero-width-livelock-repro.sh /path/to/tmux
+# ══════════════════════════════════════════════════════════════════════════════
+set -uo pipefail
+
+TMUX_BIN="${1:-./tmux}"
+[ -x "$TMUX_BIN" ] || { echo "not executable: $TMUX_BIN" >&2; exit 2; }
+
+# Resolve the tmux source tree (for offsetof probe) relative to this script.
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SRC_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+[ -f "$SRC_DIR/tmux.h" ] || { echo "tmux.h not found in $SRC_DIR" >&2; exit 2; }
+
+SOCK="repro-$$"
+log() { printf '\033[1;34m[repro]\033[0m %s\n' "$*" >&2; }
+
+command -v gdb >/dev/null || { echo "gdb required" >&2; exit 2; }
+
+log "target binary: $TMUX_BIN ($("$TMUX_BIN" -V 2>&1))"
+log "source tree:   $SRC_DIR"
+log "private socket: $SOCK  (your 'default' server is NOT touched)"
+
+# ── Auto-compute struct offsets for this exact build ──────────────────────────
+OFFSETS_PROBE="$(mktemp -d)/offsets.c"
+cat > "$OFFSETS_PROBE" <<'EOF'
+#include <stddef.h>
+#include <stdio.h>
+#define TMUX_H_INSIDE 1
+#define TMUX_NAME "__probe__"
+#include "tmux.h"
+int main(void) {
+    printf("WP_BASE=%zu\n", offsetof(struct window_pane, base));
+    printf("SCREEN_GRID=%zu\n", offsetof(struct screen, grid));
+    printf("GC_SIZE=%zu\n", sizeof(struct grid_cell));
+    printf("GC_DATA=%zu\n", offsetof(struct grid_cell, data.data));
+    printf("GC_USIZE=%zu\n", offsetof(struct grid_cell, data.size));
+    printf("GC_UWIDTH=%zu\n", offsetof(struct grid_cell, data.width));
+    printf("GC_BG=%zu\n", offsetof(struct grid_cell, bg));
+    return 0;
+}
+EOF
+OFFSETS_BIN="${OFFSETS_PROBE%.c}"
+if ! gcc -I"$SRC_DIR" -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 \
+        -DTMUX_VERSION='"probe"' -DTMUX_TERM='"xterm"' -DTMUX_MOUSE=1 \
+        -std=gnu99 -O0 "$OFFSETS_PROBE" -o "$OFFSETS_BIN" 2>/dev/null; then
+    echo "failed to compile offsets probe against $SRC_DIR" >&2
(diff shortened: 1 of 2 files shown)


nicm:
OK this could be a problem, but one can patch any sort of stuff in with `gdb`, do you actually see this in real use?

Is it possible to make a regression test that doesn't use `gdb` to set up the grid? That is, by sending text to the pane or resizing or whatever is needed.

PR #5399, opened 2026-07-15
feat(search): show current and total matches in copy mode search

felix314159:
The one thing I always disliked about tmux was how when searching it would say e.g. '53 results' and you would not know whether you are looking at result 30/53, 45/53 or 53/53. This PR fixes the behavior to match what you would expect from any popular browser:

 - "Phrase not found" for zero matches
 - "1 of 1 match" for one match
 - "N of M matches" for multiple matches

So this PR tracks the current copy-mode search match and exposes it through the `search_count_current` format variable. The current match remains accurate across forward and backward navigation, wraparound, cursor movement, resizing, and both vi and emacs copy modes. Partial counts continue to be shown as "N+ matches". I also added regression coverage for the new formatting and navigation behavior.


nicm:
Some comments:

I think `searchpositions` doesn't need to be `uint64_t` and could probably be `u_int` (obviously you need to check the positions will fit and in the event they don't, don't track).

I think you should probably shorten all the names to `searchpos` or `matchpos` since stuff like `searchpositionallocated` is very long. You have a fair few lines >80 columns that need to be shortened.

If we do an early stop, `searchpositions` is set to `NULL` but a `NULL` causes us to do a full check if we search again instead of a partial check. This is going to be slower.

felix314159:
thanks for the feedback! should be fixed now. lmk if there is anything else to do

nicm:
OK works well, but I don't like "Phrase not found", can it just say "0 matches", everyone knows what "0" means, right?

I think a helper function to do the `xreallocarray` like `window_copy_grow_searchpos` might be nice because that code is very nested now.

felix314159:
i agree, '0 matches' is a bit cleaner. i had copied what firefox does so that's where the 'phrase not found' came from in case you were wondering :) helper function is also introduced now

Issue #5223, opened 2026-06-15
When all panes in a window are minimised, `show-pane` needs the window to be qualified to identify a pane to target.


daneofmanythings:
Reproduction: open fresh tmux, `hidep; showp -t0`. Tmux will not be able to find the pane. There is some targetting logic that appears to exist to enable shortcutting pane targetting which is probably not applying in this case.

PR #5402, opened 2026-07-16
feat(master): Added `hide-pane` and `show-pane`.

daneofmanythings:
There are two small annoyances with the current implementation.
1. #5223. I want to fix this before merging if possible.
2. The last visible pane cannot be hidden. This would result in the active pane being NULL. This can be fixed, but it would require a scope of change beyond this PR. A non-NULL active pane invariant is enforced on `cmd-find.c:1004`. I believe allowing a NULL active pane would require NULL checks in all the consumers of `cmd_find_target`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -103,6 +103,7 @@ dist_tmux_SOURCES = \
 	cmd-display-message.c \
 	cmd-find-window.c \
 	cmd-find.c \
+	cmd-hide-pane.c \
 	cmd-if-shell.c \
 	cmd-join-pane.c \
 	cmd-kill-pane.c \
--- a/cmd-hide-pane.c
+++ b/cmd-hide-pane.c
@@ -1,0 +1,242 @@
+/* $OpenBSD$ */
+
+/*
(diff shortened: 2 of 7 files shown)


nicm:
For this:
~~~~
		cmdq_error(item, "can't show a pane while window is zoomed");
~~~~
Why not just unzoom? Isn't that what we normally do?

Is having this (hidden panes) going to make layout changes harder? I don't think we have a pressing need for hidden panes if they are going to cause any problems and could be done later.

daneofmanythings:
Right, yes. I don't recall why that is in there. I'll adjust that.

Adding hidden panes will make layouts a little more complicated and will necessarily make adding features take more consideration. I'm not sure as to the magnitude of the effect, but there is something I should probably audit in this PR, and that is `window_count_panes`. It was given a parameter to interact with floating panes, and will need one for hidden as well. It feels like the function should be broken up into how it is used, instead of making a flag soup. It looks like it cares about counting visible, tiled, and all panes. I couldn't see any other immediate effects.

This feature can certainly wait as its impact on complexity is non-zero!

PR #5405, opened 2026-07-16
Add kitty keyboard protocol support (finish #4912)

brynbellomy:
Follow-on to #4912 -- intends to address the remaining feedback from @nicm.


nicm:
Thanks for picking this up. I will take a look at it at some point soon.

tty-keys.c
```diff
@@ -846,6 +857,21 @@ tty_keys_next(struct tty *tty)
...
+	/* Is this a kitty keyboard protocol key? */
+	if (tty->kitty_supported_flags != 0) {
+		switch (tty_keys_kitty(tty, buf, len, &size, &key)) {
+		case 0:		/* yes */
+			goto complete_key;
+		case -1:	/* no, or not valid */
+			break;
+		case 1:		/* partial */
+			goto partial_key;
+		case -2:	/* release event, discard */
```
musjj, line 869: I hope this PR would address release events.

nicm:
Can you rebase this on to `master` and fix conflicts?

There is no need to change `KEYC_MASK_MODIFIERS` etc, there is already room to add `KEYC_SUPER` and `KEYC_HYPER`.

I don't think release should be done now, it should be a separate change on top of this because it will need more changes to be able to represent release keys.

nicm:

mgrant0:
Just looking at this, I notice you use this term 'legacy' a to refer to the plain unextended way of sending key srokes.  I would avoid this term `"force_legacy"` and negate the logic and use `extended_encoding` (or `enable_extended_encoding`) instead.

mgrant0:
@brynbellomy Are you still working on this?

PR #5407, opened 2026-07-18
Add ability to collapse commmands to copy-mode.

### LINKED ISSUE #5237: Feature request: forward OSC 133 (semantic prompt) sequences to the outer terminal by default
#### Background

Modern terminal emulators (Ghostty, kitty, WezTerm, iTerm2, VS Code's terminal,
etc.) implement [OSC 133 "semantic prompt" / shell-integration
sequences](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md)
(`OSC 133 ; A`, `; B`, `; C`, `; D ; <exit-code>`). These power features such as
prompt navigation (jump to previous/next prompt), command duration, exit-status
indicators, and output folding.

When the shell runs inside tmux, tmux sits between the shell and the outer
terminal and does not forward OSC 133 to the outer terminal. As a result, the
outer terminal's shell-integration features silently stop working inside tmux.

#### The problem with the current workaround

Today the only way to get these sequences out to the outer terminal is to set
`allow-passthrough on` (or `all`) **and** have every shell-integration script
manually wrap each sequence in tmux's DCS passthrough envelope
(`\ePtmux;<payload-with-each-ESC-doubled>\e\\`).

That means each terminal vendor has to ship tmux-aware shell integration that:

1. Detects whether it is running inside tmux (`$TMUX`).
2. Doubles every `ESC` byte in the payload.
3. Wraps the whole thing in `\ePtmux; ... \e\\`.

This is fragile and duplicated across every terminal and every shell (bash, zsh,
fish, ...). For example, to make Ghostty's fish shell integration work inside
tmux today I had to patch `ghostty-shell-integration.fish` so that its OSC 133
sequences are wrapped in the tmux passthrough envelope. That patch
lives inside the Ghostty `.app` bundle and is **wiped on every Ghostty update**,
so it has to be re-applied constantly. Every user of every terminal hits the
same papercut.  FYI: https://github.com/ghostty-org/ghostty/discussions/11433

The passthrough mechanism also has downsides as a transport for this:

* It is opt-in and off by default, so it does not "just work".
* It bypasses tmux entirely, so tmux has no chance to translate the sequence,
  scope it to the active pane, or strip it when a pane is not visible — the
  payload is forwarded verbatim regardless of pane state (except for the
  on/all visibility distinction).
* It forces the *application* to know it is inside tmux, which is exactly the
  abstraction tmux normally provides for the application.

#### Proposed feature

Have tmux understand OSC 133 natively and forward it to the outer terminal,
ideally by default (or behind a dedicated, easy-to-discover option), without
requiring applications to wrap their sequences in DCS passthrough.

Possible shapes (any one of these would solve the pain):

1. **Native pass-through of OSC 133 specifically.** tmux already special-cases a
   number of OSC sequences (e.g. OSC 52 clipboard via `set-clipboard`, OSC 4/10/11
   colors, OSC 7/8). OSC 133 could be handled the same way: parse it, associate
   it with the active pane, and forward it to the outer terminal when that pane
   is visible. A server/session option (for example a new value for an existing
   option, or a dedicated `allow-osc133`/`forward-osc133` toggle) could gate it,
   defaulting to on for capable outer terminals.

2. **A generic "forward these OSCs" allowlist.** An option such as
   `forward-osc 133,...` listing OSC codes that tmux should transparently forward
   to the outer terminal, so users/terminals don't have to wrap them in
   passthrough manually.

3. **Make `allow-passthrough` unnecessary for OSC 133** by recognizing the
   sequence and forwarding it, so terminal vendors can drop the tmux-specific
   wrapping branch from their shell-integration scripts entirely.

#### Why tmux is the right place to fix this

The wrapping logic is identical for every terminal and every shell, and it is
purely about getting a sequence *through tmux*. Putting the knowledge in tmux
once removes N×M duplicated, update-fragile workarounds (N terminals × M shells)
and lets shell-integration scripts emit the standard OSC 133 sequence
unconditionally, exactly as they do when no multiplexer is present.

#### Steps to reproduce the current limitation

1. Use a terminal that implements OSC 133 shell integration (e.g. Ghostty).
2. Confirm prompt navigation / command marks work **outside** tmux.
3. Start tmux **without** any passthrough wrapping in the shell integration.
4. Observe that the outer terminal's OSC 133 features no longer work inside
   tmux, because tmux does not forward the sequences.
5. The only fix today is `allow-passthrough on` **plus** manually wrapping every
   OSC 133 sequence in `\ePtmux;...\e\\` inside the shell-integration
   script.

### Required information

* **tmux version** (`tmux -V`): `tmux 3.6b`
* **Platform** (`uname -sp`): `Darwin arm64` (macOS, Apple Silicon)
* **Terminal in use**: Ghostty
* **$TERM inside tmux** (`echo $TERM`): `tmux-256color`
* **$TERM outside tmux** (`echo $TERM`): `xterm-ghostty`
* **Logs from tmux** (`tmux kill-server; tmux -vv new`): N/A — this is a feature
  request about intended behavior, not a crash or incorrect-behavior bug, so
  there is no faulty run to attach logs for. The behavior described above is by
  design in current tmux (OSC 133 is not forwarded; `allow-passthrough` defaults
  to `off`).

### Notes

* Relevant tmux option today: `allow-passthrough [on | off | all]`.
* Relevant prior art inside tmux: existing special handling of OSC 52, OSC 4/10/11,
  OSC 7 and OSC 8.
* This request is about OSC 133 primarily; OSC 7 (working directory) has partial
  handling already, but the same "forward to outer terminal" need applies and
  could share the mechanism.

mgrant0:
New copy-mode entry options:

  - copy-mode -c — enter with output collapsed and controls visible.
  - copy-mode -U — enter uncollapsed with controls visible.
  - copy-mode — unchanged initial view: uncollapsed, controls hidden by default.

  New copy-mode commands:

  - collapse-output — collapse output for the marked command at the cursor.
  - uncollapse-output — show output for that command.
  - Add -a to either command to affect every command output.

  Default bindings:

  - Tab — toggle the current command’s output.
  - C-Tab — toggle all command outputs.
  - Click the control/line-number area to toggle its command.

  New window options:

  - copy-mode-collapse-controls — controls visibility of the >/V column:
      - on-demand (default): show with copy-mode -c/-U, or after a collapse/uncollapse action.
      - always: always show it, even when no OSC 133 markup exists.
      - never: never show it; keyboard commands still work.

  - copy-mode-collapse, replacing collapse-output.

  It accepts one or more letters from A–D (invalid values are rejected); default is C. 

  | Letter | Hidden section |
  |---|---|
  | `A` | Prompt: A→B |
  | `B` | Entered command: B→C |
  | `C` | Command output: C→D |
  | `D` | Text after command output: D→next A |

  Examples:

```tmux
  set -w copy-mode-collapse C    # default: hide command output only
  set -w copy-mode-collapse CD   # hide command output and separator text
  set -w copy-mode-collapse CDA  # hide output, separator, and prompts; show commands only
```

  Bash Configuration (~/.bashrc)
In Bash, you use PROMPT_COMMAND to run code immediately before the prompt is drawn, and PS0 to run code immediately before a command executes. Non-printing characters in prompts are wrapped in \[ and \].

Bash

1. Output Start (C)
PS0 is evaluated right after you press Enter, but before the command runs.

```
PS0='\e]133;C\a'
```

2. Command Finished (D)
PROMPT_COMMAND runs right before PS1 is drawn, covering empty Enters.

```
__osc133_prompt_command() {
    # Capture the exit code of the last command immediately
    local exit_code=$?
    
    # Emit the D marker
    printf '\e]133;D;%s\a' "$exit_code"
    
    # (If you have custom pre-prompt output, echo it here)
}
PROMPT_COMMAND=__osc133_prompt_command
```

3. Primary Prompt (A and B)
Strip any existing markers, then wrap your prompt.
Replace '\u@\h$ ' with your actual visible prompt.

```
PS1="${PS1//\\[\\e\]133;*\\a\\]/}"
PS1='\[\e]133;A\a\]\u@\h$ \[\e]133;B\a\]'
```

4. Secondary Prompt (A;k=s and B)
Strip any existing markers, then wrap the continuation prompt.

```
PS2="${PS2//\\[\\e\]133;*\\a\\]/}"
PS2='\[\e]133;A;k=s\a\]> \[\e]133;B\a\]'
```

Zsh Configuration (~/.zshrc)

In Zsh, you use the built-in hook functions precmd (runs before the prompt) and preexec (runs before the command). Non-printing characters in prompts are wrapped in %{\ and \%\}.

1. Command Finished (D)
precmd runs before the prompt is drawn, covering empty Enters.

```
precmd() {
    # Capture the exit code of the last command immediately
    local exit_code=$?
    
    # Emit the D marker
    printf '\e]133;D;%s\a' "$exit_code"
    
    # (If you have custom pre-prompt output, echo it here)
}
```

2. Output Start (C)
preexec runs exactly after pressing Enter, before command execution.

```
preexec() {
    printf '\e]133;C\a'
}
```

3. Primary Prompt (A and B)
Strip any existing markers, then wrap your prompt. 
Using $'' allows Zsh to interpret the \e escape character.
Replace '%n@%m$ ' with your actual visible prompt.

```
PS1=${PS1//$'%{\e]133;'*'\a%}'/}
PS1=$'%{\e]133;A\a%}%n@%m$ %{\e]133;B\a%}'
```

4. Secondary Prompt (A;k=s and B)
Strip any existing markers, then wrap the continuation prompt.

```
PS2=${PS2//$'%{\e]133;'*'\a%}'/}
PS2=$'%{\e]133;A;k=s\a%}> %{\e]133;B\a%}'
```


nicm:
OK a few things:

- This is not worth adding 24 bytes to every `struct grid_line`:
  - Why are the cursor positions needed? You can't collapse less than one line.
  - `end_output`, `source_line`, `output_line` and `GRID_LINE_OUTPUT_COLLAPSED` and `GRID_LINE_OUTPUT_MEMBER` belong to copy mode, these should not live in `struct grid_line` but be maintained separately by copy mode.

- Please get rid of the `copy-mode-collapse` option. This feature really does not need options. IMO the default of `C` is OK as the only choice but if you really want the other views, have `Tab` cycle through them.

- Can you change "uncollapse" to "expand"? "uncollapse" is not a common word.

- Don't use `V` as an expanded marker. Either use actual ACS arrows like we do for the pane indicators, or use `+` and `-` like in tree mode.

- `GRID_LINE_START_SECONDARY_PROMPT` is very long, how about just `GRID_LINE_START_PROMPT2`? I am not entirely sure we need to bother caring about secondary prompts at all though, are they widely used?

- The secondary check in `input.c` probably needs to check that `k=s` is terminated by `\0` or `;`.

- I think itd be nicer if this lived on the right side of line numbers, not the left (or be part of them eg in the space between line number and screen content).

- `C-Tab` is not always supported so `BTab` (`S-Tab`) or `M-Tab` might be better.

mgrant0:
The cursor positions within the lines (hence within the grid) are needed because you need to know where the data in the line starts:

```
pre-prompt output [A] user@host ~$ [B] ls -1 (ENTER)
[C] file1.txt
file2.txt
[D]
pre-prompt output [A] user@host ~$ [B]
```

First you need to isolate where the actual command starts at B and in normal cirumstances C starts at the beginning of a line.  D also generally happens after a new line but if the program didn't output a newline, it may be after the last line of the output.  It is also possible to have text after the D and before the A known as pre-prompt output which some people use as a separator (like me).  Lastly, you have mult-line prompts that use $PS2 which look like this:

```
[A] user@host ~$ [B] for i in 1 2 3; do
[A,k=s]> [B]echo $i
[A],k=s> [B]done
[C] file1.txt
file2.txt
[D]
[A] user@host ~$ [B]
```

So we need to know the offset in the line.  However, one simplification I did was to not track multiple ABCD tags per line, so if the tty got put in that annoying mode where pressing enter didn't output a linefeed and you got:

```
user@host ~$ user@host ~$ user@host ~$ 
```

it only recognizes the first A and B tags and the last C tag on the line basically seeing such lines as a single prompt.  If there's output in the middle of that line, well, that's an edge case and ends up being part of the prompt.

So in short, each grid line can have a A, B, C, and D tags and they can be at any column in the line but you can only have one of each in the line.

Working through the rest of your comments.  Thanks.

nicm:
I don't really want to store 12 more bytes for every single line for this, plus another one for the exit code.

We could:

- Forget the x position and always fold in full lines not partial lines (so part of the output is on the prompt line it is still shown). I think this would be fine for folding but not so nice for command view where it'd be nice to copy and get the exact output.

- Make them each `uint8_t` and only support up to column 255. This might be the best compromise because then we are only using four bytes, and how many people have prompts beyond column 255?

- Store the data separately with a link via a `uint16_t` or `uint32_t` command number. This would be most efficient especially if there might be more data later, but means extra bookkeeping.

mgrant0:
This PR has been moved to this other PR: https://github.com/tmux/tmux/pull/5415

input.c
```diff
@@ -3239,25 +3239,41 @@ input_osc_133(struct input_ctx *ictx, const char *p)
...
+	int			 secondary, status;
 
 	if (line <= gd->hsize + gd->sy - 1)
 		gl = grid_get_line(gd, line);
 
 	switch (*p) {
 	case 'A':
-		if (gl != NULL)
-			gl->flags |= GRID_LINE_START_PROMPT;
+		secondary = (strstr(p + 1, ";k=s") != NULL);
```
mgrant0, line 3252: This needs to be more general.  There are other properties and the order doesn't matter and there can be multiple of them.  Probably need to just loop through for the ones we are interested in and their synonyms.

Issue #4969, opened 2026-04-02
Bug: `tty_draw_line` enters long running loop under specific circumstances.


daneofmanythings:
### Issue description

A long loop is entered in `tty_draw_line` locking up the server. I have not let it run to completion as I am not sure how long that would take. It appears there is an abnormal state entered that avoids exit conditions. I have bisected the git commit history and found this issue was introduced in commit `d7f59eccaf7f9435065017fa214a7b3c53bc631d`.

Here is how to reproduce:
- tmux.conf:
```
set status 2
set unknown-option
```
- Start tmux
- From tmux command line run `source tmux.conf`

### Required information

* tmux version: `tmux next-3.7`
* Platform: `Linux unknown` (Void linux x86_64)
* Terminal in use: `st 0.9.3`
* $TERM *inside* tmux: `tmux-256color`
* $TERM *outside* tmux: `st-256color`.
* Logs are from letting the loop briefly run. Note that I had to truncate the server log from 2.4 million lines to 250k to upload it.

[tmux-client-23322.log](https://github.com/user-attachments/files/26450498/tmux-client-23322.log)
[tmux-out-23324.log](https://github.com/user-attachments/files/26450497/tmux-out-23324.log)
[tmux-server-23324.log](https://github.com/user-attachments/files/26450534/tmux-server-23324.log)

nicm:
I can't reproduce with your instructions. Something to do with terminal or terminal dimensions maybe?

Hard to see how it would depend on it being `st` but maybe see if it happens in other terminals?

nicm:
`nx` is `-1` which is not right:
~~~~
1775173395.626893 tty_draw_line: px=0 py=1 nx=4294967295 atx=0 aty=36
1775173395.626897 tty_draw_line: drawing 0-4294967295,1 (end 4294967295) at 0,36; defaults: fg=8, bg=8
~~~~
Can you do this and see where the backtrace is and maybe poke around in the next frame up and see if anything looks odd?
~~~~
--- tty-draw.c  23 Mar 2026 08:48:32 -0000      1.5
+++ tty-draw.c  3 Apr 2026 08:08:58 -0000
@@ -143,6 +143,9 @@ tty_draw_line(struct tty *tty, struct sc
        log_debug("%s: px=%u py=%u nx=%u atx=%u aty=%u", __func__, px, py, nx,
            atx, aty);

+       if (nx > 100000)
+               abort ();
+
        /*
[...]

PR #5409, opened 2026-07-19
Avoid unsigned underflow in window_visible_ranges for off-screen panes

FLY2ED:
### Problem

The server can enter a livelock — one CPU pinned at 100% in user mode, never
returning to the event loop. `tmux ls`, `has-session`, `info` and new
`attach-session` all block on the command socket and never return; existing
clients freeze. `kill-server` and `SIGTERM` are not serviced — only `SIGKILL`
recovers the server.

First hit in normal use via tmux-resurrect / tmux-continuum (`@continuum-restore
on`): auto-restore recreates panes and resizes on attach, and one restored pane
was in copy mode. Backtrace (macOS `sample`; captured on 3.7b, the same path
exists on current master):

```
server_client_loop -> resize_window -> layout_fix_panes -> window_copy_resize
  -> window_copy_redraw_lines -> screen_write_stop -> screen_write_collect_flush
    -> window_visible_ranges -> server_client_ensure_ranges
      -> xrecallocarray -> recallocarray        (~100% of samples)
```

### Cause

In `window_visible_ranges()` (window-visible.c):

```c
if ((u_int)py >= w->sy)
        goto empty;
if (px + width > w->sx)
        width = w->sx - px;   /* width is u_int; px > w->sx wraps to ~UINT_MAX */
```

`px < 0`, `py >= w->sy` and `width == 0` are guarded, but `px >= w->sx` is not.
During a resize a pane's x offset (a pane scrollbar, or a floating pane) can
momentarily exceed the shrunk window width, so `width` wraps to a huge value and
then drives the range-splitting loop and
`server_client_ensure_ranges(r, r->used + 1)` / `xrecallocarray()` into the
livelock.

### Fix

Guard `px` against `w->sx` exactly like `py` against `w->sy` just above — a pane
starting at or past the right edge has nothing visible, so return empty.

### Reproduction / verification (built from master, HEAD cad1c81)

Isolated socket, built with `./configure --disable-utf8proc`:

```sh
TM=./tmux
$TM -Ltest -f/dev/null new-session -d -x200 -y50 sh
$TM -Ltest set -g pane-scrollbars on
$TM -Ltest set -g window-size manual
$TM -Ltest send-keys -t s:0.0 'for i in $(seq 1 20000); do printf "L%05d\n" $i; done' Enter
sleep 1
$TM -Ltest split-window -h ; $TM -Ltest split-window -h
$TM -Ltest select-pane -t 0 ; $TM -Ltest copy-mode -t 0 ; $TM -Ltest send-keys -X -t 0 history-top
$TM -Ltest new-pane -O -X 100 -Y 20 -x 40 -y 12      # x offset (100) ends up past the right edge
for W in 200 30 10 5 8 3 60 4 6 2 40 5; do for H in 50 8 3 20; do
  $TM -Ltest resize-window -x $W -y $H; $TM -Ltest refresh-client
done; done
```

- **Unpatched:** within a few seconds the server pins a CPU at 100% and a
  `resize-window` never returns; `tmux -Ltest ls` hangs; recover with `kill -9`.
- **Patched:** the same scenario completes cleanly across repeated runs, the
  server stays responsive (`tmux -Ltest ls` returns immediately), CPU returns to
  idle, and panes still render.

### Related

- #5024 / #4969 / #5397 — same "server livelock in the redraw path, only SIGKILL
  recovers" family, different function (`tty_draw_line`).
- #4938 / #4958 / #4956 — unsigned underflow during copy-mode resize in
  `window-copy.c` (same bug class, adjacent file).

### DIFF
--- a/window-visible.c
+++ b/window-visible.c
@@ -83,6 +83,8 @@ window_visible_ranges(struct window_pane *base_wp, int px, int py, u_int width,
 	w = base_wp->window;
 	if ((u_int)py >= w->sy)
 		goto empty;
+	if ((u_int)px >= w->sx)
+		goto empty;
 	if (px + width > w->sx)
 		width = w->sx - px;
 


nicm:
Oops, yes, should have done this also. Applied to OpenBSD now but I changed it to put both checks in the same `if`, will be in GitHub later. Thanks!

PR #5412, opened 2026-07-20
Add an #{A} modifier to animate the status line

fdaciuk:
This adds an opt-in `#{A/count:frames}` modifier for short status line animations -
for example a spinner shown while a background task is working. There is no new
option, and nothing changes unless it is used.

## What it does

`#{A/count:frame,frame,...}` shows one of a comma-separated list of frames. The
period is fixed at 100 ms and `count` is how many periods each frame lasts, so a
frame is shown for `count * 100` milliseconds; `#{A:frame,frame,...}` is the same
with a count of 1. The frame is chosen from the clock as
`(now / (count * 100)) % number_of_frames`:

    set -g status-right "#{A/2:|,/,-,\\} working"

**It drives its own redraws.** While one is shown in the status line, the status
line is redrawn as the frames advance, so the animation runs on its own - there is
no need to lower `status-interval` or set anything else. When nothing on screen
contains one, no redraw is scheduled, so there is no idle cost. It is normally used
conditionally - e.g. a spinner shown only while a command is running - so the
redraws happen only while it is visible. The redraw is triggered the same way `#()`
output already triggers one (`server_status_client`).

Because the frame is chosen from the clock rather than a redraw count, animations
with different counts each keep their own speed while the client redraws at the one
fixed period.

The frames are used as they are and are not expanded as formats, and a comma always
separates them. An empty list of frames expands to the empty string, and a count
which is not a positive number is taken as 1.

## Only in a status format

It expands to the empty string unless `FORMAT_STATUS` is set, so it works in the
status line and `pane-border-format` and is empty everywhere else, including inside
`#()`.

The `#()` case is the one that matters: `format_job_get()` builds the job's expand
state with a new `FORMAT_EXPAND_NOCYCLE` flag, which covers both the command and its
output. Without it, a frame embedded in a shell command would change the command
text on every frame, and the `force = 1` path in `format_job_get()` would free and
re-run the job straight away, bypassing the once-a-second throttle.

## Docs & tests

- `tmux.1`: the `A` modifier in the FORMATS section, with an example.
- `CHANGES`: a changelog entry.
- `example_tmux.conf`: a ready-to-use spinner example.
- `regress/format-animation.sh`: a status line rendered by an inner tmux is captured
  through an outer one, and the frames have to change on their own with nothing in
  the test asking for a redraw; plus the empty cases (`display-message`,
  `list-panes -F`, inside `#()`) and degenerate input.

Builds warning-clean against current `master` with the default flags and the
AddressSanitizer DEBUG build; the regression test passes.


nicm:
I don't want a new option. Maybe make negative values milliseconds

nicm:
Also can you please make your function header omment about cycle less gigantic, the syntax should be described in the man page. Function comments are key info.

fdaciuk:
Both addressed, thanks.

- Removed `status-interval-ms`; a negative `status-interval` is now a millisecond interval (e.g. `-100` = every 100 ms). Positive is still seconds and `0` still disables, so existing configs are unaffected; the 20 ms floor is kept for the negative case.
- Shrank the `format_cycle` header comment to one line and left the syntax to the manual.

One heads-up: to accept a negative `status-interval` I set its minimum to `INT_MIN`, which meant widening `options_table_entry.minimum`/`.maximum` from `u_int` to `long long`. The only reader is the `strtonum` range check in `options.c` (already `long long`) and every existing bound - including the `UINT_MAX` one - is unchanged. Happy to do it differently if you'd prefer.

nicm:
Thanks.

They should be `long long` anyway, so that is fine.

Does this need to be time-based? Could it not be measured in redraws instead? You could control the speed by setting `status-interval`.

What is the need for `now_ms`? It isn't used in your example. If it must exist, I don't see why it shouldn't work inside `FORMAT_EXPAND_TIME`.

I think the documentation could do with some work:

- It isn't necessary to add comments like "intended for animations" or "build a spinner" to either the option description or the man page. Document the behaviour and let the user decide what to use it for. So the `options-table.c` text can just be something like `Number of seconds (or milliseconds, if negative) between status line updates. The minimum is 20 milliseconds.`.

- The example is good but put it at the end after the explanation, and don't use UTF-8 because not every terminal supports it, make it use ASCII maybe `\|/-`.

- I would avoid words like "magnitude" or "monotonically". This could all be written in much simpler language, it is for the general reader who may not necessarily be a developer.

- I would not say stuff like "in particular inside #()", just say it doesn't work inside `#()`, don't say that it particularly doesn't, and don't give the reader advice not to rely on it, saying it doesn't work is enough.

fdaciuk:
Thanks - glad the `long long` widening is fine.

On redraw- vs time-based: I did consider counting redraws, but I think time is the safer basis:

1. The status line is redrawn for many reasons besides the interval timer (activity, window/session changes, etc.), so a redraw counter would advance the animation on those too - the speed would drift with unrelated activity instead of staying steady. Driving it from the clock keeps it smooth regardless of what triggered the redraw.

2. Controlling speed only through `status-interval` means every animation runs at the same rate, and no faster than the whole line refreshes. A per-format period lets each animation have its own speed, independent of the redraw rate - e.g. two spinners at different speeds, or a slow one alongside something that refreshes quickly. That independence is really what `now_ms` is for; `cycle` is just the helper built on top of it.

(That's also why `now_ms` isn't in the example - the example only needs the helper. It's empty inside `#()` on purpose, so embedding it in a shell command can't make the command re-run every frame. I'm happy to drop `now_ms` and keep only `cycle` if you'd prefer a smaller surface.)

Does that make sense, or would you still rather it count redraws? Happy to go either way. Your documentation points all make sense - I'll fold them in once we settle on the approach.

nicm:
OK, maybe just leave it time-based, was thinking you could use it outside the status line by triggering the redraw yourself, but realistically I can't see what use it would be outside `status-format` and `pane-border-format`.

I'd drop `now_ms` for the moment if it isn't needed yet.

The thing I wonder is: is this "automatic" enough? By which I mean, if you want to show an animation in the status line, OK you can set `status-interval` to -20 and add `cycle` and off it goes, but is that practical? Who would want to have their status line updating 50 times a second forever when nothing is happening?

If seems like what would be better is if I wanted to say, show the animation while a command is running, so have it conditional on `pane_command_running`, then `status-interval` would *automatically* be reduced for the period it is running. So, if the format code sees `cycle` in the format, it adds it and then just sets itself up for next time - starts a timer and redraws the status line after the right period.

We already have a special-case call to `server_status_client` in `format_job_update` for when `#()` completes, so is there any harm in having another one for when `cycle` needs it? Since I don't think there is a practical use for this except in the status lines, I think it would be fine just to trigger a status redraw.

fdaciuk:
Agreed - I'll keep it time-based and drop `now_ms` for now.
And self-scheduling is the right approach; I'll implement it that way. Thanks!

fdaciuk:
Done - implemented the self-scheduling as discussed.

- `#{cycle:...}` now drives its own redraws: while a cycle is shown in the status line it re-arms a per-client timer at its period (floored at 20 ms) and redraws, so it animates on its own; when nothing shows a cycle, nothing is scheduled, so there's no idle cost. It still picks the frame from the clock, so two cycles with different periods keep their own speeds.
- Dropped `#{now_ms}` and removed the negative `status-interval`; neither is needed now.
- Simplified the docs per your notes.

nicm:
Would it not be simpler to do this in `format.c`? You can just store the timer in `struct client`. Also can you call it `cycle_timer`, animations are the use case not the feature.

I'd probably just make `cycle` always expand to nothing if `FORMAT_STATUS` is not set.

fdaciuk:
Sounds good - I'll move it into `format.c`, keep the timer in `struct client` and call it `cycle_timer`.

One thing I'd like your opinion on first: I think `FORMAT_STATUS` on its own may not be enough to keep the cycle out of `#()`. As far as I can tell, a `#()` in the status line still has `FORMAT_STATUS` on the format tree, and what the job path drops is `FORMAT_EXPAND_TIME` on the expand state - which is what stops the cycle from changing the command on every frame and re-running the job. So I was going to expand to nothing unless `FORMAT_STATUS` is set *and* it is a time expansion, which would also make it empty inside `#()` rather than stuck on the first frame, closer to what the manual says. Does that match how you see it, or would you rather it were just the `FORMAT_STATUS` check?

The other thing I noticed is `pane-border-format`, which sets `FORMAT_STATUS` as well: a cycle there would arm the timer, but as the callback only sets `CLIENT_REDRAWSTATUS` the border isn't redrawn, so nothing re-arms it and it stops after that one redraw. My instinct is to leave it at that, since the status line is the practical case, but tell me if you'd rather the timer redrew the borders too.

Otherwise `status.c` goes back to what it was, and the timer is armed straight from `format_cycle`, only when nothing is pending or when its own period is shorter, so several cycles with different periods still animate at the shortest one.

nicm:
I'm not really sure why `#()` matters - it will only run `#()` once a second anyway, that is entirely separate, no?

fdaciuk:
Right, that's the `fj->last != t` check, and it holds as long as the command text stays the same. The bit I had in mind is just above it: if the expanded command differs from the previous one, `format_job_get` sets `force = 1`, and that path calls `job_free()` and re-runs the job straight away, without the once-a-second check. I think that is also why `FORMAT_EXPAND_TIME` is cleared before expanding the command in 1492ae11 ("Do not expand times and #() inside #()") - otherwise a `%H:%M:%S` in there would change the text every second and keep restarting the job.

`FORMAT_STATUS` is still set on the tree inside that expansion, so with only that check `#(echo #{cycle:100:a b})` in the status line would expand to a different frame each time and each redraw would kill and restart the job - and the cycle would arm the timer from in there as well, so it would be every 100 ms rather than every second.

So what I had in mind is just keeping the existing time check and adding yours on top: nothing unless `FORMAT_STATUS` is set and it is a time expansion. Does that sound reasonable?

nicm:
What use is putting `cycle` inside `#()`? Do we even need to worry about it?

If it matters I guess we could add a `FORMAT_NOCYCLE`? Not sure about overloading `EXPAND_TIME`.

fdaciuk:
No real use, no. It's only that if someone does it by mistake it turns into a process spawn loop rather than something harmless, since the job gets killed and restarted on every frame instead of once a second. But it is cheap to rule out, and I agree a dedicated flag is nicer than leaning on `EXPAND_TIME`.

The job path already builds its state with `FORMAT_EXPAND_NOJOBS`, and both flags are local to format.c, so it would just be:

```c
format_copy_state(&next, es, FORMAT_EXPAND_NOJOBS|FORMAT_EXPAND_NOCYCLE);
```

with `cycle` expanding to nothing if `FORMAT_STATUS` is not set or `FORMAT_EXPAND_NOCYCLE` is. That state is used both for the command and for its output, so both are covered, and the `EXPAND_TIME` line above stays as it is. Does that sound right?

nicm:
Yes that's fine

fdaciuk:
Done - it is all in `format.c` now.

- The timer is `cycle_timer` in `struct client`, armed from `format_cycle` and deleted next to the other client timers in `server_client_lost`. `status.c` is back to what it was.
- A cycle expands to nothing unless `FORMAT_STATUS` is set, and the new `FORMAT_EXPAND_NOCYCLE` keeps it out of the command and the output of `#()`.
- A cycle with a single frame no longer arms anything, since its value cannot change.

One correction to something I said earlier: a cycle in `pane-border-format` does animate. `CLIENT_REDRAWSTATUS` goes through `screen_redraw_screen`, which re-expands the border format, so it re-arms itself. It is not guaranteed to keep to `period_ms` there though, so the manual only promises the period for the status line.

nicm:
Could you rename `format_cycle_timer` to `format_cycle_start_timer`?

Do we really need `period_ms`? Could we just say the period is something sensible like 50 or 100 milliseconds and be done with it? That would get rid of all this stuff about calculating the timer interval, and also weirdness with multiple cycles of different periods. If you think it is necessary you could make the first argument a period counter if you think people will want to say "2 periods" or whatever.

fdaciuk:
Sure, I'll rename it to `format_cycle_start_timer`.

And yes, a fixed period sounds good to me. `format_cycle_start_timer` then just arms the timer when nothing is pending and the minimum goes away with it. I would keep a period counter as the first argument though, so that a slow animation doesn't have to repeat frames to get there: `#{cycle:2:a b c}` shows each frame for two periods, and the frame is still picked from the clock, just `(now / (count * period)) % frames`. Since every period is a multiple of the base now, a redraw always lands on a frame boundary.

I think the count is better required than optional: to make it optional I would have to decide it from a leading number, and then frames that are numbers followed by a colon, like `#{cycle:10: 9: 8:}`, would be read as a count. Required keeps the parsing exactly as it is today, only the meaning of the number changes.

Would you prefer 50 or 100 for the base? Either is fine by me - the base is both the fastest animation and the rate the status line is redrawn while a cycle is on screen, so 100 is cheaper and 50 gives finer steps. I'll use whichever you prefer.

nicm:
Let's go with 100 for the period.

I'm not sure this should be `#{cycle`, it doesn't really fit with the other modifiers which are single letter. Maybe you could use `Y` for cycle or `A` for animate? The count could be a flag so like `#{Y/5:A,B,C,D}`. Then if the 5 is omitted it can default to 1.

fdaciuk:
Done, thanks - both of those made it smaller.

- `format_cycle_start_timer` is renamed, and with the period fixed at 100 ms it just arms the redraw when nothing is pending, so the interval arithmetic is gone.
- It is a modifier now: `#{A/count:frame,frame,...}`, or `#{A:frame,frame,...}` for a count of 1. Frames are separated by commas, so a frame can contain a space. The frame is still picked from the clock, `(now / (count * 100)) % frames`.
- Making it a modifier also removed the special case I had at the top of `format_replace`: the old `cycle:` key had to be intercepted before the modifier parser ran, because its value contains colons and it starts with a `c`. That is all gone.
- I went with `A` for animate, but the internal names stayed on cycle (`cycle_timer`, `format_cycle`) since you said that is the feature - say the word if you would rather have `Y`.
- I also reverted the `minimum`/`maximum` widening in `options_table_entry`. That was only there for the negative `status-interval`, which is gone, so it was an unrelated change sitting in the diff.
- Added a regress test: an inner tmux status line is captured through an outer one and the frames have to change on their own, with nothing in the test asking for a redraw.

nicm:
Looks good, applied to OpenBSD now with some trivial changes (I set a set a maximum `count` of 100, and renamed some bits to make lines shorter), will be in GitHub later. Thanks!

nicm:
Something like this works quite well if you have OSC 133, but probably not by default:
~~~~
set -g pane-border-status top
set -gF @old_pane_border_format '#{pane-border-format}'
set -g pane-border-format '#{?#{&&:#{!:#{alternate_on}},#{pane_command_running},#{e|>:#{pane_command_duration},3}},#[acs][#{A/2:`  , ` ,  `, ` }],}#[noacs] #{E:@old_pane_border_format}'

fdaciuk:
Nice! I gave this a go on master now the sync has landed, and ran into something interesting.

Nothing schedules a redraw for the moment the `pane_command_duration` gate opens, so the animation only starts on whatever redraw happens to come next.

If the command writes output, that is enough and it turns up about when you would expect - around a second after the threshold here. But if the command is silent, the next redraw is the `status-interval` one, so with the default of 15 seconds:

- a command which keeps running: the spinner appears at about 16 seconds rather than 3
- a command which runs for 8 seconds: the spinner never appears at all

With `status-interval 1` it starts on time, and it is the same with the gate in `status-format` rather than a border format, so it is the status timer bootstrapping it. Dropping the duration threshold avoids it too, since then the condition is already true when the command starts.

Not suggesting the modifier should do anything about it - the threshold is an ordinary format condition and there is no way for it to know when it will become true. Just a bit funny that a silent command is the case where you most want the spinner.

nicm:
Yeah, it's not ideal but I don't really want to make the default `status-interval` any lower. I guess people who are annoyed by it can reduce it themselves. And who is sitting watching a window without using it anyway?

nicm:
Or you could do like:
~~~~
set-hook -gB '@status:%*:#{&&:#{pane_command_running},#{e|>:#{pane_command_duration},3}}' 'refresh -S'
~~~~
But not sure it will really save that much. If you're happy with occasional 100 millisecond redraws, why would you quibble about `status-interval 1`?

fdaciuk:
Tested it - the hook does the trick. The spinner turns up at about 4 seconds and goes away cleanly when the command finishes; without it, the same 8 second command shows nothing at all.

And you're right that it doesn't save much. The monitor timer re-arms every second for as long as there is a subscription, so the hook has the same one second heartbeat as `status-interval 1` - it just does less on each tick, evaluating one format rather than redrawing the whole status line. So `status-interval 1` is the simpler answer of the two.

nicm:
Yes, it is probably not worth it. It won't actually draw the status line anyway if it hasn't changed, it will just reevaluate all the formats.

PR #5415, opened 2026-07-21
Osc133 line collapsible

mgrant0:
Add collapsible lines to copy-mode.

If terminal decorated the output with OSC-133 codes, when in copy-mode, tab will collapse/expand the output, and shift-tab will collapse/expand all outputs.


mgrant0:
This PR originated from this other PR https://github.com/tmux/tmux/pull/5407

PR #5420, opened 2026-07-22
Do not hide cursor during synchronized redraw

### LINKED ISSUE #5419: Synchronized client redraw toggles an otherwise-visible cursor
### Issue description

During a full client redraw, tmux emits `civis` before committing synchronized output and `cnorm` immediately afterwards, although the pane cursor is visible before and after the redraw.

The tmux-only reproducer below captures both boundaries of a nested tmux:

- application output entering the inner tmux pane;
- inner tmux client output entering the outer tmux pane.

The application emits exactly one balanced `CSI ? 2026 h` / `CSI ? 2026 l` frame and no cursor visibility controls. After that frame is stable, the script starts a fresh capture and requests one inner `refresh-client` full redraw. Current master adds both `civis` and `cnorm` to that redraw:

```text
inner client emitted civis before synchronized output was committed
inner client emitted cnorm after synchronized output was committed
```

On terminals where cursor-state updates restart blinking, repeated redraws visibly disrupt the blink phase. In Windows Terminal through WSL2, the NORMAL phase of the attached manual script blinks normally; during its 50 ms `refresh-client` phase the cursor flickers rapidly and irregularly, pauses briefly, then resumes rapid flickering. The automated reproducer flags the current DECTCEM transition pair as a deterministic byte-level proxy for that behavior; it is not intended to prescribe how tmux should avoid the visible reset.

This is distinct from application-side synchronized buffering added by #4744. It is the separate tmux-client-to-terminal redraw path. It is also narrower than #2956: the problem is the observable cursor visibility/blink lifecycle during a synchronized full redraw, not merely that raw tmux output can contain redundant bytes.

#### Reproduction

1. Build current Git master.
2. Download and extract [the automated reproducer and `tmux -vv` logs](https://github.com/xz-dev/tmux/releases/download/issue-5419-evidence/tmux-sync-cursor-issue-logs.zip).
3. Run:

```sh
cd regress
TEST_TMUX=$(readlink -f ../tmux) sh /path/to/reproduce.sh
```

The script uses only POSIX shell, standard utilities, and two isolated `-L` tmux servers with `/dev/null` configuration. Both inner and outer window/pane geometries are verified as 80x24, and the inner client is verified to advertise `sync`.

Expected: a synchronized full redraw does not emit cursor visibility transitions solely to hide and restore an otherwise-visible cursor. This does not require every terminal to preserve an idle blink phase while processing the redraw's necessary cursor-position updates.

Actual on current master `7abb9af06236eb9def862bb88a82792f6c846bef`: exit 1 with both messages above, reproducibly.

Evidence:

- [Automated reproducer, output and sanitized `tmux -vv` client/server/out logs](https://github.com/xz-dev/tmux/releases/download/issue-5419-evidence/tmux-sync-cursor-issue-logs.zip)
- [Manual Windows Terminal comparison script, observation and sanitized `tmux -vv` logs](https://github.com/xz-dev/tmux/releases/download/issue-5419-evidence/tmux-sync-cursor-visual-logs.zip)

### Required information

- tmux version: `tmux next-3.8`, Git master `7abb9af06236eb9def862bb88a82792f6c846bef`
- Platform: Linux x86_64, WSL2 kernel `6.6.87.2-microsoft-standard-WSL2`
- Terminal where the visible blink disruption was observed: Windows Terminal through WSL2; outside tmux `TERM=xterm-256color`, inside tmux `TERM=tmux-256color`
- Automated byte-level reproduction: nested tmux capture harness; it does not require third-party terminal behavior
- `$TERM` inside inner tmux: `tmux-256color`
- `$TERM` for the inner tmux client: `tmux-256color`, with `sync` explicitly enabled and verified by the script
- `$TERM` outside the automated test harness: `xterm-256color`

xz-dev:
Closes #5419.

## Problem

A physical-client redraw uses synchronized output but still applies the legacy unsynchronized redraw strategy:

1. `server_client_check_redraw()` forces `TTY_NOCURSOR`;
2. `redraw_draw()` removes all cursor modes;
3. each `tty_draw_line()` temporarily hides and restores the cursor;
4. `redraw_draw()` commits synchronized output before `server_client_reset_state()` restores the final cursor position, mode, style, and visibility.

The pane can therefore require a visible cursor before and after the redraw while tmux emits `civis` before the synchronized commit and `cnorm` after it. Repeated full redraws disrupt the terminal's cursor blink lifecycle; in the issue's Windows Terminal reproduction this appears as rapid, irregular flicker.

## Fix

Treat a client redraw and its final terminal-state restoration as one physical synchronized-output transaction.

### `server-client.c`

- Add `sync` to distinguish an active physical transaction from tmux's internal `TTY_SYNCING` bookkeeping.
- Start synchronization in `server_client_check_redraw()`, after outstanding-output/blocked-client deferral and before clearing `TTY_FREEZE` or performing any pane or full-window drawing. This covers pane-only redraw paths as well as `redraw_screen()`.
- Require both `TTY_SYNCING` and `TTYC_SYNC` before relying on synchronization. `tty_sync_start()` sets the internal flag even when the terminal lacks the capability, so checking only the flag would break the unsynchronized fallback.
- Split the old combined flag assignment. Always clear `TTY_BLOCK|TTY_FREEZE`, but add temporary `TTY_NOCURSOR` only without physical synchronization.
- Do not clear a pre-existing `TTY_NOCURSOR`; explicit/semantic cursor hiding therefore remains in force.

### `screen-redraw.c`

- Keep `tty_sync_start()` in `redraw_draw()`. It is idempotent and keeps this helper safe if reached by another caller.
- Remove `CURSOR_MODES` only when `TTYC_SYNC` is unavailable. A capable terminal buffers intermediate drawing, so changing visibility/blink modes inside the transaction is unnecessary; terminals without synchronization retain the old protection.
- Remove the local `tty_sync_end()`. It committed the temporary redraw state too early. The existing `server_client_reset_state()` call immediately after `server_client_check_redraw()` restores final cursor position/mode/style/visibility and then calls `tty_sync_end()`, making that the single lifecycle endpoint.

### `tty-draw.c`

- Record whether a line is being drawn inside an active physical synchronized transaction.
- Require no pre-existing `TTY_NOCURSOR` for the fast path. A pre-existing flag is semantic state, not a temporary redraw optimization, and must remain hidden.
- Skip line-local `TTY_NOCURSOR` hide/update and restore/update only on that fast path. Otherwise execute the original pair unchanged, preserving behavior for unsynchronized terminals and explicitly hidden cursors.
- Use the same `sync` snapshot at entry and cleanup so the function cannot skip only one half of the hide/restore pair.

The resulting order is:

```text
sync start
complete client redraw
restore final cursor position, mode, style and visibility
sync end
```

instead of committing a temporarily hidden cursor and restoring it afterwards.

## Why this is safe

- `tty_sync_start()` is idempotent when `TTY_SYNCING` is already set.
- Blocked clients return before synchronization starts.
- The server loop always calls `server_client_reset_state()` immediately after `server_client_check_redraw()`, and that function already ends any active transaction after restoring final state.
- Non-sync terminals still force `TTY_NOCURSOR`, remove cursor modes during redraw, and run the original line-local hide/restore path.
- Existing `TTY_NOCURSOR` is preserved at both client-redraw and line-draw levels.
- The final mode calculation is unchanged; this patch changes only temporary redraw handling and transaction ownership.

## Regression test

`regress/sync-cursor-redraw.sh` uses two isolated tmux servers to capture both sides of the redraw:

- the application stream entering the inner pane;
- the inner tmux client's physical output entering the outer pane.

It verifies fixed 80x24 geometry, actual `sync` client capability, one exact balanced application synchronized-output frame with no cursor visibility controls, and one isolated `refresh-client` redraw. Current master deterministically emits both `civis` and `cnorm`; this branch does not.

## Validation

- Clean current master: regression fails with both cursor-transition diagnostics.
- This branch: regression passes repeatedly.
- `make -j2` passes.
- Focused redraw/mode regressions pass:
  - `tty-draw-line.sh`
  - `screen-redraw-status.sh`
  - `screen-redraw-menus.sh`
  - `screen-redraw-popups.sh`
  - `screen-redraw-cache.sh`
  - `screen-redraw-scrollbars.sh`
  - `input-modes.sh`
  - `decrqm-sync.sh`
  - `sync-cursor-redraw.sh`
- All 118 regression scripts passed: the first bounded run passed through `style-trim.sh` before its external timeout, and the remaining nine scripts all passed separately.

### DIFF
--- a/regress/sync-cursor-redraw.sh
+++ b/regress/sync-cursor-redraw.sh
@@ -1,0 +1,163 @@
+#!/bin/sh
+
+# A synchronized pane update must not gain cursor visibility changes when tmux
+# redraws it for a client that supports synchronized output.
+
+PATH=/bin:/usr/bin
+TERM=screen
+LC_ALL=C.UTF-8
+export TERM LC_ALL
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+
+DIR=$(mktemp -d) || exit 1
+TMUX_TMPDIR=$DIR
+export TMUX_TMPDIR
+
+INNER="$TEST_TMUX -Lsync-cursor-inner-$$ -f/dev/null"
+OUTER="$TEST_TMUX -Lsync-cursor-outer-$$ -f/dev/null"
+APP_BYTES=$DIR/app-bytes
+CLIENT_BYTES=$DIR/client-bytes
+EXPECTED=$DIR/expected
+CONTROL=$DIR/control
+
+MARKER=SYNC_CURSOR_MARKER
+SYNC_ON=$(printf '\033[?2026h')
+SYNC_OFF=$(printf '\033[?2026l')
+CIVIS=$(printf '\033[?25l')
+CNORM=$(printf '\033[?25h')
+
+fail()
+{
+	echo "$*" >&2
+	exit 1
+}
+
+cleanup()
+{
+	$OUTER kill-server 2>/dev/null
+	$INNER kill-server 2>/dev/null
+	rm -rf "$DIR"
+}
+trap cleanup 0 1 15
+
+wait_for_bytes()
+{
+	file=$1
+	needle=$2
+	i=0
+	while [ "$i" -lt 50 ]; do
+		grep -F -q "$needle" "$file" 2>/dev/null && return 0
+		sleep 0.1
+		i=$((i + 1))
+	done
+	fail "timed out waiting for bytes in $file"
+}
+
+wait_for_inner_client()
+{
+	i=0
+	while [ "$i" -lt 50 ]; do
+		$INNER list-clients -F '#{client_session}' 2>/dev/null |
+		    grep -Fx -q inner && return 0
+		sleep 0.1
+		i=$((i + 1))
+	done
+	fail "inner client did not attach"
+}
+
+check_geometry()
+{
+	tmux=$1
+	target=$2
+	actual=$($tmux display-message -p -t "$target" \
+	    '#{window_width}x#{window_height} #{pane_width}x#{pane_height}') ||
+	    fail "cannot read geometry for $target"
+	[ "$actual" = "80x24 80x24" ] ||
+	    fail "$target geometry is $actual, expected 80x24 80x24"
+}
+
+wait_for_stable_bytes()
+{
+	file=$1
+	previous=-1
+	stable=0
+	i=0
+	while [ "$i" -lt 50 ]; do
+		current=$(wc -c <"$file" 2>/dev/null) || current=0
+		if [ "$current" -gt 0 ] && [ "$current" -eq "$previous" ]; then
+			stable=$((stable + 1))
+			[ "$stable" -eq 3 ] && return 0
+		else
+			stable=0
+		fi
+		previous=$current
+		sleep 0.1
+		i=$((i + 1))
+	done
+	fail "client byte stream did not become stable"
+}
+
+mkfifo "$CONTROL" || exit 1
+$INNER new-session -d -s inner -x 80 -y 24 \
+    "read signal <'$CONTROL'; printf '\033[?2026h%s\033[?2026l' '$MARKER'; exec sleep 30" ||
+    exit 1
+$INNER set-option -g status off || exit 1
+$INNER set-option -g window-size manual || exit 1
+$INNER set-option -as terminal-features '*:sync' || exit 1
+$INNER pipe-pane -O -t inner:0.0 "cat >'$APP_BYTES'" || exit 1
+
+$OUTER new-session -d -s outer -x 80 -y 24 \
+    "$TEST_TMUX -Lsync-cursor-inner-$$ -f/dev/null attach-session -t inner" ||
+    exit 1
+$OUTER set-option -g status off || exit 1
+$OUTER set-option -g window-size manual || exit 1
+
+wait_for_inner_client
+check_geometry "$INNER" inner:0.0
+check_geometry "$OUTER" outer:0.0
+
+client_geometry=$($INNER list-clients -F '#{client_width}x#{client_height}') ||
+    exit 1
+[ "$client_geometry" = "80x24" ] ||
+    fail "unexpected inner client geometry: $client_geometry"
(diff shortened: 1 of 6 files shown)


nicm:
Thanks.

- I think instead of checking `TTYC_SYNC` again in `redraw_draw` and `server_client_check_redraw` it would be neater if `tty_sync_start` returned `0` or `1` to say if sync is in use?

- Why are you removing the `tty_sync_end` in `redraw_draw`?

- Can you tidy stuff like this up a bit? Use a proper `if` instead of an assignment so it is more readable. Also `flags` is not a boolean value so better not to check for zero with `!`.
~~~~
	sync = (tty->flags & TTY_SYNCING) &&
	    tty_term_has(tty->term, TTYC_SYNC) && !flags;
~~~~

xz-dev:
Thanks for review.

Addressed in 5323e303851a01568dde762de742427d1b74bcc6:

- `tty_sync_start()` now returns whether synchronized output is actually in use, and both redraw callers use that result instead of rechecking `TTYC_SYNC`.
- The `tty_draw_line()` condition now uses a normal `if`, and compares the saved flags bitmask with `flags == 0`.

-------

I removed `tty_sync_end()` from `redraw_draw()` because this change moves ownership of the synchronized transaction outward. `server_client_check_redraw()` starts it before any pane/full-screen redraw, and one check can call `redraw_draw()` more than once. The immediately following `server_client_reset_state()` then restores the final cursor position, mode, attributes and visibility before its existing `tty_sync_end()`.

Keeping the end in `redraw_draw()` would therefore close the outer transaction on the first draw, potentially split one client update into multiple synchronized commits, and leave the final terminal-state restoration outside the transaction. The intended order is now:

```
sync start
redraw(s)
restore final cursor and terminal state
sync end
```

nicm:
I don't think `screen-redraw.c` should start the sync if it is not going to finish it. If `server-client.c` now calls `tty_sync_start` it should call `end` as well.

Is this a problem in other terminals aside from Windows Terminal? It seems to me cursor on/off should probably just be part of the sync in the terminal itself.

xz-dev:
I need more time to debug this, when I find new thing I will reopen the PR, close it now

PR #5430, opened 2026-07-25
fire after-swap-window hook in swap-window

Miloud-ux:
Swap-window wasn't triggering an after-hook,unlike nearly every other window/pane command (like new-window, split-window, select-pane, and the others). Its hook name also wasn't entered in the options-table so:

set-hook -g after-swap-window 'some-command'

would result in an invalid option: after-swap-window error.

This pulls in CMD_AFTERHOOK to the swap-window command entry and adds "after-swap-window" to options-table.c mirroring the existing convention applied to other commands.

Manual tests:

set-hook -g after-swap-window 'run-shell "echo fired >> /tmp/log"'

swap-window -s test:0 -t test:1

/tmp/log then contains "fired"; previously set-hook returned invalid option.

### DIFF
--- a/cmd-swap-window.c
+++ b/cmd-swap-window.c
@@ -38,7 +38,7 @@ const struct cmd_entry cmd_swap_window_entry = {
 	.source = { 's', CMD_FIND_WINDOW, CMD_FIND_DEFAULT_MARKED },
 	.target = { 't', CMD_FIND_WINDOW, 0 },
 
-	.flags = 0,
+	.flags = CMD_AFTERHOOK,
 	.exec = cmd_swap_window_exec
 };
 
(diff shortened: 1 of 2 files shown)


nicm:
I think you need to look at this patch again because it seems to be making a lot of unrelated changes.

Miloud-ux:
> I think you need to look at this patch again because it seems to be making a lot of unrelated changes.

Sorry about that, i think the issue was in formatting in my editor, now it should be just the 2 intended lines.

nicm:
I've applied this to OpenBSD now, although it seems of limited use since you can't tell which windows were swapped. It will be in GitHub later. Thanks!

PR #4940, opened 2026-03-19
Add "box" option to pane-border-indicators

patrick-motard:
Reopening https://github.com/tmux/tmux/pull/4747 with additional fixes since the original PR.

Add a new "box" option for `pane-border-indicators` that draws a complete box border around the active pane instead of using shared separator borders between panes. Also adds "box-all" to draw borders around every pane.

https://github.com/user-attachments/assets/321b826c-2f5f-43fd-b982-469cc50ff62c

## Changes

- Add `PANE_BORDER_BOX` / `PANE_BORDER_BOX_ALL` constants and `window_pane_box_mode()` helper
- Reduce internal screen buffer and PTY by 2 in each dimension to reserve space for box borders
- Offset pane content drawing by 1 to account for box border
- Adjust cursor positioning for box mode offset in screen-write, tty, and server-client
- Draw styled box border for active pane, blank borders for inactive panes
- Skip normal pane separator borders when box mode is active
- Clear separator cells between panes that would otherwise show old borders
- Handle mouse clicks on box borders correctly
- Add viewport clipping support so box borders work correctly when the window is larger than the terminal
- Add "box-all" variant that draws borders around all panes, not just the active one
- Document both options in tmux.1

## Test plan

- [x] Single pane: box mode inactive, no borders drawn
- [x] Horizontal split: box borders on active pane, blank on inactive
- [x] Vertical split: same behavior
- [x] `box-all`: borders on all panes
- [x] Switch between `off`/`box`/`box-all` dynamically
- [x] Mouse clicks on borders and within panes work correctly
- [x] Resize terminal: borders update correctly
- [x] Viewport panning works with box borders

PR #5433, opened 2026-07-26
Add per-pane-border option

redesigndavid:
## Summary
- Add a window option `per-pane-border` that draws a full border around each tiled pane (inset geometry + redraw).
- Keep `display-panes` aligned with that layout: draw indexes using per-pane borders, fill the window while active, and hide scrollbars so the overlay is not shifted.
- Update the display-panes regress coverage for the no-scrollbar overlay sizes.

## Test plan
- [ ] `set -w per-pane-border on` with multiple tiled panes and confirm each pane gets a full border
- [ ] Toggle `pane-border-status` (top/bottom/off) with `per-pane-border` and confirm no blank/grey opposite edge
- [ ] `display-panes` / `prefix-q` with and without scrollbars; indexes should not shift when scrollbars are enabled
- [ ] `cd regress && gmake` (or run `display-panes.sh` at minimum), though a few of the tests are broken even in master.

[asciicast](https://asciinema.org/a/GEx7LDnkfGubjV85)


nicm:
Looks OK but I don't like the option name, it should be something starting with `pane-border-` to match the others.

I'm not sure about having `WINDOW_PANESMODE` rather than generic flags and if we do have it those fixes seem like they are separate?

In `window_panes_mark_per_pane_borders`, you need `{}` around the `if` bodies where they cross multiple lines.

In `screen-redraw.c`, we do not want to call `options_get_number` for every pane, this is a window option and can be cached.

redesigndavid:
How does "pane-border-surround" sound? That keeps it in the pane-border-*
family.

WINDOW_PANESMODE was a workaround for display-panes because of the
temporary zoom. With pane-border-surround, `layout_fix_panes` was still
applying the inset during that zoom and clipping the mode screen.
WINDOW_ZOOMED alone wasn't enough when the window was already zoomed. I
later reused the same flag to hide scrollbars so they didn't shift the
overlay. That probably should have been a separate change. Let me look at
dropping the mode-specific flag and then either use something more generic
or key off the mode/zoom state directly.

I'll also add the braces in window_panes_mark_pane_border_surround and
cache the window option in the redraw context. Thanks for the feedback.

nicm:
I don't like `pane-border-surround`. Feels like this should be part of `pane-border-lines` but not sure I want to double the number of choices. Leave it as it is for the moment and I'll think about it.

nicm:
Maybe `pane-border-type` (or `-model` or something?) and choices are `joined` or `separate`?

redesigndavid:
Yeah, `pane-border-type` with `joined` or `separate` sounds as good as any? Let me make that update.

nicm:
New option needs to go in `tmux.1` also.

nicm:
OK a few minor issues:

- When I have two panes side by side or one above the other, only one of the borders works to resize when dragged with the mouse, the other border does not (behaves as if inside pane).

- `selectp -U`, `-L` etc do not work.

- `{left}`, `{top}` targets are also broken, probably for the same reason (`selectp -t '{top}'`).

- `main-vertical` seem to place the right hand border of the right pane outside the window in some circumstances (seems to depend on what the width and height is). Same for `main-horizontal` with the bottom border.

- There are similar problems with placement of cells after resize, for example try this, the right pane is now outside the window:
~~~~
tmux new \; set -g pane-border-type separate \; splitw -h \; resize-pane -x 1
~~~~ 

I also think we could do with adding test coverage for this feature.

redesigndavid:
I did find another bug. When you have pane-border-status enabled, and click on x, it seems impossible to click on yes. But the bug seems to exist in master too so I didn't bring it up. Shall we include a fix for that too?

nicm:
I do not see this problem in `master`, the menu works fine. How do you reproduce?

redesigndavid:
On master, set status-position to 'top' and pane-border-status to 'top' or 'bottom'.

https://github.com/user-attachments/assets/ae17f6df-518f-4b92-8338-6729e3e938b9

nicm:
It works OK for me but the mouse position is one line off, I'll fix it.

nicm:
~~~~diff
diff --git a/server-client.c b/server-client.c
index 46315d7d..ead84c63 100644
--- a/server-client.c
+++ b/server-client.c
@@ -1682,6 +1682,9 @@ server_client_handle_menu_key(struct client *c, struct key_event *event)
        memcpy(&new_event, event, sizeof new_event);
        if (KEYC_IS_MOUSE(event->key)) {
                m = &new_event.m;
+               m->statusat = status_at_line(c);
+               m->statuslines = status_line_size(c);
+
                tty_window_offset(&c->tty, &ox, &oy, &sx, &sy);
                m->x += ox;
                if (m->statusat == 0) {

redesigndavid:
That fixed it for me.

nicm:
Good stuff, this fix should be in GitHub later.

nicm:
Is this ready to go again or are you still working on it?

nicm:
@patrick-motard had something similar to this here https://github.com/tmux/tmux/pull/4940, but it seems like it was just for the active pane (so `pane-border-indicators`). Also it was based on the old `screen-redraw.c` code.

I think having a separate option is better. I don't really see any need to have a way for just the active pane have separate borders, but if we did want to do that we could add a `pane-border-type separate-active` or something.

The display with the separate borders is quite pretty but I can't see myself using it because it wastes screen space.

patrick-motard:
My PR does have a border on the active pane, but I intentionally adding spacing around each inactive pane so that there is no jarring resizing when moving from one pane to another. It makes sense to me that a user would be able to control the active and inactive border colors. I would like to see these PRs merged or for one to be chosen to work from. I have a fix I can push that updates the feature to be compatible with the recent changes on main. Im currently working through some visual bugs that happen with pop-ups and screen resizing.

nicm:
OK so the "active only" configuration the same as just having an invisible border on the inactive panes. That seems like it should be pretty easy to add here too?

I don't really mind which PR moves forward, they are doing much the same thing so I expect they are broadly similar?

I think a separate option like this is perhaps more logical than rolling it into `pane-border-indicators`.

redesigndavid:
@patrick-motard sorry I didn't see your PR, we could have collaborated. Hope you don't mind I continue adding the requested features to my branch.

I pushed updates that adds 'separate-active' (happy to update the name if anyone could think of a better name). It uses same layout as 'separate' but hides the borders on inactive panes, even hides the border between the title and the [f][z][x] in the pane-border-status. 

Here's a quick asciicast:
[asciicast](https://asciinema.org/a/meVlcfk8lH5HhhwA)

> The display with the separate borders is quite pretty but I can't see myself using it because it wastes screen space.

I quite agree on the points about it being pretty and a bit of a waste of screen space, I say this with fingers on prefix-z at the ready. I do think 'separate-active' is prettier though and with less visual noise.

nicm:
OK this seems to work well for me. A few comments:

I think all these:
~~~~
PANE_BORDER_TYPE_IS_SEPARATE(options_get_number(w->options,
+           "pane-border-type")
~~~~
Would be neater if you just dropped `PANE_BORDER_TYPE_IS_SEPARATE` and added a helper `int window_border_type_is_separate(struct window *)`. This would get rid of a lot of wrapped lines.

Any `if` body that crosses more than one line needs `{}`, even if it is only one statement, so here:
~~~~
+               if (!fill)
+                       layout_apply_pane_border_type(w, root, lc, &wp->xoff,
+                           &wp->yoff, &sx, &sy);
~~~~
There are a couple of others too I think.

Maybe rename `layout_pane_minimum` to `layout_pane_minimum_size`?

I think some or all the new comments are probably either unnecessary or too big...

The `!blank` checks in `redraw_draw_border_span` could probably be tidied up into nested `if`s.

Can `WINDOW_MODE_HIDE_SCROLLBARS` be done as a separate PR to get it out of the way? Could add `WINDOW_MODE_FILL_WINDOW` also even if it doesn't do anything yet.

lollipopman:
I have always wanted this feature, thanks for implementing it! I just tested it out on a local build and it seems to work great, <3

redesigndavid:
Thank you, very encouraging!

redesigndavid:
@nicm, I believe we've hit all points, yes?

nicm:
I'm away but I'll look when I'm back on Monday, thanks.

nicm:
Looks pretty good. Two other minor issues:

- `splitw` can still create panes that are too small, for example `splitw -fh -l 1`. These can end up with the border overwriting other pane borders instead of resizing them, for example:

image

- The same thing can happen with `selectl -E`, for example:
~~~~
./tmux -vvLx new \; set -g pane-border-type separate \; splitw -h \; splitw -h \; splitw -h \; resizew -x10 \; selectl -E
~~~~
image

redesigndavid:
Thanks! Pushed fixes and tests to address both.

nicm:
If you do this:
~~~~
./tmux -Lx new \; splitw -fhl1 \; set -g pane-border-type separate
~~~~
You end up with the right pane wrong. Not sure if this is worth fixing if it isn't easy.

This does not work properly (the `selectp -D` does not change the pane):
~~~~
./tmux -Lx new \; set pane-border-type separate \; set pane-border-status bottom \; splitw -v \; selectp -D
~~~~

I think `redraw_draw_ctx` could do with a cached `border_type` as well for `redraw_draw_border_span`.

This may be an existing problem - this is an error:
~~~~
./tmux -Lx new \; resizew -x 5 -y 8 \; set pane-scrollbars on \; set -g pane-border-type separate \; set pane-scrollbars-style 'width=2,pad=1' \; splitw -h
~~~~
But this gets in a muddle instead:
~~~~
./tmux -Lx new \; resizew -x 5 -y 8 \; set pane-scrollbars on \; set -g pane-border-type joined \; set pane-scrollbars-style 'width=2,pad=1' \; splitw -h
~~~~
I think both cases should be an error.

redesigndavid:
Interesting, let me have a look.

redesigndavid:
Okay pushing updates:

For item 1: on `options_push_changes` for `pane-border-type`, reflow so cells under the new minimums grow.

For 2–4: fix select-pane wrap with separate + bottom status; cache `border_type` on the draw context; for always-on scrollbars, LEFTRIGHT space counts both bars so tiny splits fail cleanly instead of leaving broken pane sizes.

nicm:
OK there are still some minor things with small windows, but TBH I am not wild about `layout_rebalance_mins` and friends, they seem like overkill. I wonder if we should access some weirdness to get away from having to do this, so long as tmux does not do anything really crazy or crash.

redesigndavid:
Let me spend a few more hours later to see if I can think of a cleaner fix?

nicm:
OK thanks. TBH I think we could probably live with the borders overlapping at small sizes so long as tmux does not crash or behave incorrectly. Anyone with a window that is just too small for the panes can't really expect it to work usefully, so long as it goes back to normal when the window is big enough again.

When the window is big enough it should probably do the right thing with small panes though.

redesigndavid:
Just pushed updates that revert the layout rebalance helpers. Yeah, it is definitely an edge case (didn't want to miss the chance to use that pun).

The broken case seems to occur specifically when you do the split while in joined border type and then switch to separate. No odd borders when you do the split while already in pane-border-type separate:
```
  ./tmux -Lx new \; set -g pane-border-type separate \; splitw -fhl1
```

Chances are most users won't switch between `joined` and `separate` that often, or have enough splits in a tiny window to hit this. Happy to leave it as _accepted weirdness_ for such edge cases.

nicm:
I think this is OK but I am away next week so I will probably not look at it until I am home again. This will be a 3.9 feature in any case.

redesigndavid:
Thank you for taking care of this!

nicm:

redesigndavid:
Ahh I'm also away but will be back tomorrow. I will rebase tomorrow no problem. 👍

redesigndavid:
Rebased onto master. Had a few conflict resolutions mostly related to screen-redraw and window mode flags, plus two small post-rebase updates for the JSON layout test and pane_unzoomed_width with display-panes. Also reverted the reflow-on-separate change since it felt overkill for an edge case.

PR #5438, opened 2026-07-28
 Add command help and readable list output

Michael-RDev:
## Summary

This adds a built-in `help` command and improves the default interactive output
of tmux's most commonly used list and show commands.

The new output is intended for people reading it in a terminal. Existing
line-oriented output is preserved for scripts, pipes, control mode clients, and
commands that provide an explicit `-F` format.

## Motivation

This came out of my workflow training models on GPU clusters. I use tmux to
keep training, monitoring, data transfer, and debugging sessions alive while I
disconnect and reconnect to a cluster.

Those jobs often run across multiple weeks. Each time I needed to spin up a new
cluster, I found myself having to remember or look up the same tmux commands
and flags again. The existing command list is complete, but it is difficult to
scan when I am trying to get a cluster running quickly.

I wanted command discovery to feel closer to Git: useful top-level overview,
clear per-command usage, practical examples, and readable tables for common
status commands.

## What changed

- Add `tmux help` with a categorized overview of all commands.
- Add `tmux help <command>` with the command's purpose, usage, alias, and
  practical examples.
- Include at least one example for every registered command.
- Expand `tmux -h` with a short list of common commands.
- Add tables for:
  - `list-commands`
  - `list-sessions`
  - `list-windows`
  - `list-panes`
  - `list-clients`
  - `list-keys`
  - `list-buffers`
  - `show-options`, `show-window-options`, and `show-hooks`
  - `show-prompt-history`
- Wrap long usage text and table values based on terminal width.
- Add ANSI styling for headings, identifiers, state, warnings, and
  errors.
- Respect `NO_COLOR` and `TERM=dumb`.

## Examples

```sh
tmux help
tmux help split-window
tmux help rename-session
tmux list-commands
tmux list-sessions
tmux list-panes
tmux show-options -g
```

Detailed help now includes an example section:

```text
NAME
  rename-session - Rename a session
USAGE
  tmux rename-session [-t target-session] new-name
ALIAS
  rename
EXAMPLES
TASK              COMMAND
Rename a session  tmux rename-session -t work project
```

## Compatibility

- Redirected and piped output keeps the existing line-oriented format.
- Control mode output is unchanged.
- An explicit `-F` format always takes precedence.
- Value-only forms such as `show-options -v` remain unchanged.
- `NO_COLOR` disables ANSI styling without disabling the readable layout.

For example, these continue to be suitable for scripts:

```sh
tmux list-sessions | while IFS= read -r session; do
    # ...
done

tmux list-panes -F '#{pane_id} #{pane_current_command}'
```

## Testing

- Built with:

  ```sh
  ./autogen.sh
  ./configure --enable-debug --disable-jemalloc
  make -j4
  ```

- Added `regress/command-help.sh`, covering:
  - all registered commands having help and examples;
  - canonical command names and aliases;
  - interactive table selection;
  - narrow terminal rendering;
  - `NO_COLOR`;
  - legacy pipe output; and
  - explicit `-F` compatibility.
- Existing buffer, option, session, window, pane, control, format, input,
  target, theme, TTY, UTF-8, and hook regressions pass.
- The full regression run has one unrelated existing-result mismatch in
  `screen-redraw-menus.sh`, where a menu over a split pane is clipped
  differently from its checked-in golden scene.


nicm:
Thanks.

Help is a good idea but it should be a flag to `list-commands` rather than a new command (maybe with a default `command-alias` to map it to `help`), and the help text should be part of `struct cmd_entry`, not all in one file. And no examples please, just usage and a few sentences describing the command, like the options and mode help. Maybe a sentence for each option. We don't want to be carrying around the whole man page in the code, just an aide memoire.

The output format changes, can you make a separate PR and we can discuss? I don't think we will want them in this form but there might be scope for alternative default formats.

nicm:
I wonder if it would be nice to have a command browser mode which shows the command, usage, help and maybe a key to open the man page located at that command (although this may only be possible on systems using `mandoc`). But that would be separate, we need the help text in first.

Michael-RDev:
Thanks, that sounds good. I'll rework the PR : )

I’ll also remove the output-format changes and submit those in a separate PR for discussion.

The command browser idea sounds useful too. Once the help text is in place, that could be a follow-up. It'll be cool if we can look separately at jumping to the relevant section of the man page, potentially with `mandoc`.

nicm:
This looks good, although a couple of things:

Maybe the default should not include the full usage, like `tmux lsk -N` (or `git -h -a`), so something like:
~~~~
$ tmux lscm -h
attach-session Attach to an existing session.
bind-key       Bind a key to one or more commands.
~~~~~
If we do this then perhaps two `-h`s could show the full usage?

I was also thinking about showing the flags, so like:
~~~~
$ tmux lscm -hh attach
attach-session (attach) [-dErx] [-c working-directory] [-f flags] [-t target-session]
    Attach to an existing session.
    -c new default working directory
    -d detach any other clients
    -E do not update environment
    -f flags for new client
    -r attach read-only
    -t target session to attach
    -x send SIGHUP with -d
~~~~
That would need a lot more curation however. And I wonder if it would be both too much and too little - either we end up copying the entirety of `tmux.1`, or we put in so little they are useless? What do you think? (@mgrant0 was talking about changing the man page to list out flags like this, at least for the commands with many flags.)

I'm also thinking maybe we should just put the command browser into customize mode, although there is not really any "customization" involved. For jumping to the man page section, on OpenBSD you can do this:
~~~~
man -O tag=list-keys tmux
~~~~
On other platforms I think you need something like:
~~~~
env MANPAGER="less -p '^[[:space:]]+list-keys([[:space:]]+\\[|$)'" man tmux
~~~~
So we would need probably need to make this an option and have a default set by configure like we do for `lock-command`.

The editor API (`spawn_editor`) could be reused to open a man page interactively, it would just need an extra argument to use a different command rather than `editor`.

Michael-RDev:
Sounds good, will implement soon 👍 : )

nicm:
@Michael-RDev are you still planning to work on this? If not, let me know because I might do it.

PR #5440, opened 2026-07-29
feat: Add WINDOW_MODE_HIDE_SCROLLBARS (and FILL_WINDOW) for display-panes

redesigndavid:
## Summary
- Add `WINDOW_MODE_HIDE_SCROLLBARS` so modes can hide scrollbars while zoomed.
- Set it on `window_panes_mode` so `display-panes` does not leave scrollbars shifting the overlay.
- Also add `WINDOW_MODE_FILL_WINDOW` on the same mode for later use (unused here).

This is a split from #5433 so this plumbing can land on its own.

## Test plan
- [x] `set -g pane-scrollbars on`, split panes, generate scrollable content
- [x] Run `display-panes` (`q`) and confirm scrollbars disappear while the overlay is active
- [x] Confirm scrollbars return after display-panes exits

### DIFF
--- a/tmux.h
+++ b/tmux.h
@@ -1194,6 +1194,8 @@ struct window_mode {
 	int		 flags;
 #define WINDOW_MODE_HIDE_PANE_STATUS 0x1
 #define WINDOW_MODE_NO_STACK 0x2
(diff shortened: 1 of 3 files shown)


nicm:

PR #5441, opened 2026-07-29
Keep the background when a wide character or a tab is broken up

aymanbagabas:
## Summary
- Padding cells now carry a background colour, so the columns covered by a wide character or a tab keep their background when a write breaks it up.
- The terminal is cleared in runs of the same background, since adjacent padding can come from different characters.
- Add `regress/tab-cell-background.sh`.

## Problem
A tab is stored as a single multiple-column cell (`GRID_FLAG_TAB`), and a wide character as a cell plus padding. Padding carries no attributes of its own — `grid_padding_cell` is default fg/bg.

When a write lands *inside* one of those spans the leftover columns are cleared to `grid_default_cell`, and the terminal is sent the matching clear with `bg = 8`, so the background colour underneath is silently lost.

Minimal reproductions, in a 24 column pane:

```sh
printf '\033[44m\033[K\t\033[4GX'		# write inside a tab
printf '\033[44m\033[K\344\270\226\033[2GX'	# overwrite half a wide character
```

`capture-pane -pe` for the first, before:

```
   \033[44mX\033[49m    \033[44m
```

Columns 0-2 and 4-7 have dropped back to the default background. After:

```
\033[44m   X
```

I hit this from a TUI that uses tabs for cursor motion while horizontally scrolling a syntax highlighted diff: whitespace runs kept losing their row background a frame later. It only reproduced under tmux.

## Fix
`grid_set_padding()` and `grid_view_set_padding()` take a background, and the two callers pass the background of the cell the padding belongs to.

`screen_write_clear_cell()` clears a cell keeping its background, and is used by both places that clear padding — `screen_write_collect_end()`, which is where plain ASCII goes, and `screen_write_overwrite()`, which handles wide characters and redraws the line from the grid afterwards.

Padding cells next to each other can come from different characters, for example when a tab is partly deleted and padding from two tabs of different colours ends up together:

```sh
printf '\033[42m\033[K\t\033[43m\033[K\t\033[0m\033[4G\033[8P\rx'
```

So `screen_write_collect_end()` queues one clear per run of the same background instead of a single one with the hardcoded 8.

This also lets the `GRID_FLAG_TAB` special case in `screen_write_overwrite()` go away, as it only covered trailing padding.

## Test plan
- [x] `regress/tab-cell-background.sh` covers all three cases and fails on master
- [x] The test attaches a second server to the first, like `am-terminal.sh`, so what is sent is checked and not just what is stored — capturing the same server only shows the grid, which was already correct in the third case
- [x] Full `regress` suite: no new failures (`check-names`, `display-panes` and `screen-redraw-menus` already fail here on macOS)
- [x] Verified end to end against the original TUI: drift before, clean after

### DIFF
--- a/grid-view.c
+++ b/grid-view.c
@@ -47,9 +47,9 @@ grid_view_set_cell(struct grid *gd, u_int px, u_int py,
 
 /* Set padding. */
 void
-grid_view_set_padding(struct grid *gd, u_int px, u_int py)
+grid_view_set_padding(struct grid *gd, u_int px, u_int py, int bg)
 {
-	grid_set_padding(gd, grid_view_x(gd, px), grid_view_y(gd, py));
+	grid_set_padding(gd, grid_view_x(gd, px), grid_view_y(gd, py), bg);
 }
 
 /* Set cells. */
--- a/grid.c
+++ b/grid.c
@@ -682,9 +682,13 @@ grid_set_cell(struct grid *gd, u_int px, u_int py, const struct grid_cell *gc)
 
 /* Set padding at position. */
 void
-grid_set_padding(struct grid *gd, u_int px, u_int py)
+grid_set_padding(struct grid *gd, u_int px, u_int py, int bg)
 {
-	grid_set_cell(gd, px, py, &grid_padding_cell);
+	struct grid_cell	gc;
+
+	memcpy(&gc, &grid_padding_cell, sizeof gc);
+	gc.bg = bg;
+	grid_set_cell(gd, px, py, &gc);
 }
 
 /* Set cells at position. */
--- a/regress/tab-cell-background.sh
+++ b/regress/tab-cell-background.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+# When a write breaks up a wide character or a tab, the columns it covered must
+# keep the background colour, both in the grid and on the terminal. The second
+# server is attached to the first so that what is actually sent is checked, not
+# just what tmux has stored.
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+TMUX="$TEST_TMUX -LtestA$$ -f/dev/null"
+$TMUX kill-server 2>/dev/null
+TMUX2="$TEST_TMUX -LtestB$$ -f/dev/null"
+$TMUX2 kill-server 2>/dev/null
+
+TMP=$(mktemp)
+trap "rm -f $TMP" 0 1 15
+
+# 1. Write inside a tab. 2. Overwrite the right half of a wide character.
+# 3. Two tabs of different colours, the first partly deleted so that padding
+#    from both ends up next to each other, then written over.
+# The sleep is so that the client is attached before anything is written and
+# the changes are sent as they happen rather than as one redraw.
+$TMUX2 -f/dev/null new -d -x24 -y3 "
+	sleep 2
+	printf '\033[44m\033[K\t\033[4GX'
+	printf '\033[2;1H\033[44m\033[K\344\270\226\033[2GX'
+	printf '\033[3;1H\033[42m\033[K\t\033[43m\033[K\t\033[0m\033[4G\033[8P\rx'
+	cat" || exit 1
+$TMUX2 set -g status off || exit 1
+
+$TMUX -f/dev/null new -x24 -y4 -d "$TMUX2 attach" || exit 1
+sleep 4
+
+$TMUX capturep -pe -S0 -E2 >$TMP || exit 1
+$TMUX kill-server 2>/dev/null
+$TMUX2 kill-server 2>/dev/null
+
+printf '\033[44m   X\n X\n\033[49mx\033[42m  \033[43m             \033[49m\n' |
+	cmp - $TMP || exit 1
+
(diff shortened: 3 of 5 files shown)


nicm:
Seems like this should be done in `screen_write_overwrite` where it writes over the padding, not by adding a new function.

aymanbagabas:
That was my first attempt actually, but it doesn't work — `screen_write_overwrite` is only reached from `screen_write_cell`, and plain ASCII never gets there. `screen_write_collect_add` collects it and `screen_write_collect_end` does its own padding clearing (to `grid_default_cell`, plus CLEAR items with `bg = 8` hardcoded). Fixing overwrite alone changed nothing in my repro.

There are two cases, one for each path:

```sh
printf '\033[44m\033[K\t\033[4GX'		# collect_end
printf '\033[44m\033[K\t\033[4G\344\270\226'	# screen_write_cell -> overwrite
```

Both lose the background without the change. The wide one fails even though `screen_write_overwrite` already has the `GRID_FLAG_TAB` branch, because that only covers trailing padding — the leading padding and the head cell still get `grid_default_cell`.

So doing it in `screen_write_overwrite` means the same tab logic in `screen_write_collect_end` as well, plus passing the tab's bg into `bci->bg`/`aci->bg` instead of 8. Expanding the tab before the write seemed simpler: one place, and neither clearer sees padding it can get wrong.

Happy to do it the other way if you'd prefer, just let me know.

nicm:
Yeah you're right. But I think the two places that already do it should handle it rather than adding a third (unless we replace them both with one).

`screen_write_collect_end` and `screen_write_overwrite` would both need to change to find the non-padding cell in order to work out the background colour, but possibly that could be common?

The other option would be to have padding cells hold the colour - they'd only need the background colour. There are only two callers to `grid_view_set_padding`, so maybe that would work? Might be more efficient since they we wouldn't need to search back to find the visible cell.

aymanbagabas:
Went with the padding option, it came out much better than I expected — updated the PR.

`grid_set_padding()` and `grid_view_set_padding()` take a background now, and the two callers pass the background of the cell the padding belongs to. Then one helper does the clearing in both places:

```c
static int
screen_write_clear_cell(struct grid *gd, u_int px, u_int py)
{
	struct grid_cell	 gc;
	int			 bg;

	grid_view_get_cell(gd, px, py, &gc);
	bg = gc.bg;

	memcpy(&gc, &grid_default_cell, sizeof gc);
	gc.bg = bg;
	grid_view_set_cell(gd, px, py, &gc);

	return (bg);
}
```

`screen_write_collect_end()` and `screen_write_overwrite()` call that instead of setting `grid_default_cell` directly, and the returned background goes into `bci->bg`/`aci->bg` instead of 8. No searching back for the visible cell, and it lets the `GRID_FLAG_TAB` special case in `screen_write_overwrite()` go away, since that was only covering the trailing padding anyway. Net is smaller than what I had before.

It also fixes the same thing for wide characters, which I hadn't realised was broken:

```sh
printf '\033[44m\033[K\344\270\226\033[2GX'
```

Overwriting the right half leaves column 0 at the default background before, blue after.

`regress` is clean apart from three that already fail here on macOS (`check-names`, `display-panes`, `screen-redraw-menus`).

nicm:
Looks like the right idea, but if you have padding from two tabs next to each other, it will clear them in the same colour. This can happen if the tab itself is deleted. For example try:
~~~~
printf '\033[42m\033[K\t\033[43m\033[K\t\033[0m\033[4G\033[8P\rx\n'
~~~~
I think the clear code will need to either clear each character individually in its own background colour, or build runs of characters in the same background and clear them together. Or just write out the section from the grid which is correct.

aymanbagabas:
Good catch, thanks — that was wrong. The grid was right but the terminal got one clear for the whole run, so the green columns came out yellow.

Fixed by clearing in runs of the same background. The clearing loops don't work out the colour any more, they just clear each cell keeping its own background, and then a second pass over the range queues one `screen_write_collect_insert_clear()` per run:

```c
static void
screen_write_clear_runs(struct screen_write_ctx *ctx, u_int px, u_int nx)
{
	struct screen		*s = ctx->s;
	struct grid_cell	 gc;
	u_int			 xx, start = px;
	int			 bg = 8;

	for (xx = px; xx < px + nx; xx++) {
		grid_view_get_cell(s->grid, xx, s->cy, &gc);
		if (xx == start)
			bg = gc.bg;
		else if (gc.bg != bg) {
			screen_write_collect_insert_clear(ctx, start,
			    xx - start, bg);
			start = xx;
			bg = gc.bg;
		}
	}
	screen_write_collect_insert_clear(ctx, start, xx - start, bg);
}
```

That let `bci`/`aci` go away, `screen_write_collect_insert_clear()` already does the item. `screen_write_overwrite()` doesn't need any of this, it sets `redraw` and the line is redrawn from the grid.

Your case now comes out the same on the terminal as in the grid:

```
x\033[42m  \033[43m             \033[49m
```

I also rewrote the test — the old one only captured the pane on the same server, which is the grid, so it couldn't have caught this. It now attaches a second server to the first the way `am-terminal.sh` does, and checks what is actually sent. It covers the tab, the wide character and your two tab case, and fails on master and on the previous version of this branch.

nicm:
This looks good, applied to OpenBSD now with some trivial changes, will be in GitHub later. Thanks!

PR #5444, opened 2026-07-30
Do not let a stuck client prevent the server from exiting

bmaurer:
### Problem

If a control client's terminal has stopped accepting output (for example its ssh connection or terminal emulator has died), the server can never finish exiting. After `kill-server` or SIGTERM the server stays alive forever with `server_exit` set, and `server_accept()` then accepts and immediately closes every new connection — so every subsequent tmux command on that socket fails with `server exited unexpectedly` until the server is killed by hand.

Observed in the wild: a 3.7b server on a dev machine sat in this state for over a week, rejecting every attach, after the EternalTerminal connection under its `-CC` client died.

### Cause

Two unbounded waits in the exit handshake:

1. `server_client_check_exit()` will not send `MSG_EXIT` until `control_all_done()` — i.e. until the write bufferevent to the client's terminal has fully drained. A terminal nobody reads never drains, so the client is never told to exit and never leaves the clients list. (`control_discard()` only drops not-yet-queued pane blocks; bytes already in the write buffer, and non-pane blocks, remain.) The "too far behind" eviction in `control_check_age()` funnels into this same gate, so it does not help.
2. Even once `MSG_EXIT` is sent, the client is only removed when it closes its socket, which a client stuck writing to a dead tty may never do.

This completes the scenario from #5356 / #5357: those stopped the server from wedging on, and queueing new output to, a control client whose terminal reader died; this stops the server waiting forever on such a client at exit.

### Fix

Bound both waits with a per-client timer (`CLIENT_EXIT_TIMEOUT`, 10 seconds):

- If the output an exiting client must flush has not drained in time, discard it — it can never be delivered — via a new `control_discard_all()`, and let the exit continue.
- If the client has still not closed its connection 10 seconds after `MSG_EXIT`, drop it with `server_client_lost()`.

Suspended clients are exempted so a stopped (ctrl-Z) client is not penalized. Healthy exits are unaffected: the timer is deleted when the client closes normally, and a normal `kill-server` still completes in milliseconds.

### Reproduction / test

`regress/control-client-exit-stalled.sh` (included): attach a `-C` client with stdout on a fifo whose read end is held open but never read, flood a pane, then `kill-server`. On master the server is still alive 20+ seconds later and any new client gets `server exited unexpectedly`; with this change the server exits ~10 seconds after `kill-server`.

Existing control regress tests (`control-client-exit.sh`, `control-client-wait-exit.sh`, `control-client-sanity.sh`, `control-client-size.sh`, `control-notify-events.sh`, `control-subscriptions.sh`) pass, and the build is warning-free.

As with my previous PRs, AI assistance was used to investigate and prepare this change; I have reviewed and verified it.

### DIFF
--- a/control.c
+++ b/control.c
@@ -878,6 +878,24 @@ control_discard(struct client *c)
 	bufferevent_disable(cs->read_event, EV_READ);
 }
 
+/*
+ * Discard all output for a client, including output which has already been
+ * queued to be written. Used when the client has stopped accepting output so
+ * it is never going to be delivered.
+ */
+void
+control_discard_all(struct client *c)
+{
+	struct control_state	*cs = c->control_state;
+	struct control_block	*cb, *cb1;
+
+	control_discard(c);
+	TAILQ_FOREACH_SAFE(cb, &cs->all_blocks, all_entry, cb1)
+		control_free_block(cs, cb);
+	evbuffer_drain(cs->write_event->output,
+	    EVBUFFER_LENGTH(cs->write_event->output));
+}
+
 /* Stop control mode. */
 void
 control_stop(struct client *c)
--- a/regress/control-client-exit-stalled.sh
+++ b/regress/control-client-exit-stalled.sh
@@ -1,0 +1,77 @@
+#!/bin/sh
+
+# A control client whose terminal has stopped accepting output (for example
+# its ssh connection or terminal emulator died) never drains the output that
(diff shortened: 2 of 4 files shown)


nicm:
I have applied this to OpenBSD now, although I trimmed or deleted most of the rather excessive comments and renamed a function. Will be in GitHub later. Thanks!

PR #2, opened 2015-06-04
updated README.md with logo and headings

hackerkid:
README renamed to README.md to support markdown . Updated README.md with tmux logo. Added added suitable headings to various topics.

PR #5445, opened 2026-07-30
Support both kitty and sixel images

### LINKED ISSUE #4902: Support kitty image protocol

mgrant0:
This PR adds generalised support for both kitty and sixel images.  Basically it breaks up images into grid size rectangles.  If the terminal is kitty then it sends kitty the image using the special unicode character encoding and kitty displays it locally.  If the terminal supports sixel images, the image is sent to the terminal using sixel escape codes.  This uses the new screen-redraw scene and span drawing, so it works in and around floating panes.  Images can be scrolled into the copy buffer without issue.


senthil-instrumentl:
Was looking for kitty support in tmux and saw this PR. Not sure if i'm doing something wrong, but getting:

```
❯ ./configure --with-image-support
...
❯ make
make: *** No rule to make target `image-kitty.c', needed by `image-kitty.o'.  Stop.
```

mgrant0:
Oops sorry, forgot to add the new files.  Added now.  Please try again.

mgrant0:
image

image

Both kitty and sixel images seem to work.  Plese let me know.  There's also a rudimentary fallback to ascii in terminals with no image support at all.

mgrant0:
@senthil-instrumentl found another missing .h file which is now fixed (included inside image-kitty.h).  Please give it a try.  Feedback welcome.  Thanks!

mgrant0:
I am using it inside Windows Terminal for development.  I tried it briefly in kitty term.  I fixed a flashing issue but that was already pushed yesterday.  I can try it in wezterm probably tomorrow.

Can you attach the png file or tell me where to get it?  

Hopefully we can fix the flashing everywhere.  We will get this to work.  Thanks for your help testing it!

paranoidi:
I Deleted my previous comments. Apologies for the email noise. My previous tests were invalid.

However the PR sill does not work as expected for me. The images are rendered at incorrect position and do not move as the terminal scrolls. It also causes occasionally the statusbar to disappear when rendering or replaced by newline symbol.

.tmux.conf has

`set -ga terminal-features ',*:RGB:sixel:sync:kitty'`

Example:

image

Using chafa produces similar results.

I can not get kitty to work either. It will either do nothing or report unsupported.

```
❯ kitty +kitten icat ~/Pictures/bitwig_layer_transparency.png

~/Pictures
❯ kitty +kitten icat ~/Pictures/bitwig_layer_transparency.png
Error: This terminal does not support the graphics protocol use a terminal such as kitty, WezTerm or Konsole that does. If you are running inside a terminal multiplexer such as tmux or screen that might be interfering as well.
```

I am using WezTerm built from master couple days ago. All of these work flawlessly outside tmux.

mgrant0:
I reproduced this in wezterm locally.  I've added a fix for this to this branch, please pull it and recompile.

IMPORTANT: you should _remove_ this from your .tmux.conf
```
set -ga terminal-features ',*:RGB:sixel:sync:kitty'
```
Then fully restart the tmux server. tmux will detect each attached terminal separately:

  - WezTerm: RGB plus sixel.
  - Kitty: RGB plus Kitty placeholders.
  - Non-image terminals: text fallback.

If you have a separate reason to force sync, it should be configured only for the appropriate terminal pattern, not with *. The global line incorrectly claims every client supports both image protocols, which defeats per-client backend selection.

bptato:
I like the initiative, but there are a few bugs.
First, a minor one: HLS conversion has the usual bug where it doesn't match Sixel semantics.  (Hue is rotated -120 degrees compared to common use today, test with `img2sixel -t hls ...`).  Also, rounding.

```c
diff --git a/image-sixel.c b/image-sixel.c
index 577ba51a..595c3836 100644
--- a/image-sixel.c
+++ b/image-sixel.c
@@ -429,9 +429,9 @@ sixel_colour_to_rgb(u_int colour, u_char *r, u_char *g, u_char *b)
 	double	h, l, s, p, q;
 
 	if (type == 2) {
-		*r = ((colour >> 16) & 0x1ff) * 255 / 100;
-		*g = ((colour >> 8) & 0xff) * 255 / 100;
-		*b = (colour & 0xff) * 255 / 100;
+		*r = (((colour >> 16) & 0xff) * 255 + 50) / 100;
+		*g = (((colour >> 8) & 0xff) * 255 + 50) / 100;
+		*b = ((colour & 0xff) * 255 + 50) / 100;
 		return;
 	}
 	if (type != 1) {
@@ -448,9 +448,9 @@ sixel_colour_to_rgb(u_int colour, u_char *r, u_char *g, u_char *b)
 	}
 	q = l < 0.5 ? l * (1 + s) : l + s - l * s;
 	p = 2 * l - q;
-	*r = sixel_hue(p, q, h + 1.0 / 3) * 255;
-	*g = sixel_hue(p, q, h) * 255;
-	*b = sixel_hue(p, q, h - 1.0 / 3) * 255;
+	*r = sixel_hue(p, q, h) * 255 + 0.5;
+	*g = sixel_hue(p, q, h - 1.0 / 3) * 255 + 0.5;
+	*b = sixel_hue(p, q, h + 1.0 / 3) * 255 + 0.5;
 }
 
 /* Convert decoded SIXEL data into the protocol-neutral immutable image. */
```

Second, sixel_from_image is...  not pretty.  The way you'd normally go about this is some kind of image quantization algorithm (I've had luck with octrees), with the current nearest neighbor approach the output looks awful, as demonstrated by your own screenshot.  Or at the very least you should apply some error diffusion, [ref](https://en.wikipedia.org/wiki/Floyd%E2%80%93Steinberg_dithering).

Third, sizing is broken.  For example,

```
printf '\033P0;1q"1;1;26;26#0;2;50;50;50#1;2;0;0;0#2;2;49;49;49#3;2;33;33;33#4;2;47;47;47#5;2;44;44;44#6;2;48;48;48#2?O?C!5?S!4?__?_?__$#0?_wW!5KGwo_-#0?~~!8?!10@BB}w$#2!12?a_A?A?A?A??@E-#0?~~!7?KK~~KK!7?~~$#2!9?K!6?K-#0?F^[owowowowowowowowooS^F$#2?G!10?@@!7?G??G-\033\'
```

*should* print a 26x26 image, but it gets scaled up to the cell size, so if I have 20x10 pixel cells then it blows up to 40x30.  From a quick look at the code, `draw_rectangle` only gets sizes in cells, so perhaps that's related.

Finally, maybe I'm wrong but it looks like the cropping Kitty parameters (`X`/`Y`/`w`/`h` (edit: sorry, I meant `x`/`y`, although I use `X`/`Y` too but it's less important)) aren't respected?  Those are kind of necessary for efficient TUIs, e.g. I use them extensively in [chawan](https://chawan.net) so that I don't have to re-send images on scroll.

mgrant0:
- Correct DEC SIXEL HLS hue ordering and rounding in image-sixel.c. 
  - Preserve original pixel dimensions so partial edge cells are not stretched.
  - Adaptive median-cut colour quantization with Floyd–Steinberg error diffusion.
  - Kitty’s actual source-cropping keys, lowercase x/y/w/h, plus c/r placement sizing.
    Uppercase X/Y are destination offsets within the first cell, not crop coordinates,
    according to the Kitty protocol (https://sw.kovidgoyal.net/kitty/graphics-protocol/).
  - Cropped Kitty placements use zero-copy immutable pixel views in image.c, and row-
    strided views are correctly streamed when uploaded.

@bptato, Did this address all your concerns?  The sixel image definitely looks better to me now, thanks!

bptato:
It's better, thanks.  Some issues left:

* In my Sixel encoder I also do error diffusion on the alpha channel; the result looks a bit strange, but at least when you have an image with alpha<128 everywhere you don't end up with an invisible image.
* Kitty input is still wonky, e.g. with `cha -o buffer.images=true https://developers.google.com/speed/webp/gallery1` you'll see partial images flowing into each other etc.  At least part of this may be because the cursor is desync'ing, I think if you emulate the regular protocol with the unicode placeholders then you'll have to move the cursor back after output.  (But I may be wrong here.)

Also, even with the error diffusion there are some pretty bad artifacts ([image](https://www.gstatic.com/webp/gallery/5.webp), up: kitty -> tmux -> sixel on this branch, bottom: sixel passed through tmux master):

compare

Apples to oranges of course, as the palettizer algorithm is different.  But you can see the stripes per line, which I think happens because the palette & dither is recomputed for each line.  At least the latter must be kept.
(That aside, I'm a bit suspicious about "Sixel-per-line"'s performance characteristics, [ref](https://nick-black.com/dankwiki/index.php?title=Theory_and_Practice_of_Sprixels#Dead_ends).  Would it be possible to output the whole thing in one go?)

mgrant0:
Wow but it's really close!  I see it gets dithery above his head and up his arm.  What do you think is causing that?  I can try to figure it out but if you have some ideas that would be super helpful.  Have a look in image-sixel.c and see if it makes sense.

This is the first step to get this to paint the image in stripes because that's the way the current screen-redraw works.  It splits the screen up into a "scene" of "spans".  When each span is output, the sub-rectangle of the image is output in that stripe.  This way, we treat the image just exactly like characters on the screen.  They can be scrolled up into the copy buffer, moved around on the screen in floating panes, and partially obscured by floating panes or popups, everything just works.

In a near future revision of this, I hope to be able to join up ajoining spans into larger rectangles.  Still though, an image may be painted with multiple sub-rectangles if it is partly obscured by one or more floating panes.  But it should improve the efficiency of sending these images to the terminal.

But I don't think this is the solution to this image aberation.  It looks like either an encoder or decoder issue and it well could be the encoding into these stripes that is causing this.

Can you show me some of the other examples of images flowing into one another?  It sounds like the sub-rectangles are not properly scaled for the terminal or something like that.  

By the way, this idea of using small rectangles is based on how Kitty does images.  Each of those special unicode characters carries with it which piece of which puzzle belongs on top of that character.

paranoidi:
Rendering is pretty broken atm. Example:

image

Original:
ChatGPT Image Apr 3, 2026, 09_26_54 PM

I still can not get kitty images to show within tmux even after correcting my .tmux.conf per instructions. I tried with default config and no dice with that either.

mgrant0:
Without the set line in .tmux.conf, can you confirm the Wezterm issue is fixed? 

I'll have a look at the kitty image issue.

bptato:
> Wow but it's really close! I see it gets dithery above his head and up his arm. What do you think is causing that? I can try to figure it out but if you have some ideas that would be super helpful. Have a look in image-sixel.c and see if it makes sense.

Sorry, I wasn't really clear about it:

1. Right now, the dither starts from scratch for every (terminal grid) line, meaning the error is periodically zeroed out.  Somehow the `current`/`next` arrays should be retained.
2. Similarly, recomputing the palette for every line makes color approximation unstable; ideally it should be computed for the entire image at once (maybe store it in `struct image`?)

That should make the stripes disappear.
(Related: point 2 also makes an optimization possible where, when receiving Sixel input, you just keep a copy of the original palette in case you ever need Sixel output again.  Then sixel -> tmux -> sixel would stay lossless.)

> This is the first step to get this to paint the image in stripes because that's the way the current screen-redraw works. It splits the screen up into a "scene" of "spans". When each span is output, the sub-rectangle of the image is output in that stripe. This way, we treat the image just exactly like characters on the screen. They can be scrolled up into the copy buffer, moved around on the screen in floating panes, and partially obscured by floating panes or popups, everything just works.
>
> In a near future revision of this, I hope to be able to join up ajoining spans into larger rectangles. Still though, an image may be painted with multiple sub-rectangles if it is partly obscured by one or more floating panes. But it should improve the efficiency of sending these images to the terminal.

That makes a lot of sense, thanks.

> Can you show me some of the other examples of images flowing into one another? It sounds like the sub-rectangles are not properly scaled for the terminal or something like that.

With the aforementioned command, after scrolling around the page using Kitty 0.48.0 for a bit:

glitch

If I keep scrolling, eventually the entire layout collapses into a mess.  I get similar results with xterm, and setting `cha` output to sixel (`-o display.image-mode=sixel`) mostly alleviates the issue, hence my suspicion that it's a cursor desync in the Kitty input handler.  But that doesn't explain the stray image remnants, so there's probably more to it...

(I say mostly; there are still some minor sizing problems in Sixel too.  I'll try to post a reduced example later.   (edit: never mind, Sixel input seems to work correctly, only Kitty has issues.))

mgrant0:
@bptato let me know if this last push fixes the sixel issues.

bptato:
Well, it fixes it for Sixel input.  But quantization is still broken for kitty -> tmux -> sixel ([image](https://www.gstatic.com/webp/gallery/1.webp)):

compare

What is the algorithm based off of, anyway?

By the way, I've found the source of my Kitty cursor positioning issue: the patch doesn't respect 'C=1', which is supposed to keep the cursor where it was.  I suppose it should set a flag and suppress cursor movement in `image_write`.

Apropos `image_write`, I've noticed this:

```c
	sx = im->sx;
	if (sx > screen_size_x(s) - cx)
		sx = screen_size_x(s) - cx;
	sy = im->sy;
	if (sy > screen_size_y(s) - 1)
		sy = screen_size_y(s) - 1; // <-- ??
	if (sx == 0 || sy == 0)
		return;

	if (screen_size_y(s) - cy <= sy) {
		lines = sy - (screen_size_y(s) - cy) + 1;
		screen_write_scrollup(ctx, lines, bg);
		if (lines > cy)
			screen_write_cursormove(ctx, -1, 0, 0);
		else
			screen_write_cursormove(ctx, -1, cy - lines, 0);
		cy = s->cy;
	}
````

Surely we don't want to crop the image to the screen height when moving the cursor?

I believe what should happen here is that if the image overflows the screen, it should first scroll down by `lines`, and *then* the image should be cropped from the top so it fits into the screen (i.e. set some kind of "y origin" variable).
Except of course if cursor movement is suppressed, in that case we can crop at the bottom and skip scrolling.

**Comment by mgrant0** [comment]
@bptato Ok I made progress on the kitty images and the kitty->tmux->sixel support.  Can you please restest  this branch.

I can't seem to reproduce this issue: "scrolling, eventually the entire layout collapses into a mess".  If it's still messing up when you scroll up and down, please can you provide me with a list of instructions how to reproduce that?  I have the 1.wepb image you used.  If you used a different image, please provide me a link.

I tested in kitty with this:
```
kitty +kitten icat --scale-up --fit=both 1.webp
```
to make the image fit full width.

**Comment by bptato** [comment]
> Ok I made progress on the kitty images

C=1 indeed works now, thanks.

> and the kitty->tmux->sixel support.

From what I can see, you removed the dither, which I find puzzling because that's the only part that worked.  Now the image looks like:

new

The part that is broken is the quantization, i.e., your (?) median cut routine.  In other words, the colors are being selected incorrectly.

> I can't seem to reproduce this issue: "scrolling, eventually the entire layout collapses into a mess". If it's still messing up when you scroll up and down, please can you provide me with a list of instructions how to reproduce that? I have the 1.wepb image you used. If you used a different image, please provide me a link.

The bug(s) I was talking about manifested when viewing the webp gallery in chawan as described [above](https://github.com/tmux/tmux/pull/5445#issuecomment-5166692524).  By "scroll" I referred to scrolling inside the TUI, which in the background both scrolls the text *and* moves the placements.  (This is necessary because every terminal I tested handled scroll + Kitty placements differently.)
Anyway, layout being messed up was caused by the lack of C=1 and is now fixed.  Remnants of images still linger on the screen, which seems to be related to moved placements not being deleted (?); I'll try to reduce that too.

>
> I tested in kitty with this:
>
> ```
> kitty +kitten icat --scale-up --fit=both 1.webp
> ```

I'm pretty sure icat discriminates against tmux, so this would use the passthrough sequence by default, i.e., this doesn't really test anything.  (I've tried passing `--transfer-mode=stream --passthrough=none` but then it doesn't print anything for some reason.)

**Comment by bptato** [comment]
> I'll try to reduce that too.

```sh
#!/bin/sh
printf '\033_GC=1,s=26,v=26,p=1,q=2,i=1,a=T,f=100,m=0;iVBORw0KGgoAAAANSUhEUgAAABoAAAAaCAYAAACpSkzOAAACmUlEQVR4XmOgF2CEWVRfX8/w//9/BkZGRgy7YeKNjY1kuwtsam1tLQMTE5MG0MBsoEWW6KYBxf8DxScD6UVNTU1kWcZYU1MD0qjLzMx8EmgYJy5TgJb8/ffvn+2vX7+Od3Z2kmwZC9AnIE0FIEuAhl0H4ulA9k9kk4Bi4UB1TkA8H2iRYXl5+Xd2dna8loGCG2Q2LLgZQXED1PEIaLjs379/g4Ds9VDLwQYBxRiAPpFmZWW9DFQjCDRgLxDfJeQloJpdQHPWA9X9A1kGs+gZ0BBJoKQtUOIIeqSDgheoKR6IF5ASZkAH3gKaCzLzFQsxGltaWhjq6uoWguIJ6hi82oCG8wIV+AEdpga0LBfIrmUh1oWg1AYM5iVA9UuI1JMGVDcTiCNIsghkOChIQ0NDwfasXr0ap33QeL8GUgD0HTeIZmIgAUAtYQZqYYZZSKx2oi2CWaKpqbkIhEm1jOg40tLSAjleEBgUUSAGkJ8PpN8Q6yOCFoF8AvQBTvOAqZHh+vXrDPjijGAcIcXJUqBPQOXda5iNQP5rkBiQv5SYYMTrIxUVFZC5IkDCCk8QWQHVgdS8xBeMeBNDe3s7A7CwfQl0uSmwKDIBYhdQGQbCIDZIDCQHUgNSS7ZFII2gUgEU6UADzwJz+kUgDa6zoOyzIDmoGgaKLAJpBheKkArxC7BIeQDCQPEvIEuJrQyJzkdQA38ADbcHYaBFP0ipcVGqCWDYBwPxOkYs1TkDCeA/pEkQBDRnLZD9GKhVjgUkCGTsBgomAekWIJYG4p8MlAFQrZgJMgJo/m4QzQJ1fT9QIBLI1gTiSRRaAm/gAM38DjSvH2QeuHECLW1xNk7IsRhoyXGgJVOBem+AExPMEHzNLTItYiAlVTJQCwAAJxkMkgbHo2UAAAAASUVORK5CYII=\033\\\n'
sleep 1
printf '\033_GC=1,s=26,v=26,p=1,q=2,i=1,a=p;\033\\\n'
```

The first `printf` should draw the icon, then the second one should move it to the next line.
So I guess it's happening because the patch doesn't handle placement ids (in this case, `p=1`). [ref](https://sw.kovidgoyal.net/kitty/graphics-protocol/#display-images-on-screen)

mgrant0:
I've reverted ab591ae4302f52f6a9775dce4965419f16dd7ea6 and hopefully now I have fixed the quantization, sorry about that misunderstanding.  Hopefully the quantization bug is fixed.  

I tested both using icat in kitty and displaying the image in tmux in ms terminal which output using sixel and visa versa I cat'ed out a sixel image in ms term and it displayed in kitty.  scrolling the tmux window works but that isn't quite what you were doing.

I will see if i can reproduce what you describe tomorrow.  Meanwhile, tell me if I got the quantization bug this time.  I feel we're getting closer.  Thanks for help testing, very helpful!

mgrant0:
@bptato I ran your shell script:
image
it didn't "move" the icon down, it waited a second then it printed a second icon.  Is that what you expected?

mgrant0:
@bptato Theicat test was useful for the new Kitty→tmux→SIXEL interoperability because we now deliberately intercept that wrapped placeholder form and convert it into shared tmux image cells.

I pushed another fix to hopefully fix or improve the scrolling in chawan, pleases let me know if it's any better or worse.

mgrant0:
After my last fix, your script indeed makes the little icon folder move down (I see only one icon now).

image

bptato:
> I will see if i can reproduce what you describe tomorrow. Meanwhile, tell me if I got the quantization bug this time.

It seems to be working correctly now, thanks.

> I pushed another fix to hopefully fix or improve the scrolling in chawan, pleases let me know if it's any better or worse.

Something is still off.  I can't reliably reproduce this, but following script:

```sh
#!/bin/sh
reset
tput cup 0 0
printf 'test\n'
tput cup $(($(tput lines) - 1)) 0
printf 'hi world\n'
printf '\033_GC=1,s=26,v=26,p=1,q=2,i=1,a=T,f=100,m=0;iVBORw0KGgoAAAANSUhEUgAAABoAAAAaCAYAAACpSkzOAAACmUlEQVR4XmOgF2CEWVRfX8/w//9/BkZGRgy7YeKNjY1kuwtsam1tLQMTE5MG0MBsoEWW6KYBxf8DxScD6UVNTU1kWcZYU1MD0qjLzMx8EmgYJy5TgJb8/ffvn+2vX7+Od3Z2kmwZC9AnIE0FIEuAhl0H4ulA9k9kk4Bi4UB1TkA8H2iRYXl5+Xd2dna8loGCG2Q2LLgZQXED1PEIaLjs379/g4Ds9VDLwQYBxRiAPpFmZWW9DFQjCDRgLxDfJeQloJpdQHPWA9X9A1kGs+gZ0BBJoKQtUOIIeqSDgheoKR6IF5ASZkAH3gKaCzLzFQsxGltaWhjq6uoWguIJ6hi82oCG8wIV+AEdpga0LBfIrmUh1oWg1AYM5iVA9UuI1JMGVDcTiCNIsghkOChIQ0NDwfasXr0ap33QeL8GUgD0HTeIZmIgAUAtYQZqYYZZSKx2oi2CWaKpqbkIhEm1jOg40tLSAjleEBgUUSAGkJ8PpN8Q6yOCFoF8AvQBTvOAqZHh+vXrDPjijGAcIcXJUqBPQOXda5iNQP5rkBiQv5SYYMTrIxUVFZC5IkDCCk8QWQHVgdS8xBeMeBNDe3s7A7CwfQl0uSmwKDIBYhdQGQbCIDZIDCQHUgNSS7ZFII2gUgEU6UADzwJz+kUgDa6zoOyzIDmoGgaKLAJpBheKkArxC7BIeQDCQPEvIEuJrQyJzkdQA38ADbcHYaBFP0ipcVGqCWDYBwPxOkYs1TkDCeA/pEkQBDRnLZD9GKhVjgUkCGTsBgomAekWIJYG4p8MlAFQrZgJMgJo/m4QzQJ1fT9QIBLI1gTiSRRaAm/gAM38DjSvH2QeuHECLW1xNk7IsRhoyXGgJVOBem+AExPMEHzNLTItYiAlVTJQCwAAJxkMkgbHo2UAAAAASUVORK5CYII=\033\\'
printf '\n'
sleep 1
printf '\033_GC=1,s=26,v=26,p=1,q=2,i=1,a=p;\033\\\n'
sleep 1
```

*sometimes* leaves the word "hi" under the icon, like:

hi

Keyword sometimes; in about 50% the cases it displays correctly.  Happens in xterm, wezterm, yafy (but not in kitty, so it's probably a sixel output issue.)

Edit: although thinking of it more, this is probably unrelated to the (well, another) problem I still see with `cha`, as the script depends on scroll.  I'll investigate...

> Theicat test was useful for the new Kitty→tmux→SIXEL interoperability because we now deliberately intercept that wrapped placeholder form and convert it into shared tmux image cells.

Well, you added that after my post :)
Anyway, what is your rationale for it?  Personally I would expect a passthrough sequence to, well, pass things through.
(Such a kludge will also break things that depend on full support for the protocol, which this patch doesn't do.)

mgrant0:
> Well, you added that after my post :)
> Anyway, what is your rationale for it? Personally I would expect a passthrough sequence to, well, pass things through.
> (Such a kludge will also break things that depend on full support for the protocol, which this patch doesn't do.)

Ok, i see what you mean, let me reverse that.  What do you think should happen on a simultaneously connected sixel compatible terminal (not kitty)?  Should we try to passively look at the passthrough image and if possible try to create a sixel shadow image of it for the sixel capable terminal (for example Microsoft Windows Terminal)?  Because otherwise, those other terminals will just see a gaping hole.  This won't support everything you could throw at kitty like animations but it would support the majority of the static images which seems useful.  Not the easiest thing to do but it should be possible.  What's your suggestion in this case?

bptato:
> Ok, i see what you mean, let me reverse that. What do you think should happen on a simultaneously connected sixel compatible terminal (not kitty)?

I'd say "nothing".  We shouldn't incentivize apps to stick to an obsolete hack by adding another hack on top of it.

The right way to fix this is to patch terminal libraries (that don't already do so) to query for kitty support *without* passthrough as well.  Then, if they get a regular response, they can just send the image as normal, and optionally fall back to passthrough for older tmux versions as long as that's relevant.

mgrant0:
ahh i see, I didn't realise this passthrough was an obsolete method.  Ok I have put it back to the passthrough behaviour you suggest now.

mgrant0:
I ran your script which prints hi behind the folder icon.  I see it fails for me about 1 in 5 times.  I just pushed a possible fix, let me know if that solves it.  

What else is currently broken, is this usable yet?  What else is needed?

bptato:
>  I just pushed a possible fix

That seems wrong, Sixel (or Kitty) isn't supposed to clear text beneath itself.  (Try printing a partially transparent image on top of text to see what I mean.)
The point is that text appears on scroll that shouldn't have been there in the first place; further, the instability implies there must be a race somewhere (or just some evil invalidation bug?)

> What else is currently broken, is this usable yet? What else is needed?

1. The above scroll bug
2. As I said before, Sixel should dither alpha too so translucent pixels aren't just cut off at alpha<128; in my encoder I use floyd-steinberg for it where "empty pixel"=alpha:0 and "filled pixel"=alpha:255, so for e.g. [this](https://upload.wikimedia.org/wikipedia/commons/thumb/7/77/Hue_alpha_falloff.svg/960px-Hue_alpha_falloff.svg.png) it should produce something like [this](https://github.com/tmux/tmux/pull/4149#issuecomment-2391049699)
3. Maybe there's another scroll bug too, not sure, will try to reduce that too if the current one is properly fixed and that still persists

Well, at least that's it for the subset my program needs (unless I find another bug).

mgrant0:
I added a patch Nic gave me to fix an underlying bookkeeping bug which hopefully will fix the scroll bug with aparent race condition.  I have now, hopefully, fixed the alpha bug.  Let me know if I missed anything!  Thanks!

bptato:
> I added a patch Nic gave me to fix an underlying bookkeeping bug

That was it, thanks.  Alpha dither seems to work well too.

I've tested it again, two major bugs left:

1. Cropping a Kitty image at the top seems to change its width too for some reason.  See [script](https://github.com/user-attachments/files/30830189/test.sh).  (The image is large, sorry.)

2. Kitty image & text mixing semantics are still incorrect, e.g.,

   ```sh
   #!/bin/sh
   reset
   printf 'test\r'
   sleep 1
   printf '\033_GC=1,s=26,v=26,p=1,q=2,i=1,a=T,f=100,m=0;iVBORw0KGgoAAAANSUhEUgAAABoAAAAaCAYAAACpSkzOAAACmUlEQVR4XmOgF2CEWVRfX8/w//9/BkZGRgy7YeKNjY1kuwtsam1tLQMTE5MG0MBsoEWW6KYBxf8DxScD6UVNTU1kWcZYU1MD0qjLzMx8EmgYJy5TgJb8/ffvn+2vX7+Od3Z2kmwZC9AnIE0FIEuAhl0H4ulA9k9kk4Bi4UB1TkA8H2iRYXl5+Xd2dna8loGCG2Q2LLgZQXED1PEIaLjs379/g4Ds9VDLwQYBxRiAPpFmZWW9DFQjCDRgLxDfJeQloJpdQHPWA9X9A1kGs+gZ0BBJoKQtUOIIeqSDgheoKR6IF5ASZkAH3gKaCzLzFQsxGltaWhjq6uoWguIJ6hi82oCG8wIV+AEdpga0LBfIrmUh1oWg1AYM5iVA9UuI1JMGVDcTiCNIsghkOChIQ0NDwfasXr0ap33QeL8GUgD0HTeIZmIgAUAtYQZqYYZZSKx2oi2CWaKpqbkIhEm1jOg40tLSAjleEBgUUSAGkJ8PpN8Q6yOCFoF8AvQBTvOAqZHh+vXrDPjijGAcIcXJUqBPQOXda5iNQP5rkBiQv5SYYMTrIxUVFZC5IkDCCk8QWQHVgdS8xBeMeBNDe3s7A7CwfQl0uSmwKDIBYhdQGQbCIDZIDCQHUgNSS7ZFII2gUgEU6UADzwJz+kUgDa6zoOyzIDmoGgaKLAJpBheKkArxC7BIeQDCQPEvIEuJrQyJzkdQA38ADbcHYaBFP0ipcVGqCWDYBwPxOkYs1TkDCeA/pEkQBDRnLZD9GKhVjgUkCGTsBgomAekWIJYG4p8MlAFQrZgJMgJo/m4QzQJ1fT9QIBLI1gTiSRRaAm/gAM38DjSvH2QeuHECLW1xNk7IsRhoyXGgJVOBem+AExPMEHzNLTItYiAlVTJQCwAAJxkMkgbHo2UAAAAASUVORK5CYII=\033\\'
   sleep 1
   printf '\r'
   sleep 1
   printf 'test\n'
   sleep 1
   printf 'test\n'
   ```

   On Kitty, this a) prints "test", then b) the icon above the text, then c) nothing changes, d) nothing changes, and finally e) prints "test" in the *under* the image.  In the end, you get "test (newline) test" and an image above (both remain intact).

   Meanwhile, tmux a) prints "test", then b) destroys part of the text without printing the image (?), then c) *sometimes* it prints the image too (??), then d) prints "test" again destroying the upper half of the image, and finally e) prints "test" on the next line, destroying the lower half.

   Notably, b) seems very strange (maybe the image isn't flushed?), c) is non-deterministic, and all steps except for a) differ from Kitty (breaking invalidation logic in TUIs).

mgrant0:
Nic provided a patch for #1.  I ran your script in both ms terminal and kitty:
image
Notice the difference.  But at least it's consistent, i don't see it sometimes happening in either.  Which one is expected?

I've fixed #2, the image in kitty stays the same now.  Try this modified version of your script:
[bptato-test-3.sh](https://github.com/user-attachments/files/30834978/bptato-test-3.sh)
it progressively crops the image.  You can see as it gets shorter, it stays the same width.  I ran this in tmux in both ms term and kitty term.

mgrant0:
ok i fixed the descrepency, now both kitty and sixel look like this when your script is run:
image

bptato:
> Which one is expected?

Of course, it would be what Kitty produces *without* tmux:

expected

(which is also what I attempted to describe, unsuccessfully it seems...)

Please refer to the [page](https://nick-black.com/dankwiki/index.php?title=Theory_and_Practice_of_Sprixels#Transparency_and_z-ordering) I linked above describing the semantics of both protocols; it's the best resource on this topic that I'm aware of.

> I've fixed 2, the image in kitty stays the same now. Try this modified version of your script:

~~Sorry, I forgot to mention, the bug only appears in Kitty (and it still persists).
It seem it's somehow related to the cell size; for a better demonstration, try [this version](https://github.com/user-attachments/files/30836310/test.sh).~~ Edit: never mind, sorry, it's fixed now, I just forgot to restart the tmux server.

mgrant0:
ok i feel like an idiot, yes, of course, what kitty produces outside tmux.  I've corrected that now, please check.

mgrant0:
@nicm

> * A lot of functions are missing header comments. They should all have them.

Added.

> * `sixel_from_image` is clearly a disaster and needs to be tidied up. Could probably do with some comments.

done.

> * Some of the naming could be better `image_rect` would be much neater than `image_rectangle`. Maybe `sixel_hgram` not `sixel_histogram`.

Ok shortened these up, also dimensions renamed to size.  histogram to hgram in the struct, and hg in var, but left it histogram in the flags.

> * Instead of `image_tty_is_graphical` and `image_tty_scrolls` I would move the flags into `tmux.h` and add `image_backend_flags`.

Exposed image backend_flags now.

> * `image_cmp`/`RB_GENERATE_STATIC` should be at the head of the file.

relocated.

> * There is some crazy casting going on here eg in `image_size_in_cells`.

u_int64 was necessary in case of overflow which is highly unlikely.  removed cast all together now, using  width / xpixel + (width % xpixel != 0).

> * SIXEL should definitely not imply `Sync`.

 SIXEL and synchronized updates must now be enabled independently.
 
> * `tty_draw_line` should not know so much about images. The `ENABLE_IMAGES` bit in `tty-draw.c` should call a image helper to get the cell.

image cell drawing now moved into image.c

bptato:
OK, now Kitty seems to work right, but Sixel broke.  (It's why I posted that link, but anyway...)
The semantics differ for the two.  With Kitty (at least the original protocol - not sure about Unicode placeholders...), the images are a separate layer from text, and neither affects the other.
Meanwhile with Sixel, printing any text on top of an image erases it from the cells affected.  Note that the inverse is *not* true, i.e., printing images on top of text does not remove the text:

```sh
#!/bin/sh
reset
printf '\033P0;1q"1;1;26;26#0;2;50;50;50#1;2;0;0;0#2;2;49;49;49#3;2;33;33;33#4;2;47;47;47#5;2;44;44;44#6;2;48;48;48#2?O?C!5?S!4?__?_?__$#0?_wW!5KGwo_-#0?~~!8?!10@BB}w$#2!12?a_A?A?A?A??@E-#0?~~!7?KK~~KK!7?~~$#2!9?K!6?K-#0?F^[owowowowowowowowooS^F$#2?G!10?@@!7?G??G-\033\'
tput cup 0 0
printf 'this line damages the first image\n'
printf '\n\n'
printf 'but on this line, the image will appear on top the text\r'
printf '\033P0;1q"1;1;26;26#0;2;50;50;50#1;2;0;0;0#2;2;49;49;49#3;2;33;33;33#4;2;47;47;47#5;2;44;44;44#6;2;48;48;48#2?O?C!5?S!4?__?_?__$#0?_wW!5KGwo_-#0?~~!8?!10@BB}w$#2!12?a_A?A?A?A??@E-#0?~~!7?KK~~KK!7?~~$#2!9?K!6?K-#0?F^[owowowowowowowowooS^F$#2?G!10?@@!7?G??G-\033\'
printf '\n\n'
```

mgrant0:
Here is what I see:
image
If you look closely you can see that the folder icon is transparent but the edge looks like it is acting as you describe in both cases.  What am I missing?

mgrant0:
wait, i think i see what you mean, the second line is right, the first line is wrong.

mgrant0:
Try the latest push please.

mgrant0:
@bptato I was thinking about this last night, the current patch is not perfect, i may be able to do better.  The issue is that tmux stores what the tty looks like internally so that when another client connects it can redraw it.  tmux does not store the history of a cell (the temporal state of the cell).  Storing a complete history of a cell (back to when it was empty) does not seem worth it.  The current implementation does not store the state of overlapping sixel images.  For example, if you try to composit 2 sixel images on top of one another, tmux would draw it, but if you reconnect or redraw the screen, the lower image would go away because it's no longer in the cell.  The cell tracks which image (which part of which image) goes there, not a bit image of the cell.  

So what can I do?  I can leave it as it is now which is once a cell has been overwritten by a character, any bit image in that cell is gone.  This is definitely different than outside of tmux on ms term and probably other terminals.  Or I can track a single image layer that basically tracks if there were bits written there before, don't clear first and just drop the text on top of it.  That will make your test look correct but it won't work for more complex things like image1...image2...text where image2 has some transparent parts or does not fully cover image1 and would get composited on image1.

Since we already store a single image ref and character for each cell, I am willing to store an order flag in the cell:

  - flag clear: the retained image was drawn after the cell’s text, so redraw text then image;
  - flag set: the text was drawn after the retained image, so redraw image then text.

And if multiple images overlap, only the most recent image reference is retained for that cell. Any earlier image underneath is no longer represented there.  This I think is quite easy to implement.

I realise this isn't perfect but it seems like trying to preserve arbitrary z-ordering compositing of multiple images is going to add a ton of complexity for something which will never be used.  What do you think?  

(edit) I have prepared a version that does the above.  It wasn't super easy and it affected a bunch of files outside image-sixel.c.  Please let me know how important is it to preserve this semantic even partially?   My preference is the current behaviour because it is simple, but it is clearly not identical to other terminals. 

(edit 2) when there is both an image ref and a character in the cell, it was always drawing them both but first the character then the image.  with what i just did, it preserves an order and more specifically does not clear the cell before ploping a character in it so that any remnant there remains there on purpose.  It also fixed another bug.  I am going to push this now, after some thinking, this really is the better of the two worlds, but it only preserves an image layer and text layer and which one was written first, not n-layers.  if preserving n-layers is really important, let me know.

bptato:
> The issue is that tmux stores what the tty looks like internally so that when another client connects it can redraw it. tmux does not store the history of a cell (the temporal state of the cell). Storing a complete history of a cell (back to when it was empty) does not seem worth it.

You don't have to store cell history.  You "simply" have to store the images in the order they were received.  (I know it's not simple...)

> The current implementation does not store the state of overlapping sixel images. For example, if you try to composit 2 sixel images on top of one another, tmux would draw it, but if you reconnect or redraw the screen, the lower image would go away because it's no longer in the cell. The cell tracks which image (which part of which image) goes there, not a bit image of the cell.

Yeah, at this point I'm skeptical of any approach that directly associates images with cells.  At the very least, that association must be an ordered list, and that gets expensive (or complex) very fast.

(Of course, it's a tradeoff, storing images in a separate list makes Sixel invalidation complex.  Really the core issue is that Sixel was designed to be dumped onto the terminal's framebuffer, while Kitty intentionally decouples images from text, so combining them into a single model is difficult whatever you do.)

> So what can I do? I can leave it as it is now which is once a cell has been overwritten by a character, any bit image in that cell is gone. This is definitely different than outside of tmux on ms term and probably other terminals. Or I can track a single image layer that basically tracks if there were bits written there before, don't clear first and just drop the text on top of it. That will make your test look correct but it won't work for more complex things like image1...image2...text where image2 has some transparent parts or does not fully cover image1 and would get composited on image1.
>
> Since we already store a single image ref and character for each cell, I am willing to store an order flag in the cell:
>
> * flag clear: the retained image was drawn after the cell’s text, so redraw text then image;
> * flag set: the text was drawn after the retained image, so redraw image then text.
>
> And if multiple images overlap, only the most recent image reference is retained for that cell. Any earlier image underneath is no longer represented there. This I think is quite easy to implement.
>
> I realise this isn't perfect but it seems like trying to preserve arbitrary z-ordering compositing of multiple images is going to add a ton of complexity for something which will never be used. What do you think?

Well, it's broken, and I have no idea how the 'z' Kitty parameter can ever be implemented under these constraints (AFAICT right now it isn't?)
Although it might work in practice in most cases, but as you mention that's only because the internal state does not match what the terminal actually sees.  So I'm definitely not a fan.

> The implementation now distinguishes a newly placed image cell from one later overwritten by text. On a SIXEL terminal, damaged cells are omitted from image output; on Kitty they remain ordinary positional placements.

This makes no sense.  Image/text mixing semantics are decided by the input format, not the output.  The app sending the image has no idea about the outer terminal's output format.

mgrant0:
Perhaps cell history is not the proper way to describe it.  Do I need to store each layer is really what I meant.  By  history, I mean that if you don't specify a z-index, the next image is above the previous one and if there are transparent bits you see through the one on top, down to older ones or ones lower in the z index.  I can definitely do this and my question to you was is this necessary? and it sounds like from your throwing up your arms in disgust the answer sounds like yes.  I was hoping to avoid doing this but if i have to i will.

Your test cases have been super helpful.  If you could prepare one or more that exercise the zindex more that would be helpful and I hope to get closer.  I don't know how close I can get to working for you but hopefully I can get close enough.

> This makes no sense. Image/text mixing semantics are decided by the input format, not the output. The app sending the image has no idea about the outer terminal's output format.

Is this a comment again about the z-index?  I thought by implementing this single pane I had more or less fixed this but it looks like I am going to have to go all the way and implement multiple and it's not the output dictating it, it's the input and z-index.  Since sixel images do not support any sort of z-index placement, they are temporal, as in, the next one is on top of the previous.  I have that info already, it's just a matter of wiring that to the cells.  That should also make kitty z-index work.  This all doesn't feel impossible, just a bit fiddly.  Maybe I'm deluded?

bptato:
> Do I need to store each layer is really what I meant.

What do you want to achieve?

1. If you just want to dump images to the CLI and marvel at the result, the patch is fine as it is.
2. If you want to implement Kitty *correctly* so that, for example, TUI apps that use it work well, then yes, to my knowledge, this is what the protocol requires.

As for me, my main priority is, selfishly, that *my* TUI app does not work much worse after this patch than before.  Admittedly, this doesn't affect many people, but it's a weaker version of 2, so I thought it'd be a useful direction for now.
In any case, I don't know if what you're *proposing* would cause significant problems in practice (again, for me...), because the thing you've *implemented* is broken in a different way:

> Is this a comment again about the z-index?

No, it is a comment about how, according to the commit message, text & image interaction now depends on the outer terminal, and from a quick test, the commit message appears to be accurate.  I'm saying this makes no sense, because the app that sent the image does not have information about the outer terminal.
For example, in the scripts I posted above, the result ought to be the same no matter if the outer terminal supports Sixel or Kitty.

Hint: I think the image struct should have a "received from Sixel" flag that affects what happens after some text is printed over it.

Edit: re-reading the conversation, maybe I wasn't clear enough about this: Sixel cannot be displayed *behind* text.  If you print any text over a Sixel image, then the image disappears from the cells where you printed the text.  So any logic of whether the image is "behind" the text is redundant; if it's behind the text, then it's already gone.
(I think it's because the original DEC terminals just printed the image to the framebuffer directly, which would preserve text in the background.  But when you printed text over images, the (presumably) generic text drawing routine would clear the cell and *then* draw the text.)

> Since sixel images do not support any sort of z-index placement, they are temporal, as in, the next one is on top of the previous. I have that info already, it's just a matter of wiring that to the cells.

...by creating an array of images, then assigning an id to that array, and linking the array to the cell?
I mean, yes, it can work, I'm just not entirely convinced this is any simpler than handling images separately from the grid :P

mgrant0:
z-index hopefully now works.  Thanks, your hint was right. The grid now retains every image placement as a separate ordered layer, represented sparsely as spans on the affected grid lines rather than as an array per cell. Each placement records its input protocol, IDs, z-index, and order.

Previously, an image cell contained only one image reference. If image B overlapped image A, the cell could remember only B; tmux lost the fact that A should still be drawn underneath through B’s transparent pixels.

Now the grid does not store an image list inside every cell. Instead:

- An image placement is one logical object: “this image is placed here,” with its Kitty/SIXEL origin, image and placement IDs, Kitty z-index, and creation order.
- For every affected grid row, that placement has a horizontal span—for example, columns 20–39 on row 8.
- Those spans attach to the grid lines the image crosses. A 10×5-cell image therefore has five spans, not 50 per-cell image records.
- Multiple placements may have spans over the same cells, so the full stack is retained.

On redraw, tmux finds the spans touching the area being rendered and orders them:

1. Kitty images with negative z-index: behind text.
2. SIXEL images: temporal framebuffer-style layer.
3. Kitty images with zero or positive z-index: above text.

Within compatible layers, creation order breaks ties. This lets transparent overlapping images retain the layers below them, while avoiding an expensive per-cell linked list or array.

Text writes now damage only SIXEL-origin spans. Kitty-origin placements retain their z semantics, including overlapping placements; the outer client’s Kitty/SIXEL capability no longer decides the logical image/text ordering.  Hoping that makes sense now.

The unavoidable output limitation is that SIXEL cannot draw an image beneath a text glyph. When rendering the same logical scene to SIXEL, tmux preserves text in those cells rather than emitting the underlying image portion.

I installed Chawan to test tmux image scrolling and found that I needed -o buffer.images=true before inline images were fetched. Is keeping this opt-in intentional? Since it controls downloading images even when they cannot be displayed, I can see the bandwidth/privacy rationale; but would enabling it by default, or making it more discoverable, better match the expected browser experience?

I tested the normal, non-passthrough path with Chawan on the WebP gallery, including scrolling, resizing, and Kitty/SIXEL clients attached to the same server. Could you please retry the original gallery-scrolling case?

One separate Microsoft Terminal observation: its large SIXEL redraws need the independently advertised sync terminal feature to avoid visible intermediate repainting. This remains separate from SIXEL support; SIXEL does not imply Sync.

Here is what I did in Windows Terminal which inhibits the flashing you might have seen:
```
./tmux -Lchawan-ms start-server \; set -as terminal-features ',xterm-256color:sync' \; new
```
then in tmux:
```
cha -o buffer.images=true https://developers.google.com/speed/webp/gallery
```
It scrolls smoothly directly in Windows Terminal and Kitty (outside of tmux).  I also tested this in Wezterm.  Albeit with this single web gallery page.  How close are we to getting this working?  Please do let me know if you find some issues.

bptato:
> z-index hopefully now works. Thanks, your hint was right. The grid now retains every image placement as a separate ordered layer, represented sparsely as spans on the affected grid lines rather than as an array per cell. Each placement records its input protocol, IDs, z-index, and order.
>
> Now the grid does not store an image list inside every cell. Instead:
>
> * An image placement is one logical object: “this image is placed here,” with its Kitty/SIXEL origin, image and placement IDs, Kitty z-index, and creation order.
> * For every affected grid row, that placement has a horizontal span—for example, columns 20–39 on row 8.
> * Those spans attach to the grid lines the image crosses. A 10×5-cell image therefore has five spans, not 50 per-cell image records.
> * Multiple placements may have spans over the same cells, so the full stack is retained.
>
> On redraw, tmux finds the spans touching the area being rendered and orders them:
>
> * Kitty images with negative z-index: behind text.
> * SIXEL images: temporal framebuffer-style layer.
> * Kitty images with zero or positive z-index: above text.
>
> Within compatible layers, creation order breaks ties. This lets transparent overlapping images retain the layers below them, while avoiding an expensive per-cell linked list or array.

That sounds like a correct solution, thanks.

> I installed Chawan to test tmux image scrolling and found that I needed -o buffer.images=true before inline images were fetched. Is keeping this opt-in intentional? Since it controls downloading images even when they cannot be displayed, I can see the bandwidth/privacy rationale; but would enabling it by default, or making it more discoverable, better match the expected browser experience?

Yes, it's horrible UX.  I wish it was enabled by default too, but I couldn't do that for a multitude of reasons - some still relevant, others less so.
There are other similar discoverability issues too; eventually I hope to alleviate these by adding a setup screen to the first launch.
(Do note that you don't have to pass the flag every time, you can also set it to true in the config.)

> I tested the normal, non-passthrough path with Chawan on the WebP gallery, including scrolling, resizing, and Kitty/SIXEL clients attached to the same server. Could you please retry the original gallery-scrolling case?

Indeed, it seems to be working correctly now.  Thank you!

> One separate Microsoft Terminal observation: its large SIXEL redraws need the independently advertised sync terminal feature to avoid visible intermediate repainting. This remains separate from SIXEL support; SIXEL does not imply Sync.

I also see it with xterm (which has no Sync), it's what I alluded to with the "performance characteristics" comment before.
But I think some jank is acceptable for now, as long as the semantics are correct.

mgrant0:
Oh great, glad it works finally!

mgrant0:
@bptato some major changes in the 4902-image-support branch.  Would you mind retesting your application and let me know if everything still works?  Basically this branch is based on the redraw-damage-rectangles branch which tracks damage to the screen and redraws only the necessary spans.  It should substantially reduce the amount of flashing of images.

bptato:
It seems to work fine, thank you.

(Although if the goal was to eliminate flashing, that still happens with Sixel, except when Sync hides it - in xterm it's visible even when I just dump an image to the shell with `img2sixel` and scroll by repeatedly hitting return.)

paranoidi:
How can I compile this correctly, seems like  --with-image-support no longer works:

`"configure: WARNING: unrecognized options: --with-image-support"`

Currently when I test this PR sixel graphics do not show up at all.

Based on logs the `--enable-sixel` is deprecated as well.

mgrant0:
Use --enable-images.  Option changed for consistency.

mgrant0:
@bptato without sync, i can't seem to get it to completely eliminate such flashing.  Depending on the terminal, one can't depend on the terminal actually scrolling a sixel image so it needs to be redrawn and I bet this is the cause of the flash you are seeing.  Unless you know of some way to know if a terminal actually moves pixels around that are associated with characters in the grid.  However, with kitty images, it's different.  I added something hopefully to help the flash when displaying images in kitty, you can let me know if that helps at all.  It may not be possible to fix all the flashing.

j4james:
> Depending on the terminal, one can't depend on the terminal actually scrolling a sixel image

Can you give a specific example of a terminal that _doesn't_ scroll sixel images? Because I've tested quite a lot of terminals over the years, and I don't recall that ever being an issue.

I know there was a stage in the past where some terminals had implemented sixel display mode (`DECSDM`) backwards, so if you were toggling that mode, it might end up have the opposite effect to what you expected (i.e. turning off scrolling when you thought you were turning it on). But I can't believe that's still a problem today. And even if it is, you should still be fine as long as you leave it with the default setting.

bptato:
Yeah the only terminal I know that fails to scroll Sixels is tmux (prior to this PR at least).
It gets messy with translation between the protocols, but if you're just scrolling a Sixel image by hitting return in the shell, that should be possible to replicate - presumably tmux already scrolls to avoid repainting the whole screen every frame, so the invalidation algorithm "just" has to take that into account.

(As for Kitty, I hadn't tested that, but I also did not even consider it would flash - why would it, moving images is trivial there, it's baked into the protocol.)

paranoidi:
@mgrant0 thanks. It seems to enable the images. However there are some issues. 
- Sometimes the image fails to render
- Images may disappear after first render
- The performance is significantly slower than native terminal emulators. In fact I think current master may be somewhat more performant (but obviously bugged).

I have analyzed byte output via LLM developed tooling and the issue seemed to be within tmux. It was able to propose a patch that seems to resolve these issues completely.

You may want to check out the commits on my fork and pick up from there.

https://github.com/paranoidi/tmux/commit/98cc26d83eda622b507643ba595fc415f2b2f3aa

https://github.com/paranoidi/tmux/commit/55da37f9c3482deca2323dfeaa71f5e42f23760d

I do not have enough experience / time to dive into this, but hopefully the testing and LLM based fixes provides at least some value instead of being unwanted noise.

mgrant0:
@paranoidi thank you, i have tried to merge in this or use this to help fix the issue you reported.  Can you please test it to see if I got your case, I can't use codex at the moment so i can't check it.  I'm still working on this, I have some other fixes that I hope will fix some of the other flashing issues reported above.

mgrant0:
@j4james @bptato I made some progress on the scrolling.  Please can you check if this fixes that issue?  Hopefully it should be smooth now.

j4james:
Testing commit 7ff30a2 in Windows Terminal I see the images being deleted when the page scrolls. Although this only applies to certain operations. Scrolling up with newline and scrolling down with `RI` will delete the image. Simulating the scrolling with `DL` and `IL` seems to leave the images intact. Technically all operations are performing a complete repaint of the page - it's just that some of them don't repaint the images when they do that.

mgrant0:
@j4james I tested a few different terminals now, the only one I saw working was Wezterm.  

@bptato, the issue isn't simple scrolling images using a linefeed in terminal and pushing an image upwards. tmux doesn't do that because unless you turn it off, there's a status bar at the bottom of the screen.  It uses region scrolling.  I could special case it to use a LF to scroll the terminal if no status bar, no scrollbar, no splits, no pane-border...etc, but it just doesn't seem worth it.

The issue is that it's unpredictable if a terminal can actually scroll a region with an image in it.  It looks like in the handfull of terminals I tested (including Windows Terminal), when you scroll a region with a sixel, it just disappears.  Kitty just smudges it.

There is currently a terminal feature which is on by default, you can run tmux like this to disable it:

```
tmux set -g image-region-scrolling off
```

I am inclined to remove this and go back to always redrawing the image.  I could default this to off, but it seems like it's more rare that scrolling images in scrolling regions is actually supported.  It seems like it's not worth the added complexity in tmux when this doesn't work generally.

I have another idea I am working on, might take me a day or so.  I feel there aught to be a way we can minimise flickering here but there will still have to be some redrawing.  It just weems we are doing a bit too much drawing.

nicm:
If it stays it needs to be a capability not an option, but if there is a method that will work on all terminals, even if it is suboptimal on some, better to just do that.

j4james:
> It looks like in the handfull of terminals I tested (including Windows Terminal), when you scroll a region with a sixel, it just disappears.

@mgrant0 I can assure you that sixel scrolling within margins is supported in Windows Terminal. Here's a simple test case (this obviously needs to be run _outside_ tmux):
```sh
echo -e '\e[?69h\e[6;19r\e[21;60s\e[6;21H'; cat snake.six
```

But I can understand if you don't want to deal with the hassle of different code paths for different terminals with different capabilities. I just got the impression that your latest revision was supposed to have made this work. And from what I could see of your output (Windows Terminal has a debug monitor that shows exactly what escape sequences you're sending) you weren't even trying to scroll the window - you were just redrawing the whole view but without the image content, and that's why images were "disappearing".

mgrant0:
@j4james wow that really was a good obsservation.  I spent hours working on this, went down several dead ends, discovered a few things.  Please try the latest push on windows terminal but I want you to run tmux like this:

```
./tmux -v -L wt-test set -as terminal-features ',*:imagescroll'\; new
```
or add the `set -as` line to your .tmux.conf.  I moved imagescroll to a terminal feature now.  Unfortunately I couldn't find an acceptable way to identify Windows Terminal to be able to automatically add this to a tty-features table, so you will need to add this manually yourself and then tmux should use scrolling regions with images without flashing.  Without this, it redraws on each scroll flashing a bit.

Can you tell me if this removes the signicant flashing you have been seeing in this branch?  Are there other cases I need to work on?

Currently working on a bug in the kitty code now, hopefully fixed shortly.

j4james:
> Please try the latest push on windows terminal but I want you to run tmux like this:

I've tested commit 386799e, both with your suggested command line and with no command line. In both cases I could scroll without losing images, but I could see the `imagescroll` version was more efficient, since it wasn't redrawing the images on every scroll. It was still redrawing the text, though, which seemed weird to me. Like it trusts the terminal to scroll images reliably, but isn't sure it can handle scrolling text!? Still, that's not a huge concern.

What did worry me was the way it was accumulating images. If you output two sixel images on top of each other like this:
```sh
echo -e '\e7\eP9;1q#1!100~\e\\\e8\eP9;1q#2!100~\e\\'
```

You've got a blue bar that's been completely replaced by a red bar. Those blue pixels are never going to be seen ever again. But you continue to render _both_ of those images whenever the screen is repainted. The most painful example of this is when you're watching a sixel video or animation, and it just accumulates hundreds of layers on top of each other.

Sixel isn't like KGP, where you can have multiple image layers that are independent of each other. Semantically it's a single bitmap onto which you can plot pixels. When you plot a red pixel on top of a blue pixel, that blue pixel is gone for good. So when sixel sequences overlap each other, you should really be merging the most recent one into the earlier one. There's no need to keep them separate.

> I moved imagescroll to a terminal feature now. Unfortunately I couldn't find an acceptable way to identify Windows Terminal

That's fine. You shouldn't be whitelisting features based on terminal names anyway. It incentivizes the wrong behaviour. But that's an argument for another day. As long as it still works both with and without the feature, I don't really care.

Issue #4897, opened 2026-02-28
menu-border-lines none doesn't work


joshua-cooper:
Setting `menu-border-lines none` seems to have the same behavior as `menu-border-lines padded`. Reading the docs, I'd expect setting it to `none` would remove the borders entirely, like it does with `popup-border-lines`.

```
> uname -sp && ./tmux -V && echo $TERM
Linux unknown
tmux next-3.7
xterm-ghostty
```

nicm:
Yeah it just draws them as empty, it doesn't remove them. It would need additional changes to `menu.c` to adjust the menu size, like it does in `popup.c`.

nicm:
Added to [todo list](https://github.com/tmux/tmux/wiki/Contributing#todo-list).

PR #5447, opened 2026-07-31
Remove padding from borderless menus

Mr-Neutr0n:
## Summary

- size and position menus without frame rows or columns when `menu-border-lines` is `none`
- keep hidden titles from reserving width in borderless menus
- use the same geometry for rendering, cursor placement, resize, and mouse hit testing

Addresses #4897.

## Testing

- `make -j4`
- focused `screen-redraw-menus.sh` no-border regression
- `regress/menu-mouse.sh`
- full regression run: related tests pass; three unrelated existing macOS/timing failures remain in `copy-mode-selection-scroll.sh`, `format-mouse.sh`, and the `menu-over-split` golden

Prepared with OpenAI Codex assistance; I reviewed and tested the changes.

### DIFF
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -482,7 +482,7 @@ cmd_display_menu_exec(struct cmd *self, struct cmdq_item *item)
 	enum box_lines		 lines = BOX_LINES_DEFAULT;
 	char			*title, *cause = NULL;
 	int			 flags = 0, starting_choice = 0;
-	u_int			 px, py, i, count = args_count(args);
+	u_int			 px, py, sx, sy, i, count = args_count(args);
 	struct options		*o = target->s->curw->window->options;
 	struct options_entry	*oe;
 
@@ -531,9 +531,6 @@ cmd_display_menu_exec(struct cmd *self, struct cmdq_item *item)
 	}
 	if (menu->count == 0)
 		goto out;
-	if (!cmd_display_menu_get_menu_pos(tc, item, args, &px, &py,
-	    menu->width + 4, menu->count + 2))
-		goto out;
 
 	value = args_get(args, 'b');
 	if (value != NULL) {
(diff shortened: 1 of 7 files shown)


nicm:
Looks good but try this:
~~~~
tmux new
tmux set menu-border-lines none
tmux neww
tmux display-menu -t%0 a b c
tmux selectw -t%0
~~~~
The menu has lines even though that window is configured with `none`.

Mr-Neutr0n:
Thanks, I reproduced the exact sequence. The issue was that display-menu read menu-border-lines from the current window after neww instead of the explicitly targeted window (%0), so it saw the default border setting. I updated it to use the target window options, added a regression case for this sequence, and verified the rebuilt binary renders the borderless menu. Commit 2776e2ea is pushed for recheck.

Mr-Neutr0n:
Rebased and resolved the conflicts.

nicm:
I think this looks good now, thanks. This will be for after 3.8 and I will look at it again after that is released.

nicm:

Mr-Neutr0n:
Rebased onto master `d215280b` - clean, no conflicts (`bcfed825`).

Verified on the rebased head:
- `make -j2` builds with 0 warnings
- `regress/screen-redraw-menus.sh` and `regress/menu-mouse.sh` pass
- Full `regress`: only 4 failures (`floating-pane-geometry`, `options-values`, `prompt-words-history`, `screen-redraw-floating`); all reproduce byte-identically on a pristine master build, so they're from the new rounded pane-border-lines tests rather than this change

nicm:
I have applied this to OpenBSD now with some trivial changes, will be in GitHub later. Thanks!

Mr-Neutr0n:
Thanks for applying it, and for the careful review along the way - much appreciated!

Issue #4127, opened 2024-09-19
Feature request: add support to querying the cursor color (OSC12)


aymanbagabas:
### Issue description

It would be nice to add support for querying the cursor color. As of Tmux 3.4, users can query the fg/bg colors using OSC10/OSC11, but not OSC12. The [Xterm control sequences specs](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) also uses `OSC 12 ?` to query the cursor color.

Here's a picture from Ghostty with the cursor color changed

<img width="358" alt="image" src="https://github.com/user-attachments/assets/9e1eaee7-0d4a-4e5c-bbc1-44936127ef60">

### Required information

Please provide the following information:

* tmux version (`tmux -V`). `tmux 3.4`
* Platform (`uname -sp`). `Darwin arm`
* $TERM inside and outside of tmux (`echo $TERM`). `tmux-256color`
* Logs from tmux (`tmux kill-server; tmux -vv new`). `N\A`

nicm:
OSC 12 already works if you change the cursor colour in tmux.

aymanbagabas:
> OSC 12 already works if you change the cursor colour in tmux.

Ah, I see. I updated the title and description accordingly 🙂

nicm:
What is the use case for this? Do applications use it and what for?

aymanbagabas:
For example, Neovim supports setting the cursor color advertised by the `Cs` terminfo capability. I would like to use different colors for different Vim modes. For example, green during normal mode, blue in insert mode, etc. Right now, you could use Tmux passthrough DCS sequence to change the color _globally_, and would be neat to make this work on a pane to pane basis.

See this comment https://github.com/neovim/neovim/blob/ff7832ad3fce55671ac4032716164ada0350b0ec/src/nvim/tui/tui.c#L2348

nicm:
I'll add this to the todo list but I'm not sure I would expect it soon.

nicm:
If anyone wants to add it, it is a matter of duplicating how OSC 10 and 11 work - look at `tty_send_requests` and `tty_keys_colours`.

PR #5448, opened 2026-07-31
Query the outer terminal cursor colour

Mr-Neutr0n:
## Summary

- request the outer terminal cursor colour with OSC 12 alongside the existing OSC 10 and 11 queries
- store the reported colour separately from cursor colours applied by tmux
- answer an inner OSC 12 query with the outer colour when no explicit tmux cursor colour is configured

Addresses #4127.

## Testing

- `make -j4`
- `regress/input-requests.sh` (new outer-terminal to inner-application round-trip)
- `regress/input-osc.sh`
- `regress/theme-report.sh`
- `regress/format-variables.sh`
- full regression run: all related tests pass; the same three unrelated existing macOS/timing failures remain in `copy-mode-selection-scroll.sh`, `format-mouse.sh`, and the `menu-over-split` golden

Prepared with OpenAI Codex assistance; I reviewed and tested the changes.

### DIFF
--- a/input.c
+++ b/input.c
@@ -3163,6 +3163,8 @@ static void
 input_osc_12(struct input_ctx *ictx, const char *p)
 {
 	struct window_pane	*wp = ictx->wp;
+	struct window		*w;
+	struct client		*loop;
 	int			 c;
 
 	if (strcmp(p, "?") == 0) {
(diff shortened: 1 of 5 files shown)


nicm:
This looks OK but what is the use case... what applications use this?

Mr-Neutr0n:
Addressed the review: the concrete consumer I found is Bubble Tea v2. It exposes RequestCursorColor, sends OSC 12, and returns CursorColorMsg; its v2 example queries foreground, background, and cursor colours together before changing them. Through tmux that same program gets OSC 10/11 replies but not OSC 12. I did not find a named end-user application that currently depends on it, so this is framework parity/theme preservation rather than a known app breakage. Happy to close if that is not sufficient. https://github.com/charmbracelet/bubbletea/discussions/1374

nicm:
I think let's leave it until someone who uses it comes along. Thanks.

Issue #4910, opened 2026-03-04
Feature request: vertical window list as an alternative to the status bar


junta1kanazawa:
### Problem

When working with many windows (10+), the horizontal status bar becomes
insufficient for displaying window names. Long window names get truncated,
and it becomes difficult to identify and switch between windows at a glance.

The `choose-tree` command already provides an excellent vertical window list,
but it's a modal overlay — you have to invoke it, pick a window, and it
disappears. There's no way to keep a persistent vertical listing of windows
visible alongside the working pane.

### Proposal

Support a vertical window/session list that can be persistently displayed
at the left or right edge of the terminal, similar to how the status bar
persistently displays information at the top or bottom.

This could work as a narrow pane-like area managed by tmux itself (not a
regular pane), showing the window list vertically — one window per line.

### Use case

- Users who regularly work with 5+ windows per session and rely on window
  names to navigate
- Workflows where `automatic-rename-format` produces descriptive but longer
  window names that don't fit well in a single horizontal line
- Sessions with multiple active processes where quick visual scanning of
  all windows is important

### Current workarounds

- `choose-tree` (modal, not persistent)
- Using a dedicated pane running a script to display `tmux list-windows`
  (fragile, wastes a pane)
- Community plugins like Tabby (external dependency, limited integration)

### Notes

I understand this would be a significant addition. I'm raising this to
gauge interest and see if it aligns with tmux's direction. Happy to
discuss alternative approaches or provide more detail on the use case.

nicm:
I would accept something like this if someone else wrote it, but I'm not personally going to do it.

It cannot just be a "window list", I don't think that would be flexible enough. It needs to be a vertical status line, so:

- I don't mind if it is instead of the existing status line or in addition to.
- It needs to be configurable whether it is on the right or the left.
- It needs to accept a format for the info to show, like the existing status line does with `status-format`, meaning at least:
  - You would need to add some way to express "new line here", so you could use a `W` loop like `status-format` for windows.
  - Ranges would need to be updated to somehow include a y position as well as an x range, so the mouse could work.
  - All the stuff to handle trimming and keeping the current window visible would need to be adapted to work for a vertical status line instead of horizontal.

If you want something that works right now, I expect you could fairly easily show more windows by using multiple status lines.

ghost:
Thank you for the detailed guidance — this is really encouraging.

[...]

PR #5456, opened 2026-08-01
Add a vertical (side) status line

dewyze:
Implements the vertical status line described by @nicm in #4910.

## What this does

A vertical status line ("side status line") at the left or right of the terminal, drawn from formats like the existing status line:

```
0:editor                $ ...window content...
1:server*
2:logs-
```

Three new session options:

- `side-status` (`off`/`left`/`right`, default `off`) — enable and position.
- `side-status-width` (default 24) — fixed width in columns.
- `side-status-format[]` — format array; each member may produce multiple rows. The default shows one row per window, using the existing `window-status-format`/`window-status-current-format`.

It reuses `status-style` and `status-interval` rather than adding parallel options, and works instead of or in addition to the horizontal status line (`status off` + `side-status left` gives vertical only).

## Against the requirements from #4910

- **"a way to express new line here"** — a new `nl` style (`#[nl]`), plus literal newlines in the expanded string. Because `#[...]` passes through format expansion untouched, `#[nl]` can be emitted conditionally inside `#{W:...}`/`#{P:...}` loops, so a loop produces one row per item. Colours, attributes and any open range carry over the break (an open range is closed at the end of the row and reopened on the next, giving one range per row).
- **"ranges would need a y position as well as an x range"** — `struct style_range` gains a `y` field, stamped with the row a range was drawn on. The side status keeps one flat y-aware range list; the horizontal status line's per-line lists are unchanged.
- **"trimming and keeping the current window visible"** — rows drawn under `list=on` form a vertical list section trimmed to the available height and scrolled to keep the `list=focus` row centred; rows drawn under `list=left-marker`/`list=right-marker` replace the first/last visible list row when rows are hidden above/below. Same syntax as the horizontal list, transposed.
- **Mouse** — clicks on the bar resolve ranges by column and row and fire the existing `MouseDown1Status`/`...StatusDefault`/etc. bindings (no new key names); `mouse_status_range`, `mouse_status_line`, `mouse_x`, `mouse_y` work with bar-relative positions. Window-area clicks subtract the bar's columns everywhere the status lines are already subtracted from rows.

## Commit structure

The first three commits are behaviour-preserving and independently useful (y on ranges, the `nl` keyword, `format_draw_lines`); one small fix (default-aligned lists were silently dropped by `format_draw`); then options, geometry, rendering, mouse, docs. Each commit builds cleanly.

## Known tradeoff

With any side bar no pane is ever full-width, so the linefeed/scroll fast paths fall back to region redraws on terminals **without** DECSLRM left/right margin support. Terminals with margins (xterm, iTerm2, wezterm, recent VTE) are unaffected. This is the same cost vertically-split panes already pay, but the bar imposes it on full-screen panes too.

Naming (`side-status-*` vs something else) and the default format are easy to change if you'd prefer different colours on this bikeshed.

## QA / testing

Regress: `regress/side-status.sh` (golden scenes for left/right/coexistence/width/off, regenerate with `GENERATE=1`). The full regress suite passes except `screen-redraw-menus.sh`, which fails identically on unpatched master on macOS.

Quick start:

```
./tmux -Ltest -f/dev/null new -A -s test
set -g side-status left            # bar appears, window shrinks
set -g side-status right           # moves to the right
set -g side-status-width 30        # resizes
set -g status off                  # vertical only
set -g mouse on                    # click a window row to select it
```

Worth exercising interactively:

- Create/kill/rename windows: the bar updates immediately; more windows than rows scrolls the list around the current window with `^`/`v` marker rows.
- Both status lines on (`status 2` + `side-status left`), messages and the command prompt (they overdraw the bar and restore on the next redraw), `display-popup` and menus over the bar.
- Mouse: click window rows (select), pane clicks/drags land on the right cells with a left bar, border drag-resize, `split-window`/`join-pane` mouse targets.
- Resize the terminal below `side-status-width` + 1: the bar disables itself; above: it returns.
- `window-size latest` with two clients of different widths attached to the same session; zoomed panes; a window bigger than the client (panning).
- Scroll a full-screen pane (e.g. `yes`) on a terminal without margin support to feel the redraw cost described above.

https://claude.ai/code/session_017EBimQYoNdaRG5SvwSSmPJ

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,12 @@
 CHANGES FROM 3.7b TO 3.8
 
+* Add a vertical status line at the left or right of the terminal, enabled
+  with the side-status option, sized with side-status-width and drawn from
+  the side-status-format array. Formats may produce multiple rows with the
+  new nl style, the window list is scrolled to keep the current window
+  visible, and ranges know the row they were drawn on so the mouse works as
+  on the status line.
+
 * Build with jemalloc on macOS to avoid what appears to be a bug in the system
   calloc(3) (issue 5385).
 
--- a/cmd-display-menu.c
+++ b/cmd-display-menu.c
@@ -313,7 +313,16 @@ cmd_display_menu_get_menu_pos(struct client *tc, struct cmdq_item *item,
 
 	ft = format_create_from_target(item);
 	if (event->m.valid) {
-		mouse_x = event->m.x + ox;
+		if (event->m.sideat == 0) {
+			if (event->m.x >= event->m.sidecols)
+				mouse_x = event->m.x - event->m.sidecols + ox;
+			else
+				mouse_x = ox;
+		} else if (event->m.sideat > 0 &&
+		    event->m.x >= (u_int)event->m.sideat)
+			mouse_x = ox + sx - 1;
+		else
+			mouse_x = event->m.x + ox;
 		if (event->m.statusat == 0) {
 			if (event->m.y >= event->m.statuslines)
 				mouse_y = event->m.y - event->m.statuslines + oy;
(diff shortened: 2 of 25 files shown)


dewyze:
Closing while I finish QA locally; will resubmit when ready.

PR #5457, opened 2026-08-02
Add pane_private_modes format listing enabled DECSET private modes

gnachman:
Report the DECSET private mode numbers currently enabled for a pane as a space-separated list (for example "7 25 1004 2004 1006"), so a control mode client attaching to an existing session can initialize its terminal state to match the pane's current state. Live changes already pass through in %output, but a newly attached client never sees the modes set before it connected; MODE_FOCUSON (1004) in particular had no format variable at all.

Alternately, we could add a new format for focus reporting, but this nicely futureproofs us against any future DECSETs.

### DIFF
--- a/format.c
+++ b/format.c
@@ -2101,6 +2101,60 @@ format_cb_synchronized_output_flag(struct format_tree *ft)
 	return (NULL);
 }
 
+/* Callback for pane_private_modes. */
+static void *
+format_cb_pane_private_modes(struct format_tree *ft)
+{
+	static const struct {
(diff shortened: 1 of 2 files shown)


nicm:
This looks good but most of the lists are comma separated rather than space, so that might be better?

gnachman:
You're right. Changed to commas

nicm:
I think you forgot to push but I just changed it to commas anyway. Applied to OpenBSD now, will be in GitHub later. Thanks!

PR #5458, opened 2026-08-03
Never emit control mode notifications inside a %begin/%end block

gnachman:
## Background

Control mode notifications must be standalone lines and must never appear inside the `%begin`/`%end` guard block that wraps a command's output. Since notifications became event-based they fire synchronously while a command runs, so notifications generated during command execution (for example %sessions-changed and %session-changed at attach) were written between that command's `%begin` and `%end.` iTerm2 treats lines inside a guard as the command response, so it never processes them as notifications and fails to finish setting up the connection, leaving no window open.

This was introduced by commit d29aa121. Before that, the notify system deferred delivery until the running command had finished, so notifications came out after `%end`. Delivering events synchronously moved them inside the block.

## The change

Track the open guard depth and defer notifications while a block is open, flushing them once the outermost block closes. Guard lines are only ever produced by `cmdq_guard,` so the depth is maintained in a dedicated `control_write_guard`.

Route all notification emitters (control-notify.c, `%subscription-changed,` `%pause`, `%continue` and `%config-error`) through `control_notify_write`.

### DIFF
--- a/cfg.c
+++ b/cfg.c
@@ -226,7 +226,7 @@ cfg_print_causes(struct cmdq_item *item)
 
 	for (i = 0; i < cfg_ncauses; i++) {
 		if (c != NULL && (c->flags & CLIENT_CONTROL))
-			control_write(c, "%%config-error %s", cfg_causes[i]);
+			control_notify_write(c, "%%config-error %s", cfg_causes[i]);
 		else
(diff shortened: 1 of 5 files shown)


nicm:
Whoops, yes, this was an oversight. Applied to OpenBSD now, will be in GitHub later. Thanks!

PR #5459, opened 2026-08-03
Set exit status on server socket failure

### LINKED ISSUE #5452: start-server and new-session exit 0 after failing to create the server socket
### Issue description

When the tmux server cannot create its listening socket, tmux prints the error to stderr but exits with status **0**. Scripts that check `$?` see success and carry on as if a server were running.

#### Steps to reproduce

Make the socket's parent directory non-writable, then start a server in it:

```sh
mkdir -p /tmp/tmuxbug/ro
chmod 500 /tmp/tmuxbug/ro
tmux -S /tmp/tmuxbug/ro/s start-server; echo "exit: $?"
tmux -S /tmp/tmuxbug/ro/s new-session -d; echo "exit: $?"
```

Observed:

```
error creating /tmp/tmuxbug/ro/s (Permission denied)
exit: 0
error creating /tmp/tmuxbug/ro/s (Permission denied)
exit: 0
```

Expected: a non-zero exit status, as with every other failure to reach a server. For contrast, on the same directory these are correct:

```sh
tmux -S /tmp/tmuxbug/ro/s has-session;  echo "exit: $?"   # exit: 1
tmux -S /tmp/tmuxbug/ro/s kill-server;  echo "exit: $?"   # exit: 1
```

Cleanup: `chmod 700 /tmp/tmuxbug/ro && rm -rf /tmp/tmuxbug`

A read-only directory is just the most convenient trigger; anything that makes `server_create_socket()` fail (`bind`/`listen` failing, socket path too long for `sun_path`, filesystem full, sandbox/MAC denial) produces the same result. This matters for scripted use, e.g. `tmux new-session -d ... || exit 1` never takes the error branch, and CI jobs that pre-create a socket directory with the wrong mode report success while nothing was started.

#### Cause

`server_start()` in `server.c` handles the failure differently depending on whether there is a client peer. At `server.c:240`:

```c
	if (cause != NULL) {
		if (c != NULL) {
			c->exit_message = cause;
			c->flags |= CLIENT_EXIT;
		} else {
			fprintf(stderr, "%s\n", cause);
			exit(1);
		}
	}
```

The `CLIENT_NOFORK` branch (`c == NULL`) correctly exits 1. The normal branch sets `exit_message` and `CLIENT_EXIT` but leaves `c->retval` at its initial 0.

`server_client_check_exit()` then packs that untouched `retval` in front of the message (`server-client.c:2314`):

```c
		size = (sizeof c->retval) + msize;
		data = xmalloc(size);
		memcpy(data, &c->retval, sizeof c->retval);
		if (c->exit_message != NULL)
			memcpy(data + sizeof c->retval, c->exit_message, msize);
		proc_send(c->peer, MSG_EXIT, -1, data, size);
```

and `client_dispatch_exit_message()` (`client.c:596`) copies that 0 into `client_exitval`. So the error text is delivered over the wire and the exit status is not.

Setting `c->retval = 1;` next to `c->exit_message = cause;` looks like it would be enough, and would match the `c == NULL` branch.

### Required information

* tmux version (`tmux -V`): `tmux next-3.8`, built from Git master `31dccb6bc9521b0ea46307974d071ad7f09f0e9b` (`./configure --enable-debug --disable-jemalloc`)
* Platform (`uname -sp`): `Darwin arm` (macOS 26.6, arm64)
* Terminal in use: not relevant — reproduces from any terminal and from a non-interactive shell; verified with `TERM=xterm-256color`
* `$TERM` *inside* tmux: not applicable, no session is ever created
* `$TERM` *outside* tmux: `xterm-256color`
* Logs from tmux (`-vv`, relevant lines):

client:

```
client started (98677): version next-3.8, socket /tmp/tmuxbug/ro/s, protocol 8
using libevent 2.1.13-stable select
flags are 0x10010000
socket is /tmp/tmuxbug/ro/s
trying connect
connect failed: No such file or directory
lock file is /tmp/tmuxbug/ro/s.lock
open failed: Permission denied
didn't get lock (-1)
got lock (-1)
trying connect
connect failed: No such file or directory
add peer 0x102b42560: 6 (0x0)
...
cmd_pack_argv: argv[0]=start-server
sending message 200 to peer 0x102b42560 (17 bytes)
client loop enter
client_signal: Child exited: 20
peer 0x102b42560 message 203
client loop exit
```

server:

```
server started (98679): version next-3.8, socket /tmp/tmuxbug/ro/s, protocol 8
using libevent 2.1.13-stable select
...
cmdq_next <global>: exit (empty)
sending message 203 to peer 0x102b45b70 (57 bytes)
server loop exit
```

The 57-byte `MSG_EXIT` is 4 bytes of `retval` (0) followed by the 53-byte message — `error creating /tmp/tmuxbug/ro/s (Permission denied)` is 52 characters plus its terminating NUL. The error text is on the wire; the status is zero.

cyphercodes:
## Summary

When the server could not create its listening socket, it sent the error message to the client with the default zero return value. Set the client return value to 1 before exiting and add regression coverage for both `start-server` and detached `new-session`.

## Testing

- `make clean && make -j$(nproc)`
- `regress/server-socket-error.sh` as an unprivileged user
- `regress/control-client-exit.sh`
- `regress/control-client-wait-exit.sh`
- `regress/new-session-no-client.sh`
- `sh -n regress/server-socket-error.sh`
- `git diff HEAD^ HEAD --check`

Fixes #5452

### DIFF
--- a/regress/server-socket-error.sh
+++ b/regress/server-socket-error.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+# Server socket creation failures must produce a nonzero client exit status.
+
+PATH=/bin:/usr/bin
+TERM=screen
+export TERM
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
+
+# Root bypasses directory mode bits, so it cannot trigger this failure.
+[ "$(id -u)" -eq 0 ] && exit 0
+
+TMP=$(mktemp -d) || exit 1
+SOCKET="$TMP/ro/socket"
+
+cleanup()
+{
+	chmod 700 "$TMP/ro" 2>/dev/null
+	rm -rf "$TMP"
+}
+
+fail()
+{
+	echo "$1" >&2
+	cleanup
+	exit 1
+}
+
+mkdir "$TMP/ro" || exit 1
+chmod 500 "$TMP/ro" || exit 1
+trap cleanup 0 1 15
+
+check_failure()
+{
+	output=$($TEST_TMUX -S "$SOCKET" -f/dev/null "$@" 2>&1)
+	retval=$?
+
+	[ "$retval" -ne 0 ] || fail "tmux $* exited zero: $output"
+	case "$output" in
+	"error creating $SOCKET ("*) ;;
+	*) fail "tmux $* produced unexpected error: $output" ;;
+	esac
+}
+
+check_failure start-server
+check_failure new-session -d
+
+exit 0
--- a/server.c
+++ b/server.c
@@ -240,6 +240,7 @@ server_start(struct tmuxproc *client, uint64_t flags, struct event_base *base,
 	if (cause != NULL) {
 		if (c != NULL) {
 			c->exit_message = cause;
+			c->retval = 1;
 			c->flags |= CLIENT_EXIT;
 		} else {
 			fprintf(stderr, "%s\n", cause);


nicm:

Issue #3693, opened 2023-09-13
Feature request , pane names for select-pane -T 'newName


freddieventura:
Hi,

This is not an issue rather a feature request (which I dont know if it exactly goes here)

So as per newer versions of tmux, we have got the ability to change the height of the status bar , making it span more than 1 line height, such as

`set-option -g status 2`

1st feature request) , the ability to distribute the whole status bar on these new-lines , seems that adding 3 lines height , the remaining lines from the 2nd one are just for padding.
So a comprehensive way to code an customize the status bar with these new lines. (so for instance if we have a big session with loads of windows opened , maybe the window names wont fit on the status bar, making it span 2 lines height even 3 would be awesome)

2nd feature request) as when setting status bar height 2 , looks like the second line shows an abstract of the current panes that exists within a window. They are show in the following format 
0 [199x55] 1 [49x55] etc...

Seeming that there is no `rename-pane` option such as `rename-window` , the trick to get pane names is to
`select-pane -T 'newName'`
And this pane Name can be shown when using the following 
```
set -g pane-border-status top
set -g pane-border-format " [ ###P #T  ] "
```

(when using that pane-border parameter)

So it would be nice to embed that to the second line of the status bar.

Or make a plugging that can do this.

Thank you

nicm:
Change your `status-format[1]` to include `#{pane_title}`?

freddieventura:
Your answer is in the right direction.

Ok I can get  the current pane index and name on the status bar by doing
```
set-option -g status 2
set-option -g status-format[1] "###P #T"
```

But how do I get a list of all them available for that window?

```
set-option -g status-format[1] "#{pane[0]} ###P #T #{pane[1]} ###P #T "
```

I know this doesnt work , but you get my idea?

nicm:
Look at what the default does...

freddieventura:
Yeah I was thinking.
Where can I look at thatone? Is it in the codebase? Any commands to check it?

freddieventura:
Capito :D
```
:show -g status-format[1]
status-format[1] "#[align=centre]#{P:#{?pane_active,#[reverse],}#{pane_index}[#{pane_width}x#{pane_height}]#[default] }"
```

freddieventura:
set-option -g status-format[1] "#[align=centre]#{P:#{?pane_active,#[reverse],}#{pane_index}[#{pane_title}]#[default] }"

Is a good way to get the functionality I was looking for

Issue #1227, opened 2018-01-08
Pane Titles


2E0PGS:
We really need a way to set a pane title just like a window title.

Allow the user to hit a key combo to enable pane titles and then another combo to trigger input for setting a title for that selected pane.

Terminator has this function and its useful if I am running many `tail -f logfile` or other bash commands that don't set a title. This also allows for easier viewing at a glance.

This is the main reason I cannot switch from Terminator to Tmux fully. There is just no proper pane titles yet! Yes I have tried using and setting current_pane_command but that is no good. I must be able to assign a name manually for this to be useful.

2E0PGS:
Terminator example:
![image](https://user-images.githubusercontent.com/15659301/34695329-57c0ba8c-f4c2-11e7-8b80-f58a28ab8ef7.png)

nicm:
select-pane -T

2E0PGS:
@nicm That does not help. So your suggesting I have to switch around to each pane every time to see it say "bash" in the bottom left or see it say "tail". This does not provide a easy to see overview of whats running at a glance. Also if I am running many tail commands I cannot see which tail is following which logfile since it just shows "tail" in the bottom left not tail -f /var/log...

Also I want the option to name the pane what I want. I can assign more info than just a running programs name without flags. Maybe I have a pane specifically for running general commands so a program name as the pane title is no good if it need to know which one I use as the hotdesk so to speak.

Hopefully this clears things up? Look at my terminator screenshot. It shows a nice overview of whats running and I can name the pane whatever I want. I don't think it would take much to implement this in tmux. It has a way to set a pane title in the border in newer versions but not a way to set that manually.

nicm:
You set it manually with select-pane -T.


On 11 Jan 2018 10:04 am, "Peter Stevenson" <notifications@github.com> wrote:

@nicm <https://github.com/nicm> That does not help. So your suggesting I
have to switch around to each pane every time to see it say "bash" in the
bottom left or see it say "tail". This does not provide a easy to see
overview of whats running at a glance. Also if I am running many tail
commands I cannot see which tail is following which logfile since it just
shows "tail" in the bottom left not tail -f /var/log...

Also I want the option to name the pane what I want. I can assign more info
than just a running programs name without flags. Maybe I have a pane
specifically for running general commands so a program name as the pane
title is no good if it need to know which one I use as the hotdesk so to
speak.

Hopefully this clears things up? Look at my terminator screenshot. It shows
[...]

Issue #807, opened 2017-03-05
Feature Request: add "set-pane-title" command


mahescho:
It would be nice to have a "rename-pane" command. The solution with using "send-keys" and a escape sequens is neither elegant nor straight forward nor institutive :-) This wold really help when scripting Tmux.

nicm:
Panes don't have names.



On 5 Mar 2017 2:35 pm, "Matthias Henze" <notifications@github.com> wrote:

It would be nice to have a "rename-pane" command. The solution with using
"send-keys" and a escape sequens is neither elegant nor straight forward
nor institutive :-) This wold really help when scripting Tmux.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/tmux/tmux/issues/807>, or mute the thread

mahescho:
Ok .... set-pane-title .... better?

https://github.com/tmux/tmux/issues/680

nicm:
Did you try the patch in the issue you linked to?

Issue #618, opened 2016-10-28
rename-pane ?


DavidGamba:
Is there a way to manually label your panes inside tmux?

I am running tmux 2.3 on linux and have `set -g pane-border-status bottom` on my `~/.tmux.conf`.

snehesht:
Try prefix + ,

DavidGamba:
yes, I am taking about pane titles. Is there a way to manually override the title? I SSH into multiple QA servers that get rebuilt all the time and they have no bashrc settings and they all have localhost as their hostnames so they all look the same within tmux.

I end up working with them as windows instead, but with the new `pane-border-status` I was hoping for a better workflow that would allow me to use them in panes with sane identifiers just like windows.

nicm:
You can change the title from within the terminal with the escape sequences, but tmux itself does not have a way to change it.

DavidGamba:
Can this issue be used to track it as a feature request? or is this something you wouldn't consider adding to tmux?

nicm:
Pane titles are not really tmux's to play with, they are for the
application inside tmux to set.

Can't you get whatever you would put in them into pane-border-format a
different way?

vdun:
I use the command `tmux send-keys -t x ...` often in my scripts. Changing the target pane number `x` in my scripts is annoying.
An alternative name could be added to panes (like windows).

micheal-w-wells:
2019 here, option to rename panes would still be appreciated.

PiyushDeshmukh:
2020 here, option to rename panes would still be appreciated.

nicm:
I don't mind having a pane name or label as well as the title, create a PR. It needs: a member in struct window_pane (like title and path); a command to change it, I would suggest select-pane -N like select-pane -T; a format like pane_name; and some code in cmd-find.c to allow it to be used as a target (-t :.mypane).

If you just want a quick label to view rather than all of that, then you can always set a pane user option (`set -p @mylabel foo`) and use that in pane-status-format or wherever with `#{@mylabel}`.

PR #5461, opened 2026-08-03
Add a key binding to change the pane title

FouadWahabi:
`select-pane -T` can already set a pane's title, but unlike sessions and windows there is no default key binding for it, so it is only reachable by typing the command out at the command prompt.

This adds `C-b T`, giving panes the key binding sessions and windows already have:

|         | key |
| ------- | --- |
| session | `$` |
| window  | `,` |
| pane    | `T` |

The prompt is pre-filled with the current title via `#T`, the same way `$` uses `#S` and `,` uses `#W`. `T` was previously unbound in the prefix table.

There is no new command and no new concept here — this only makes the existing `select-pane -T` reachable by a key, which is what the requests in #618, #807, #1227 and #3693 were each reaching for.

### What it looks like

`C-b T`, with the prompt pre-filled from the current title:

```
──0 "build"────────────────────────────────────────────────────[f][z][x]

──1 "logs"─────────────────────────────────────────────────────[f][z][x]

(select-pane) build
```

### Testing

Built with `--enable-utf8proc` on macOS; no new compiler warnings, and `mandoc -Tlint tmux.1` is clean.

Checked end to end by driving a real attached client from an outer server, the same way `regress/menu-mouse.sh` does:

- `C-b T` changes the pane title
- the prompt is pre-filled, so pressing Enter straight away leaves the title unchanged
- Escape cancels without changing the title
- with a split, only the active pane is affected
- clearing the prompt and accepting it clears the title

The `regress` suite shows no change: four tests fail here, and all four fail identically on unmodified master on this platform (`check-names`, `screen-redraw-fill-character`, `screen-redraw-menus` consistently; `prompt-mechanics` and `tty-draw-line` are timing-flaky on both).

### Changes since first version

- dropped the pane menu item, per review
- the binding note and man page entry now say "change the current pane title" rather than "rename", since panes have titles and not names

### One note

The prompt reads `(select-pane)`, since `command-prompt` falls back to the command name — the same mechanism that makes `,` read `(rename-window)`. It is consistent, but less self-explanatory. Happy to add an explicit `-p` if you would prefer something clearer.

I left `CHANGES` alone, assuming you would rather fold this in at release time.

### DIFF
--- a/key-bindings.c
+++ b/key-bindings.c
@@ -421,6 +421,7 @@ key_bindings_init(void)
 		"bind -N 'Spread panes out evenly' E { select-layout -E }",
 		"bind -N 'Switch to the last client' L { switch-client -l }",
 		"bind -N 'Clear the marked pane' M { select-pane -M }",
+		"bind -N 'Change the current pane title' T { command-prompt -I'#T' { select-pane -T '%%' } }",
 		"bind -N 'Enter copy mode' [ { copy-mode }",
 		"bind -N 'Paste the most recent paste buffer' ] { paste-buffer -p }",
 		"bind -N 'Create a new window' c { new-window }",
--- a/tmux.1
+++ b/tmux.1
@@ -356,6 +356,8 @@ Force redraw of the attached client.
 Select a new session for the attached client interactively.
 .It t
 Show the time.
+.It T
+Change the current pane title.
 .It w
 Choose the current window interactively.
 .It x


nicm:
I don't think this is a common enough operation to warrant appearing on the pane menu (already pretty big) or having a default key binding.

nicm:
I think we can add a key binding but no menu to start with, but please don't use the word "rename", panes do not have names. Say "change title" instead.

FouadWahabi:
Thanks — both done. Dropped the menu item, and the binding note and man page entry now say "change the current pane title" instead of "rename". Also reworded the commit message so "rename" doesn't leak into the tree.

nicm:

Issue #5258, opened 2026-06-22
floating panes and zoom


nicm:
A big difference between floating panes and popups is that because floating panes are fully panes, they can be zoomed. And there can only be one zoomed pane, so you cannot have a zoomed tiled pane plus floating panes on top of that.

In order to replace popups we need a way to zoom a tiled pane but allow a floating pane to float on top of it.

The main use case for this is for internal popups like the editor in buffer mode, but its likely other people are using popups in ways that mean it will be needed.

It would be sufficient to have a flag on the pane meaning "this pane cannot be zoomed and floats over whatever pane is zoomed". But I'm not tied to that if you have better ideas.

This is not necessarily needed short term.

Edit: the conclusion seems to be we need modal floating panes. We would want to support all the behaviours of popups like key-to-close.

@daneofmanythings @mgrant0

nicm:
Perhaps we simply need a way to say "this floating pane is modal" - meaning it floats over all other panes, and captures all keys and mouse clicks.

It would be modal with respect to the *window*, not the *client* (like popups), but that is fine.

nicm:
The modal idea also relates to https://github.com/tmux/tmux/pull/5256.

mgrant0:
If panes in this mode will only be used to replace modal popups and menus, how about a flag pane_popup which implies floating and is in highest z-axis so it behaves like the erstwhile overlays.

nicm:
The term popup needs to go away too to avoid confusion but yes I'd be happy with a `PANE_MODAL` flag.

polothy:
Sorry if this is noise, but with `tmux` 3.7b, macOS, Ghostty:

1. Have multiple tiled panes.
2. Zoom one of the panes.
3. Run `Prefix+*` for a new floating pane.

Result: it appears to crash/exit `tmux`.

I tried without my user config - behavior is a little different, but still breaks.

EDIT: should have said first, the floating panes feature looks really cool! Looks like it works with inline image rendering too - thanks and appreciated!

nicm:
@polothy I see this but only when I close the floating pane (`exit`), and it seems to be fixed in `master`. Does this fix it for you with 3.7?
~~~~
--- cmd-split-window.c~ 2026-06-26 18:55:30.000000000 +0100
+++ cmd-split-window.c  2026-07-02 22:23:34.737141048 +0100
@@ -93,6 +93,8 @@
                is_floating = !args_has(args, 'L');
        else
                is_floating = 0;
[...]

PR #5464, opened 2026-08-04
Prevent crashing when splitting a zoomed-in floating pane

ccjmne:
Currently, trying to `splitw` from a zoomed floating window causes tmux to crash.

Reproduction:

```tmux
newp -O
resizep -Z
splitw
```

Here I propose to save its layout cell `saved_lc` and ensure both `SPAWN_FLOATING` and `SPAWN_SPLIT` are used, so that we get the same split as we would otherwise, had the floating pane not be zoomed in.

### DIFF
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -96,6 +96,8 @@ cmd_split_window_exec(struct cmd *self, struct cmdq_item *item)
 	enum pane_lines		 lines;
 	u_int			 count = args_count(args);
 
+	window_unzoom(w, 1);
+
 	if (cmd_get_entry(self) == &cmd_new_pane_entry)
 		is_floating = !args_has(args, 'L');
 	else {


ccjmne:
Ah, I just now read through #5258...
I can see how my solution here would only be temporary.
Still though, considering that the problem incurred is an outright crash, losing all the current sessions... it may be worth considering, for those who can't wait for the stable release ;-)

nicm:
I thought we disabled creating floating panes when zoomed, I guess this is an oversight? @daneofmanythings?

daneofmanythings:
This is an oversight and should be disallowed. Either we kick the layout out of the zoomed state, or return an error. I'll take a look tonight.

daneofmanythings:
Seems like the zoomed state check got lost in the shuffle. Splitting a zoomed pane is not meant to be supported. @ccjmne, would you mind changing this pr to add a state check at the top of `cmd_splitw_window_exec` that errors out if the window is zoomed instead, please?

I'd also like to get a sense of the use case for this feature if you have time. Is this a feature you would like to see in tmux? If so, what is your workflow that utilizes it?

Thanks!

ccjmne:
@daneofmanythings, sure! I just use floating windows for many things. For example, here's what happened to me a couple of times over the last couple of weeks:

1. I have a binding to open Vim with a particular file where I have my "WIP notes": I can quickly pull it up, and close it out.
2. Sometimes I want to maximise it (and I should perhaps use `<C-b>!` instead of `<C-b>z`, but heh) when some note ends up requiring a bit more attention than a quick jotting-down.
3. I also use tmux's popups as quick menus for stuff, invoking `fzf` (some fuzzy-search TUI), and doing something with the result: I notably can browse the `man` pages like this, and have a binding there to either pop the page open in a new window, or split into a new pane in the current window.

When editing my notes and splitting the pane to consult some quick `man` page, currently this specific workflow results in the crash I'm trying to avoid.

--

I have the same-ish workflow with some TUI RSS reader: just pop it open, no news/no big news is not too disruptive, but a new message on mailing list thread of a recent patch of mine and suddenly I need the big mode ;-)

To be clear, in this case, the fix I propose here isn't what is ultimately most adequate for me: as I implemented it just now it will, as you put it, _kick the layout out of the zoomed state_ and "split" open a new floating window next to the one that was zoomed out a moment prior.

- It's probably not what would be best in general, which I think #5258 is attempting to explore/achieve, and
- it doesn't exactly make it work best for me: I should just pop my floating-become-fullscreen panes into their own windows, or even sessions with specific names...

But until I polish my set-up, at least it avoids crashing in the meantime.

--

I have no problem making the change to have `tmux` error out, however! Will get to it tonight.
Thank you!

nicm:
Ultimately we will want to support floating panes that can sit above a zoomed layout but at the moment that is not supported. This is the main thing preventing them fully replacing popups (although it is likely popups will go away even without that).

But for the moment we should either forbid opening a floating pane when zoomed, or unzoom first.

ccjmne:
I believe my proposed change is a roundabout way to "unzoom first". But I think that @daneofmanythings's suggestion to simply have it error out is most adequate, and will adjust this PR in that way tonight :-)

daneofmanythings:
Yeah, `breakp` seems like it will work better for you given how you're using the panes. I find that unzooming and continuing to split is a jarring experience and not what the user wanted when trying to make a new pane when zoomed.

ccjmne:
I made the change discussed. Note that I had it affect *all zoomed-in windows*, it doesn't have anything to do with the floating panes specifically anymore...

I figured I would relocate the test accordingly, to `pane-ops.sh`. Reading through it, it seems quite clear that at some point (just last month) we intentionally wanted an _implicit unzoom on split_, see f645f0ab (_A few more tests._, 2026-07-03).

Did I misunderstand something? Did you mean for me to only prevent the _implicit unzoom_ when it specifically is a *floating pane* that's zoomed in?

Considering that I agree that the implicit unzoom is jarring, I went ahead with what I understood we want now.

nicm:
This is not quite right because you need to check `is_floating` as well.

But I don't really understand why we can't just call `window_unzoom`? Does this not fix the problem:
~~~~
--- cmd-split-window.c  21 Jul 2026 12:28:43 -0000      1.146
+++ cmd-split-window.c  6 Aug 2026 09:02:39 -0000
@@ -104,6 +104,9 @@ cmd_split_window_exec(struct cmd *self,
                flags |= SPAWN_SPLIT;
        }

+       if (is_floating && (w->flags & WINDOW_ZOOMED))
+               window_unzoom(w, 1);
+
        if (args_has(args, 'O')) {
                if (!is_floating) {
                        cmdq_error(item, "modal pane must be floating");
~~~~
It isn't ideal but better than either a crash or an error?

This is what we do if you split a tiled pane.

ccjmne:
Yes, anything is better than an crash.

As I mentioned several times, my original implementation had it essentially unzoom first, then split. It was simple and obvious, but less than ideal: #5258 will do the work of making it "ideal": I just want it to *not crash* until then.

@daneofmanythings said that he found the automatic preemptive unzoom to be jarring—not specifically regarding floating panes—and explicitly asked for:

> a state check at the top of `cmd_splitw_window_exec` that errors out if the window is zoomed

@nicm, you had deferred to him just prior, and didn't express anything contrary then.

As I documented, I was very surprised that we wanted to prevent splitting even from a zoomed-in "tiled" pane, but , so I figured that this was our collective stance going forward.

--

I'll go back to my original design, though using your suggested, explicit call to `window_unzoom` :-)

However, calling `window_unzoom` there, as the single change, would still result in a crash:

1. `is_floating` is `0` when we are zoomed in:
```c
	if (cmd_get_entry(self) == &cmd_new_pane_entry)
		is_floating = !args_has(args, 'L');
	else {
		is_floating = window_pane_is_floating(wp);
		flags |= SPAWN_SPLIT;
	}
```
This is because `window_pane_is_floating` checks for the `LAYOUT_CELL_FLOATING` flag in the current cell, but zooming a window uses a temporary *new layout cell* that doesn't have this flag. The floating one is saved in `wp->saved_layout_cell`.

https://github.com/tmux/tmux/blob/c9ae3f6a0207444292076c9ab74054115504858f/window.c#L2813-L2821

2. If we try to unzoom like that, we do not enter this branch:

```c
	if (is_floating)
		flags |= SPAWN_FLOATING;
```

3. and therefore end up selecting the wrong cell type there:

```c
	if (flags & SPAWN_FLOATING) {
		lc = layout_get_floating_cell(item, args, lines, w, wp, flags,
		    &cause);
	} else
		lc = layout_get_tiled_cell(item, args, w, wp, flags, &cause);
```

This is all inside of `cmd_split_window_exec`.

**Maybe it can be as simple as "unzoom first", but I don't know exactly where.
What I can do, is add a couple of lines so that `is_floating` in that method actually look into the `saved_layout_cell` in case we're currently zoomed in.**

**This was the bulk of my original implementation in a8ddf0f.** Looking back at it, I still think it's the cleanest, though I could do without the `saved_lc` pointer to keep it shorter, maybe.

I pushed a new commit that does what I believe is the minimum amount of work to have the following *not crash*:

```diff
+       if (is_floating && (w->flags & WINDOW_ZOOMED))
+               window_unzoom(w, 1);
+
```

Thank you for your patience, I know I type a lot but I assure you it's generally worthwhile.
I just really want no crashes, and would love to help out in this lovely project. Cheers.

nicm:
Yes I think we should do this, we unzoom for tiled split so until we can fix it properly we should just unzoom for floating also.

ccjmne:
Is my original implementation https://github.com/tmux/tmux/commit/a8ddf0f7bc9a9d9b774aef27c4025d70527d8162 adequate?
It does just that, though not "very explicitly".

nicm:
Why not just call `window_unzoom`? I don't really know what your change is doing that is better?

ccjmne:
Just calling `window_unzoom` in that function still creates a crash: it relies on selecting the right layout cell:

```c
	if (flags & SPAWN_FLOATING) {
		lc = layout_get_floating_cell(item, args, lines, w, wp, flags,
		    &cause);
	} else
		lc = layout_get_tiled_cell(item, args, w, wp, flags, &cause);
```

and we get the *WRONG* one, the `tiled` one, because currently we don't consider a *zoomed-in floating layout* to be `is_floating`.

This `lc` is somehow inadequate and makes the whole thing go down. I haven't looked in what way exactly, I stopped when I saw we entered the `else` branch here.

I figured I could just fix that "bug" instead, and have everything work automatically.

nicm:
Not sure why it doesn't work if the floating pane is zoomed, I suspect `is_floating` is not set. But we always want to unzoom before split right now, so we may as well just do this:
~~~~
--- a/cmd-split-window.c
+++ b/cmd-split-window.c
@@ -96,6 +96,8 @@ cmd_split_window_exec(struct cmd *self, struct cmdq_item *item)
        enum pane_lines          lines;
        u_int                    count = args_count(args);

+       window_unzoom(w, 1);
+
        if (cmd_get_entry(self) == &cmd_new_pane_entry)
                is_floating = !args_has(args, 'L');
        else {

ccjmne:
I believe I've outlined fairly clearly why it doesn't work if the floating pane is zoomed in [that comment](https://github.com/tmux/tmux/pull/5464#issuecomment-5204540488).

You are correct in your suspicion: `is_floating` then is unset. We could explicitly `window_unzoom`, but:
1. **that change *alone* would still result in a _crash_**
2. as long as I've been using `tmux`, we have been unzooming the current window as we split. I looked at the implementation right now: we *are* already calling `window_unzoom` later, in both tiled-zoomed and floating-zoomed paths, via `window_push_zoom(wp->window, 1, (flags & SPAWN_ZOOM))` in both `layout_get_floating_cell` and `layout_get_tiled_cell`
3. invoking `window_unzoom` another time prior to that, doesn't change anything to the behaviour. In fact, if I'm reading this right, its effects would be overridden by the later calls to it that are part of the very `layout_get_{tiled,floating}_cell`.

The way I see it, if we just invoke `layout_get_floating_cell` instead of `layout_get_tiled_cell` when we're floating-zoomed, everything is back in order.

**We already unzoom on break-pane**, I could *directly* trace back the place we call it at, all the way to c71844de (Add resize-pane -Z, 2013-03-24). **We always have.** The problem now is that we misidentify the "unzoomed" cell as a tiled one rather than a floating one, do something quite illegal and ram ourselves into a `SEGFAULT`.

nicm:
How do you make it crash with this change? It works OK for me.

ccjmne:
My apologies, I don't know how I could miss that.
You're correct, just unconditionally calling it there is plenty fine. Whew!

ccjmne:
I updated the change to the minimal change you proposed. I could bring back the test I'd come up with, but the change is so trivial maybe there's no point.
Thanks again for your patience. Cheers!

nicm:
I've applied this to OpenBSD now, will be in GitHub later. Thanks!

PR #5468, opened 2026-08-04
Add a vertical (side) status line

dewyze:
Hello, I saw the vertical status line issue and decided to take a shot at it with an LLM. I would call this "Human directed, LLM Written." I am not a C dev, I'm a ruby dev, but did my best to direct it based on previous art, some QA, and asking it to do its best to act like @nicm in PR review.

I know that LLM generated PRs created a bigger burden for maintainers when the PR author doesn't have full context, and I acknowledge if this missed the mark, or is too much work in that area, let me know.

That being said, happy to address any and all feedback you have to help this land. I include a screenshot of a separate plugin I am working on and will publish after this to get the "herder" type feel, that is ultimately my goal here, I don't want to leave tmux.

Let me know how I can help! -- John

CleanShot 2026-08-04 at 14 55 34@2x

--- BELOW THIS LINE IS LLM GENERATED --- 

Implements the vertical status line described by @nicm in #4910.

## What this does

A vertical status line ("side status line") at the left or right of the terminal, drawn from formats like the existing status line:

```
0:editor                $ ...window content...
1:server*
2:logs-
```

Three new session options:

- `side-status` (`off`/`left`/`right`, default `off`) — enable and position.
- `side-status-width` (default 24) — fixed width in columns.
- `side-status-format[]` — format array; each member may produce multiple rows. The default shows one row per window using the window status *styles* (not the window status *formats*, which lay text out for a horizontal line); with more than one session it shows a clickable row per session with the current session's windows indented below it, so clicking a collapsed session switches to and expands it.

It reuses `status-style` and `status-interval` rather than adding parallel options, and works instead of or in addition to the horizontal status line (`status off` + `side-status left` gives vertical only).

## Against the requirements from #4910

- **"a way to express new line here"** — a new `nl` style (`#[nl]`), plus literal newlines in the expanded string. Because `#[...]` passes through format expansion untouched, `#[nl]` can be emitted conditionally inside `#{W:...}`/`#{P:...}` loops, so a loop produces one row per item. Colours, attributes and any open range carry over the break (an open range is closed at the end of the row and reopened on the next, giving one range per row).
- **"ranges would need a y position as well as an x range"** — `struct style_range` gains a `y` field, stamped with the row a range was drawn on. The side status keeps one flat y-aware range list; the horizontal status line's per-line lists are unchanged.
- **"trimming and keeping the current window visible"** — rows drawn under `list=on` form a vertical list section trimmed to the available height and scrolled to keep the `list=focus` row centred; rows drawn under `list=left-marker`/`list=right-marker` replace the first/last visible list row when rows are hidden above/below. Same syntax as the horizontal list, transposed.
- **Mouse** — clicks on the bar resolve ranges by column and row and fire the existing `MouseDown1Status`/`...StatusDefault`/etc. bindings (no new key names); `mouse_status_range`, `mouse_status_line`, `mouse_x`, `mouse_y` work with bar-relative positions. Window-area clicks subtract the bar's columns everywhere the status lines are already subtracted from rows.

## Commit structure

The first three commits (y on ranges, the nl keyword, the default-aligned list fix) stand alone and can go in as a separate change first if you prefer — happy to split the PR. The first three commits are behaviour-preserving and independently useful (y on ranges, the `nl` keyword, `format_draw_lines`); one small fix (default-aligned lists were silently dropped by `format_draw`); then options, geometry, rendering, mouse, docs. Each commit builds cleanly.

## Known tradeoff

With any side bar no pane is ever full-width, so the linefeed/scroll fast paths fall back to region redraws on terminals **without** DECSLRM left/right margin support. Terminals with margins (xterm, iTerm2, wezterm, recent VTE) are unaffected. This is the same cost vertically-split panes already pay, but the bar imposes it on full-screen panes too.

Naming (`side-status-*` vs something else) and the default format are easy to change if you'd prefer different colours on this bikeshed.

## QA / testing

Regress: `regress/side-status.sh` (golden scenes for left/right/coexistence/width/off, regenerate with `GENERATE=1`). The full regress suite passes except `screen-redraw-menus.sh`, which fails identically on unpatched master on macOS.

Quick start:

```
./tmux -Ltest -f/dev/null new -A -s test
set -g side-status left            # bar appears, window shrinks
set -g side-status right           # moves to the right
set -g side-status-width 30        # resizes
set -g status off                  # vertical only
set -g mouse on                    # click a window row to select it
```

Worth exercising interactively:

- Create/kill/rename windows: the bar updates immediately; more windows than rows scrolls the list around the current window with `^`/`v` marker rows.
- Both status lines on (`status 2` + `side-status left`), messages and the command prompt (they overdraw the bar and restore on the next redraw), `display-popup` and menus over the bar.
- Mouse: click window rows (select), pane clicks/drags land on the right cells with a left bar, border drag-resize, `split-window`/`join-pane` mouse targets.
- Resize the terminal below `side-status-width` + 1: the bar disables itself; above: it returns.
- `window-size latest` with two clients of different widths attached to the same session; zoomed panes; a window bigger than the client (panning).
- Scroll a full-screen pane (e.g. `yes`) on a terminal without margin support to feel the redraw cost described above.


mgrant0:
What happens when there's more than one status line?   e.g. set -g status 3

nicm:
This looks like the right idea on a quick look.

I don't think we should support more than one side status line.

I'm not that wild about the name "side" but it will do for now.

What is the point in `side-status-format` being an array when you can only have one side status line?

I think the default green background is too bold for the right status line and instead it should probably be green-on-black with a line on the right (ACS `x`).

The default width of 24 is far too wide, make it 10.

Do not use `^` and `v` for up and down markers, they look horrible, use ACS arrow characters (`-` and `.`).

I like the default tree view but it'd look nicer if the child items also had arrows, like tree mode (I just changed it to use an ACS arrow also).

Functions in `status.c` should have a `status_` prefix so `side_status_size` should be `status_side_size` etc.

dewyze:
@nicm Thanks, all pushed:

- status.c functions renamed to `status_side_*`.
- Default width 10.
- ACS up/down arrows for the list markers; the sessions view uses the
  tree mode branch markers and arrow (`tq`/`mq` + `+`).
- Default style is green on black with an ACS `x` line on the edge next
  to the window area, taken out of `side-status-width`. I read "line on
  the right" as the separator against the window area; easy to flip if
  you meant otherwise.

Width: 10 as asked, but the line and branch markers leave five columns
for a window name, so most truncate. 14 leaves nine. Not attached,
screenshots below. Thoughts on 14?

Style: `status-style` couldn't give green on black here, so I added
`side-status-style`. The alternative is colors in the default format
string, but blank rows below the content would still fill from
`status-style`. Happy to switch if you prefer.

Array: each member adds its rows below the previous one, so a plugin
can append rows without clobbering the user's format. If you don't see
value there for multiple plugins to be able to contribute I'll fold it to a single string.

Width: 10

side_status_width10

Width 14:

side_status_width14

dewyze:
@mgrant0 

The horizontal status rows keep the full terminal width; the bar fills
the remaining rows above (or below with `status-position top`). So
`status 3` plus `side-status left` gives three full-width rows at the
bottom and the bar in the rest. Screenshot: `status 3`, a popup, and
more windows than rows (the arrow is the scroll marker).

tmux_vert

nicm:
I think a single option is fine.

Width 14 or even 16 is OK.

dewyze:
Done: side-status-format is a single option and the default width is 14.

dewyze:
One question I have, is does it make more sense for a side bar (is "vertical status bar or side bar" a better name?) to actually extend to the bottom and the bottom status line stops at the side bar? I think in most other apps, the side bar wins, but I also don't want to break years of tradition on this one.

I don't have a strong opinion, I imagine most people can customize their setup and if they want both on at the same time they can adjust their lines to be what they want.

nicm:
We don't call it the "status bar", it is the "status line". So really this should be "status column" but that is a bit weird...

I personally would keep it so it does not shorten the status line, I think it would look odd since for most people the bottom row will be empty.

dewyze:
Maybe "status panel" Is the most accurate? It's not a line or a sidebar (despite it being on the side), it's the idea of more space than a line?

Either way, are the next steps just time to review? Otherwise let me know anything else I can help with.

nicm:
I think "side status" is probably OK for now, but I'll think about it.

I will need to take another look at it but I am away all next week so it will not be until after that.

This will be a 3.9 feature in any case, so it will wait until 3.8 is tagged before it goes in.

nicm:

dewyze:
Rebased on current master (post 3.8-rc). Only conflict was the CLIENT_ flag bit, so CLIENT_SIDESTATUSOFF has moved again. Everything else is unchanged from what you last looked at.

While I was in there, CLIENT_CONTROL_DISCARD is defined as 0x1000000000ULL, which is the same value as CLIENT_BRACKETPASTING. I assume it was meant to be 0x10000000000ULL, so I skipped that bit for ours.

https://github.com/tmux/tmux/blob/e880cf63e0a9fe095d7c5d313761520fb1a8653c/tmux.h#L2303-L2307

juanibiapina:
I found a refresh edge case when the side status is enabled without the horizontal status:

```tmux
set -g status off
set -g status-interval 1
set -g side-status left
set -g side-status-format '#(printf x >> /tmp/tmux-side-timer; printf x)'
```

The sidebar renders correctly, but its periodic refresh stops:

```text
After attach:             1 execution
After 3 seconds:          1 execution
After enabling status:    5 executions after 3 more seconds
```

`status_timer_start()` currently starts the timer only when `status` is enabled:

```c
if (s != NULL && options_get_number(s->options, "status"))
    status_timer_callback(-1, 0, c);
```

The callback sets `CLIENT_REDRAWSTATUS`, which already maps to both `REDRAW_STATUS` and `REDRAW_SIDE_STATUS`. It seems the timer condition should also include `side-status`:

```c
if (s != NULL &&
    (options_get_number(s->options, "status") != 0 ||
     options_get_number(s->options, "side-status") != 0))
    status_timer_callback(-1, 0, c);
```

`options_push_changes()` may also need to call `status_timer_start_all()` when `side-status` changes, so enabling the sidebar at runtime starts the timer.

The existing regression test verifies that `status off` and `side-status left` render together, but it does not cover periodic format updates in that configuration.

nicm:
@dewyze could you have a look at @juanibiapina's problem?

dewyze:
Thanks, addressed. The timer only started when status was on, and turning side-status on didn't start it either. Fixed both and added a test for side-status with status off.

Also rebased over the popup removal. Since nothing can sit over the side status now, it's drawn directly like the status line. Added a scene with a floating pane next to it to cover that. Let me know if you'd prefer another approach.

PR #5469, opened 2026-08-04
Bugfix, fix expanding every format callback even when logging was off.

mgrant0:
format_log_debug() was expanding every format callback even when logging was off. Each
wheel event triggered a hook path that evaluated #{mouse_word}; on the unbroken base64
line, that scanned and copied the entire huge word repeatedly.

regression test included


nicm:

PR #5472, opened 2026-08-07
Fix source-file nesting depth leak when sourcing from a hook

### LINKED ISSUE #5437: `source-file` from a hook leaks the nesting-depth counter → spurious "too many nested files"
# Issue description

The issue has been extensively studied by Claude code by analying the phenomenon combined with source code. The problem is reproducible.

# `source-file` from a hook leaks the nesting-depth counter → spurious "too many nested files"

Status: open (reported against tmux 3.6a; code path unchanged on master)

## Summary

When `source-file` is run from a **hook**, its nesting-depth counter is incremented on
one variable and decremented on another. The counter therefore grows without bound.
After 50 hook-driven `source-file` invocations the client is permanently poisoned and
every subsequent `source-file` — including ones from ordinary key bindings — fails with:

```
too many nested files
```

despite no actual nesting being involved. The sourced file is never read, so whatever
the configuration was supposed to do silently stops working.

## Affected versions

Reproduced on tmux 3.6a. The relevant code in `cmd-source-file.c` is byte-identical on
current master (only the CVS `$OpenBSD$` header line differs), so master is affected too.

## The defect

`cmd-source-file.c` keeps two separate depth counters — a file-static one used when the
command has no client, and a per-client one:

```c
33: #define CMD_SOURCE_FILE_DEPTH_LIMIT 50
34: static u_int cmd_source_file_depth;
```

`cmd_source_file_exec()` picks a counter based on the client of the **`source-file`
command item**:

```c
164:    struct client *c = cmdq_get_client(item);
...
173:            if (cmd_source_file_depth >= CMD_SOURCE_FILE_DEPTH_LIMIT) {
174:                    cmdq_error(item, "too many nested files");
177:            cmd_source_file_depth++;
...
180:            if (c->source_file_depth >= CMD_SOURCE_FILE_DEPTH_LIMIT) {
181:                    cmdq_error(item, "too many nested files");
184:            c->source_file_depth++;
```

The matching decrement happens in a completion callback that picks its counter based on
the client of a **different, newly created queue item**:

```c
 64: cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
 66:    struct client *c = cmdq_get_client(item);
 69:            cmd_source_file_depth--;
 72:            c->source_file_depth--;
```

That callback item is created by `cmdq_get_callback()` and inserted with
`cmdq_insert_after(cdata->after, new_item)` (line 91), so its client is inherited from
`cdata->after` — the last item produced while loading the file — not from the original
`source-file` item.

For a hook-dispatched `source-file` these two clients disagree: the `source-file` item
carries the attached client, while the completion item ends up client-less. The result
is `c->source_file_depth++` paired with `cmd_source_file_depth--`.

Both counters are wrong afterwards:

* `c->source_file_depth` grows by one per hook-driven `source-file` and is never reduced.
  Once it reaches 50, **every** `source-file` issued by that client fails — including
  ones from plain key bindings that never leaked anything.
* the file-static `cmd_source_file_depth` is decremented below zero and, being `u_int`,
  wraps to `UINT_MAX`.

The counts are balanced in number but not in target, which is what makes this easy to
miss: for every increment there *is* a decrement, just on the wrong variable.

## Reproducer

Fully self-contained; requires only `tmux` and `script(1)`. It starts a private server on
its own socket and drives a real attached client through a pty, because the bug needs a
long-lived client — dispatching commands with `tmux <cmd>` from the shell creates a fresh
short-lived client each time and will not show the leak.

```bash
#!/bin/bash
set -u
T=$(mktemp -d); cd "$T"

echo 'set -g @dummy 1' > leaf.conf

cat > repro.conf <<'EOF'
set -g prefix C-b
EOF
echo "set-hook -g session-window-changed \"source-file $T/leaf.conf\"" >> repro.conf

mkfifo keys
tmux -L repro -f repro.conf -vv new-session -d -s s -x 80 -y 24 "sleep 100000"
tmux -L repro new-window -d -t s: "sleep 100000"

# a real attached client, driven through a pty
setsid script -qfc "tmux -L repro attach -t s" /dev/null < keys >/dev/null 2>&1 &
sleep 1.5
exec 3> keys

for i in $(seq 60); do
    printf '\002n' >&3          # prefix + n == next-window
    sleep 0.06
done
sleep 1

LOG=$(ls tmux-server-*.log | head -1)
echo "increments : $(grep -c 'cmd_source_file_exec: depth now' "$LOG")"
echo "decrements : $(grep -c 'cmd_source_file_complete_cb: depth now' "$LOG")"
grep -h "depth now" "$LOG" | head -6 | sed 's/^[0-9.]* /  /'
echo "errors     : $(grep -c 'too many nested files' "$LOG")"

tmux -L repro kill-server
```

### Observed

The status line starts showing `too many nested files` from the 51st window switch on,
and the `-vv` server log shows the two counters diverging from the very first switch:

```
increments : 50
decrements : 50
  cmd_source_file_exec: depth now 1
  cmd_source_file_complete_cb: depth now 4294967295
  cmd_source_file_exec: depth now 2
  cmd_source_file_complete_cb: depth now 4294967294
  cmd_source_file_exec: depth now 3
  cmd_source_file_complete_cb: depth now 4294967293
errors     : 10
```

50 increments, 50 decrements, and yet one counter climbs to the limit while the other
underflows past zero.

### Expected

The depth returns to 0 after each hook fires, and no error is ever produced — the hook
sources one file, one level deep, 60 times in a row.

## Secondary symptom: unrelated key bindings break

Because the poisoned counter is the *client's*, the failure is not confined to the hook.
With a config that both sources from a hook and sources from a key binding:

```
bind -n S-Up source-file /path/leaf.conf
set-hook -g session-window-changed "source-file /path/other.conf"
```

where the hook chain is two files deep (so it leaks 2 per window switch), pressing `S-Up`
after N window switches gives:

```
S-Up after  0 switches -> ok
S-Up after 20 switches -> ok
S-Up after 25 switches -> ERROR 'too many nested files'
S-Up after 30 switches -> ERROR 'too many nested files'
```

The key binding never leaks anything itself — 40 consecutive `S-Up` presses with the hook
removed keep the counter at a high-water mark of 1 — it simply inherits the exhausted
budget. Since `source-file` returns `CMD_RETURN_ERROR` without reading the file, the
binding silently stops doing its job.

## Scope

* Only the **hook** path leaks. `source-file` from a key binding, from the command
  prompt, or from `tmux source-file` on the command line is correctly balanced.
* The damage is per client: detaching and reattaching allocates a fresh `struct client`
  and clears `source_file_depth`. This makes the bug look intermittent — it disappears on
  reattach and comes back after enough hook invocations.
* Any hook that sources a file is affected; `session-window-changed` merely makes it
  obvious because it fires on every window switch.

## Suggested fix

Make the decrement target the same counter the increment used, rather than re-deriving it
from the completion item's client. Carrying the client (or a flag saying which counter was
bumped) in `struct cmd_source_file_data` and passing it to the callback through
`cmdq_get_callback()`'s `data` argument would be sufficient.

## Workaround

Do not call `source-file` directly from a hook. Wrapping it so it runs as its own
short-lived client avoids the mismatched counters entirely:

```
# leaks
set-hook -g session-window-changed "source-file ~/path/to/file.conf"

# does not leak
set-hook -g session-window-changed "run-shell -b 'tmux source-file ~/path/to/file.conf'"
```

Measured over 30 window switches with a two-file-deep hook chain: the direct form reaches
a depth of 60 and starts erroring at switch 25, the `run-shell` form stays at a high-water
mark of 2 with 0 errors, and the sourced files still take effect identically.

# Required information

* tmux version (`tmux -V`). tmux 3.6a
* Platform (`uname -sp`). Linux x86_64
* Terminal in use (xterm, rxvt, etc). mobaxterm
* $TERM *inside* tmux (`echo $TERM`).  xterm-256color
* $TERM *outside* tmux (`echo $TERM`). xterm-256color

MsfPablo:
Fixes #5437.

`cmd_source_file_exec()` increments either the file-static `cmd_source_file_depth` (when there is no client) or `c->source_file_depth`. The completion callback re-derived its client from `cmdq_get_client()` on a *different* queue item — the callback item created by `cmdq_get_callback()` and inserted after the last item produced while loading the file — so it could decrement the other counter.

The mismatch is easy to hit in practice: `file_create_with_client()` clears the client when it is attached, so the client threaded through `file_read()` -> `cmd_source_file_done()` -> `cmd_source_file_complete()` is NULL for an attached client, while `cmd_source_file_exec()` saw the real client. For a hook-driven `source-file` the increment therefore lands on `c->source_file_depth` and the decrement on the static counter.

Consequences, per the issue report:

* `c->source_file_depth` grows by one per hook-driven `source-file` and never comes back down. Once it reaches 50, *every* `source-file` from that client fails with "too many nested files" — including ones from plain key bindings that never leaked anything, which then silently stop working.
* the file-static counter is decremented below zero and, being `u_int`, wraps to `UINT_MAX`.

The fix stores the client in `struct cmd_source_file_data` at exec time and passes it to the callback through the `cmdq_get_callback()` data argument, so the decrement always targets the counter that was incremented. A reference is taken on the client for the lifetime of the callback item.

## Verification

Built on macOS with the reproducer from #5437 (driven through a pty so the client is long-lived), 60 window switches with `set-hook -g session-window-changed "source-file leaf.conf"`:

Before:

```
increments : 50
decrements : 50
  cmd_source_file_exec: depth now 1
  cmd_source_file_complete_cb: depth now 4294967295
  cmd_source_file_exec: depth now 2
  cmd_source_file_complete_cb: depth now 4294967294
  cmd_source_file_exec: depth now 3
  cmd_source_file_complete_cb: depth now 4294967293
errors     : 10
```

After:

```
increments : 60
decrements : 60
  cmd_source_file_exec: depth now 1
  cmd_source_file_complete_cb: depth now 0
  cmd_source_file_exec: depth now 1
  cmd_source_file_complete_cb: depth now 0
  cmd_source_file_exec: depth now 1
  cmd_source_file_complete_cb: depth now 0
errors     : 0
```

Also checked that the unaffected paths still behave: sourcing from the startup config, nested `source-file` from a config file, repeated `tmux source-file` from the command line, and the error message for a missing file.

### DIFF
--- a/cmd-source-file.c
+++ b/cmd-source-file.c
@@ -50,6 +50,7 @@ const struct cmd_entry cmd_source_file_entry = {
 
 struct cmd_source_file_data {
 	struct cmdq_item	 *item;
+	struct client		 *c;
 	int			  flags;
 
 	struct cmdq_item	 *after;
@@ -61,9 +62,9 @@ struct cmd_source_file_data {
 };
 
 static enum cmd_retval
-cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
+cmd_source_file_complete_cb(struct cmdq_item *item, void *data)
 {
-	struct client	*c = cmdq_get_client(item);
+	struct client	*c = data;
 
 	if (c == NULL) {
 		cmd_source_file_depth--;
@@ -71,6 +72,7 @@ cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
 	} else {
 		c->source_file_depth--;
 		log_debug("%s: depth now %u", __func__, c->source_file_depth);
+		server_client_unref(c);
 	}
 
 	cfg_print_causes(item);
@@ -78,8 +80,9 @@ cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
 }
 
 static void
-cmd_source_file_complete(struct client *c, struct cmd_source_file_data *cdata)
+cmd_source_file_complete(struct cmd_source_file_data *cdata)
 {
+	struct client		*c = cdata->c;
 	struct cmdq_item	*new_item;
 	u_int			 i;
 
@@ -88,7 +91,9 @@ cmd_source_file_complete(struct client *c, struct cmd_source_file_data *cdata)
 		    c != NULL &&
 		    c->session == NULL)
 			c->retval = 1;
-		new_item = cmdq_get_callback(cmd_source_file_complete_cb, NULL);
+		if (c != NULL)
+			c->references++;
+		new_item = cmdq_get_callback(cmd_source_file_complete_cb, c);
 		cmdq_insert_after(cdata->after, new_item);
 	}
 
@@ -127,7 +132,7 @@ cmd_source_file_done(struct client *c, const char *path, int error,
 	if (n < cdata->nfiles)
 		file_read(c, cdata->files[n], cmd_source_file_done, cdata);
 	else {
-		cmd_source_file_complete(c, cdata);
+		cmd_source_file_complete(cdata);
 		cmdq_continue(item);
 	}
 }
@@ -187,6 +192,7 @@ cmd_source_file_exec(struct cmd *self, struct cmdq_item *item)
 
 	cdata = xcalloc(1, sizeof *cdata);
 	cdata->item = item;
+	cdata->c = c;
 
 	if (args_has(args, 'q'))
 		cdata->flags |= CMD_PARSE_QUIET;
@@ -249,7 +255,7 @@ cmd_source_file_exec(struct cmd *self, struct cmdq_item *item)
 		file_read(c, cdata->files[0], cmd_source_file_done, cdata);
 		retval = CMD_RETURN_WAIT;
 	} else
-		cmd_source_file_complete(c, cdata);
+		cmd_source_file_complete(cdata);
 
 	free(cwd);
 	return (retval);


nicm:
Is this different from my patch?

MsfPablo:
Same idea, your version is cleaner — you should land yours.

Approach is identical: stash the client in `cdata`, use it in `complete_cb` instead of re-deriving from the queue item. Differences:

* You extracted `cmd_source_file_free_data()` (cleans up cdata + does `server_client_unref`) — single place for the unref.
* You bump `c->references` once in `cmd_source_file_exec()` and rely on the unref in free_data.
* I bumped it inside `cmd_source_file_complete()` and unref'd inside `complete_cb`, before `free(cdata)` at the old call site — ref dance happens in two places, which is uglier.

Yours is the right shape. Closing this one in favor of your diff.

MsfPablo:
Closing in favor of the maintainer's patch from the linked issue (#5437 x.diff), which addresses the same defect with a cleaner extraction of `cmd_source_file_free_data()`. Thanks nicm.

PR #5474, opened 2026-08-08
file: report the real error from a failed async file write

### LINKED ISSUE #5451: save-buffer reports success (exit 0) when writing the file fails, silently destroying the target file
### Issue description

`save-buffer` exits 0 and prints nothing when the write to the destination file fails. The destination has already been truncated by `open(O_TRUNC)` at that point, so its previous contents are gone and the buffer is not written either — data is lost in both directions with no indication that anything went wrong.

Reproducer, using `ulimit -f` to make the write fail part-way through. Any real write failure behaves the same (`ENOSPC`, `EDQUOT`, `EIO`, `EFBIG`, `EPIPE`):

```sh
rm -rf /tmp/tmuxbug && mkdir -p /tmp/tmuxbug && cd /tmp/tmuxbug

tmux -Ltest kill-server 2>/dev/null

# 100 KB of data in a paste buffer
yes X | head -c 100000 > big.txt
tmux -Ltest -f/dev/null new-session -d
tmux -Ltest load-buffer -b big big.txt

# a file that already contains data we care about
printf 'IMPORTANT EXISTING DATA\n' > target.txt
echo "target.txt before: $(wc -c < target.txt) bytes"

# make writes to target.txt fail part-way through (RLIMIT_FSIZE = 20 KB)
( ulimit -f 20; tmux -Ltest save-buffer -b big target.txt; echo "save-buffer exit status: $?" )

echo "target.txt after : $(wc -c < target.txt) bytes (buffer was 100000 bytes)"
tmux -Ltest kill-server
```

Output:

```
target.txt before:       24 bytes
save-buffer exit status: 0
target.txt after :    20480 bytes (buffer was 100000 bytes)
```

The original 24 bytes are gone, only 20480 of the 100000 buffer bytes were written, and `save-buffer` reported success.

It also happens when the destination is a pipe whose reader goes away:

```sh
mkfifo p
( head -c 8 < p > got.txt ) &
sleep 1
tmux -Ltest save-buffer -b big p; echo "save-buffer exit status: $?"
wait
echo "delivered: $(wc -c < got.txt) bytes"
```

```
save-buffer exit status: 0
delivered:        8 bytes
```

For contrast, these two failures *are* reported correctly today, so it is specifically the write-error path that is swallowed:

```
$ tmux -Ltest save-buffer -b big nowrite/x.txt     # open() fails
Permission denied: /tmp/tmuxbug/nowrite/x.txt      # exit 1

$ tmux -Ltest load-buffer -b r /nonexistent        # read path
No such file or directory: /nonexistent            # exit 1
```

Expected behaviour: `save-buffer` should report the error and exit non-zero, the way it already does when `open()` fails.

### Where it goes wrong

`file_write_error_callback()` in `file.c` tears the stream down but never tells
the server anything:

```c
/* file.c:531 */
static void
file_write_error_callback(__unused struct bufferevent *bev, __unused short what,
    void *arg)
{
	struct client_file	*cf = arg;

	log_debug("write error file %d", cf->stream);

	bufferevent_free(cf->event);
	cf->event = NULL;

	close(cf->fd);
	cf->fd = -1;

	if (cf->cb != NULL)
		cf->cb(NULL, NULL, 0, -1, NULL, cf->data);
}
```

Compare `file_read_error_callback()` (`file.c:678`), which does send the error back to the server, which is why `load-buffer` reports read failures:

```c
	msg.stream = cf->stream;
	msg.error = (what & EVBUFFER_ERROR) ? EIO : 0;
	proc_send(cf->peer, MSG_READ_DONE, -1, &msg, sizeof msg);
```

There is no equivalent for writes, so the server never learns the write failed and `cmd_save_buffer_done()` runs with `error == 0`. Once the callback has set `cf->event = NULL`, the remaining `MSG_WRITE` data is quietly discarded by
`file_write_data()` (`file.c:634`), and `file_write_close()` (`file.c:654`) just frees the stream.

The `-vv` client log for the failing run shows exactly that — the error is noticed, then the rest of the data is dropped and nothing is reported:

```
1785564371.850712 peer 0x10316e560 message 303
1785564371.850728 open write file 4 /tmp/tmuxbug/target.txt
1785564371.850803 sending message 305 to peer 0x10316e560 (8 bytes)
1785564371.850815 write check file 4
1785564371.850900 peer 0x10316e560 message 304
1785564371.850907 write 16364 to file 4
1785564371.850940 write check file 4
1785564371.850947 peer 0x10316e560 message 304
1785564371.850951 write 16364 to file 4
1785564371.850972 peer 0x10316e560 message 304
1785564371.850976 write 16364 to file 4
1785564371.850984 write error file 4
1785564371.851147 peer 0x10316e560 message 304
1785564371.851154 write 16364 to file 4
1785564371.851166 peer 0x10316e560 message 304
1785564371.851169 write 16364 to file 4
1785564371.851187 peer 0x10316e560 message 304
1785564371.851191 write 16364 to file 4
1785564371.851197 peer 0x10316e560 message 304
1785564371.851201 write 1816 to file 4
1785564371.851204 peer 0x10316e560 message 306
1785564371.851207 close file 4
1785564371.851211 peer 0x10316e560 message 203
1785564371.851213 client loop exit
```

Note that `ulimit -f` also caps the `-vv` log files themselves, so the limit is set high enough above for the log to survive.

### Required information

* tmux version (`tmux -V`): `tmux next-3.8`, built from Git master at `31dccb6bc9521b0ea46307974d071ad7f09f0e9b`
* Platform (`uname -sp`): `Darwin arm` (macOS, arm64); libevent 2.1.13-stable, utf8proc 2.11.3, ncurses 6.6
* Terminal in use: not involved — the reproduction is entirely non-interactive and no client ever attaches to the session
* `$TERM` *inside* tmux: `tmux-256color`
* `$TERM` *outside* tmux: the reproduction runs from a non-interactive shell with no tty
* Logs from tmux: the relevant `-vv` client log is quoted in full above; the server log contains nothing beyond the normal `save-buffer` command dispatch.

MsfPablo:
Fixes #5451.

### Root cause

`file_write_error_callback()` is the bufferevent error callback invoked when a file write fails partway through (`ENOSPC`, `EDQUOT`, `EIO`, `EFBIG`, `EPIPE`, ...). It unconditionally called the completion callback with `error=0`:

```c
if (cf->cb != NULL)
    cf->cb(NULL, NULL, 0, -1, NULL, cf->data);
```

`cmd_save_buffer_done()` only reports an error when `error != 0`, so it silently reported success even though the destination had already been truncated (`O_TRUNC`) and only partially written — data lost in both directions with no indication anything went wrong, as described in #5451.

### Fix

Capture `errno` at the point the write failed (the callback runs synchronously off the failing write, so `errno` is still valid) and pass it through instead of a hardcoded 0, falling back to `EIO` if it's somehow unset.

### Verification

I confirmed the bug by code inspection — this is the only caller of the completion callback on the error path, and it's the only one that hardcodes 0. I was not able to reproduce the exact repro from #5451 locally: it uses `ulimit -f` to force a write failure, but on macOS this delivers `SIGXFSZ` and kills the writing process outright rather than returning `EFBIG` from `write()`, so it doesn't exercise this exact code path the same way it does on Linux (where the issue was filed). The fix itself is a straightforward propagation of an already-available `errno` value at the one place it was being dropped.

### DIFF
--- a/file.c
+++ b/file.c
@@ -533,6 +533,7 @@ file_write_error_callback(__unused struct bufferevent *bev, __unused short what,
     void *arg)
 {
 	struct client_file	*cf = arg;
+	int			 error = errno;
 
 	log_debug("write error file %d", cf->stream);
 
@@ -542,8 +543,10 @@ file_write_error_callback(__unused struct bufferevent *bev, __unused short what,
 	close(cf->fd);
 	cf->fd = -1;
 
+	if (error == 0)
+		error = EIO;
 	if (cf->cb != NULL)
-		cf->cb(NULL, NULL, 0, -1, NULL, cf->data);
+		cf->cb(NULL, NULL, error, -1, NULL, cf->data);
 }
 
 /* Client file write callback. */


nicm:
Is this the same as my fix?

MsfPablo:
Different (related) fix — your 742efe989f38 fixed the read path (file_read_error_callback and fread()). This one only touches the write path (file_write_error_callback), which is the only other completion-callback caller that hardcodes 0.

So the two patches together close the same class of bug on both sides; mine is the write half that your read-side commit didn't cover. Worth pulling both.

nicm:
I don't think you read the original issue or in fact wrote that reply yourself.

PR #5478, opened 2026-08-08
prompt: fix off-by-one heap overflow deleting a character

Uzaaft:
Found while using antithesis/bombadil to torture tmux.


nicm:

PR #5482, opened 2026-08-10
WIP: Select the buffer affected by set-clipboard/OSC 52

Lattay:
Hello,

I would like tmux to use the primary selection instead of the clipboard because it matches the behavior of other software on linux: select with a mouse drag, paste with middle-click.

To make this possible in a compatible manner, I propose to introduce a new global option `clipboard-target` that could have the values `"clipboard", "primary", "selection"`. This maps easily to how OSC 52 works. Since the functions leading to the OSC 52 escape sequence already took a clip parameter, it was straightforward to make this option work.
This has been successfully tested in the foot terminal.

No AI code btw.

Now, this PR is a working proof of concept of what I am trying to achieve when it comes to tmux->primary (set-clipboard), unfortunately I am still figuring how the primary->tmux direction (get-clipboard).

Feedback welcome! (also if anything in here is wrong for the project please tell me, I am happy to rework).

### DIFF
--- a/cmd-load-buffer.c
+++ b/cmd-load-buffer.c
@@ -61,6 +61,7 @@ cmd_load_buffer_done(__unused struct client *c, const char *path, int error,
 	size_t				 bsize = EVBUFFER_LENGTH(buffer);
 	void				*copy;
 	char				*cause;
+	const char			*clip;
 
 	if (!closed)
 		return;
@@ -76,8 +77,10 @@ cmd_load_buffer_done(__unused struct client *c, const char *path, int error,
 			free(copy);
 		} else if (tc != NULL &&
 		    tc->session != NULL &&
-		    (~tc->flags & CLIENT_DEAD))
-			tty_set_selection(&tc->tty, "", copy, bsize);
+		    (~tc->flags & CLIENT_DEAD)) {
+			clip = window_copy_clipboard_target();
+			tty_set_selection(&tc->tty, clip, copy, bsize);
+		}
 		if (tc != NULL)
 			server_client_unref(tc);
 	}
--- a/cmd-refresh-client.c
+++ b/cmd-refresh-client.c
@@ -182,6 +182,7 @@ cmd_refresh_client_exec(struct cmd *self, struct cmdq_item *item)
 	const char		*errstr;
 	u_int			 adjust;
 	struct args_value	*av;
+	const char			*clip;
 
 	if (args_has(args, 'c') ||
 	    args_has(args, 'L') ||
@@ -235,7 +236,8 @@ cmd_refresh_client_exec(struct cmd *self, struct cmdq_item *item)
 	}
 
 	if (args_has(args, 'l')) {
-		tty_clipboard_query(&tc->tty);
+		clip = window_copy_clipboard_target();
+		tty_clipboard_query(&tc->tty, clip);
 		return (CMD_RETURN_NORMAL);
 	}
 
--- a/cmd-set-buffer.c
+++ b/cmd-set-buffer.c
@@ -61,6 +61,7 @@ cmd_set_buffer_exec(struct cmd *self, struct cmdq_item *item)
 	char			*bufname = NULL, *bufdata = NULL, *cause = NULL;
 	const char		*olddata;
 	size_t			 bufsize = 0, newsize;
+	const char		*clip;
 
 	if (args_get(args, 'b') != NULL) {
 		bufname = xstrdup(args_get(args, 'b'));
@@ -124,8 +125,10 @@ cmd_set_buffer_exec(struct cmd *self, struct cmdq_item *item)
 		cmdq_error(item, "%s", cause);
 		goto fail;
 	}
-	if (args_has(args, 'w') && tc != NULL)
- 		tty_set_selection(&tc->tty, "", bufdata, bufsize);
+	if (args_has(args, 'w') && tc != NULL) {
+		clip = window_copy_clipboard_target();
+		tty_set_selection(&tc->tty, clip, bufdata, bufsize);
+	}
 
 	return (CMD_RETURN_NORMAL);
 
--- a/input.c
+++ b/input.c
@@ -3587,6 +3587,7 @@ input_add_request(struct input_ctx *ictx, enum input_request_type type, int idx)
 	struct client		*c = NULL, *loop;
 	struct input_request	*ir;
 	char			 s[64];
+	const char		*clip;
 
 	if (wp == NULL)
 		return (-1);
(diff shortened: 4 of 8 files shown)


mgrant0:
I use osc52 to go from tmux -> local, but in the other direction local -> tmux, I shift-rightclick in putty or ctrl-shift-V in windows, or xterm I think maybe just right-click as long as it's not sent to tmux.  osc52 only works in one direction.  The other option you could do is to run xsel or xclip and enable X11 forwarding (ssh -X).  Honestly I found that more cumbersome to set up than just using shift-ctrl-V.

I haven't tried this, but are you trying to get tmux so that if something is selected, it goes to the local host via osc52 immediatly?  What about trying this instead of your PR?

```
bind-key -T copy-mode MouseDragEnd1Pane send-keys -X copy-selection-no-clear
```

This works for drag end, you'd also need some things for double-click and triple-click too but that's easy enough to add.

Lattay:
> I use osc52 to go from tmux -> local, but in the other direction local -> tmux, I shift-rightclick in putty or ctrl-shift-V in windows, or xterm I think maybe just right-click as long as it's not sent to tmux.

Well my point is that I would appreciate not to have to change my muscle memory depending of if I am clicking in tmux or elsewhere.

> osc52 only works in one direction.

Do you mean in tmux specifically? Because in general, OSC52 can be used in both directions (neovim does that for example).

> The other option you could do is to run xsel or xclip and enable X11 forwarding (ssh -X).

I use various machines, that I do not always control, and connect to various servers that I rarely control. So while these solutions work in some cases (I currently use xsel/wl-copy in local tmux to have the primary set), it's not an ideal solution in my opinion.

While there are workarounds, it looks to me like all of the machinery is present in tmux already and very little work is missing to make it work in a more general fashion.

OSC-52 is very appealing for tmux compared to xsel/xclip/wl-paste because:
1. it's bidirectional (xsel and xclip are, but wl-paste/wl-copy is a pair)
2. it does not require external binary
3. it is supported by all the good terminals that are required to have tmux works at full potential anyway (kitty, ghosty, foot, st, alacritty...)
4. it is already implemented for the pass-through behavior (using `"*yy` in neovim with OSC-52 enabled will copy the current line to the primary buffer even through tmux+ssh)
5. it works through ssh even without X forwarding

Anyway, I'll make this work and use it. If it's unsuitable for the project I'll just keep my fork.

Lattay:
Looking at the problem carefully, I think this PR might already been done, not because it solves my problem completely, but because it solves the part of the problem that can be solved without introducing a bigger feature to tmux:

when it comes to pasting, the current state of this PR already handles querying the buffer in `refresh-client -l`. Now I would like basically to be able to do that refresh automatically on every `paste` call, but as this is an async problem, it might be out of scope.

What do you think?
It seems something following the pattern of `tty_clipboard_query_callback` may be the way forward...

nicm:
The empty string means terminal default which in xterm is `s 0`. The meaning of `s` can be configured with the `selectToClipboard` resource. Are you sure the terminal you are using does not have a similar option? You are not alone in preferring CLIPBOARD nowadays.

I think if you want `refresh-client -l` to automatically paste, I would either make it accept a command to fire when the clipboard is received. Or (perhaps better) simply fire an event you can then use to trigger `paste-buffer` with a hook. The latter may need to have the clipboard client as payload (or its active pane at the time the request was sent) so you can paste in the right place (or you could just accept whatever is the active pane now - it may be unlikely you'd be able to change it rapidly enough without trying hard).

Lattay:
> The empty string means terminal default which in xterm is s 0. The meaning of s can be configured with the selectToClipboard resource.

As far as I can tell it's pretty common for terminal to actually hardcode empty string to clipboard. Here is the relevant line concerning foot: https://codeberg.org/dnkl/foot/src/commit/a635e0a196d9adbe1088d03e08d3e312206c640d/osc.c#L33

> I think if you want refresh-client -l to automatically paste

Actually I see it the other way around, I want `paste-buffer` to trigger `refresh-client -l` and way for it.

One partial solution to my problem is (on top of the changes I already made) to have this in my config:
```tmux
set -g set-clipboard on
set -g get-clipboard request
set -g clipboard-target primary
set -g focus-event on
set-hook -g pane-focus-in 'refresh -l'
# that next one is default
# bind-key -T root MouseDown2Pane select-pane -t = \; if-shell -F "#{||:#{pane_in_mode},#{mouse_any_flag}}" { send-keys -M } { paste-buffer -p }
```

This is not perfect, as if the pane-focus-in is caused by a middle click, the paste races the refresh, but it is partially emulating the intended behavior.

nicm:
I think the meaning of the default is something that should be configured in the terminal. I don't think we should add an option for this.

What happens if you define `Ms` to push `%p1` but never pop it? So something like `\E]52;%p1p;%p2%s\a`. Does that work?

If not, then maybe we should have tmux try say `Ms1` if it exists before `Ms`. `Ms1` could accept only `%p1` with the data.

Lattay:
> I think the meaning of the default is something that should be configured in the terminal.

I see that my initial proposal was changing tmux default behavior, I fixed that. Now the default is an empty string again.

> What happens if you define Ms to push %p1 but never pop it? So something like \E]52;%p1p;%p2%s\a. Does that work?

> If not, then maybe we should have tmux try say Ms1 if it exists before Ms. Ms1 could accept only %p1 with the data.

Honestly, I am not sure if I understand what you are saying. Are you proposing to introduce an extra terminal capability that can recycle the parameter of the query at the response time?

nicm:
I'm saying instead of adding a new `clipboard-target` option you should try to do this by changing the `Ms` capability and if you can't (because there is no way to avoid including the first parameter) then it may be better to add an `Ms1` capability.

Lattay:
I apologize, but I really don't understand how a new, more restrictive, capability would be better than using the existing one with the parameters it already has. The changes I made are just about setting that existing parameter to a user-defined value in the few cases where tmux previously hard-coded the value.

That being said, if the former is what you'd rather have in tmux, I am not the right person to implement it because I don't know anything about terminal capabilities. So I will just close this now.

nicm:
The reason is that it seems better to use the existing `terminal-overrides` option to do this rather than adding a separate option to change just one capability. It is unfortunate that `tiparm_s` requires both parameters to exist or it would be trivial, that's why I'm wondering if there is a way to do it anyway, by for example including `%p1` to push the first parameter but then not actually using it. But I'm not able to test until next week and if you're not interested in carrying on with this then it is irrelevant for the moment anyway.

As for adding a new capability, there is nothing magic about `Ms`, I invented it to let tmux set the clipboard, there is no reason we can't add a new one for a different purpose (even if it isn't *very much* different).

Lattay:
I think I am starting to understand where you want to go, but for me `terminal-overrides` are just opaque incantations, so I'd rather not deal with them at all.

nicm:
Not doing anything at all with parameter 1 doesn't work, but if you insert it with a field width of zero it is enough to get past the `tiparm_s` checks without actually inserting the parameter. So this works for me:
~~~~
set -s terminal-overrides '*:Ms=\E]52;%p1%0sp;%p2%s\a'
~~~~

PR #5485, opened 2026-08-11
tty: detect synchronized updates

mgrant0:
Query `sync` terminal feature from terminal and enable if terminal has this feature.


nicm:
Applied to OpenBSD with some minor changes, thanks!

PR #5486, opened 2026-08-11
tty: allow terminal features to be disabled

mgrant0:
Allow terminal features to be disabled like:

set -as terminal-features ',xterm-256color:-sync:-sixel'

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -16,6 +16,7 @@ concurrency:
 jobs:
   action:
     runs-on: ubuntu-latest
+    if: github.repository == 'tmux/tmux'
     steps:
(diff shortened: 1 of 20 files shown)


nicm:
I think this syntax should be `sixel@` to match `terminal-capabilities`.

mgrant0:
Use an '@' suffix to disable the feature like this:
 
set -as terminal-features ',xterm-256color:sync@:sixel@'

The man page now says that trailing @ disables a feature even if it is in terminfo or auto-detected, and explicitly notes this matches terminal-overrides.

nicm:
I don't like the way this makes `tty-features.c` reach back into the client, features are a property of the `struct tty_term`.

PR #5488, opened 2026-08-12
Fix #5453: log unknown client messages

### LINKED ISSUE #5453: Client hangs forever with no output when the server sends an unrecognised message type
### Issue description

If a tmux client receives an imsg whose type it does not recognise, it silently discards the message and keeps waiting. Nothing is printed, no error status is returned, and there is no timeout, so the client blocks forever. The same happens if the header announces a payload length that never fully arrives.

This is the residual case of #2189 / #2659. The version-mismatch hang described there is fixed, because `peer_check_version()` in `proc.c` replies with `MSG_VERSION` and `client_dispatch_wait()` handles it. But a message with a matching protocol version and an unhandled type falls off the end of the `switch` in `client_dispatch_wait()`, which has no `default:` label, so the client is left in the same untimed silent wait.

#### Steps to reproduce

A trivial stand-in server is enough to trigger it. Save this as `fake.py`:

```python
import os, socket, struct, sys, time

PROTOCOL_VERSION = 8
IMSG_HEADER_SIZE = 16          # type, len, peerid, pid (4 x uint32, native)
BOGUS_TYPE = 999               # not a member of enum msgtype

path, mode = sys.argv[1], sys.argv[2]
try:
    os.unlink(path)
except FileNotFoundError:
    pass

s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.bind(path)
s.listen(1)
c, _ = s.accept()

if mode == "close":                 # control case
    c.close()
elif mode == "unknown-type":        # valid header, unhandled type
    c.sendall(struct.pack("=IIII", BOGUS_TYPE, IMSG_HEADER_SIZE,
                          PROTOCOL_VERSION, os.getpid()))
elif mode == "short-payload":       # announces 4096 bytes, sends none
    c.sendall(struct.pack("=IIII", BOGUS_TYPE, IMSG_HEADER_SIZE + 4096,
                          PROTOCOL_VERSION, os.getpid()))

time.sleep(600)
```

Then, for each mode:

```sh
python3 fake.py /tmp/sock <mode> &
sleep 1
tmux -S /tmp/sock kill-server; echo "exit: $?"
```

Results:

| mode | behaviour |
| --- | --- |
| `close` | exits in under a second, status 1, prints `server exited unexpectedly` |
| `unknown-type` | still running after 60s, status never returned, zero bytes on stdout and stderr |
| `short-payload` | still running after 60s, status never returned, zero bytes on stdout and stderr |

The `close` case shows the client does detect a dead peer correctly, so the problem is specific to a peer that stays connected but says something the client does not understand. The hung client also does not respond to being backgrounded or to a subsequent server start, and has to be killed.

Any command behaves the same way; `kill-server` is just short. In a script such as `tmux -S "$sock" has-session` this is worse than an error, because the caller blocks indefinitely rather than failing.

#### Cause

`client_dispatch_wait()` in `client.c:622` dispatches on `imsg->hdr.type` with no `default:` case:

```c
	switch (imsg->hdr.type) {
	case MSG_EXIT:
	case MSG_SHUTDOWN:
	...
	case MSG_OLDSTDERR:
	case MSG_OLDSTDIN:
	case MSG_OLDSTDOUT:
		fprintf(stderr, "server version is too old for client\n");
		proc_exit(client_proc);
		break;
	}
}
```

An unhandled type therefore returns to the event loop with no state change, and since nothing else will ever wake the client it waits forever. `client_dispatch_attached()` at `client.c:718` has the same shape.

The short-payload variant never reaches the dispatcher at all: `imsgbuf_read()` keeps the partial message buffered waiting for the announced length, and the client keeps polling.

Treating an unrecognised type the way `MSG_OLDSTDERR` is already treated, that is printing a message and calling `proc_exit(client_proc)`, would seem to cover it. Even just logging and exiting non-zero would be better than blocking.

### Required information

* tmux version (`tmux -V`): `tmux next-3.8`, built from Git master `31dccb6bc9521b0ea46307974d071ad7f09f0e9b` (`./configure --enable-debug --disable-jemalloc`)
* Platform (`uname -sp`): `Darwin arm` (macOS 26.6, arm64)
* Terminal in use: not relevant, reproduces from any terminal and from a non-interactive shell; verified with `TERM=xterm-256color`
* `$TERM` *inside* tmux: not applicable, the client never attaches
* `$TERM` *outside* tmux: `xterm-256color`
* Logs from tmux (`-vv`, `unknown-type` mode, identify messages elided):

```
client started (8740): version next-3.8, socket /tmp/sock, protocol 8
using libevent 2.1.13-stable select
flags are 0x10000
socket is /tmp/sock
trying connect
add peer 0x10529a580: 6 (0x0)
sending message 111 to peer 0x10529a580 (8 bytes)
...
sending message 106 to peer 0x10529a580 (0 bytes)
cmd_pack_argv: argv[0]=kill-server
sending message 200 to peer 0x10529a580 (16 bytes)
client loop enter
peer 0x10529a580 message 999
```

The log ends there. `peer ... message 999` is the last line ever written, and no `client loop exit` follows; the process was still alive 60 seconds later and had to be killed. The `short-payload` mode produces the same log without even the `message 999` line.

MsfPablo:
Fixes #5453

When the client receives an unknown message type, it currently silently
waits. This adds a debug log so unknown messages are visible at debug
level while keeping the client alive for forward compatibility (as
requested by the maintainer @nicm).

Both dispatch switch statements (client_dispatch_wait and
client_dispatch_attached) now have a default case calling log_debug
with the unknown message type.

### DIFF
--- a/client.c
+++ b/client.c
@@ -712,6 +712,10 @@ client_dispatch_wait(struct imsg *imsg)
 		fprintf(stderr, "server version is too old for client\n");
 		proc_exit(client_proc);
 		break;
+	default:
+		log_debug("%s: unknown message type %#x", __func__,
+		    imsg->hdr.type);
+		break;
 	}
 }
 
@@ -797,6 +801,10 @@ client_dispatch_attached(struct imsg *imsg)
 
 		system(data);
 		proc_send(client_peer, MSG_UNLOCK, -1, NULL, 0);
+		break;
+	default:
+		log_debug("%s: unknown message type %#x", __func__,
+		    imsg->hdr.type);
 		break;
 	}
 }
--- a/cmd-source-file.c
+++ b/cmd-source-file.c
@@ -50,6 +50,7 @@ const struct cmd_entry cmd_source_file_entry = {
 
 struct cmd_source_file_data {
 	struct cmdq_item	 *item;
+	struct client		 *c;
 	int			  flags;
 
 	struct cmdq_item	 *after;
@@ -61,9 +62,9 @@ struct cmd_source_file_data {
 };
 
 static enum cmd_retval
-cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
+cmd_source_file_complete_cb(struct cmdq_item *item, void *data)
 {
-	struct client	*c = cmdq_get_client(item);
+	struct client	*c = data;
 
 	if (c == NULL) {
 		cmd_source_file_depth--;
@@ -71,6 +72,7 @@ cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
 	} else {
 		c->source_file_depth--;
 		log_debug("%s: depth now %u", __func__, c->source_file_depth);
+		server_client_unref(c);
 	}
 
 	cfg_print_causes(item);
@@ -78,8 +80,9 @@ cmd_source_file_complete_cb(struct cmdq_item *item, __unused void *data)
 }
 
 static void
-cmd_source_file_complete(struct client *c, struct cmd_source_file_data *cdata)
+cmd_source_file_complete(struct cmd_source_file_data *cdata)
 {
+	struct client		*c = cdata->c;
 	struct cmdq_item	*new_item;
 	u_int			 i;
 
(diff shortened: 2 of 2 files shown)


nicm:
Applied a slightly different version of this, thanks!

PR #5489, opened 2026-08-12
Add default terminal features for Rio

raphamorim:
Rio replies to XTVERSION with `Rio x.y.z`. This adds it to the detected terminals with the features it implements, the same set as ghostty minus extkeys and margins.


aymanbagabas (approved):
Great to see Rio terminal here!

nicm:

raphamorim:
thank you!

PR #5496, opened 2026-08-14
vi: skip trailing whitespace after e/E word-end motion

### LINKED ISSUE #5491: `e` and `E` motions in vi copy mode do not cross line boundaries
### Issue description

`e` and `E` motions in vi copy mode do not cross line boundaries. In vi(1), pressing `e` at the last word of a line advances to the end of the first word on the next line. In tmux, the cursor stops and does not move.

`w`/`W` and `b`/`B` correctly cross line boundaries. Only `e`/`E` (next-word-end / next-space-end) are affected.

**Steps to reproduce:**

Minimal config:

```
set -g mode-keys vi
```

1. Have multiple lines of text visible in a pane:
   ```
   hello world
   foo bar baz
   ```
2. Enter copy mode
3. Position cursor on the `w` of "world" (last word of first line)
4. Press `e` — cursor moves to the `d` at end of "world" (correct)
5. Press `e` again

**Expected:** Cursor moves to the end of "foo" on the next line (matching vi behavior).

**Actual:** Cursor does not move. It stays stuck on the `d` of "world".

Pressing `w` from the same position correctly moves to "foo" on the next line. `b` also correctly traverses lines in the reverse direction.

**Root cause (from reading source):**

`grid_reader_cursor_next_word_end()` in `grid-reader.c` relies on `grid_reader_handle_wrap()` to cross lines, but `handle_wrap` only fires when `cx > xx`. At the last word of a line there is no trailing whitespace to push `cx` past `xx`, so the outer `while` loop exits without advancing to the next line.

By contrast, `grid_reader_cursor_next_word()` works because its whitespace-skip phase happens *after* passing the word boundary, naturally pushing `cx` past end-of-line. And `grid_reader_cursor_previous_word()` explicitly calls `grid_reader_cursor_up()` to cross lines.

### Required information

* **tmux version:** tmux 3.7b
* **Platform:** Darwin arm
* **Terminal:** Alacritty
* **$TERM inside tmux:** tmux-256color
* **$TERM outside tmux:** xterm-256color
* **Logs:** (attach output from `tmux kill-server; tmux -vv new`, reproduce the issue, then provide the log from `/tmp/tmux-*/`)

MsfPablo:
Fixes #5491.

grid_reader_cursor_next_word_end 'return's from the inner
do/while loops without continuing the outer wrap loop. Its sibling
grid_reader_cursor_next_word breaks out of the inner loops and then
runs a trailing WHITESPACE-skip loop; this function should match that
pattern. Without the trailing skip, e/E in vi copy-mode land on the
first character of the next word rather than on the last character of
the current one whenever the next word is on a fresh, wrapping line.

Fall through from the inner branches with 'break' instead of 'return',
preserve the outer wrap guard, and add the trailing WHITESPACE-skip
loop. Declare 'width' locally for the skip loop.

AI-assisted: implementation, commit message, and PR text prepared with AI assistance.

### DIFF
--- a/grid-reader.c
+++ b/grid-reader.c
@@ -234,7 +234,7 @@ grid_reader_cursor_next_word(struct grid_reader *gr, const char *separators)
 void
 grid_reader_cursor_next_word_end(struct grid_reader *gr, const char *separators)
 {
-	u_int	xx, yy;
+	u_int	xx, yy, width;
 
 	/* Do not break up wrapped words. */
 	if (grid_get_line(gr->gd, gr->cy)->flags & GRID_LINE_WRAPPED)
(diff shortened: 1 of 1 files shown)


nicm:
I don't think this is the right place to fix this.

PR #5511, opened 2026-08-20
Use correct X position when combining characters

frosas:
## Context

With `variation-selector-always-wide` enabled, a variation-selector sequence written below the first row of a right-hand pane can be replaced by a space. For example, `A✏️B` becomes `A B`.

## Problem

`screen_write_combine()` passes a pane-local X position to `window_visible_ranges()`, which expects window coordinates. In a pane with a nonzero X offset, the visibility check can therefore inspect another pane and treat the combined character as obscured.

## Solution

Add the pane X offset when checking the combined character's visible range, matching the coordinate conversion already used by `screen_write_cell()`.

Add an attached-client regression test with a horizontally split window. The test fails on unmodified `master` with:

```
line 2: expected 'A✏️B', got 'A B'
```

## Testing

- The new regression passes with the fix.
- Focused checks preserve `ℹ️`, `⚠️`, and `✏️` in a right-hand pane.
- The full regression suite passes except `input-replies.sh` and `screen-redraw-menus.sh`; both failures reproduce on unmodified `master`.

## AI assistance

This issue was diagnosed and the patch was developed with OpenAI/OpenCode assistance. I reviewed the change and verified it locally.

### DIFF
--- a/regress/tty-draw-line.sh
+++ b/regress/tty-draw-line.sh
@@ -105,6 +105,21 @@ sleep 1
 CLIENT=$($TMUX2 list-clients -F '#{client_name}' | head -1)
 [ -n "$CLIENT" ] || fail "no inner client"
 
+# A variation selector which widens the previous character must use window
+# coordinates when checking visibility in a pane with a nonzero x offset.
+$TMUX2 set -s variation-selector-always-wide on || exit 1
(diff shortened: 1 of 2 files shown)


nicm:

This shows the problem:
~~~~
./tmux new \; set -s variation-selector-always-wide on \; splitw -h 'printf "\nA✏️B"; sleep 100'
~~~~

PR #5516, opened 2026-08-23
Redraw damage rectangles

mgrant0:
1. layout — only mark scrollbar for redraw when the pane actually changed
  2. screen-redraw, screen-write — damage-rectangle tracking and composition
  3. window — floating-pane damage/redraw helpers
  4. server-client — consume window damage during the normal redraw pass
  5. window — redraw only borders/status on active-pane change
  6. cmd-join-pane, cmd-split-window, cmd-resize-pane — use shared floating-pane redraw
  7. popup — damage-based redraw during drag/resize


darlingm:
Thanks for working on this. I am exploring composite views of live panes that remain in their existing sessions, so the narrower redraws in this PR are useful for that work.

I tested the current head (`f66eeef8a`) and pushed four follow-up commits here:

https://github.com/tmux/tmux/compare/redraw-damage-rectangles...darlingm:pr5516-regression-fixes

They add regression coverage for:

- damage-only redraw passes
- full and regional screen-write fallbacks
- popup move and resize damage
- damage shared by multiple clients

They also make two related changes.

First, the fallback added in `f66eeef8a` covers a case that already occurs, rather than only a possible future case. `screen_write_redraw_cb()` can record window damage without setting client redraw flags. More generally, damage belongs to the window while redraw flags belong to each client. The paired regression test reproduces this with a panned client. The follow-up keeps the same behavior but handles damage once after the flag-specific redraw work.

Second, the screen-write callback currently records only the cursor row even when the failed operation needs a full or regional redraw. The follow-up passes the affected row range to the callback. This fixes incomplete alternate-screen restoration and the other full and regional fallback paths.

All four tests pass on the follow-up branch.

I have not opened another PR yet. I can open a companion PR against `redraw-damage-rectangles`, or you can cherry-pick or fold the commits into this branch as you prefer.

darlingm:
Separate integration note: the current conflict with master is in `window_set_active_pane()`.

I think the combined logic should track whether unzoom actually occurred:

```c
int		    did_unzoom = 0;

/* ... */

if ((w->flags & WINDOW_ZOOMED) && !window_pane_is_visible(wp)) {
	window_unzoom(w, 1);
	did_unzoom = 1;
}

/* ... */

if (did_unzoom)
	server_redraw_window(w);
else {
	server_redraw_window_borders(w);
	server_status_window(w);
}
```

Master preserves zoom when switching to a visible `-A` or modal floating pane. This PR needs a full redraw only when unzoom changed the pane geometry. Tracking the actual unzoom preserves both behaviors.

I left this resolution out of the follow-up branch so its diff remains limited to the tests and fixes above.

darlingm:
I need to correct my earlier note. The first four follow-up commits were valid, but they were not complete. I had already reproduced five more regressions and failed to carry them into the branch or comment.

I updated the same companion branch:

https://github.com/tmux/tmux/compare/redraw-damage-rectangles...darlingm:pr5516-regression-fixes

It now adds five more fix-and-test commits for:

- switching windows within the same session without redrawing the client
- damage edges splitting double-width characters
- damage redraw replacing pane prompts with pane content
- popup movement leaving old borders on bottom or multi-line top status lines
- floating-pane geometry changes leaving status formats stale

Each new test passes current master (`e5db7598b`) and fails `f66eeef8a` plus the original four follow-up commits. The fixes are folded into the corresponding test commits. All nine focused tests pass on the updated branch.

I preserved the exact four-commit version I linked earlier as [`pr5516-regression-fixes-v1`](https://github.com/darlingm/tmux/tree/pr5516-regression-fixes-v1).

The branch still starts at `f66eeef8a`. It intentionally does not duplicate master’s existing `b2de80387` synchronized-output fix or resolve the master conflict; those should come from rebasing onto master. This keeps the companion diff limited to PR 5516 review findings.

Two design concerns are not included in this tested patch set: window-global deferred damage can make fast clients repeat work while another client is blocked, and pane-global `PANE_NEWSTATUS` can cache client-specific status content. Those need a separate design decision rather than a narrow fix.

mgrant0:
Thanks for your tests, I've included them all now.  I ran things through claude this morning.  There was an issue with redraw-multiclient.sh which was failing.  Here is what it said and did:

Summary of the change to redraw-multiclient.sh for your reply:

  ---

  Change: Switched the test to ASCII pane borders (pane-border-lines with plain ascii characters, e.g. -,|, and +) instead of the default UTF-8 box-drawing borders.

  Why: The original test reliably "failed" — but I (Claude) tracked it down and it wasn't a real bug. This test nests a real tmux
  client inside another tmux's pane to get a genuine terminal to capture from. That nested-tmux relay has a rendering
  artifact: when a cell that previously held a multi-byte UTF-8 border character (┌, ─, etc.) gets overwritten later by plain
  content, the outer tmux instance's own interpretation of that byte stream mis-renders it — independent of anything the inner
  tmux actually sends (verified the inner server's escape sequences byte-for-byte correct in both the passing and failing
  cases). Confirmed it's a harness artifact, not a tmux bug, by replaying the identical mouse-drag sequence against a real
  terminal (xterm) instead of nested tmux — never reproduces there. ASCII borders sidestep the multi-byte-overwrite path
  entirely, since +/-/| are single-byte.

  Mechanical changes needed to support that:
  - wait_outer_has ... '┌' → wait_outer_has ... '+'
  - assert_scene's frame-count check: grep -o '┌' | wc -l expecting 1 → grep -o '+' | wc -l expecting 4, since ASCII borders
    use the same + for every corner/junction (no distinct top-left glyph like ┌), so one rectangular frame always draws
    exactly 4 of them instead of 1.

  Still verified to catch the real bug: reintroduced the actual "damage consumed by only one client" defect (made the damage
  list get freed by whichever client draws first, instead of after all clients have had a turn) and confirmed the test fails
  reliably against that, with the exact symptom the test's own comment describes — so it still does its job, just without the
  nested-terminal false positive.

mgrant0:
All 9 regression tests now pass. Summary of what I did with each of darlingm's 4 fixes:

  Took essentially as-is (2 of 4):

  - switch-client-redraw fix (server-client.c + new redraw_client_has_window() in screen-redraw.c/tmux.h) — this is a real bug
    in code I wrote: old->curw != s->curw is always false when old == s, since both sides read the same already-mutated
    field. Darlingm's fix (compare against the client's own cached scene instead) is the correct, minimal way to sidestep that
    aliasing. No changes made.
  - popup-drag-status-line fix (popup.c + two small screen-redraw.c tweaks) — reuses a flag (CLIENT_REDRAWSTATUSALWAYS) that already existed in the codebase, unused, with a name that exactly matches what's needed here. The clip-to-pane-area logic correctly generalizes to status-at-top/bottom/off, which the original popup_damage() never handled for the bottom case. Took it as-is.

  Took the approach, reimplemented simpler (1 of 4):

  - Wide-character clipping fix — darlingm's redraw_damage_expand_span() walks the grid cell-by-cell with a switch over span type (PANE/STATUS/MENU), fetching each type's own screen/offset. I replaced it with redraw_damage_grow_span_clip():
    unconditionally widen the clip by one cell on each edge, clamped to the span's own bounds — no grid lookup, no type
    switch, ~8 lines instead of ~50. This works because tmux's grid caps any character at 2 columns wide, so a flat ±1 margin
    is always sufficient to pull back a split character, regardless of what's actually there. I verified this isn't just
    theoretically sound — I disabled it, confirmed popup-drag-wide-character.sh fails with the exact original symptom (torn
    CJK glyphs), then re-enabled it and confirmed the fix.

  Took the approach, applied as-is with no changes (1 of 4):

  - Pane-prompt-after-damage fix (redraw_make_pane_prompt() extracted as a shared helper, redraw_damage_draw_pane_prompt() added) — this is a legitimate dedup with correct logic (the row-selection math exactly mirrors the existing full-redraw path, and it correctly respects overlay clipping). Nothing to improve here.

  All 4 previously-failing tests pass now (3/3 runs each), the other 5 still pass, and the two pre-existing floating-pane tests (floating-pane-drag-ghost.sh, floating-pane-drag-scrollbar-strip.sh) still pass.

nicm:
Try this:
~~~~
$ cat <<EOF >x.sh
left=$(./tmux -Lx new-window -P -F '#{pane_id}' 'printf "LEFT\n"; sleep 600')
./tmux -Lx split-window -h -t "$left" 'printf "RIGHT\n"; sleep 600'
./tmux -Lx set-option -w -t "$left" window-style 'bg=red'
./tmux -Lx set-option -w -t "$left" window-active-style 'bg=blue'
./tmux -Lx bind-key -n x attach-session -t "$left"
EOF
$ ./tmux -Lx new
$ sh x.sh # inside tmux
~~~~
Then press `x`. The active pane changes but the colours don't change. `C-b r` and they change.

mgrant0:
@nicm your x.sh test should be fixed now, thanks.

PR #5520, opened 2026-08-24
Fix memory leak when losing client

blmarket:
there's no reason to keep `path` not to be freed while we free title, cwd.

---

## Again, LLM generated reproduction script

```
#!/usr/bin/env bash
#
# Shows the leak of a lost client's path, fixed on branch
# fix-path-leak: whatever runs in a pane may report its directory with
# OSC 7, the active pane's path is copied into each client by
# server_client_set_path() when set-titles is on, and
# server_client_lost() freed every other string the client owns --
# title, set right beside path, included -- but not path, so the last
# one a client held went with it. One string leaks per lost client, and
# input-buffer-size lets the pane make it a megabyte.
#
# The repro: a pane announces a path with OSC 7, then N (default 10)
# clients in turn attach -- picking the path up on their first redraw --
# and detach. Under AddressSanitizer the exit report of an unfixed
# server holds
#
#     Direct leak of 400 byte(s) in 10 object(s) allocated from:
#         #2 server_client_set_path server-client.c:2607
#         #3 server_client_check_redraw server-client.c:2547
#
# stack passes through server_client_set_path. Other leaks the server
# may have are ignored, so the answer is to this defect and not to
# whatever else the exit report happens to hold.
#
# Measured at N=10:
#
#   c1f947a3 (origin/master)            10 objects   LEAKING
#
#
#
# Usage: ./demo-path-leak.sh [asan-tmux-binary]  (default: $BIN, else ./tmux)
#        N=30 ./demo-path-leak.sh ./tmux

set -u

BIN=${1:-${BIN:-./tmux}}
N=${N:-10}

	exit 2
}

	exit 2
fi

DIR=$(mktemp -d "${TMPDIR:-/tmp}/dcp.XXXXXX") || exit 2
SOCK=$DIR/s
	exit 2
fi

CLIENT_PID=
cleanup() {
	rm -rf "$DIR"
}
trap cleanup EXIT

# The pane announces its path with OSC 7 and stays; every client that
# attaches later finds it on the active pane and copies it.
tm -f /dev/null new-session -d -x 80 -y 24 \
    "printf '\\033]7;file://host/leaked/by/every/lost/client\\033\\\\'; sleep 600" \
    2>"$DIR/start.err" || {
	cat "$DIR/start.err" >&2
	exit 2
}

# server_client_set_path only runs when titles are being set.
tm set-option -g set-titles on
tm set-option -g status off

# Wait for OSC 7 to have been parsed into the pane.
for i in $(seq 50); do
	[ -n "$(tm display-message -p '#{pane_path}' 2>/dev/null)" ] && break
	sleep 0.1
done
[ -n "$(tm display-message -p '#{pane_path}')" ] || {
	echo "pane path was not set" >&2
	exit 2
}

printf 'clients  %s attached and lost, each holding "%s"\n\n' "$N" \
    "$(tm display-message -p '#{pane_path}')"

# Each round attaches a client on a terminal from script(1), lets it
# draw once so server_client_set_path runs for it, and detaches it, so
# server_client_lost sees a client whose path is set.
for i in $(seq "$N"); do
	CLIENT_PID=$!
	for j in $(seq 50); do
		sleep 0.1
	done
	tm refresh-client 2>/dev/null
	sleep 0.2
	for j in $(seq 50); do
		sleep 0.1
	done
	CLIENT_PID=
done

tm kill-server 2>/dev/null
for i in $(seq 100); do
	sleep 0.1
done
	exit 2
fi

LOG=$DIR/asan.$SERVER_PID

# through server_client_set_path -- the copy this defect never freed.
awk -v RS= -v ORS='\n\n' '/Direct leak/ && /server_client_set_path/' \
    "$LOG" >"$DIR/hit"

if [ -s "$DIR/hit" ]; then
	cat "$DIR/hit"
	printf 'LEAKING: %s lost client(s) took their path string with them.\n' \
	    "$objects"
	exit 1
fi
printf 'clean: losing a client frees its path.\n'
exit 0
```

### DIFF
--- a/server-client.c
+++ b/server-client.c
@@ -524,6 +524,7 @@ server_client_lost(struct client *c)
 	input_cancel_requests(c);
 
 	free(c->title);
+	free(c->path);
 	free((void *)c->cwd);
 
 	evtimer_del(&c->repeat_timer);


nicm:

PR #5523, opened 2026-08-24
Fix memory leak of collected items left on a screen

blmarket:
Simply speaking, it's not guaranteed that `screen_write_citem` does not remain in screen->write_list[]->items. Currently they are simply forgotten, resulting memory leak.

This can happen only when `screen_write_collect_flush_line` fail to flush, which can happen when screen width is zero (and few more cases).

This commit fix by iterate remaining items, which is safe no-op when items is empty.

---

## LLM generated reproduction script

```
#!/usr/bin/env bash
#
# Shows the leak of collected items left on a screen, fixed on branch
# fix-screen-items-leak: screen_write_collect_flush_line() writes out
# the part of each collected item that is visible and leaves the rest
# collected, so a line can still hold items when the screen is freed --
# there is nothing in front of the flush that promises it emptied the
# list. screen_write_free_list() freed the lines' text and the array and
# dropped whatever was still on them.
#
# A zero-width screen is the easy way to see it: format_draw() sizes its
# eight screens by the length of the string, so drawing an empty format
# gives each one the clear item screen_write_clearendofline() collected
# and no column to write it in.
#
# The repro: a client attaches with status-format[0] set to the empty
# string and the status is redrawn N (default 10) times. Under
# AddressSanitizer the exit report of an unfixed server holds
#
#     Direct leak of 384 byte(s) in 4 object(s) allocated from:
#         #2 screen_write_get_citem screen-write.c:73
#         #3 screen_write_init screen-write.c:358
#         #4 screen_write_start screen-write.c:401
#         #5 format_draw format-draw.c:748
#
# stack passes through both screen_write_get_citem and format_draw.
# Other leaks the server may have are ignored, so the answer is to this
# defect and not to whatever else the exit report happens to hold.
#
# The count is not N: the free list is reused between draws, so what is
# lost is the items left on the screens, not one set per redraw.
#
# Measured at N=10:
#
#   c1f947a3 (origin/master)                     7 objects   LEAKING
#   + the sibling unterminated-style fix, alone  7 objects   LEAKING
#   + this fix                                   0 objects   clean
#
# The middle row is the check that this is the leak of items left on a
# screen and not the sibling leak of an unterminated style: there is no
# unterminated style here, format_draw() runs to its ordinary end, and
# the items are still lost when the screens are freed.
#
#
#
# Usage: ./demo-screen-items-leak.sh [asan-tmux-binary]  (default: $BIN, else ./tmux)
#        N=30 ./demo-screen-items-leak.sh ./tmux

set -u

BIN=${1:-${BIN:-./tmux}}
N=${N:-10}

	exit 2
}

	exit 2
fi

DIR=$(mktemp -d "${TMPDIR:-/tmp}/dsl.XXXXXX") || exit 2
SOCK=$DIR/s
	exit 2
fi

CLIENT_PID=
cleanup() {
	rm -rf "$DIR"
}
trap cleanup EXIT

	cat "$DIR/start.err" >&2
	exit 2
}

# The whole line has to be empty, not just status-left: format_draw()
# sizes its screens by the string it is given, and only an empty string
# leaves them no column to flush an item into.
tm set-option -g status on
tm set-option -g 'status-format[0]' ''

CLIENT_PID=$!
for i in $(seq 50); do
	sleep 0.1
done
	echo "no client attached" >&2
	exit 2
}

printf 'status   an empty format redrawn %s times\n\n' "$N"

for i in $(seq "$N"); do
	tm refresh-client -S 2>/dev/null
	sleep 0.15
done

for i in $(seq 50); do
	sleep 0.1
done
CLIENT_PID=

tm kill-server 2>/dev/null
for i in $(seq 100); do
	sleep 0.1
done
	exit 2
fi

LOG=$DIR/asan.$SERVER_PID

# The verdict is only the Direct leak blocks of collected items taken
# under format_draw -- the ones this defect dropped with the screen.
awk -v RS= -v ORS='\n\n' \
    '/Direct leak/ && /screen_write_get_citem/ && /format_draw/' \
    "$LOG" >"$DIR/hit"

if [ -s "$DIR/hit" ]; then
	cat "$DIR/hit"
	printf 'LEAKING: %s collected item(s) were dropped with the screen.\n' \
	    "$objects"
	exit 1
fi
printf 'clean: freeing a screen hands its collected items back.\n'
exit 0
```

### DIFF
--- a/screen-write.c
+++ b/screen-write.c
@@ -338,10 +338,18 @@ screen_write_make_list(struct screen *s)
 void
 screen_write_free_list(struct screen *s)
 {
-	u_int	y;
+	struct screen_write_cline	*cl;
+	struct screen_write_citem	*ci, *ci1;
+	u_int				 y;
 
-	for (y = 0; y < screen_size_y(s); y++)
-		free(s->write_list[y].data);
+	for (y = 0; y < screen_size_y(s); y++) {
+		cl = &s->write_list[y];
+		TAILQ_FOREACH_SAFE(ci, &cl->items, entry, ci1) {
+			TAILQ_REMOVE(&cl->items, ci, entry);
+			screen_write_free_citem(ci);
+		}
+		free(cl->data);
+	}
 	free(s->write_list);
 }
 


nicm:

PR #5532, opened 2026-08-25
security(workflows): pin third-party Actions to a commit SHA

andife:
## Summary
- Pin `dessant/lock-threads@v6`, `actions/checkout@v4`, and `actions/upload-artifact@v4` in `.github/workflows/{lock,regress}.yml` to their current tag's full commit SHA, keeping the tag as a trailing comment.
- Tags are mutable — a compromised maintainer account (or a supply-chain attack on the action's own repo, as happened to `tj-actions/changed-files` in 2025) can retarget a tag your workflow already trusts, silently changing what code runs with that job's token.
- `lock.yml` runs weekly with an `issues: write`, `pull-requests: write`, `discussions: write` token, so a hijacked `lock-threads` tag would have real reach even though the workflow never runs on untrusted PR code.

## Test plan
- [x] Both modified files parse as valid YAML.
- [x] Diffed to confirm only the three `uses:` lines changed.

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'tmux/tmux'
     steps:
-      - uses: dessant/lock-threads@v6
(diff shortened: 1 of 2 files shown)


nicm:
Applied, thanks!

PR #5549, opened 2026-08-29
cmd-resize-pane: fix use-after-free when resizing a zoomed pane

Uzaaft:
cmd_resize_pane_exec reads the target pane's layout cell into a local variable at the top of the function, then calls server_unzoom_window. For a zoomed window that frees the entire zoomed layout tree — including that cell — and restores the saved layout. The stale cell is then dereferenced again near the end of the function when fixing up offsets, a use-after-free.

Re-read the layout cell from the pane after unzooming. Reproduce by splitting a window, zooming it (resize-pane -Z), then resize-pane -x. Found with AddressSanitizer, driven by Bombadil.

### DIFF
--- a/cmd-resize-pane.c
+++ b/cmd-resize-pane.c
@@ -93,6 +93,12 @@ cmd_resize_pane_exec(struct cmd *self, struct cmdq_item *item)
 	}
 	server_unzoom_window(w);
 
+	/*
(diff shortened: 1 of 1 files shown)


nicm:
Applied to OpenBSD now (although I made the comment shorter), will be in GitHub later. Thanks!

PR #5554, opened 2026-08-30
Fix memory leak during command parse

blmarket:
`cmds` is local variable and should be freed before return.

---

## LLM generated repro script

```
#!/usr/bin/env bash

set -u

BIN=${1:-${BIN:-./tmux}}
N=${N:-20}

        exit 2
}

        exit 2
fi

DIR=$(mktemp -d "${TMPDIR:-/tmp}/dapl.XXXXXX") || exit 2
SOCK=$DIR/s
        exit 2
fi

cleanup() {
        rm -rf "$DIR"
}
trap cleanup EXIT

tm -f /dev/null new-session -d -s alias-leak 'sleep 300' 2>"$DIR/start.err" || {
        cat "$DIR/start.err" >&2
        exit 2
}

tm set-option -s 'command-alias[100]' 'leak=display-message -p ALIAS' || {
        echo "could not install command alias" >&2
        exit 2
}
for i in $(seq "$N"); do
        tm leak >/dev/null || {
                echo "alias invocation failed at iteration $i" >&2
                exit 2
        }
done

printf 'commands %s of command-alias expansion\n\n' "$N"

tm kill-server 2>/dev/null
for i in $(seq 100); do
        sleep 0.1
done
        exit 2
fi

LOG=$DIR/asan.$SERVER_PID
[ -s "$LOG" ] || : >"$LOG"
awk -v RS= -v ORS='\n\n' \
        '/(Direct|Indirect) leak/ && /cmd_parse_expand_alias/' \
        "$LOG" >"$DIR/hit"

if [ -s "$DIR/hit" ]; then
        cat "$DIR/hit"
        printf 'LEAKING: %s allocation(s) came from command-alias expansion.\n' \
                "$objects"
        exit 1
fi
printf 'clean: no command-alias expansion leak was reported.\n'
exit 0
```

### DIFF
--- a/cmd-parse.y
+++ b/cmd-parse.y
@@ -802,6 +802,7 @@ cmd_parse_expand_alias(struct cmd_parse_command *cmd,
 	if (last == NULL) {
 		pr->status = CMD_PARSE_SUCCESS;
 		pr->cmdlist = cmd_list_new();
+		cmd_parse_free_commands(cmds);
 		return (1);
 	}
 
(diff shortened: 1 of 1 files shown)


nicm:

Issue #1019, opened 2017-07-25
Question / Feature Request: Ability to disable the dropping of output when there is a lot of output being outputted? 


heymoe:
I recently noticed that my ability to cat a somewhat large text file so I can cut-n-paste it elsewhere has stopped working. Basically if a text file contains say 1000 lines of text for example and I cat it and then scroll back in my terminal so I can select the output and then cut-n-paste it somewhere else, I end up with a lot few lines then the actual text file contained.

I'm currently running tmux v2.5 and noticed in the change log these 2 changes:

> * Handle slow terminals and fast output better: when the amount of data
>   outstanding gets too large, discard output until it is drained and we are
>   able to do a full redraw. Prevents tmux sitting on a huge buffer that the
>   terminal will take forever to consume.
> 
> * Do not redraw a client unless we realistically think it can accept the data -
>   defer redraws until the client has nothing else waiting to write.

To see if that is what broke my ability to cut-n-paste, I compiled tmux v2.4 and had to the same results. I didn't see anything in the change log for v2.4 that might be causing this except for maybe the general:

> * There have been substantial performance improvements.

Any ways, I then compiled v2.3 since I know I was able to do this at some point and it started working again.

Based on the v2.5 change log entries I noted above, I totally get why the dropping of output make sense. Trust me, there have been many times where I catted a file or ran some other command that had huge amounts of output and I just had to walk away until my Ctrl-C finally stopped it or the command finally finished. That being said, there are cases where you do want to do something that outputs a lot of data that you want to see completely like my cut-n-paste scenario. 

So my question is, is there a way to disable the  dropping of output that I can enable when needed or via the command line in versions greater than v2.3? I did look at the man pages and didn't see any command options jump out at me that would do this.

If not, may I submit a feature request where this feature can be turned off or maybe better be able to define how many lines of output can be fully displayed during a fast output scenario before the dropping starts.

Thanks for your time.

nicm:
It is impossible for tmux to guarantee the consistency of the terminal scrollback and we are not going to try and do so with half measures.

nicm:
The recommended way to copy and paste from tmux is to use copy mode and if it falls short it should be improved.

heymoe:
As an added note, I'm not using a Linux shell or terminal. I'm running putty from windows and SSHing into a Linux box where I then use tmux to manage multiple shells instead of having to run multiple putty sessions. 

[...]

PR #5557, opened 2026-08-30
Add scroll-passthrough, so every scrolled line reaches the client

ejc3:
### The problem

With `smcup`/`rmcup` removed via `terminal-overrides`, tmux stays on the outer
terminal's primary screen and lines scrolling off a full-window pane land in that
terminal's own scrollback. This is a long-standing way to get native scrollback
(swipe/wheel on a phone, for instance) instead of copy-mode.

It doesn't work reliably: scroll back and a large fraction of the output simply
isn't there, and in its place you find rows of whatever the application had
pinned at the bottom of the screen.

This is **not** the `TTY_BLOCK` backoff from #1019 — `client_discarded` is 0 throughout.

### Cause

Two things in the write collector, both correct for a pane and both lossy for a
client whose terminal keeps its own scrollback:

1. `screen_write_collect_scroll()` discards the pending write for the row being
   scrolled out, and `screen_write_collect_flush()` emits the scroll *before*
   painting the collected writes — so the row the terminal files away is stale.

2. `screen_write_collect_flush_scrolled()` clamps a batch of pending scrolls to
   the height of the scroll region and drops the excess:

   ```c
   if (ctx->scrolled > s->rlower - s->rupper + 1)
           ctx->scrolled = s->rlower - s->rupper + 1;
   ```

The pane is unaffected either way — the grid history has the lines and copy-mode
shows them. But a terminal keeping its own scrollback only ever keeps what tmux
actually sends it.

### The change

A window/pane flag `scroll-passthrough`, **off by default**.

When set, `screen_write_linefeed()` replays the outgoing row with
`screen_write_collect_flush_line()` and emits that one scroll, instead of
flushing the whole screen — the visible rows are still painted by the ordinary
flush at the end of the batch. `screen_write_scrollup()`, a much rarer path,
takes a full flush before the scroll instead.

Both run before the background is latched: a full flush resets `ctx->bg` to 8, and
the scroll would otherwise go out with the default background rather than `bg`.

### Measurements

51x29 window, an application redrawing the way a full-screen TUI does (rewind
with `CSI nA`, rewrite in place, keep a footer pinned at the bottom), writing
~4000 lines in bursts:

| `scroll-passthrough` | content lines never sent | footer rows wrongly pushed into scrollback |
|---|---|---|
| `off` (current behaviour) | 2067 | 55 |
| `on` | **0** | **0** |

Eight consecutive runs with the option on were clean; one earlier run left 3
footer rows, which I have not isolated.

A plain `cat` of a 3000-line file, same window: 283–1359 lines lost with the
option off (varies by run), 0 with it on.

Replaying just the outgoing row rather than flushing the whole screen costs about
9% fewer bytes on that test — 80kB against 88kB across three runs — and leaves
background handling identical to the option being off (97 of 98 scrolls carrying
the active background in both cases).

With the option off, behaviour is unchanged from stock, verified by running the
same harness against an unpatched build of the same tree.

### Cost and limitations

- Extra output per scrolled line, hence off by default.
- No effect while a pane is in copy mode — `input.c` deliberately uses a
  pane-less write context there, so the guard is false.
- No effect for output inside a synchronized update: `screen_write_collect_flush()`
  discards queued items under `MODE_SYNC` and `CSI ?2026l` schedules a full
  redraw. Preserving those would need outgoing rows retained and replayed across
  the update.
- A tall SIXEL image still scrolls in `screen_write_sixelimage()` without
  consulting the option.

### DIFF
--- a/options-table.c
+++ b/options-table.c
@@ -1731,6 +1731,24 @@ const struct options_table_entry options_table[] = {
 	  .text = "Style of matched characters in switch mode."
 	},
 
+	{ .name = "scroll-passthrough",
+	  .type = OPTIONS_TABLE_FLAG,
+	  .scope = OPTIONS_TABLE_WINDOW|OPTIONS_TABLE_PANE,
+	  .default_num = 0,
+	  .text = "Whether to paint pending output to the client before each "
+		  "line scrolls off the top. tmux normally batches writes and "
+		  "emits the scroll before painting them, discarding the "
+		  "pending write for the outgoing line and clamping a batch of "
+		  "scrolls to the height of the scroll region. Neither is "
+		  "visible in the pane, but a client whose terminal keeps its "
+		  "own scrollback (smcup and rmcup removed with "
+		  "terminal-overrides) only keeps what tmux sends, so those "
+		  "lines never reach its history. With this set the outgoing "
+		  "row is replayed and its scroll emitted one at a time, which "
+		  "costs extra output. No effect in copy mode or inside a "
+		  "synchronized update."
+	},
+
 	{ .name = "synchronize-panes",
 	  .type = OPTIONS_TABLE_FLAG,
 	  .scope = OPTIONS_TABLE_WINDOW|OPTIONS_TABLE_PANE,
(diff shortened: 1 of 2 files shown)


nicm:
Thanks but we are not going to do this, it is fundamentally incompatible with having multiple windows and panes.

PR #5559, opened 2026-08-31
Add a pane output generation format

darlingm:
`pane_last_output_time` has one-second resolution, so an external poller cannot always tell whether output arrived since its previous check. Waiting a second between polls does not avoid this: later output in the same wall-clock second as the previously recorded output leaves the value unchanged.

This adds `pane_output_generation`, a per-pane 64-bit value that changes whenever tmux processes a nonempty pane-output buffer. A periodic pane-content saver can remember this value and, after also checking pane identity and geometry, avoid capturing panes that received no output. This enables a more efficient external design than repeatedly capturing every pane, as snapshot-based tools such as tmux-resurrect with tmux-continuum do.

The value is an opaque change token. tmux processes output in batches, so the numerical difference is not a count of bytes, characters, writes, lines or screen changes. Control-only output changes it, while `capture-pane` does not. Resizes, respawns and other changes unrelated to pane output must be tracked separately.

The implementation adds one 64-bit value per pane and one increment per nonempty buffer, with no per-character work. Saving policy and storage remain external to tmux.

The regression checks independent panes, non-mutating capture, visible output and control-only output.

### DIFF
--- a/format.c
+++ b/format.c
@@ -2276,6 +2276,17 @@ format_cb_pane_last_output_time(struct format_tree *ft)
 	return (NULL);
 }
 
+/* Callback for pane_output_generation. */
+static void *
+format_cb_pane_output_generation(struct format_tree *ft)
+{
+	struct window_pane	*wp = ft->wp;
+
+	if (wp != NULL)
+		return (format_printf("%llu", (unsigned long long)wp->output_generation));
+	return (NULL);
+}
+
 /* Callback for pane_last_prompt_time. */
 static void *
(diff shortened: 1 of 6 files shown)


nicm:
"tmux" in tmux.1 needs to be .Nm, but I would just rephrase the description to avoid it. Also please don't wrap the line in the middle of `return` in `format_cb_pane_output_generation`.

darlingm:
Thanks. I fixed both.

nicm:

PR #5560, opened 2026-08-31
Add formats for pane history change counters

darlingm:
tmux already maintains counters that allow copy mode to update its saved view without repeatedly copying the complete pane history. This exposes them as formats:

- `history_added`: lines scrolled into pane history
- `history_collected`: oldest history lines discarded at the history limit
- `history_generation`: changes when pane history is cleared or reflowed

An external scrollback saver can use these values to retain unchanged history, capture newly added rows and account for rows removed by tmux. Together with separate pane-output tracking, this allows periodic pane-content saving without recapturing all retained scrollback every interval, unlike the full-snapshot approach used by tmux-resurrect with tmux-continuum.

This only exposes existing bookkeeping. It does not change the grid counter types or their update paths, so it adds no work while history is being updated. The formats read the values only when requested.

These values do not describe every possible `history_size` change. Changing pane height or using `resize-pane -T` may change `history_size` without changing them. Applications updating a saved copy should check that the changes agree with `history_size` and take a new complete copy if they do not.

The regression covers scrolling, collection at the history limit, clearing and reflow.


nicm:
Please don't wrap the line in the middle of return like this, add a temporary to avoid it.

Also the man page change is too verbose, please remove the extra paragraph about these, they are too obscure to need anything more than the entry in the table.

darlingm:
Thanks. I fixed both.

nicm:

Issue #5298, opened 2026-06-29
Regression in tmux 3.7: fzf fullscreen redraw/input lag in iTerm2; worked fine in 3.6b


lczerepuk:
### Issue description

After upgrading tmux from 3.6b to 3.7, fullscreen TUI applications such as `fzf` became noticeably slower inside tmux when running in iTerm2.

This worked without any noticeable lag in tmux 3.6b.

The issue is visible as a redraw/input lag when `fzf` uses the whole tmux pane. The same `fzf` command is fast outside tmux in iTerm2. Also, running `fzf` in a tmux popup is fast, so the issue seems specific to normal pane rendering/fullscreen redraw inside tmux.

I have restarted the tmux server after the upgrade with:

```sh
tmux kill-server
```

#### Steps to reproduce

Start tmux with a minimal config:

```sh
tmux kill-server
tmux -L fzf-test -f /dev/null new
```

Inside tmux, run:

```sh
seq 2000 | FZF_DEFAULT_OPTS= fzf --no-sort --no-border --no-info
```

Observe that fullscreen `fzf` redraw/input feels laggy.

For comparison, run the same command outside tmux in iTerm2:

```sh
seq 2000 | FZF_DEFAULT_OPTS= fzf --no-sort --no-border --no-info
```

Outside tmux it is fast.

Also, inside tmux, running `fzf` in a popup is fast:

```sh
seq 2000 | FZF_DEFAULT_OPTS= fzf --tmux center,80%,40% --no-sort --no-border --no-info
```

So this does not appear to be an `fzf` filtering performance issue; it seems related to normal tmux pane rendering.

#### Additional observations

The lag is more noticeable with a large terminal size. My tmux client size was:

```text
361x80
```

`client_discarded` did not increase during the test:

```text
discarded=0 written=1920138 term=xterm-256color size=361x80
```

So tmux does not appear to be discarding output, but the interactive redraw still feels delayed.

I also tested with a minimal tmux config using `tmux -f /dev/null`, and the issue still reproduced.

### Required information

tmux version:

```text
tmux 3.7
```

Previous tmux version where this worked fine:

```text
tmux 3.6b
```

Platform:

```text
╰─ uname -sp
Darwin arm
```

Terminal in use:

```text
iTerm2
```

`$TERM` inside tmux:

```text
screen-256color
```

tmux client terminal:

```text
xterm-256color
```

`$TERM` outside tmux:

```text
╰─ echo $TERM
xterm-256color
```

[...]

PR #5561, opened 2026-08-31
Hold client redraws until resized panes repaint

Nowaker:
## Problem

Resizing a pane could make the whole viewport repaint line by line for 15 to
60 seconds, with the terminal effectively unusable until it finished. It was
reproducible in Terminal.app on macOS against a tmux server on another
machine, and this change removes that stall there.

The server applies new pane geometry immediately, but the application inside
each pane learns its new size only when `TIOCSWINSZ` delivers `SIGWINCH`.
Drawing between those two events sends the new layout filled with the old
pane contents, followed by one correction per pane as each application
catches up. Scrolling within a pane stayed fast throughout; only changing
geometry produced the burst.

The same workflow still has an unresolved slow path in iTerm2, so this PR
does not claim to fix iTerm2. That residue was observed on a tree that
already contains `e06207c9` ("Cache scrollbar options in window to avoid a
slow lookup..."), the fix for #5298, so it is something else again. I have
not root-caused it and it is not addressed here.

## Reproduction

Two full-screen applications using synchronized updates (DEC mode 2026), side
by side, with a client whose output drains more slowly than the server
generates redraws - in practice a Mac terminal attached over ssh. Resizing the
split repeatedly produces the stall; scrolling inside either pane does not.

`regress/resize-sync-redraw.sh` is that reproduction made deterministic. It
drives real applications through an attached PTY client and measures the bytes
the client receives:

- unpatched: 4259 bytes reach the client before either application has
  repainted, and the test fails at its first assertion;
- patched: zero bytes until both applications complete, then one synchronized
  transaction carrying both frames.

I do not have a video, and the applications above are synthetic emitters
rather than a real program. The byte counts are what the regression asserts,
so they can be reproduced by running it against either binary.

## Change

A window-scoped resize redraw barrier:

- mark the window dirty before a pre-resize synchronized frame can flush;
- establish each pane's resize epoch only after a successful `TIOCSWINSZ`;
- release immediately when post-resize frames complete;
- after 150 ms, stop waiting for capable panes that never started a frame,
  while preserving any frame already in progress;
- force a bounded release after one additional second;
- recheck the barrier after tty backpressure and immediately before redraw;
- carry a pane's participation to its destination when it moves between
  windows, without discarding the application's synchronized-update
  capability lease.

## This is a bounded exception to "redraw always wins"

On #4744 you wrote that an application should not be able to prevent tmux
redrawing a pane, and that redraw should cancel the sync flag rather than the
other way round. This patch is an exception to that, and I would rather name
it than let it be found.

The exception is narrow in three ways:

- **It applies only to a resize tmux itself performed**, from the geometry
  change until every pane in that window has been told its new size. Nothing
  else defers a redraw, and an application cannot enter the state on its own.
- **It is bounded by tmux, not by the application**: 150 ms, then a further
  1 s, both fixed. A pane that never answers costs the soft timeout once and
  is then dropped from the epoch.
- **`screen-redraw.c` is untouched.** `screen_redraw_draw_pane()` still stops
  sync and clears the dirty flag before drawing, so once the barrier commits,
  a redraw cancels an application's frame exactly as it does today.

What is deferred is not the application's frame but tmux's own: after tmux
changes geometry, it declines to draw the frame it knows to be wrong - new
layout, old contents - which it created itself and which nothing else will
correct.

If that trade is unacceptable I would rather know now than argue it; the
alternative I considered was drawing the stale frame immediately and
suppressing the corrections, which loses more.

## Interaction with the resize debounce (issue 2677)

`server_client_check_pane_resize()` has a branch for several resizes ending at
the same size, which deliberately sends one resize, leaves one queued, and
re-arms at 10 ms so the application is forced to redraw. Draining the queue at
barrier time would have deleted that timer and collapsed the two resizes into
one, which is exactly the case cb2943fa / 84e22168 / cf6034da fixed.

`resize_timer_second` marks that deliberate 10 ms timer so a forced drain
leaves it alone; ordinary 250 ms debounce timers are still drainable. The
regression asserts two `SIGWINCH` callbacks at least 5 ms apart with the
expected widths in order.

## Ending a synchronized update the barrier interrupted

`server_client_reset_state()` returns early while the barrier is active, so
that cursor placement and mode updates are not applied at the new geometry
before the client has seen the new frame. That return still calls
`tty_sync_end()` first, which is load-bearing rather than tidy.

`tty_sync_start()` is reached from `tty_cmd_syncstart()` on the ordinary pane
write path, and `server_client_reset_state()` is the only place that closes
what that opens. A synchronized update can be opened by pane output and the
barrier engage later in the same libevent batch - I measured 1.1 ms between
the two - and leaving it open would stop the terminal presenting anything
until the barrier released. tmux's own `screen_write_start_sync()` gives an
unfinished update just one second, which is shorter than the barrier's 1.15 s
worst case, so the terminal would time out mid-transaction.

## Regression coverage

`regress/resize-sync-redraw.sh` covers frames open before a resize, prompt
completion, frames spanning the soft timeout, no-start soft release, repeated
DECSET at the hard fallback, repeated resize supersession, pane exit,
primary-screen updates, a late DECSET parsed after soft expiry while client
output is backpressured, cross-window pane transfer during both `WAIT_START`
and `IN_FRAME` including an equal-sized swap, and a synchronized update opened
in the same event-loop batch as the resize that engages the barrier. The PTY
answers terminal feature discovery before any timing assertion begins, so
tmux's five-second startup fallback cannot be mistaken for a barrier redraw.

**It is timing-sensitive and I would rather you did not take it as it
stands.** It asserts that genuine completion releases well before the 150 ms
soft timeout - a real property of the change, but a 20 ms margin under ASan on
a shared runner, and you have already had one synchronized-output regression
from me fail in CI. Widening the deadline would stop it proving prompt
release, so there is no version of it that is both meaningful and generous.
Say the word and I will drop the second commit and keep the script somewhere
you can run it by hand.

## AI authorship

This patch was written with AI assistance, so to address the copyright point
you raised on #4744. The work used OpenAI and Anthropic models, and both
vendors assign ownership of model output to the user:

- OpenAI, Terms of Use, "Ownership of content"
  (https://openai.com/policies/terms-of-use/): "you ... own the Output. We
  hereby assign to you all our right, title, and interest, if any, in and to
  Output."
- Anthropic, Consumer Terms of Service, section 4
  (https://www.anthropic.com/legal/consumer-terms): "Subject to your
  compliance with our Terms, we assign to you all of our right, title, and
  interest-if any-in Outputs."

I own this code and license it under the ISC licence tmux uses.

### DIFF
--- a/cmd-break-pane.c
+++ b/cmd-break-pane.c
@@ -99,6 +99,7 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 	struct session		*dst_s = target->s;
 	struct window_pane	*wp = source->wp;
 	struct window		*w = wl->window, *old_w = w;
+	struct window_pane_resize_sync_token token = { 0 };
 	char			*cause, *cp;
 	int			 idx = target->idx, before, old_idx = wl->idx;
 	const char		*template, *name = args_get(args, 'n');
@@ -150,6 +151,7 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	}
 
+	window_pane_resize_sync_detach(w, wp, &token);
 	TAILQ_REMOVE(&w->panes, wp, entry);
 	TAILQ_REMOVE(&w->z_index, wp, zentry);
 	server_client_remove_pane(wp);
@@ -164,6 +166,7 @@ cmd_break_pane_exec(struct cmd *self, struct cmdq_item *item)
 	TAILQ_INSERT_HEAD(&w->z_index, wp, zentry);
 	w->active = wp;
 	w->latest = tc;
+	window_pane_resize_sync_attach(wp, &token);
 
 	free(w->name);
 	if (name == NULL)
--- a/cmd-join-pane.c
+++ b/cmd-join-pane.c
@@ -417,6 +417,7 @@ cmd_join_pane_exec(struct cmd *self, struct cmdq_item *item)
 	struct winlink		*src_wl, *dst_wl;
 	struct window		*src_w, *dst_w;
 	struct window_pane	*src_wp, *dst_wp;
+	struct window_pane_resize_sync_token token = { 0 };
 	const char		*s;
 	char			*cause = NULL;
 	int			 flags = 0, dst_idx;
@@ -485,8 +486,9 @@ cmd_join_pane_exec(struct cmd *self, struct cmdq_item *item)
 		return (CMD_RETURN_ERROR);
 	}
 
+	if (src_w != dst_w)
+		window_pane_resize_sync_detach(src_w, src_wp, &token);
 	layout_close_pane(src_wp);
-
 	server_client_remove_pane(src_wp);
 	window_lost_pane(src_w, src_wp);
 	TAILQ_REMOVE(&src_w->panes, src_wp, entry);
@@ -502,6 +504,8 @@ cmd_join_pane_exec(struct cmd *self, struct cmdq_item *item)
 		TAILQ_INSERT_AFTER(&dst_w->panes, dst_wp, src_wp, entry);
 		TAILQ_INSERT_AFTER(&dst_w->z_index, dst_wp, src_wp, zentry);
 	}
+	if (src_w != dst_w)
+		window_pane_resize_sync_attach(src_wp, &token);
 	layout_assign_pane(lc, src_wp, 0);
 	colour_palette_from_option(&src_wp->palette, src_wp->options);
 
--- a/cmd-swap-pane.c
+++ b/cmd-swap-pane.c
@@ -66,6 +66,8 @@ cmd_swap_pane_exec(struct cmd *self, struct cmdq_item *item)
 	struct cmd_find_state	*target = cmdq_get_target(item);
 	struct window		*src_w, *dst_w;
 	struct window_pane	*tmp_wp, *src_wp, *dst_wp;
+	struct window_pane_resize_sync_token src_token = { 0 };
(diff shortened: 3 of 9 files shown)


nicm:
Can you explain this and how it works in your own words? These big 300 line AI-generated explanations are too hard to follow.

Using `tcgetpgrp` for anything to do with redraw or resize is definitely not on.

Nowaker:
@nicm I've no idea how it works. I can tell you what the observed problem was and how this fixes it.

iTerm2 + my Tmux setup (pretty complicated setup: multiple panes, windows; left, right, top and bottom bars with system status indicators etc.) became unusable. I first thought it was my desktop computer (the server) being slow, until I sat down at it, and it was totally fine. After some poking around, I realized it's a MacOS problem. iTerm and Terminal.app were affected in the same way. And it didn't require me to ssh to my desktop to observe it - even locally executed Tmux had the same issue.

The AI looked hard, and built a performance fix for OpenCode (optimize window resize to be much faster), and for Tmux (this patch). **This one fixed my Terminal.app.**  iTerm is still extremely slow - while still trying to resolve it, Terminal.app saves my day and I can steer my system remotely again.

output

nicm:
OK well it may have been better if you had opened an issue explaining the problem rather than opening a PR you are not able to review.

How do you reproduce? You run something with fast output using synchronized updates and then resize the pane? Or do you need to resize the whole terminal?

nicm:
I am not all that convinced by this, I am not sure how a stale redraw of ~4000 bytes can make tmux unresponsive for 15 seconds let alone a full minute.

Maybe you can see if you can get Claude to come up with a Python or sh test program to run instead of OpenCode which will show the problem?

Issue #5558, opened 2026-08-31
Feature request: opt-in Unicode-aware word-like and non-whitespace selection in copy mode


hongyi-zhao:
#### Feature request

I would like to propose opt-in copy-mode primitives for:

1. selecting a Unicode-category-aware word-like run; and
2. selecting a run delimited by Unicode whitespace.

Tentative command names used by my proof of concept are:

```text
send-keys -X select-word-unicode
send-keys -X select-token-unicode
```

The names and interface are provisional. Existing `select-word`, `word-separators`, word-motion commands, default bindings, and mouse behaviour should remain unchanged.

This is not a request for [UAX #29 word segmentation](https://www.unicode.org/reports/tr29/). It is a smaller, deterministic grid-cell classification using the Unicode Character Database's [`General_Category`](https://www.unicode.org/reports/tr44/#General_Category_Values) and binary [`White_Space`](https://www.unicode.org/reports/tr44/#White_Space) properties.

#### Current limitation

`word-separators` already supports literal UTF-8 characters, following [#1551](https://github.com/tmux/tmux/issues/1551). However, it remains a [literal string option](https://github.com/tmux/tmux/blob/40381bdcd91dac4deb177e1ed1a5794969d97d39/options-table.c#L1245-L1253); it cannot express Unicode properties, categories, or their complements.

In the relevant copy-mode word-selection and word-motion paths, tmux's [`WHITESPACE`](https://github.com/tmux/tmux/blob/40381bdcd91dac4deb177e1ed1a5794969d97d39/tmux.h#L669) constant contains only TAB and U+0020. [`grid_in_set`](https://github.com/tmux/tmux/blob/40381bdcd91dac4deb177e1ed1a5794969d97d39/grid.c#L1710-L1729) uses literal cell membership, while [`utf8_cstrhas`](https://github.com/tmux/tmux/blob/40381bdcd91dac4deb177e1ed1a5794969d97d39/utf8.c#L960-L976) compares an item from the configured set with the complete grid cell.

Consequently, the current model has no direct way to express the following as Unicode-property predicates without exhaustive enumeration:

- every Unicode Letter, Mark, Number, or Connector_Punctuation cell as word-like;
- every Unicode `White_Space` character as a token boundary;
- all remaining non-whitespace cells as a separate class;
- category-based handling of a cell containing multiple code points.

The existing membership helper also treats wide-character padding columns specially rather than resolving either display column to the leading cell.

[...]

PR #5562, opened 2026-08-31
Treat space as Unicode whitespace in word handling

hongyi-zhao:
## Summary

- treat an ASCII space in a grid membership set as any character with the Unicode `White_Space` property, using an in-tree switch for the 25 Unicode 17.0 code points
- inspect every code point stored in a cell, while preserving TAB and wide-cell padding widths
- update the copy-mode word/space documentation and add regressions for both emacs and vi mode keys

This keeps the implementation independent of utf8proc. The default `word-separators` value has no space, so its punctuation behavior is unchanged; a space in a custom value now means any Unicode whitespace.

## Testing

Passed with both `./configure --disable-utf8proc` and `./configure --enable-utf8proc`:

- `make -j4`
- `regress/copy-mode-unicode-whitespace.sh`
- `regress/copy-mode-test-emacs.sh`
- `regress/copy-mode-test-vi.sh`
- `regress/utf8-test.sh`

The new regression covers ASCII space and TAB, all 11 `White_Space` ranges (including NBSP and U+2029), U+3000 leading and padding cells, a mixed cell containing zero-width U+0085, custom literal separators, and soft/hard wrap behavior.

Also passed `git diff --check`, `sh -n`, and ShellCheck.

Addresses the whitespace portion of #5558.

### DIFF
--- a/grid.c
+++ b/grid.c
@@ -1711,21 +1711,28 @@ grid_in_set(struct grid *gd, u_int px, u_int py, const char *set)
 {
 	struct grid_cell	gc, tmp_gc;
 	u_int			pxx;
+	int			has_tab, has_space;
 
 	grid_get_cell(gd, px, py, &gc);
-	if (strchr(set, '\t')) {
-		if (gc.flags & GRID_FLAG_PADDING) {
-			pxx = px;
-			do
-				grid_get_cell(gd, --pxx, py, &tmp_gc);
-			while (pxx > 0 && tmp_gc.flags & GRID_FLAG_PADDING);
-			if (tmp_gc.flags & GRID_FLAG_TAB)
-				return (tmp_gc.data.width - (px - pxx));
-		} else if (gc.flags & GRID_FLAG_TAB)
-			return (gc.data.width);
-	}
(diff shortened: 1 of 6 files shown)


nicm:
There are only 25 codepoints here so I think a table is overkill, why not just have a switch like in `utf8_should_combine`?

hongyi-zhao:
Done, thanks — I replaced the range table with an explicit 25-case switch matching `utf8_should_combine` in eee41c83. I kept the per-codepoint cell traversal so combined cells are still handled, and the Unicode whitespace regression passes with and without utf8proc.

nicm:
Looks good - applied to OpenBSD now with some trivial changes, will be in GitHub later. Thanks!

hongyi-zhao:
> Looks good - applied to OpenBSD now with some trivial changes, will be in GitHub later. Thanks!

BTW, do you use OpenBSD as your main OS for work?

nicm:
I use Linux for work work and OpenBSD and Linux for tmux work.

hongyi-zhao:
Are you ready for merging this into master on GitHub?

nicm:
This was added in 8f3101d last week.

PR #5564, opened 2026-08-31
Add code contribution guidance.

darlingm:
I am going to use this guidance locally when preparing future tmux changes and wanted to offer it upstream in case it would also be useful to other contributors and AI-assisted editors.

This adds a root `CONTRIBUTING.md` covering recurring guidance on code style, design, documentation, testing, licensing, and AI use. It was assembled from the current contributing wiki, relevant parts of OpenBSD style(9), the tmux source, and maintainer comments across many issues and pull requests.

The two `CONTRIBUTING.md` files have separate purposes. The existing `.github/CONTRIBUTING.md` remains focused on issue and support reports and links to the new root file for code and documentation changes.

`AGENTS.md` is deliberately only a one-line pointer, so tools that load it automatically do not spend tokens on the longer guide unless they are making changes.

`AGENTS.md` is currently listed in `.gitignore`. Once the file is tracked, that entry no longer has any effect, so this removes it. I do not know whether the entry is intended to prevent accidental submissions or to allow a private maintainer file. In the latter case, tracking this path would be awkward, and an existing ignored file could be overwritten when checking out the branch. One option would be to ignore `AGENTS.local.md` instead and add `If AGENTS.local.md exists, read it.` to the tracked file. I am also happy to omit `AGENTS.md`, but then users' AI tools would usually not automatically read `CONTRIBUTING.md`.

I developed and reviewed this with substantial assistance from OpenAI's ChatGPT 5.6 Sol Pro and Codex. ChatGPT 5.6 Sol Pro performed an extended research pass over the source, style guidance, and a large set of maintainer review comments. Codex and I then reviewed and refined the resulting guidance.

I may have generalized some preferences incorrectly or worded them differently than intended. I am happy to adjust, shorten, or remove anything. I wanted to offer the result in case repository-level guidance would be helpful.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,3 +1,8 @@
+# Reporting issues
+
+See [Code contributions](../CONTRIBUTING.md) for code and documentation
+changes. The guidance below covers bug reports and support requests.
+
 ## What should I do before opening an issue?
 
 Before opening an issue, please ensure that:
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@
 .deps/
 .dirstamp
 .serena/
-AGENTS.md
 Makefile
 Makefile.in
 aclocal.m4
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+When making changes, read and follow `CONTRIBUTING.md`.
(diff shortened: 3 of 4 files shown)


nicm:
I think this is too much, I think anything pertinent should be rolled into this section https://github.com/tmux/tmux/wiki/Contributing#code-contributions and the file should just link to that.

Issue #5553, opened 2026-08-30
Control mode: command replies are buffered without limit, server RSS grows unboundedly


llvm-x86:
### Summary

A control-mode client that issues reply-producing commands faster than it reads the replies makes the tmux **server** buffer the entire backlog — one byte of anonymous memory per byte of reply, with no ceiling. `control.c` has flow control for `%output`, but command replies bypass all of it.

The memory is reachable and is freed when the client drains or disconnects, so it is not a leak and LeakSanitizer finds nothing. But glibc does not return the arena afterwards, so a long-lived server's RSS stays near its high-water mark. For a server driven continuously by an automated control client, the effect is hard to distinguish from a leak.

Affected: **3.7c** (built from the release tarball, unmodified) and current **master**. Line numbers below are from the 3.7c tarball.

### Reproducer

```sh
tmux -L leak new-session -d -x 200 -y 50
tmux -L leak set -g history-limit 2000
tmux -L leak send-keys 'for i in $(seq 1 2000); do printf "%-180s\n" "line $i"; done' Enter
sleep 6
pid=$(tmux -L leak display-message -p '#{pid}')

# A control client that commands faster than it reads. ~19 kB reply each.
python3 - "$pid" <<'EOF'
import subprocess, sys, threading, time
pid = int(sys.argv[1])
def rss():
    return next(int(l.split()[1]) for l in open("/proc/%d/status" % pid)
                if l.startswith("VmRSS"))
p = subprocess.Popen(["tmux", "-L", "leak", "-C", "attach"],
                     stdin=subprocess.PIPE, stdout=subprocess.PIPE)
time.sleep(1)
base = peak = rss()
stop = threading.Event()
def reader():                       # drains, but slower than tmux produces
    while not stop.is_set():
        if not p.stdout.read1(8192):
            return
        time.sleep(0.002)
threading.Thread(target=reader, daemon=True).start()
for _ in range(9000):
    p.stdin.write(b"capture-pane -p -S -\n"); p.stdin.flush()
    peak = max(peak, rss())
print("server RSS %d -> peak %d MB" % (base//1024, peak//1024))
stop.set(); p.kill()
EOF
tmux -L leak kill-server
```

[...]

PR #5565, opened 2026-09-01
Bound the memory used by buffered control mode command replies, GitHub issue 5553.

llvm-x86:
Fixes the server-side memory growth in #5553. It replaces the patch posted in that issue.

`control.c` meters `%output` — `CONTROL_BUFFER_HIGH`, the per-pane block queues, `control_check_age()` — but command replies bypass all of it, so a client that issues reply-producing commands faster than it reads them makes the server buffer the entire backlog with no ceiling. On master, 10000 under-read `capture-pane -p -S -` commands take the server from 5 MB to 187 MB, and the peak scales linearly with the command count.

The backlog is not in `all_blocks`: `control_write_line()` writes straight to the bufferevent when `all_blocks` is empty, which it always is for a client receiving no `%output`, so the bytes accumulate in the write event's output evbuffer. When there is pane output as well, replies queue as zero-size blocks. The bound counts both.

When the total exceeds 64 MB the client gets the existing "too far behind" exit, as `control_check_age()` already does for a client that cannot keep up with `%output`. Reads are never paused: pausing them can deadlock, because control I/O is the client's own stdin and stdout with 64 kB of pipe capacity, and `control_check_age()` walks only `cp->blocks`, so nothing would break the cycle.

Measured on a private socket with 10000 commands and a client reading slower than tmux produces, the server peaks at 89 MB rather than 187 MB — the 64 MB bound plus evbuffer slack and allocator overhead — and the client exits with `%exit too far behind`. A client that drains normally is unaffected: reply stream identical to the byte (57112708 B, 3001 `%end` blocks over 3000 commands), same drain time (8.7 s before, 8.5 s after), no kill. With `%output` interleaved so replies take the block-queue path instead, framing is identical (2001 `%begin`/`%end` pairs) and neither build kills the client. A client that never reads at all has every command accepted until the bound trips, then exits cleanly; before this patch it grows without bound.

64 MB is a guess at "no legitimate client is this far behind"; a single `capture-pane` reply in the reproducer is 19 kB. Reusing `CONTROL_BUFFER_HIGH` would be wrong, since one ordinary `capture-pane` reply already exceeds 8192. Change the constant if you want a different number.


nicm:
Should `control_write_line` stop adding more output once the maximum is hit?

llvm-x86:
@nicm 

Yes, it should - it didn't. Nothing in the command queue looks at CLIENT_EXIT, so the commands already read from the client kept running and their replies kept being buffered after the client had been flagged to exit. In the never-reading test that was another 187 replies (about 1.7 MB) written after the kill.

Updated: the check is now done once at the top of control_write_line() and the line is dropped if the client is over the limit. This works without any extra state because control_discard() leaves the write buffer and the queued reply blocks alone, so the total stays over the limit until the client goes away. The check is deliberately on the limit rather than on CLIENT_EXIT: a plain detach-client also sets CLIENT_EXIT between its %begin and %end, and gating on the flag would leave that %begin unterminated.

Retested: 0 lines buffered after the kill (was 187), detach-client and kill-server still produce a complete %begin/%end and then %exit, and a client that drains its output gets byte-identical output to the unpatched build (3000 capture-pane replies, 3001 %end).

llvm-x86:
Rebased on master (no overlap with the 5570 change) and added a regress test in a second commit, regress/control-reply-buffer-bound.sh, following control-client-exit-stalled.sh: a control client with unread output down a fifo is sent 1000 capture-pane commands with its stdin held open, and must be dropped by the server within the exit timeout while the server survives. Passes in about 12 seconds with the fix; on master the client is never dropped.

nicm:
>  This works without any extra state because control_discard() leaves the write buffer and the queued reply blocks alone, so the total stays over the limit until the client goes away. 

Couldn't the buffer drain below the limit again if a command returns `CMD_RETURN_WAIT`? I think you need to set a new flag as well a `CLIENT_EXIT` so dropping output is permanent.

Also, what if `line` is longer than 64 MB itself? That will not be detected until the next line is written. Maybe that is not a real concern...

llvm-x86:
@nicm

You're right on both.

The wait case is real: with a run-shell queued behind the flood and the client draining 40 MB while it slept, the buffer dropped below the limit and 164 more replies were written to the client after it had been killed. Added CLIENT_CONTROL_DISCARD (the unused 0x800000000ULL bit), set alongside CLIENT_EXIT and checked first in control_write_line(), so once over the limit nothing more is buffered for that client regardless of what the buffer does afterwards. Same scenario now writes 0 lines after the kill.

For the long line, the check now includes the size of the line about to be added, so a line that would itself take the total over the limit is dropped and the client killed before it is buffered, rather than on the next line. That also means strlen() is done once instead of twice.

Retested: detach-client and kill-server still get a complete %begin/%end and %exit, a draining client gets byte-identical output to master (3000 replies, 3001 %end), and with a pane flooding %output so replies go through all_blocks the framing is identical to master (2001 %begin/%end) with no kill. The regress test passes.

nicm:
Looks good - applied to OpenBSD now, will be in GitHub later. Thanks!

llvm-x86:
Thank you, it was a pleasure helping.

nicm:
Just look at the change list for the files you changed https://github.com/tmux/tmux/commit/a3129249d19b74439f0c25c72124f72e47ddd9eb

PR #5571, opened 2026-09-03
permit selecting by target window in new-window

### LINKED ISSUE #5563: support automatic select-window on new-window if index exists
### Issue description
I wanted to change the behaviour of `<prefix><number>`  to automatically create a new window with that index if it didn't already exist. Saves me from running `:neww` on new instances.

This is the solution I came up with:
```tmux
# autocreate windows if they don't exist
bind 0 if-shell 'tmux select-window -t :0' '' 'new-window -t :0 -c "#{pane_current_path}"'
bind 1 if-shell 'tmux select-window -t :1' '' 'new-window -t :1 -c "#{pane_current_path}"'
bind 2 if-shell 'tmux select-window -t :2' '' 'new-window -t :2 -c "#{pane_current_path}"'
bind 3 if-shell 'tmux select-window -t :3' '' 'new-window -t :3 -c "#{pane_current_path}"'
bind 4 if-shell 'tmux select-window -t :4' '' 'new-window -t :4 -c "#{pane_current_path}"'
bind 5 if-shell 'tmux select-window -t :5' '' 'new-window -t :5 -c "#{pane_current_path}"'
bind 6 if-shell 'tmux select-window -t :6' '' 'new-window -t :6 -c "#{pane_current_path}"'
bind 7 if-shell 'tmux select-window -t :7' '' 'new-window -t :7 -c "#{pane_current_path}"'
bind 8 if-shell 'tmux select-window -t :8' '' 'new-window -t :8 -c "#{pane_current_path}"'
bind 9 if-shell 'tmux select-window -t :9' '' 'new-window -t :9 -c "#{pane_current_path}"'
```

This seems rather obtuse for such a "simple" feature. I was hoping I could do something like `tmux new-window -At :3`, similar to how I would script this with `new-session -A` in my ssh aliases. Documentation says that `-S` works in this way, but it only matches on `window-name`. 

While my configuration works, I think the approach leaves room for improvement. From reading the source code (`cmd-new-window.c:91-117`), it seems like `-S -n <name>` works, but `-S -t <number>`  will never work. Would it make sense to add support for `-S -t :<number>` here? I'd be happy to make a PR if this makes sense.

Related question: Why is this called `-S` and not `-A` consistent with `new-session`? I didn't fully understand the documentation until I read the source code.

### Required information

* `tmux -V`: `tmux 3.5a`
* `uname -sp`: `Linux AMD Ryzen 9 5950X 16-Core Processor`
* Terminal in use: kitty
* $TERM *inside* tmux (`echo $TERM`): `tmux-256color`
* $TERM *outside* tmux (`echo $TERM`): `xterm-kitty`
* Logs from tmux (`tmux kill-server; tmux -vv new`): n/a

mazunki:
As discussed in #5563, this adds support for `-S -t` in `new-window`.

- implemented the feature
- added regression tests for the feature
- explicitly document what `-n` does on `new-window`
- compiled and passed regression tests

It feels a bit weird that `-S` without either `-n` or `-t` is allowed (since the flag becomes a no-op), but I haven't touched that behaviour as to not introduce any regressions.

Closes: https://github.com/tmux/tmux/issues/5563

### DIFF
--- a/cmd-new-window.c
+++ b/cmd-new-window.c
@@ -74,8 +74,10 @@ cmd_new_window_exec(struct cmd *self, struct cmdq_item *item)
 	}
 
 	/*
-	 * If -S and -n are given and -t is not and a single window with this
-	 * name already exists, select it.
+	 * If -S is given, select an existing window instead of creating a
+	 * new one: preferring -t if it already points at a window, or
+	 * otherwise -n if one exists with that name. If neither matches,
+	 * fall through and create a window as normal.
 	 */
 	name = args_get(args, 'n');
 	if (name != NULL) {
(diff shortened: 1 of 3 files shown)


nicm:
Try this:
~~~~
./tmux -Lx kill-server
./tmux -Lx new -dnA \; neww -dnB \; neww -S -t :+1
./tmux -Lx display -p '#{window_index}:#{window_name}'
~~~~
This shows `0:A` when it should be selecting window 1. Same with `-t:+2` which should be creating index 2 but also selects 0.

mazunki:
Thanks, I missed that. Made the conditionals a bit cleaner too.

nicm:

Issue #5543, opened 2026-08-28
Vulnerability — uncontrolled recursion in layout_construct


entropy1337:
layout-custom.c layout_construct recurses for every nested { (LEFTRIGHT) or [ (TOPBOTTOM) child in a layout string, with no depth limit. A deeply nested layout string therefore drives unbounded C recursion → stack exhaustion → SIGSEGV (tmux server crash). Entry: layout_parse (layout-custom.c:174); reached from cmd-select-layout.c:127 (select-layout / select-layout -E), and also when a layout string is loaded from a config or a restored session.

Reachability: a crafted layout string (deeply nested cells) passed to select-layout, a config file, or a restored/saved session. A client able to run select-layout on the session, or able to supply a config/saved session that is loaded, can trigger it.
Impact: stack-exhaustion crash of the tmux server → DoS of all attached sessions. CVSS 3.1 5.5 (AV:L/AC:L/PR:L/UI:N/S:U/C:N/I:N/A:H).
Confirmation: AddressSanitizer stack-overflow on a crafted nested layout string. Full write-up + PoC available on acknowledgement.
Affected version: HEAD c1f947a3 (2026-08-21); the recursive layout_construct has no depth bound and is longstanding — please help confirm the first-affected release.

nicm:
Can you give an example of a layout string or a script that produces one so we can reproduce?

nicm:
This should be fixed in OpenBSD should be fixed in `master` later on. Thanks!

PR #5572, opened 2026-09-04
layout-custom: bound layout_construct's recursion depth

afonsojanu:
Follows up on #5543 with a fix.

layout_construct recurses once for every nested `{` (LEFTRIGHT) or `[` (TOPBOTTOM) group in a layout string, and nothing caps how deep that goes. A layout string with enough nested groups drives unbounded C recursion until the stack is exhausted and the server crashes. This is reachable from `select-layout`, from a config file that sets a layout, and from a restored or saved session, so anything able to get tmux to parse a layout string it doesn't fully control (a crafted config, a malicious or corrupted saved session) can trigger it.

I built tmux with ASan/UBSan and confirmed this directly: a layout string nested 100000 levels deep crashes with "Thread stack size exceeded due to excessive recursion" in `layout_construct` at layout-custom.c:404, matching the reachability described in #5543. With the fix in place the same input is rejected as an invalid layout instead, and the server keeps running.

The fix adds a `LAYOUT_MAX_DEPTH` constant (1000, well past anything a real layout would ever need) and threads a depth counter through `layout_construct`'s recursive calls, returning an error once it's exceeded rather than recursing further.

Added `regress/layout-nesting-limit.sh`, checking that a layout nested past the limit is rejected and that the server and an existing session survive the attempt. Ran the full regress suite locally alongside it, no other failures.

### DIFF
--- a/layout-custom.c
+++ b/layout-custom.c
@@ -28,7 +28,17 @@ static u_short			 layout_checksum(const char *);
 static int			 layout_append(struct layout_cell *, char *,
 				     size_t);
 static int			 layout_construct(struct layout_cell *,
-				     const char **, struct layout_cell **);
+				     const char **, struct layout_cell **,
+				     u_int);
+
+/*
+ * Maximum nesting depth for a layout string. This exists purely to bound
+ * layout_construct's recursion: without it, a layout string with enough
+ * nested { or [ groups drives unbounded recursion and exhausts the stack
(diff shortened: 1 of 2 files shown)


nicm:
There is no point in fixing this yet because this code is going to be replaced soon.

nicm:
I have applied this to OpenBSD now with some minor changes for the new code layout, will be in GitHub later. Thanks!

PR #5573, opened 2026-09-05
clear-on-attach off: keep the terminal's scrollback correct

ejc3:
With `clear-on-attach off` the terminal keeps its own scrollback and tmux draws on the primary screen. On master three things break that scrollback. This is three commits, one per defect, in dependency order. The first is also open on its own as #5580 in case that piece is wanted separately.

## 1. A full-screen program destroys the scrollback

A full-screen program in a pane scrolls the terminal's scrollback away. tmux paints the pane's alternate screen onto the terminal's primary screen with line feeds, so every page `less` shows is scrolled into the scrollback and stays there after `less` quits. Paging through a file left 140 rows of it in the scrollback.

The commit sends `smcup` to the client when the active pane fills the window, is not in a mode, and enters its alternate screen, and `rmcup` when it leaves it. The program then draws on the terminal's alternate screen and the terminal restores the primary screen and scrollback itself on exit: 0 rows, with the output from before `less` intact.

The switch is made in `server_client_check_redraw()` after the tty is unfrozen, inside the same synchronized update as the redraw that follows, and forces a full redraw since the new screen has no status line. A tty flag, `TTY_ALTSCREEN`, records it. `tty_stop_tty()` leaves the alternate screen if the flag is set, so a detach or suspend does not leave the terminal in it, and a reattach while the program is still running enters it again. Both `smcup` and `rmcup` must be present, so a terminal with one but not the other is never put on a screen it cannot leave.

## 2. Most scrolled lines never reach the scrollback

The write collector emits a scroll before painting the row it pushes out. `screen_write_collect_scroll()` drops the pending write for that row. `screen_write_collect_flush_scrolled()` clamps a run of scrolls to the region height and discards the rest. `cat` of 300 lines leaves 109 of them in the scrollback. Claude Code expanding a 400-line tool result with ctrl-o delivers 118 of the 400 lines; the other 282 are in its pane history but never reach the terminal.

The commit paints the outgoing row before each scroll, in `screen_write_linefeed()` and `screen_write_scrollup()`, when the active pane fills the window. It is the existing collector flushing one call earlier at the two places it scrolls, about 40 lines; no new path. Painting it can leave the cursor past the last column when the scroll is sent, so `tty_cursor()` now reaches column 0 with CR rather than an absolute move: shorter, and it clears the pending wrap on terminals that keep it across cursor moves, the class `tty_region()` already works around. The terminal then receives every line in order: all 300, and all 400. No option: with `clear-on-attach off` the scrollback is the point, so the lines are delivered.

Output that arrives a line at a time is not affected either way; a streamed Claude Code reply reached the terminal in full on master too. It is bursts of more than a screenful in one write, such as build or test output, that lose lines.

The cost is one cursor-position and erase sequence per scrolled row. Per line that actually arrives it is close to nothing: the ctrl-o expansion above is 3,814 bytes for 118 lines on master and 13,286 for 400 with the change, 32 against 33 bytes per delivered line. Throttling the client to 4, 8, 16 and 32 kB/s still delivered 300 of 300.

## 3. After a window switch the scrollback shows the previous window

The terminal has one scrollback and the session has one history per window, so after a switch the terminal shows the new window's screen above the previous window's history.

The commit adds a window option, `scroll-replay`, the number of history lines to replay. On a window change tmux clears the terminal's scrollback and writes the new window's history and visible rows, inside the same synchronized update as the redraw, so the switch is one frame. The replay is deferred while the terminal is on its alternate screen (commit 1) and happens when the pane returns to the primary screen. A window with 829 lines of history replayed with none lost or duplicated; 2000 lines is about 90 kB per switch. It erases whatever the terminal's scrollback held, including output from before tmux was attached, so it is off by default.

## Scope

None of the three changes anything with `clear-on-attach on`, in a split, or in copy mode; the first also needs `smcup` and `rmcup`, the third `E3`. Measured with a pty harness at 51x29 as `xterm-256color`, status line off. With a status line the pane scrolls inside a region, and whether a terminal adds region scrolls to its scrollback is up to the terminal. The regress suite gives the same results as master at each commit.


nicm:
I don't think we are going to do this. It is an invasive change that few people will use. Replaying history on every new window is clever but is an awful bodge: it adds a lot of data which might work for you but won't for many, and only works on terminals with E3. And it is *still* incompatible with panes - a key tmux feature.

The supported way to access a pane's history is to use copy mode. You have not given any convincing use case for this, or any reason copy mode doesn't work for you. It would be a very high bar to reach to convince me to accept something that is evidently such a narrow solution in any case.

You could also use a terminal capable of integrating with tmux's history itself like iTerm2 with control mode.

mgrant0:
I want to jump in here and second what Nic is saying. I already went down the path you are going down.  I started with gnu-screen and PuTTY.  I wrote some scripts that can repaint the scrollback buffer when i logged back in.  I had a key binding to prefix c-l that would repaint a window so I could switch windows and then repaint the local buffer, then use the local scrollbar to scroll back up and copy/paste things which were larger than the screen.  It mostly worked but was klunky, it lost all the colouring.  I got something similar working in tmux which I don't use anymore, I can send it to you if I can find it.

But in the end, what I really wanted was a way to select stuff with the mouse, be able to scroll around and select large sections of the screen's scrollback, and to have a scrollbar that functioned like my local one.  In essence, I wanted a remote window pane that acted a lot like a local one.  

I moved to tmux a few years ago because it was way more configurable than screen.  I added scroll bars to tmux and with the mouse selection stuff in copy mode, it now works amazingly well.  And in fact, it works better than trying to repaint the local window's scroll back buffer.  You can log in from multiple places and it works.  You can switch panes and it works.  It works when there's multiple panes on the window split in any direction.  It works with floating panes sitting on top of what you are selecting, copying, and pasting.  It works with sixel and kitty images (or will work once the image PR 4902-image-support gets put into tmux).

I abandoned PuTTY for a tty that supports osc-52 to be able to copy things to my local computer's paste buffer and enabled that in tmux and now copy-pasting works in both directions, albeit you need to hold the shift-key while pasting back into the tty.

Unless I completely misunderstand what you are after, it seems like what you are trying to do, tmux already does better if you let it.  This is what's in my .tmux.conf related to this (you may not need the last line):

set -g mouse on
set -g pane-scrollbars on
set -s set-clipboard external
set -as terminal-features ',*-256color:256:clipboard'

If I have misunderstood your goal, let us know, or let me know if you want my old repaint hack scripts and bindings if that's useful.

ejc3:
Thank you for your thoughtful feedback.

The terminal I feel this most acutely is Prompt, which doesn’t support the mouse wheel on iOS, which makes non-native scrolling untenable.

The other place I run tmux in full screen is on cmux and the native Mac scrolling is smoother than using tmux to manage the buffer (but not nearly as important). Given cmux has its own window manager, running tmux full screen with native window scrolling feels like a logical way to run things. 

I agree that when you split, this fix doesn’t help, but I don’t find it that bad since the main screen still renders okay and alt mode apps are still fine. It’s also opt in, so if folks want buttery smooth multi-window tmux, they can have it and otherwise, can allow tmux tradition scroll handling work.

mgrant0:
I don't use any of those.  Have you tried tmux in control mode in iTerm2?  That's pretty cool and it's going to get better.  There are other programs that also support tmux control mode (tmux -CC) like wezterm.  I don't know about cmux.

ejc3:
> I don't use any of those. Have you tried tmux in control mode in iTerm2? That's pretty cool and it's going to get better. There are other programs that also support tmux control mode (tmux -CC) like wezterm. I don't know about cmux.

cmux added control mode too (I actually did a bunch of fixups). I agree it’s a good model, but still want the native scrolling:)

nicm:
I do appreciate your work, and the idea of integrating with terminal scrollback is attractive, even just for single-pane windows. But I don't think we want to add a not-insignificant second path here through input processing that we will need to maintain both code and the promise of support forever, I expect hardly anyone will use, and where the implementation both requires giving up the alternate screen and doesn't actually solve the problem completely. Not on the strength of you wanting to use a terminal which doesn't support the mouse wheel (!) or because it feels logical.

It seems to me that the practical solutions here are to add mouse wheel support to the terminal you want to use or switch to a different one which has support already, and if you have requirements not met by copy mode to work on it so it does what you want.

Or, if you are must have this approach, maybe you could maintain it externally as a patch for a while and if we find it is getting a lot of uptake and we get a better feel for the use cases, then we can talk about it again?

If there was a protocol which allowed tmux to interact fully with terminal scrollback it might be a different matter. I don't know it would need to be that big - some way to have multiple scrollbacks by ID, to populate one, to switch between them, to delete one. And some way to detect it and to turn it on even in the alternate screen. Maybe this is something you'd like to design and see if you can get consensus on among terminal developers?

mgrant0:
I really think the way forward here is to use control mode and a compatible terminal if you really want to do this locally.

If you wanted to repaint all the scrollback when you switched windows or reattached, couldn't you just replay the copy buffer?  It's what I was doing previously with PREFIX c-l.  Could you use sync to stop it from flashing lots of text scrolling by?

I have never seen (or never noticed) any escape sequences that would let you manipulate up into the local terminal's scroll buffer.  The most I have seen is that you can push lines up into it by scrolling them off the top of the screen when the tty is not in alternate screen mode.

Again, I want to reiterate, this only works in somewhat narrow circumstances.  No window splits, no pane-border-status, status line at the bottom or none at all, no floating panes of course, no pane-scrollbar, mouse=off, perhaps other things too.  

Could this be achieved without any modification to tmux I wonder?  What if you used these hooks in tmux: client-attached, session-window-changed and maybe also pane-mode-changed and active-pane-changed to repaint the scrollback?  Something like this:

```
set-hook -g client-attached 'run-shell "tmux capture-pane -p -S - -E - > #{client_tty}"'
set-hook -g session-window-changed 'run-shell "tmux capture-pane -p -S - -E - > #{client_tty}"'
```

I have felt your pain.  What you gain by being able to do this you lose not being able to use tmux to its fullest extent.

ejc3:
@mgrant0 I tried the hook approach before writing any C. It works for the window switch if you fix three things. You need `ESC[3J` before the capture, otherwise each switch stacks the new history under the old one. With two windows of 100 and 60 lines and four switches between them, the terminal's scrollback went 111, 181, 292, 362 rows; with the erase first it went 111, 70, 111, 70, the current window each time. You need `capture-pane -e` or you lose the colour. And `run-shell` can't put the write inside tmux's synchronized update, so you watch the whole history scroll past. Two things it can't do at all. It runs after tmux has already redrawn the new window, so a switch is two frames. And nothing fires when a pane leaves its alternate screen, which matters once #5580 is in: after `less` quits, the terminal's scrollback still shows the previous window. That's why I put it in tmux, next to the code that already scrolls the old screen away on attach when `clear-on-attach` is off.

Passthrough can't be a hook. The lines go missing inside the write collector. `screen_write_collect_scroll()` drops the pending write for the row that scrolls out, and `screen_write_collect_flush_scrolled()` caps a burst of scrolls at the region height and throws the rest away. Nothing outside tmux ever sees those lines. The fix is about 40 lines: the same collector flushing one call earlier at the two places it scrolls. No new path through input processing.

@nicm understood, and thanks for spelling out the reasons.

Copy mode scrolls with the mouse wheel, and on iOS there mostly isn't one. [Blink](https://github.com/blinksh/blink/issues/2006) and [Termius](https://termius.com/blog/new-touch-terminal-on-ios) turn a swipe into arrow keys, Prompt scrolls its own buffer, and only [Moshi](https://getmoshi.app/articles/best-ios-terminal-app-coding-agent) turns a swipe into wheel events. The other common way to run a shell from a phone is mosh, which has no scrollback of its own ([its issue #2](https://github.com/mobile-shell/mosh/issues/2), 82 comments), and the answer there has always been to run tmux inside it.

While measuring all this I found `clear-on-attach off` on master has the same problems by itself. `cat` 300 lines and 109 of them reach the terminal's scrollback; Claude Code expanding a 400-line tool result delivers 118 of the 400. Run `less` full-window and it scrolls the scrollback away, leaving 140 rows of the file behind. Switch windows and the scrollback still shows the old one. So I've restructured this PR as three commits, one per problem, and rewritten the description around that. The alternate-screen one is also open by itself as #5580 if you'd rather take it alone. The replay one is capped at `scroll-replay` lines, needs `E3`, does nothing in a split, and waits while the terminal is on its alternate screen.

nicm:
`clear-on-attach` is not necessarily meant to control whether tmux puts new lines into the scrollback, it was added for a narrow purpose - to prevent tmux using `clear` and preserve what was on the terminal before tmux started so that if tmux is a login shell they are not destroyed. What tmux does after it is started is not really related, in fact I suspect it would probably be closest to the intent if we didn't put any tmux content at all into the scrollback (although we are not going to make that promise).

It seems far from ideal that iOS terminal emulators are not able to send scroll wheel events for swipe, even as an option, but that seems like something that should be fixed in the terminal emulator not worked around in tmux.... tmux supports 11 mouse buttons plus the wheel so there would be plenty of choice.

Anyway, I'm going to close this issue for the moment. I'll take a look at your other ones. Thanks!

ejc3:
@nicm thank you for the thoughtful response and considering #5580. I have marked it not draft now.

I do want reiterate the benefit of this pull request. Without clients having to make **any** changes, we can give users an experience where lines aren't dropped, window switching works, and native scroll just works.

The attached screen recording is *such* a good experience. In less than a second, I attach to the tmux session, my scrollback is restored, and native scroll / find, etc just works.

I still think it's worth a flag to enable this feature. We don't have to reuse the one you just added (I realize I didn't understand the purpose), but letting people opt into a perfect full screen experience seems supremely beneficial (there are edge cases still to be worked through).

https://github.com/user-attachments/assets/d57a12c1-8954-4181-b6a5-2dfe832aa883

PR #5574, opened 2026-09-05
Osc133 aware commands

### LINKED ISSUE #5417: commands to do stuff with osc133 marked up output
With output between the OSC133 C (end of prompt, start of output) and D (end of output), do some useful things that mimic the things you can do with selections:

Add select-output to select the output of a command in copy-mode if the output is marked up with osc133.  if collapsed, expand it first.

Add copy-output, copy-pipe-output, pipe-output to match the selection commands, but operate on current output rather than selection.

Add open-output which call spawn_editor() to open selection or current output in an editor. If there is an argument they should  use that command (so "open-selection /usr/bin/vi"), if no argument they should use the editor option like spawn_editor() does by default.

current output = output under cursor, if no output, do nothing

mgrant0:
1. Added top-level commands:
   - select-output
   - copy-output
   - copy-pipe-output
   - pipe-output
   - open-output
   - open-selection

3. Added -a to above commands support to operate on the entire pane/copy-mode buffer, including panes without OSC 133 markers.

4. Top-level commands use the most recent OSC 133 output when outside copy mode.
   - select-output enters copy mode and selects the output; other top-level operations use a temporary copy-mode snapshot.

5. Added default copy-mode editor bindings:
   - Opens the selection when present, otherwise the current command output.
      - emacs: e
      - vi: M-e
   - Select output:
      - emacs copy mode: C-o → select-output
      - vi copy mode: M-o → select-output

Some suggested key bindings useful for testing:
```
bind -n C-o select-output
bind -n M-o copy-output
bind -n M-e open-output
bind -n M-W pipe-output 'wc -l >~/tmux-output-lines'
bind -Tcopy-mode M-o send -X copy-output
bind -Tcopy-mode M-e send -X open-output
bind -Tcopy-mode M-W send -X pipe-output 'wc -l >~/tmux-output-lines'
bind -Tcopy-mode M-s send -X open-selection
```


mgrant0:
Closing this PR because it is now superceded by https://github.com/tmux/tmux/pull/5597 and (https://github.com/tmux/tmux/pull/5415

PR #5576, opened 2026-09-06
cmd-find: reject invalid relative target offsets

cuishuang:
Invalid relative target offsets such as `+0`, `-0`, `+foo`, and `-foo` are currently accepted by target resolution.

The relative window and pane target parsers call `strtonum()` without checking its error result. When parsing fails, the value returned by `strtonum()` is zero. Passing zero to the relative target helpers leaves the target unchanged, so the invalid target silently resolves to the current window or pane.

I check the `strtonum()` error result when parsing relative window and pane offsets. Return an unresolved-target error when the offset is invalid.


nicm:

PR #5582, opened 2026-09-08
Keep floating panes inside the window when it shrinks

### LINKED ISSUE #5581: Floating pane is left outside the window when the window shrinks
A floating pane keeps the geometry it was created with when the window is
resized, so shrinking the window can leave it partly or entirely outside,
where it cannot be seen or reached.

Reproduced on `next-3.8` at 578e07fcbc66dc60822b55b88ba12f518df57374 (also
present on master; `layout.c` is unchanged there in the relevant part).

### Repro

```
$ tmux -f/dev/null new-session -d -x 120 -y 40
$ tmux new-pane -x 30 -y 10 -X 85 -Y 25
$ tmux resize-window -x 60 -y 20
$ tmux list-panes -F '#{pane_width}x#{pane_height} at #{pane_left},#{pane_top}'
60x20 at 0,0
28x8 at 86,26      <- window is 60x20
```

The float is wholly outside the window. It stays there for the rest of the
window's life — nothing else revisits its geometry — so this is reachable by
just detaching and reattaching from a smaller terminal, or moving a client
between a monitor and a laptop screen.

### Where

`layout_resize` adjusts tiled cells only: `layout_resize_adjust` walks the
tree through `layout_resize_check`, which counts `is_leaf && !is_floating`
(`layout.c:263-265`), and `layout_fix_panes` then copies each floating cell
through to its pane unchanged.

I realise #5135 defers floating-pane resize *policy*. This is narrower than
that: whatever the policy turns out to be, a pane the user cannot reach is a
defect, and this is the one part that does not depend on the policy.

### Patch

Clamping at the end of `layout_resize` is enough. Two things that are not
obvious and cost me a build to find:

- the size parameters have to be used rather than `w->sx`/`w->sy` — `resize.c`
  resizes the layout *before* the window ("Resize the layout first",
  `resize.c:63`), so `w->sx` is still the old size in here;
- a floating cell is the *inside* of the pane, so the border has to be counted
  the way `layout_floating_args_parse` counts it on creation (`-x` → `sx -= 2`,
  `-X` → `ox += 1` unless `pane-border-lines` is `none`), or a bordered float
  ends up one cell over the edge.

```diff
--- a/layout.c
+++ b/layout.c
@@ -789,6 +789,60 @@
 	layout_free_cell(w->layout_root, only_nodes);
 }
 
+/*
+ * Move, and if necessary shrink, floating panes so they stay inside the
+ * window. layout_resize_adjust walks only tiled cells, so a floating cell
+ * keeps the geometry it was given when it was created: after the window
+ * shrinks it can be left partly or entirely outside, where it cannot be seen
+ * or reached.
+ *
+ * A floating cell is the inside of the pane, so the borders around it have to
+ * be counted here as they are when it is created (see
+ * layout_floating_args_parse).
+ */
+static void
+layout_clamp_floating_panes(struct window *w, u_int sx, u_int sy)
+{
+	struct window_pane	*wp;
+	struct layout_cell	*lc;
+	u_int			 pad, avail, csx, csy;
+
+	TAILQ_FOREACH(wp, &w->z_index, zentry) {
+		lc = wp->layout_cell;
+		if (lc == NULL || (~lc->flags & LAYOUT_CELL_FLOATING))
+			continue;
+		if (window_pane_get_pane_lines(wp) == PANE_LINES_NONE)
+			pad = 0;
+		else
+			pad = 1;
+
+		csx = lc->g.sx;
+		avail = (sx > 2 * pad) ? sx - 2 * pad : 0;
+		if (csx > avail)
+			csx = (avail > PANE_MINIMUM) ? avail : PANE_MINIMUM;
+		csy = lc->g.sy;
+		avail = (sy > 2 * pad) ? sy - 2 * pad : 0;
+		if (csy > avail)
+			csy = (avail > PANE_MINIMUM) ? avail : PANE_MINIMUM;
+		if (csx != lc->g.sx || csy != lc->g.sy)
+			layout_set_size(lc, csx, csy, lc->g.xoff,
+			    lc->g.yoff);
+
+		if (lc->g.xoff + lc->g.sx + pad > sx) {
+			if (lc->g.sx + 2 * pad >= sx)
+				lc->g.xoff = pad;
+			else
+				lc->g.xoff = sx - lc->g.sx - pad;
+		}
+		if (lc->g.yoff + lc->g.sy + pad > sy) {
+			if (lc->g.sy + 2 * pad >= sy)
+				lc->g.yoff = pad;
+			else
+				lc->g.yoff = sy - lc->g.sy - pad;
+		}
+	}
+}
+
 /* Resize the entire layout after window resize. */
 void
 layout_resize(struct window *w, u_int sx, u_int sy)
@@ -840,6 +894,7 @@
 
 	/* Fix cell offsets. */
 	layout_fix_offsets(w);
+	layout_clamp_floating_panes(w, sx, sy);
 	layout_fix_panes(w, NULL);
 }
 
```

### Verified

Built with the patch and checked:

- the repro above now gives `28x8 at 31,11` — flush inside the 60x20 window
  (outer box 30x10 at 30,10);
- with `-B none` the same float lands at `30,10` instead, so the border inset
  is doing real work rather than being a constant;
- shrinking progressively (40x15, 20x8, 12x5) keeps it inside at each step,
  down to `10x3 at 1,1`;
- growing the window back does not move a float that already fits;
- `#{window_layout}` is byte-identical to the unpatched build across 20
  float-free cases (4 tiled layouts x 5 sizes), so nothing changes for windows
  without floats.

I could not run `regress/` on this host — every test sets
`PATH=/bin:/usr/bin`, which does not exist on NixOS — so the above is manual.
Happy to send this as a PR with a `regress/` case if you would prefer that
shape.

noamsto:
Fixes #5581.

`layout_resize_adjust` walks only tiled cells — `layout_resize_check` counts
`is_leaf && !is_floating` — and `layout_fix_panes` then copies each floating
cell through to its pane unchanged. So a floating pane keeps the geometry it
was created with, and shrinking the window can leave it partly or entirely
outside, where it cannot be seen or reached. Detaching and reattaching from a
smaller terminal is enough to hit it.

This clamps floating cells at the end of `layout_resize`, per the discussion on
the issue: force them visible for now, ahead of any layout-hints work.

Two things that are not obvious from the surrounding code, and are why the
patch looks the way it does:

- the `sx`/`sy` parameters are used rather than `w->sx`/`w->sy`, because
  `resize.c` resizes the layout *before* the window ("Resize the layout
  first", `resize.c:63`), so `w->sx` is still the old size in here;

- a floating cell is the *inside* of the pane, so the border has to be counted
  the way `layout_floating_args_parse` counts it on creation (`-x` → `sx -= 2`,
  `-X` → `ox += 1` unless `pane-border-lines` is `none`), or a bordered float
  ends up one cell over the edge.

A pane is shrunk only when it cannot fit at its current size, and never below
`PANE_MINIMUM`; otherwise it is only moved.

### Test

Built on `master` at 1b87cb75 with no new warnings. The repro from the issue,
against the same build with and without the patch:

```
$ tmux -f/dev/null new-session -d -x 120 -y 40
$ tmux new-pane -x 30 -y 10 -X 85 -Y 25
$ tmux resize-window -x 60 -y 20
$ tmux list-panes -F '#{pane_width}x#{pane_height} at #{pane_left},#{pane_top}'
```

before — the float is wholly outside the 60x20 window:

```
62x20 at 0,0
30x10 at 85,25
```

after — moved and shrunk to fit, with the border accounted for
(31+28 = 59 <= 60, 11+8 = 19 <= 20):

```
60x20 at 0,0
28x8 at 31,11
```

### DIFF
--- a/layout.c
+++ b/layout.c
@@ -789,6 +789,60 @@ layout_free(struct window *w, int only_nodes)
 	layout_free_cell(w->layout_root, only_nodes);
 }
 
+/*
+ * Move, and if necessary shrink, floating panes so they stay inside the
+ * window. layout_resize_adjust walks only tiled cells, so a floating cell
+ * keeps the geometry it was given when it was created: after the window
+ * shrinks it can be left partly or entirely outside, where it cannot be seen
+ * or reached.
+ *
+ * A floating cell is the inside of the pane, so the borders around it have to
+ * be counted here as they are when it is created (see
+ * layout_floating_args_parse).
+ */
+static void
+layout_clamp_floating_panes(struct window *w, u_int sx, u_int sy)
+{
+	struct window_pane	*wp;
+	struct layout_cell	*lc;
+	u_int			 pad, avail, csx, csy;
+
+	TAILQ_FOREACH(wp, &w->z_index, zentry) {
+		lc = wp->layout_cell;
+		if (lc == NULL || (~lc->flags & LAYOUT_CELL_FLOATING))
+			continue;
+		if (window_pane_get_pane_lines(wp) == PANE_LINES_NONE)
+			pad = 0;
+		else
+			pad = 1;
+
+		csx = lc->g.sx;
+		avail = (sx > 2 * pad) ? sx - 2 * pad : 0;
+		if (csx > avail)
+			csx = (avail > PANE_MINIMUM) ? avail : PANE_MINIMUM;
+		csy = lc->g.sy;
+		avail = (sy > 2 * pad) ? sy - 2 * pad : 0;
+		if (csy > avail)
+			csy = (avail > PANE_MINIMUM) ? avail : PANE_MINIMUM;
+		if (csx != lc->g.sx || csy != lc->g.sy)
+			layout_set_size(lc, csx, csy, lc->g.xoff,
+			    lc->g.yoff);
+
+		if (lc->g.xoff + lc->g.sx + pad > sx) {
+			if (lc->g.sx + 2 * pad >= sx)
+				lc->g.xoff = pad;
+			else
+				lc->g.xoff = sx - lc->g.sx - pad;
+		}
+		if (lc->g.yoff + lc->g.sy + pad > sy) {
+			if (lc->g.sy + 2 * pad >= sy)
+				lc->g.yoff = pad;
+			else
+				lc->g.yoff = sy - lc->g.sy - pad;
+		}
+	}
+}
+
 /* Resize the entire layout after window resize. */
 void
 layout_resize(struct window *w, u_int sx, u_int sy)
@@ -809,8 +863,11 @@ layout_resize(struct window *w, u_int sx, u_int sy)
 	 * out proportionately - this should leave the layout fitting the new
 	 * window size.
 	 */
-	if (lc->type == LAYOUT_WINDOWPANE && (lc->flags & LAYOUT_CELL_FLOATING))
+	if (lc->type == LAYOUT_WINDOWPANE && (lc->flags & LAYOUT_CELL_FLOATING)) {
+		layout_clamp_floating_panes(w, sx, sy);
+		layout_fix_panes(w, NULL);
 		return;
+	}
 	xchange = sx - lc->g.sx;
 	xlimit = layout_resize_check(w, lc, LAYOUT_LEFTRIGHT);
 	if (xchange < 0 && xchange < -xlimit)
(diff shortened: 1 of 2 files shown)


nicm:
Looks good with new panes but try this:
~~~~
./tmux -Ltest new -d -x120 -y40 \; break-pane -W -x30 -y10 -X85 -Y25 \; resize-window -x60 -y20 \; a

nicm:

PR #5585, opened 2026-09-09
Add dump/restore commands to save and recreate session layouts

asRizvi888:
## Summary
- Adds `dump [-f path]` to write the layout of every session (sessions,
  windows, panes, splits, active pane/window, and each pane's current
  working directory) to a text file as a sequence of tmux commands.
- Adds `restore [-f path]` to recreate that layout by sourcing the file.
  This does not relaunch whatever process was previously running in each
  pane - only the shape of the layout and where each pane's shell starts.
- Default file location is derived from the directory of the last-loaded
  config file (falling back to $HOME), named .tmux.dump.
- dump records each session's base-index and pins it before emitting
  indexed commands, so restore places windows correctly even when the
  restoring environment's config sets a different base-index, then
  restores the prior global value afterward.
- Pane working directories are captured live (via the same mechanism
  as the pane_current_path format variable), not the pane's original
  spawn directory, so restore reflects where you actually left off.

## Test plan
- [x] Dump/restore round-tripped across sessions with multiple windows,
      splits, and custom layouts; verified pane cwds match exactly.
- [x] Verified restore works correctly when the restoring environment's
      base-index differs from the dump-time value.
- [x] make (clean build, no new warnings)


nicm:
Thanks - this is a reasonable start but I think:

- It should contain a lot more and the intent is for this to use JSON now that we have a parser.
- Output and input should go via the `file.c` API like `load-buffer`/`save-buffer`.
- The commands should have more tmux-typical names.

Also can you please put your real name in the copyright header.

nicm:
I don't think you got the point...

nicm:
I think this should be entirely driven by JSON, it should not generate tmux commands.

Could you maybe write up how you see this working? All the stuff that would be saved, restored, how the JSON would look, etc. So we can agree the design before writing the code.

asRizvi888:
Thanks for the pointer - I think I understand now. Here's how I'd like this to work, before I write any more code:

**Scope (v1):** for every session - name, base-index, current window; for every window - index, name, layout tree (splits/sizes/active pane, the same JSON the new layout format already produces); for every pane - index, live working directory. Not included: window/pane options, session environment, pane titles, zoomed state, or relaunching whatever was running in each pane. I'd rather ship this minimal version first and add any of that afterwards if you want it - let me know if you'd rather see more of that up front.

**JSON shape:**

```json
{
  "version": 1,
  "sessions": [
    {
      "name": "work",
      "base_index": 0,
      "current_window": 1,
      "windows": [
        {
          "index": 0,
          "name": "zsh",
          "active_pane": 0,
          "layout": { "V": 2, "L": { "t": "p", "w": 80, "h": 24, "x": 0, "y": 0, "a": true, "i": 0, "I": "%0" } },
          "panes": [ { "index": 0, "cwd": "/home/user" } ]
        }
      ]
    }
  ]
}
```

`layout` is exactly what `layout_dump()` already emits for a window, embedded directly rather than escaped into a string.

**save-layout:** unchanged from the current code - walks the live session/window/pane trees and writes the JSON above via `file_write()`.

**load-layout:** this is what I'll rework. Right now it parses the JSON and then generates a script of tmux commands (`new-session`, `new-window`, `select-layout`, ...) and runs that through `load_cfg_from_buffer()`. I understand now that's not what you want - instead it should read the file via `file_read()`, parse it with `json_parse()`, and for each session/window/pane call the internal creation functions directly: `session_create()` for the session, `spawn_window()` (the same primitive `new-session`/`new-window` already use internally, with `sc.idx` set to the recorded index so each window lands at the right place without needing any base-index tricks) for each window's first pane, and `layout_parse()` on the embedded layout JSON to build out the rest of that window's panes and geometry. No `cmd_parse`/`load_cfg_from_buffer` anywhere in the restore path.

Let me know if this matches what you had in mind, or if I'm still off - happy to adjust before writing the code.

nicm:
OK I think this is fine but I'm not sure about `base_index`, what do you need it for? If you are saving the options you should save them all probably (or at least the selection you want to save for now) in an `options` section - but mostly a session will use the global option here, and I assume we are not saving that.

PR #5591, opened 2026-09-12
Do not unzoom in window_destroy, it destroys the window twice

rfrancoise:
window_destroy() calls window_unzoom() so that the zoomed layout is put back before the panes are destroyed (GitHub issue 3717). Since events replaced notifications, window_unzoom() also resizes the panes through layout_fix_panes() and window_pane_resize() fires pane-resized with the window in the payload. The payload takes a reference on a window whose count is already zero; releasing it fires window-closed a second time and calls window_destroy() again from inside the first call, which frees the panes that the outer layout_fix_panes() is still walking.

Reproducer (server exits unexpectedly on the last command):

    tmux -f /dev/null new -d -s m
    tmux split-window -d -h -t m:0
    tmux resize-pane -Z -t m:0
    tmux new -d -s d
    tmux kill-session -t m

ASan: heap-use-after-free in layout_fix_panes (layout.c:477), from window_unzoom <- window_destroy <- winlink_remove <- session_destroy; freed by window_pane_free <- window_destroy_panes <- window_destroy <- window_remove_ref <- event_payload_free_value.

Both layouts are freed by window_destroy() anyway, so just clear the zoom flags on the window and its panes instead of calling window_unzoom(): nothing gets resized and no event fires, and a later unzoom while the panes are destroyed is still prevented. This is the same problem bad8d0fd20 fixed in 2013 before the call came back with 36e1ac6556.

Add a regress test for kill-session and kill-server with a zoomed window.

### DIFF
--- a/regress/kill-session-zoomed.sh
+++ b/regress/kill-session-zoomed.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# Killing a session whose window is zoomed must not crash the server:
+# window_destroy used to unzoom the window, which resized the panes and
+# fired pane-resized with the window in the payload, and releasing that
+# reference destroyed the window a second time.
+
+PATH=/bin:/usr/bin
+TERM=screen
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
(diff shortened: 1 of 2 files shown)


nicm:

Issue #4808, opened 2026-01-11
Contributing to tmux!


nicm:
Code and documentation contributions to tmux are welcome.

There are some [notes on the development process here](https://github.com/tmux/tmux/wiki/Contributing).

The [same page](https://github.com/tmux/tmux/wiki/Contributing#code-contributions) has a list of improvements and feature requests roughly categorized by effort (small, medium or large).

There is both the [documentation in the wiki](https://github.com/tmux/tmux/wiki) and the [man page](https://man.openbsd.org/OpenBSD-current/man1/tmux.1) (source code [here](https://github.com/tmux/tmux/blob/master/tmux.1)).

If you would like to make a contribution, please feel free to open an MR or ask questions in this issue.

peripherium:
Hi, I've already implemented several feature requests from the list. However, two of these entries are still on the wiki page:

"(#3503) Move panes or windows by dragging with mouse."

I think my PR #4766 provides the desired feature.

"(#3655) Ability to change popup styles after it is open."

#4678 allows popups to be modified after they've been created.

Should I have requested the project page be changed? If so, I apologize for the misunderstanding.

nicm:
It's OK I remove stuff as I notice it's done but sometimes I miss them, I'll remove them

marcindulak:
I posted a related question about the development process https://github.com/orgs/tmux/discussions/4809 - please have a look.

ghost:
Is there a tool that formats written code according to tmux standards?

nicm:
Not to my knowledge although it is broadly OpenBSD [style(9)](https://man.openbsd.org/style) so possibly someone has come up with a `clang-tidy` file that will get close.

Edit: The main thing that we don't do from.style(9).is "Put a tab after the first word, i.e., use ‘int^Ix;’ and ‘struct^Ifoo *x;’." The tab goes after the struct tag, not before.

There is [indent(1)](https://man.openbsd.org/indent) of course, I forgot about it... no idea how well it works.

ethanuppal:
> Use of AI tools is acceptable but please see [this section](https://github.com/tmux/tmux/wiki/Contributing#use-of-ai).

Do you happen to know the last commit to `tmux` known to not have any code attributed to LLMs?

eMPee584:
> Do you happen to know the last commit to `tmux` known to not have any code attributed to LLMs?

[...]

PR #5449, opened 2026-07-31
Add manual-or-smallest window sizing

Mr-Neutr0n:
## Summary

- add the `manual-or-smallest` choice proposed in #2671
- use the manual dimensions as a cap while still shrinking for smaller attached clients
- preserve the mode across `resize-window` and expose its cap through the existing manual-size formats
- document the new choice and cover client attach, cap changes, and detach

Addresses #2671.

## Testing

- `make -j4`
- `regress/window-size-manual-smallest.sh` (fails before the implementation, passes after)
- `regress/session-group-resize.sh`
- `regress/new-session-size.sh`
- `regress/new-session-no-client.sh`
- `regress/screen-redraw-cache.sh`
- `regress/format-variables.sh`
- `regress/options-values.sh`

Prepared with OpenAI Codex assistance; I reviewed and tested the changes.

PR #5592, opened 2026-09-12
Add window-size manual-or-smallest

### LINKED ISSUE #2671: Can no longer set max width or height limit after `force-width` and `force-height` options removed
### Issue description

I find it really annoying that the `force-width` and `force-height` options got removed from version 2.9. This means I can no longer set the maximum limit of the width or height and still have it reduce when reducing the window size below the limit.

It says in the changelog the `resize-window` option does the same thing but it doesn't. It sets `window-size` to manual which I don't want. There needs to be `limit-width` and `limit-height` options that do the same as what the `force-width` and `force-height` options did.

### Required information

* tmux version: 3.0a
* Platform: Ubuntu.
* $TERM inside of tmux : `screen` . Outside of tmux: `xterm-256color`.

wizzseen:
## Summary
- Add a `window-size manual-or-smallest` choice: keep the manual size unless an attached client is smaller.
- Width and height are capped independently.
- Documents the option and adds regress coverage.

Fixes #2671.

## Test plan
- [x] `TEST_TMUX=$(pwd)/tmux sh regress/window-size-manual-or-smallest.sh`
- [x] `TEST_TMUX=$(pwd)/tmux sh regress/options-values.sh`
- [x] Comment on #2671 / #4808 so this does not collide with other work

### DIFF
--- a/format.c
+++ b/format.c
@@ -3215,7 +3215,8 @@ format_cb_window_manual_height(struct format_tree *ft)
 
 	if (w == NULL)
 		return (NULL);
-	if (options_get_number(w->options, "window-size") != WINDOW_SIZE_MANUAL)
+	if (!WINDOW_SIZE_USES_MANUAL(options_get_number(w->options,
+	    "window-size")))
 		return (xstrdup(""));
 	return (format_printf("%u", w->manual_sy));
 }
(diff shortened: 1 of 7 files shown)


nicm:
Someone else already did this in https://github.com/tmux/tmux/pull/5449.

wizzseen:
Thanks — I missed https://github.com/tmux/tmux/pull/5449 and opened this as a duplicate. Closing in favor of that PR.

PR #5593, opened 2026-09-12
Prevent live pane output from overwriting menus

atomicstack:
Window menus are drawn above panes during scene redraws, but incremental pane output does not account for their position. Text updates, line clears, scrolling, and other terminal operations can overwrite an open menu.

This change excludes the menu rectangle from pane-visible ranges and treats overlapping menus as obstructions when choosing redraw paths. Whole-line fallbacks and insert-character operations use the scene renderer when a menu is present.

Pane contents continue updating underneath the menu, and uncovered areas remain live. `capture-pane` still returns the underlying pane contents without clipping or menu decorations.

### Validation

- Added a nested-client regression covering text updates, clears, scrolling, character and line edits, wide characters, insert mode, viewport clipping, top status positioning, clipboard delivery, and menu closure.
- Confirmed failures before the corresponding fixes and passing results afterward.
- Passed existing menu, popup, TTY drawing, synchronized-output, and input editing/scrolling regressions.
- Separately verified that capturing a background pane includes newly updated content beneath an open menu.

### DIFF
--- a/regress/menu-live-output.sh
+++ b/regress/menu-live-output.sh
@@ -1,0 +1,108 @@
+#!/bin/sh
+
+# check the actual terminal image while a pane writes underneath a menu.
+# an outer tmux captures the display of an attached inner tmux client.
+
+PATH=/bin:/usr/bin
+TERM=screen
+LC_ALL=C.UTF-8
+export TERM LC_ALL
+
+[ -z "$TEST_TMUX" ] && TEST_TMUX=$(readlink -f ../tmux)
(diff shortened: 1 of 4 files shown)


nicm:
Hah, I thought I fixed this, but clearly not well enough.

nicm:
I don't think this is quite right. Why do you need these checks in `tty.c`? If `tty_draw_pane` needed to check for a menu, it would also need to check for a floating pane. This should be handled at a higher layer.

PR #5594, opened 2026-09-13
Write bracketed paste directly to popup job

Pajn:
Send pasted text to the popup job buffer as received instead of encoding it key by key, matching `window_pane_paste()`. Otherwise keys inside the paste are sent as escape sequences when extended keys are enabled.

This matches the previous fix for panes 17bab327943824ae6c4b6c0f76cf8d2fd0055d12.

AI notice:
The bug was identified, the original patch written and the previous commit found, with help by Opus 5. I have manually verified the fix, and read the code, which lead to a fix on the original patch.

Opus has originally placed this check inside the `if (pd->job != NULL) {` block but I moved it to be before the `event->key == '\033' || event->key == ('c'|KEYC_CTRL)` check so that a past containing ESC or CTRL-C does not close the popup.


nicm:
Thanks, popups are going to be removed very shortly so I don't really want to take any more changes for them.

PR #5610, opened 2026-09-17
manpage: clarify that window/pane spawning commands use `default-shell`

LunarLambda:
`/bin/sh` is used for `if-shell`, `run-shell`, and `#(command)` formats.

`default-shell` is used for everything else, from my testing.

Also mentioned `display-popup` in the list of commands that allow multi-argument command specifications that run without the shell.


LunarLambda:
If different choice of markup or a rewording is desired I'd be happy to tweak it. I went for the least noisy possible change first.

nicm:
I think this is confusing because `default-shell` is not a command. Better to leave it as `/bin/sh` and explain it in the text.

LunarLambda:
> I think this is confusing because `default-shell` is not a command. Better to leave it as `/bin/sh` and explain it in the text.

I have changed the example back to `/bin/sh` and added a brief sentence mentioning that `default-shell` is used.

nicm:

PR #5611, opened 2026-09-18
Replay scrolls deferred during synchronized updates instead of redrawing the whole scroll region

bmaurer:
### Problem

Since 565db46a a synchronized update (DECSET 2026) only redraws the lines that changed when it ends, but any scroll inside the update marks the entire scroll region dirty (d33d5b7d). A TUI that streams text into scrollback through a scroll region inside the same synchronized update as its frame — codex does this on every frame while a response is streaming — therefore has the whole region redrawn for every frame.

Bytes tmux writes to the outer terminal per frame (200x50 pane, one frame wrapped in `ESC[?2026h … ESC[?2026l`, measured with a pty harness around the client):

| frame contents | 3.5a (ignores 2026) | master | master + this PR |
|---|---|---|---|
| spinner cell rewritten | 49 B | 75 B | 75 B |
| `\n` on the bottom row (whole-screen scroll) + new line | 76 B | 8,070 B | 105 B |
| DECSTBM region above the last row, one line scrolled into it | 149 B | 8,375 B | 176 B |
| EL/ED + rewrite | 117 B | 151 B | 151 B |

### Change

When the screen scrolls while `MODE_SYNC` is set, `screen_write_collect_flush()` now records the scroll (region and count) on the pane instead of marking the region dirty: dirty bits already set inside the region move up with it and the newly exposed lines are marked dirty. When the update ends, `screen_write_flush_dirty()` sends the scroll to the clients with `tty_cmd_scrollup()` (inside the client's own synchronized update) and then redraws only the dirty lines — the same output as without synchronized updates.

Only one region is deferred per update. If a different region scrolls, the scroll covers the whole region, the pane is resized or obscured, or a scrollbar overlay is visible, it falls back to marking everything dirty / `PANE_REDRAW` as before. Inserted/deleted lines, reverse index and scroll down are not deferred and still mark their range dirty.

### Testing

- New `regress/sync-output-scroll.sh`: fails on master (`unchanged line redrawn after synchronized scroll`), passes with this change.
- `sync-output-atomic.sh`, `sync-output-cursor.sh`, `decrqm-sync.sh`, `input-scroll.sh`, `input-raw-scroll.sh`, `copy-mode-redraw.sh` pass; the full `regress/` run has the same pass/fail set as master (the remaining failures are environmental and identical on both).
- Final screen contents of the outer client (captured through a second, known-good tmux) compared against 3.5a and against the same build without 2026, for: spinner only; whole-screen scroll; scroll region; IL/DL; EL/ED; writes before and after a scroll in one frame; several scrolls per frame; two regions in one frame; a scroll larger than the region; region scroll followed by IL/DL; CSI S; reverse index after a scroll. All identical.

Draft: opened for discussion — the deferred-scroll state lives on `struct window_pane` next to `sync_dirty`; happy to move it or restructure.


nicm:
Hmm yes I think this makes sense but I'm not sure if there is a better way.

nicm:
I think the idea here is fine.

The comments are excessively verbose.

Are you sure it is right to always set the background to 8 in `screen_write_sync_replay_scroll`?

I don't like the name `screen_write_sync_replay_scroll`, it is not replaying it, it is just doing it. How about `screen_write_sync_apply_scroll`?

I'd probably call `screen_write_sync_scrolled` something like `screen_write_sync_scroll_dirty` so it is more "active" and it matches up with the other "dirty" functions. I might rename `flush_dirty` and `clear_dirty` to `sync_flush_dirty` and `sync_clear_dirty` so it is obvious they are related to sync.

bmaurer:
Thanks — all addressed in the updated commit (rebased on master):

- Renamed as suggested: `screen_write_sync_scroll_dirty`, `screen_write_sync_apply_scroll`, `screen_write_sync_flush_dirty`, `screen_write_sync_clear_dirty`.
- Background: you were right to question it. The scroll's background (`ctx->bg`) is now recorded with the deferred scroll (`wp->sync_bg`) and passed to `tty_cmd_scrollup`, so it behaves exactly like the immediate path; a scroll with a different background in the same update falls back to marking everything dirty, same as a different region.
- Comments cut down to one line each.
- One fix from re-review: `sync_apply_scroll` no longer sets `PANE_REDRAW` for the scrollbar-overlay case. It can run inside one client's redraw (via `screen_write_stop_sync` from `redraw_draw`), and a flag set there is cleared at the end of `server_client_check_redraw` before other clients see it. It now marks the region dirty instead, like the obscured case.

Re-verified: bytes/frame unchanged from the PR description (scroll 8070 → 105, region 8375 → 176), final screens identical to 3.5a and to the same build without 2026 across the spinner/scroll/region/IL-DL/EL-ED/mixed cases plus coloured-background scrolls, `sync-output-*.sh` and `decrqm-sync.sh` pass, full `regress/` pass/fail set matches master.

While testing this I found a separate, pre-existing bug in the sparse redraw (`tty_draw_line` trusting the wrapped-line shortcut when the cursor is on another row): #5625. It is independent of this change but shows up more once scrolls are deferred.

nicm:
There is a one problem:

- `server_client_loop` loops over the clients and calls `server_client_check_redraw` on each which may call `redraw_draw`.

- `redraw_draw` calls `screen_write_stop_sync`. This can set `PANE_REDRAW` via `tty_cmd_scrollup`.

- But it the second (or subsequent) client sets `PANE_REDRAW`, that is too late. Previous client redraws are finished so it will not redraw the pane.

- The `PANE_REDRAW` flags is lost in the "clear the flags" loop in `server_client_loop`.

bmaurer:
You're right, thanks. I reproduced it: two clients, a pane that is not full width (so neither client can scroll it), and the update still open when only the second client is redrawn. The first client kept the old screen. The new `regress/sync-output-scroll-clients.sh` fails on the previous version and passes on master and on the update.

Fixed by not relying on `PANE_REDRAW` when the deferred scroll is sent:

- `screen_write_sync_apply_scroll` gives `tty_cmd_scrollup` its own redraw callback. A client that cannot scroll the region (large region, obscured) marks the region dirty instead, and `screen_write_sync_flush_dirty` redraws those lines for every client straight afterwards.
- The scrollbar redraw now happens when the scroll is recorded (during input, like the immediate path), so sending the scroll sets no pane flags at all.
- Clients with `CLIENT_REDRAWWINDOW` are already skipped by `tty_client_ready`, so `screen_write_set_client_cb` cannot set `PANE_REDRAW` from here either.

Re-tested on current master: bytes per frame are unchanged (scroll 7987 → 105, region 8300 → 169), screens are identical to 3.5a across the earlier cases, and a full `regress/` run has the same results as master.

nicm:
Looks good, applied to OpenBSD now with some trivial changes, will be in GitHub later. Thanks!

PR #5612, opened 2026-09-18
release_3.7d: only redraw changed lines when a synchronized update ends (backport)

bmaurer:
Backport of the synchronized-output (DECSET 2026) fixes from master to the 3.7 release branch.

### Problem

3.7 added support for mode 2026. When the application ends a synchronized update, 3.7b/3.7c/release_3.7d set `PANE_REDRAW` (input.c) and redraw the whole pane, even if only a few cells changed. Modern TUIs (codex, ratatui applications, neovim, …) wrap every frame in `ESC[?2026h … ESC[?2026l`, so a one-line spinner animation makes tmux resend the entire pane on every frame — very visible over SSH, and a regression from 3.5a which ignored 2026 and just forwarded the delta.

Bytes tmux writes to the outer terminal per frame for a pane program that rewrites only a spinner cell inside a 2026 bracket (200x50 pane, measured with a pty harness around the client):

| tmux | spinner frame |
|---|---|
| 3.5a (ignores 2026) | 49 B |
| 3.7b / release_3.7d | 9,800 B (whole pane, every frame) |
| release_3.7d + this PR | 74 B |
| master | 75 B |

### Changes

Cherry-picks from master (all by nicm), in order:

- 11b6e784 When MODE_SYNC is on, all terminal updates need to deferred, not just scrolling (GitHub issue 4983)
- 724f85d2 Do not draw to tty if PANE_REDRAW has been set since it is just going to be replaced
- 565db46a Instead of fully redrawing the pane when sync ends, only redraw lines that have been changed
- c5a73099 Dirty ED 2 correctly (GitHub issue 5322)
- d33d5b7d When sync is enabled, we need to mark the whole scroll region dirty if there has been any scrolling (GitHub issue 5330)
- b2de8038 Flush output before ending sync (GitHub issue 5495)
- 57a13664 Do not allow cursor on/off to escape synchronized updates

Conflicts resolved by hand where the release branch predates the screen-redraw.c rewrite (95afd754): `screen_write_clear_dirty()` is called from `screen_redraw_draw_pane()`, `screen_redraw_get_visible_ranges()` keeps its old name, the SIXEL blocks in linefeed/reverse index are kept, and the `PANE_REDRAW` on DECRST 2026 in input.c is dropped since `screen_write_stop_sync()` now flushes the dirty lines. The screen-redraw.c half of 57a13664 does not apply (this branch already ends the client sync in `server_client_reset_state()`).

### Testing

- Final screen contents of the outer client (captured through a second, known-good tmux) compared against 3.5a and against the same build without 2026, for spinner, whole-screen scroll, scroll region, IL/DL, EL/ED and mixed edge-case scenarios: identical.
- `regress/` suite: same pass/fail set as the unpatched branch (the failures are environmental and identical on both).

Scrolling inside a synchronized update still redraws the whole scroll region, as on master; #5611 addresses that on master.


nicm:
I think this is too big for 3.7 particularly with 3.8 so close.

PR #5600, opened 2026-09-14
Add Kitty keyboard protocol support

mgrant0:
Add kitty keyboard protocol support (KKP).
  - Clean implementation based on the published wire protocol
    (https://sw.kovidgoyal.net/kitty/keyboard-protocol/); no Kitty source code was
    used.
  - Integrated through extended-keys-format kitty.
  - Supports flags 1 and 8, Super (s-) and Hyper (H-).
  - Separate input-kitty.c and tty-kitty.c.
  - One saved protocol state per main/alternate screen.
  - Eager outer-terminal negotiation, restored on detach.
  - Release events are safely discarded.
  - Existing key masks remain unchanged.
  - Xterm and Kitty negotiation are mutually exclusive.
  - New files use tmux’s ISC license with the configured Git identity.

This PR follows #5405 (Add kitty keyboard protocol support (finish #4912).  All the comments were hopefully addressed.

This work adds the new extended-keys-format `kitty` which now supports `xterm`, `csi-u`, and now `kitty`.

This PR uses the emacs style naming convention:
  - `S-` = Shift
  - `C-` = Control
  - `M-` = Meta/Alt
  - `s-` = Super
  - `H-` = Hyper

PR #5615, opened 2026-09-19
Add Kitty Keyboard Protocol support with autodetect

mgrant0:
This is the same as https://github.com/tmux/tmux/pull/5600 but with autodetection.

What changed

Detection, done like sync:

- ESC[?u goes to every VT100-like terminal at attach, whatever the options.
- A reply turns on a new kittykeys terminal feature. Users can force it on or off with terminal-features, including kittykeys@ to turn it off.
- Changing extended-keys now takes effect immediately for attached clients (options.c).

Use, per client, when extended-keys is on: Kitty for clients that have kittykeys, modifyOtherKeys for clients with extkeys, VT10x otherwise.

Pane side:

- Applications can request either protocol; Kitty wins if one asks for both.
- Each application gets what it asked for, whichever client typed the key.
- ESC[?u is answered truthfully.
- The per-client gate is gone. window.c and popup.c are simply master's versions again.

extended-keys always is back to its master meaning, forcing modifyOtherKeys mode 1 only. Forcing Kitty on every program would have sent bash ESC[97;5u for Ctrl-A.

extended-keys-format kitty is still accepted, but now behaves the same as csi-u.


pdietl:
With only the disambiguate flag granted, `input_key_kitty` returns -1 for Enter, Tab and Backspace whatever the modifiers, so Shift+Enter, Ctrl+Enter, Alt+Enter and Shift+Tab reach the pane as legacy bytes. Claude Code and nvim both request exactly flag 1 (`#{pane_key_mode}` = `Kitty 1`), so in both Shift+Enter is indistinguishable from Enter; the spec's exception covers the unmodified keys only, so a shell stays usable after a crash. Gating the exception on `modifiers == 0` fixes it: pdietl/tmux@7c6ee83d on top of this branch. `kitty-keys.sh`, `kitty-keys-input.sh` and `kitty-keys-clients.sh` still pass, and Shift+Enter inserts a newline in Claude Code under kitty with it. Cherry-pick if it looks right.

Lesser: under flag 8, a key that reaches tmux without shift information (`send-keys A`, or a client on a non-kitty terminal) is encoded `CSI 65u`; the spec wants the unshifted codepoint plus the shift modifier (`CSI 97;2u`), at least for A-Z where the mapping does not depend on layout.

Written with Claude's assistance. 🤖 Pete Dietl

luisdavim:
Will this be in 3.8?

